PR #126, opened 2026-06-15
fix(sparq-hdt): cap unbounded pre-alloc in HDT loader (OOM-DoS) — sq-tzwa [OPUS-4.8]

jeswr:
> 🤖 _SPARQ agent_

Implements bead **sq-tzwa** (security, kind:hardening): cap unbounded pre-allocation in untrusted on-disk-file loaders to prevent OOM-DoS. Scoped run over **sparq-vectors, sparq-hdt, sparq-text, sparq-geo** (sparq-core already done by sq-f5jh; sparq-zk excluded — in flight on another branch).

## The bug pattern
A loader reads a count/length from an **attacker-controlled** file, then feeds it straight to `Vec::with_capacity(n)` / `vec![0u8; n]` **before** validating `n` against the bytes that exist. A hostile file with a huge declared count triggers an uncatchable OOM-abort (DoS). Fix: bound the allocation against the remaining slice / actually-delivered bytes, failing closed with the loader's existing error type — mirroring sparq-core's sq-f5jh (`read_str_bounded` / `load_pred_stats`).

## Per-crate audit result

| Crate | Vulnerable loader sites | Action |
|---|---|---|
| **sparq-hdt** | `decode.rs`: `decode_section` `num_strings`; `read_bitmap_words` `num_bits` body; triples `num_triples` | **Capped (3 sites)** |
| **sparq-vectors** | none — `.spqv`/`.spqg` `open` already validate the file count against `map.len()` via checked arithmetic **before** any count-indexed alloc (`open_validated`, `DiskAnnIndex::open`) | none |
| **sparq-text** | none — no on-disk binary loader; index built from a live `Graph`, `with_capacity` sites use in-RAM `.len()` | none |
| **sparq-geo** | none — no on-disk/mmap loader; geometries from query-time WKT (parser-validated) | none |

## Capped loaders (all `crates/sparq-hdt/src/decode.rs`)

1. **`decode_section`** — `Dict::with_capacity(n)` / `Vec::with_capacity(n)` where `n = sect.num_strings`. Upstream `DictSectPFC::read` parses `num_strings` and `packed_length` as **independent** VBytes (their CRC8 does not relate them), so a hostile section can declare `num_strings = 2^60` with a tiny `packed_data`.
   **Bound:** `cap = n.min(packed_data.len())` — every decoded string consumes ≥1 byte (its null terminator), so a section holds ≤ `packed_data.len()` strings.
2. **`read_bitmap_words`** — `vec![0u8; full_byte_amount]` where `full_byte_amount` derives from the `num_bits` VByte. A generic streaming `BufRead` has no cheap remaining-length, so instead of pre-allocating the declared size we read **through** `reader.take(full_byte_amount).read_to_end(&mut buf)` (grows only as bytes actually arrive) and then **verify the full count was delivered** (fail-closed, same semantics as the prior `read_exact`).
   **Bound:** allocation tracks bytes actually delivered, never the declared `num_bits`.
3. **triples `Vec::with_capacity`** — `num_triples = sequence_z.entries`. By here `Sequence::read` has already consumed + CRC-validated the backing words, so `entries` is bounded-by-construction; capped anyway as defence-in-depth.
   **Bound:** `num_triples.min(sequence_z.data.len() * 64)` (the tight entry-count ceiling of the read backing).

## Corruption-oracle tests (`decode.rs`)
- `read_bitmap_words_rejects_huge_num_bits_without_oom` — CRC8-valid bitmap header declaring **2^60 bits** with a truncated body ⇒ clean `Err`.
- `decode_section_rejects_huge_num_strings_without_oom` — `DictSectPFC` declaring **2^60 strings** with 1 packed byte ⇒ clean `Err`.
- `load_reader_rejects_inflated_dict_count_without_oom` — end-to-end: inflate a real fixture's shared-section `num_strings` VByte and repair the section CRC8 so the corrupted count reaches the decoder ⇒ `load_reader` returns `Err` (asserted no OOM, no panic, via `catch_unwind`).

A regression that dropped a cap would not return `Err` here — it would OOM-abort the test process (the DoS being guarded).

## Gate (all pass)
`cargo test` + `cargo clippy --all-targets --all-features -- -D warnings` green for all four in-scope crates. New code is `cargo fmt`-clean (remaining fmt diffs are the repo's pre-existing deferred-reformat backlog, untouched).

## sparq-zk deferred
sparq-zk is being modified on another branch and is **excluded** from this run; its proof/key-blob on-disk loader audit is filed as a separate **sq-tzwa follow-up bead** (id in the PR thread).

### DIFF
--- a/crates/sparq-hdt/src/decode.rs
+++ b/crates/sparq-hdt/src/decode.rs
@@ -122,8 +122,29 @@ fn read_bitmap_words<R: BufRead>(reader: &mut R) -> Result<RawBitmap, Error> {
 
     // All but the last word are full 8-byte words; the last word is byte-aligned.
     let full_byte_amount = if num_bits == 0 { 0 } else { ((num_bits - 1) >> 6) * 8 };
-    let mut full_words = vec![0u8; full_byte_amount];
-    reader.read_exact(&mut full_words)?;
+    // [OPUS-4.8] sq-tzwa (OOM-DoS hardening): `num_bits` is a VByte read straight from an
+    // attacker-controlled `.hdt` file, and `full_byte_amount` is derived from it with NO check
+    // against the bytes remaining in the stream. The previous `vec![0u8; full_byte_amount]`
+    // ALLOCATED (and zeroed) the full declared size BEFORE the `read_exact` that would discover
+    // the bytes don't exist — so a hostile `num_bits ≈ 2^60` forces a multi-exabyte allocation
+    // → an uncatchable OOM abort (DoS). The reader is a generic streaming `BufRead`, so there is
+    // no cheap "remaining bytes" to bound against up front (sparq-core's slice-backed loaders
+    // can; this one can't). Instead read THROUGH a `take(full_byte_amount)` adaptor with
+    // `read_to_end`, which grows the buffer only as bytes ACTUALLY arrive (doubling, never a
+    // single declared-size pre-allocation), then verify the full count was delivered — failing
+    // closed exactly as the prior `read_exact` did, but without ever trusting the declared
+    // length enough to pre-allocate it. Same fail-closed intent as sq-f5jh's bounded reads in
+    // sparq-core (`read_str_bounded` / `load_pred_stats`).
+    let mut full_words: Vec<u8> = Vec::new();
+    let read = reader
+        .take(full_byte_amount as u64)
+        .read_to_end(&mut full_words)?;
+    if read != full_byte_amount {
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 _SPARQ agent_
Follow-up bead filed for the deferred sparq-zk audit: **sq-3wco** — "Audit sparq-zk proof/key-blob on-disk loaders for unbounded pre-alloc (OOM-DoS) — sq-tzwa follow-up" (P2). Copilot review requested.

PR #136, opened 2026-06-15
docs(agents): background sub-agent dispatch — never block the turn on a batch

jeswr:

Strengthens *Orchestration cadence — background by default* to explicitly cover **sub-agent dispatch** (not just shell commands). Blocking the turn on a synchronous batch idles the orchestrator for the slowest agent's wall-clock AND makes the session unresponsive (the user must ctrl+b to interject). New rule: fire agents with `run_in_background`, keep fanning out / merging / responding, and act on each completion notification.

Captures a maintainer correction on orchestration behavior.


jeswr:
> 🤖 **SPARQ agent** — Fixed in `1752534`. Dropped the environment-specific `(ctrl+b)` aside in favour of generic phrasing ("force-backgrounding the running batch"), and standardised "subagent-level" → "sub-agent-level" to match the doc-wide hyphenation. Meaning intact; `[OPUS-4.8]` marker preserved. Resolving the thread.

PR #163, opened 2026-06-15
feat(fuzz): coverage-guided SHACL-validation fuzz target (sq-o4pi)

jeswr:
> 🤖 **SPARQ agent** — implementing bead **sq-o4pi**: add a coverage-guided cargo-fuzz target for SHACL validation to the sq-ovnf `fuzz/` harness.

## What

Adds `fuzz/fuzz_targets/validate_shacl.rs`, the fifth cargo-fuzz / libFuzzer target alongside the existing parser + mmap-loader targets. It drives hostile bytes through `sparq_shacl::validate(&data, &shapes)` — the SHACL Core + SHACL-SPARQL entry point — so a hostile corpus must never make the validator panic / OOM / overflow-abort / UB; only a clean parser `Err` or a **bounded** `ValidationReport`.

It exercises both halves of the validator:
- `ShapesModel::parse(shapes)` — target discovery, constraint-component + `sh:path` parsing, `sh:and`/`sh:or`/`sh:node`/`sh:property` shape-graph construction;
- `eval::validate_graph` — focus-node traversal, the **(focus, shape) recursion guard** that keeps cyclic `sh:node`/`sh:property` references from overflowing the stack (a known-hardened area — see the `cyclic_*` regression tests), constraint evaluators, and the `sh:select`/`sh:ask` routing through `sparq-engine`.

## Input-split design

`validate` only runs the traversal once a shapes graph parses, so the target feeds the corpus two ways; the **low bit of the first byte** picks the mode, so libFuzzer explores both:

- **SPLIT mode** — carve the body into a shapes chunk + a data chunk at a fuzz-chosen split point (remaining control bits pick the fraction over the **borrowed** slice), parse **each** as Turtle, validate only the pairings where **both** parse. Models two separately-supplied files.
- **SELF mode** — parse the **whole** body once as the shapes graph and validate the document against **its own** shapes. A SHACL shapes doc is itself valid RDF, so as the corpus grows toward real SHACL Turtle this reaches the deepest shape-parse + traversal code with no split-luck dependence.

Bytes are fed as `&str` via `from_utf8_lossy` over the **borrowed** `data` (no full-size `to_vec()` copy — a large mutated input never gets a harness-side copy before a byte reaches the parser). `Graph` is not `Clone` (it owns a WAL / file handles), so SELF mode re-parses the (cheap, small) buffer rather than cloning.

## Wiring

- `fuzz/Cargo.toml`: new `[[bin]] validate_shacl` + the opt-in `sparq-shacl` path dep (pulled in **only** for this target, never by the workspace default build — the sparq-reason isolation pattern holds).
- **CI fuzz lane** (`.github/workflows/fuzz.yml`): no edit needed — the lane enumerates targets via `cargo fuzz list`, so `validate_shacl` is auto-discovered and runs in the same bounded budget as the others: SMOKE (per-PR / merge_group, 30s), MAIN (push to main, 120s), NIGHTLY (600s), each capped by `-max_total_time` + `-rss_limit_mb=2048`.

## Verification

- Workspace gate: `cargo clippy --workspace --exclude sparq-py --all-targets -- -D warnings` → **clean (exit 0)**, and it does **not** compile the fuzz crate (`fuzz` is workspace-`exclude`d; `cargo build -p sparq-fuzz` errors "did not match any packages"). Stable gate unaffected.
- Built on nightly: `cargo +nightly fuzz build validate_shacl` → OK.
- Ran it: `cargo +nightly fuzz run validate_shacl -- -max_total_time=25 -rss_limit_mb=2048` → **231,316 execs in 26s, exit 0, NO crash**, peak RSS 551 MB (well under cap), 2,574 new coverage units (it is reaching meaningfully into the validator). **No sparq-shacl robustness bug surfaced** — no minimal sparq-shacl fix was needed.

## Scope

Touches only `fuzz/` (new target + `Cargo.toml`/`Cargo.lock`). No `.beads/` edits, no sparq-shacl changes (no bug found).

Bead: **sq-o4pi** (atop **sq-ovnf**).

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -733,7 +733,21 @@ dependencies = [
  "libfuzzer-sys",
  "sparq-core",
  "sparq-engine",
+ "sparq-shacl",
  "tempfile",
+]
+
+[[package]]
+name = "sparq-shacl"
+version = "0.1.0"
+dependencies = [
+ "oxrdf",
+ "oxttl",
+ "regex",
+ "rustc-hash",
+ "spargebra",
+ "sparq-core",
+ "sparq-engine",
 ]
 
 [[package]]
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -35,8 +35,10 @@ libfuzzer-sys = "0.4"
 sparq-core = { path = "../crates/sparq-core", features = ["mmap", "parallel"] }
 sparq-engine = { path = "../crates/sparq-engine" }
 # SHACL validation under test (bead sq-o4pi). Opt-in crate (the sparq-reason isolation
-# pattern); pulled in here ONLY for the `validate_shacl` target, never by the workspace
-# default build. [OPUS-4.8]
+# pattern). Used by the `validate_shacl` target. Like every entry here it is a
+# package-wide dependency — Cargo deps are not scoped per [[bin]] target — but it stays
+# out of the workspace default build because this fuzz crate is excluded from the root
+# workspace (see the header note). [OPUS-4.8]
 sparq-shacl = { path = "../crates/sparq-shacl" }
 # Temp store directories for the `Graph::open` hostile-bytes target.
 tempfile = "3"
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -34,6 +34,10 @@ libfuzzer-sys = "0.4"
 # store) and `parallel` (for `load_reader_parallel`'s pipelined N-Triples path).
 sparq-core = { path = "../crates/sparq-core", features = ["mmap", "parallel"] }
 sparq-engine = { path = "../crates/sparq-engine" }
+# SHACL validation under test (bead sq-o4pi). Opt-in crate (the sparq-reason isolation
+# pattern); pulled in here ONLY for the `validate_shacl` target, never by the workspace
+# default build. [OPUS-4.8]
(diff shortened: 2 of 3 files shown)


jeswr:
> 🤖 SPARQ agent

Addressed all three Copilot review threads (surgical fixes only — fuzz invariant and coverage unchanged):

1. **SPLIT-mode split calc** — replaced `(ctrl >> 1) * body.len() / 128` with `(ctrl >> 1) % (body.len() + 1)`. This is overflow-safe under `overflow-checks = true` (the old multiply could overflow `usize` for fuzz-sized inputs and abort as a false-positive crash) and now reaches both `split == 0` (empty shapes) and `split == body.len()` (empty data), so the data-empty extreme the comment promised is actually reachable. Comment corrected to match.
2. **SELF mode double-parse** — `validate` borrows both graphs immutably and does not mutate, so the buffer is now parsed once and the same `&Graph` is passed as both `data` and `shapes`. Redundant second parse removed.
3. **`fuzz/Cargo.toml` comment** — reworded so it no longer implies per-target dependency isolation; clarified that `sparq-shacl` is a package-wide dep (like every other) but stays out of the workspace default build because the fuzz crate is excluded from the root workspace.

Verification (EC2 box — timing non-canonical): `cargo +nightly fuzz build validate_shacl` succeeds; `cargo +nightly fuzz run validate_shacl -- -max_total_time=15 -rss_limit_mb=2048` exited 0, no crash (148,930 runs).

PR #168, opened 2026-06-15
MPC: in-MPC bit-decomposition of the secret-shared sum in disclose_threshold_verdict (sq-g7t5)

jeswr:
> 🤖 SPARQ agent

Bead **sq-g7t5** — remove the in-process-simulation local reconstruct from `disclose_threshold_verdict` and compare via secret-shared bit-decomposition, opening ONLY the verdict bit.

## The shortcut that was removed
`crates/sparq-mpc/src/compare.rs` previously did, at the top of `disclose_threshold_verdict`:
```rust
let sum = backend.reconstruct(sum_shares)?;   // <-- local reconstruct of the secret sum
```
then re-dealt the cleartext sum's bits via `secure_threshold`/`share_bits`. In a real multi-party protocol that opens the sum. **That call is gone** — there is no reconstruct-the-sum path left.

## Security tier implemented — honest-majority, **semi-honest** (NOT malicious)
Inherits the `ShamirBackend` model exactly. Every multiplication in the decomposition + comparison chain routes through `degree_reduce`, which has no in-protocol check on a deviating party's re-sharing — so this is semi-honest-only. **No malicious security is claimed.** `OperatorClass::Comparison` continues to report `semi_honest_only` (unchanged).

## Bit-decomposition approach + magnitude bound
Masked-open bit-decomposition (Damgård et al. TCC'06), specialised to this crate's Shamir-over-`F_p` primitives:
1. Deal fresh **dealer-random** solved-bits `([r], [r_0..r_{L-1}])`, `r` uniform `L=60`-bit (the mask is masking material, not any party's input).
2. **Masked open** `c = sum + r` — the ONLY opening; statistically hides the sum (advantage `2^-κ`, `κ = 40`). The sum's shares are never all brought together; `reconstruct(sum_shares)` is never called.
3. Recover the sum's bits via a secret-shared **bitwise-subtraction borrow chain** `c ⊖ [r]` (each step local affine ops + one secure mult through `degree_reduce`), then feed the public-threshold comparator and open ONLY the verdict bit.

**Magnitude bound (documented + fail-closed):** sum/threshold must be `< 2^DECOMP_VALUE_BITS = 2^20 = 1_048_576` (mask 60 bits − statistical gap κ=40). Much tighter than the cleartext-operand `COMPARE_BITS=60`, because `p = 2^61-1` has no headroom for a statistically-hiding mask above ~20 value bits. It still covers the four-flatmates use case (£10^6 < 2^20). The masked open provably does not wrap `p`: `(2^20−1)+(2^60−1) < p`.

The verdict semantics are unchanged: strict `sum > threshold`.

## Privacy test (the deliverable's proof)
Mirrors the sq-km34.1 α-never-reconstructed style:
- `masked_open_is_independent_of_the_sum` — the only opened intermediate `c = sum + r` is consistent with a different sum under a different legal mask, so it reveals nothing about which sum it came from.
- `verdict_output_does_not_distinguish_sums_on_the_same_side` — two very different sums on the same side of the threshold disclose a **byte-identical** partial. A reconstruct-the-sum implementation could not satisfy this; these are the **regression guards that FAIL if a local reconstruct is reintroduced.**
- `disclose_threshold_in_mpc_only_the_verdict_bit_is_disclosed` — the rendered partial never contains the sum integer.

## Test matrix
- `bit_decompose_recovers_the_true_bits_without_opening_the_sum` — recovered shared bits reconstruct to the true (never-opened) sum bits across `0, 1, 42, 99_999, 100_000, 100_001, 250_000, 2^20-1`, for n ∈ {3,5,7}.
- `disclose_threshold_in_mpc_matches_plaintext_across_boundary` — verdict == plaintext `sum > t` for sum `< t`, `== t`, `== t-1`, `== t+1`, `>> t`, `== 0`, max-range, threshold 0, n ∈ {3,5,7}.
- `disclose_threshold_in_mpc_multi_holder_sum_boundary` — aggregate of multiple holders' shared inputs at the boundary.
- field-edge/overflow safety, valid degree-`t` verdict sharing, `n < 2t+1` fail-closed, updated out-of-range-threshold test to the new `2^20` bound.

## Gate
`cargo build -p sparq-mpc`, `cargo clippy -p sparq-mpc --all-targets -- -D warnings`, `cargo nextest run -p sparq-mpc` (**216 passed**), `cargo fmt` — all clean. (Timing on this EC2 box is non-canonical.)

## Follow-up bead filed
**sq-mnv5** (discovered-from sq-g7t5) — deployment-grade version: (1) real random-bit sub-protocol so no party knows `r`; (2) wider magnitude via a larger field or a non-masked-open comparison (Rabbit eprint 2021/119); (3) malicious security via IT-MACs through the chain (design-doc Hole 4).

### DIFF
--- a/crates/sparq-mpc/src/compare.rs
+++ b/crates/sparq-mpc/src/compare.rs
@@ -133,6 +133,46 @@ pub const COMPARE_BITS: usize = 60;
 /// The exclusive upper bound an operand must respect: `2^60`.
 pub const COMPARE_MAX_EXCLUSIVE: u64 = 1u64 << COMPARE_BITS;
 
+// =============================================================================
+// [OPUS-4.8] sq-g7t5 — in-MPC bit-decomposition of an EXISTING secret-shared sum
+// (replacing the local-reconstruct shortcut in `disclose_threshold_verdict`).
+// =============================================================================
+
+/// Statistical security parameter `κ` for the masked-open bit-decomposition
+/// (sq-g7t5). The mask `r` is drawn uniformly from `[0, 2^`[`DECOMP_MASK_BITS`]`)`,
+/// and the value being decomposed is bounded by `2^`[`DECOMP_VALUE_BITS`]`)`, so the
+/// opened `c = value + r` carries at most `2^{-κ}` statistical advantage about the
+/// value (standard statistical-masking argument; Damgård et al. TCC'06). `κ = 40`
+/// gives a `2^{-40}` distinguishing bound — the same order as the crate's other
+/// statistical claims. NOT cryptographic-strength (`2^{-80+`); see the module-level
+/// "magnitude bound" doc for why `p = 2^61−1` forces this trade-off and the
+/// follow-up bead for a wider-field / square-root-random-bit path.
+pub const DECOMP_STAT_SECURITY_BITS: usize = 40;
+
+/// Bit-width of the random mask `[r]` in the masked-open bit-decomposition. The
+/// mask must be wider than the value by the statistical-security gap `κ`
+/// ([`DECOMP_STAT_SECURITY_BITS`]) AND `value + r` must not wrap `p = 2^61−1`. We
+/// fix the mask at `2^60` (uniform 60-bit), so `value + r < 2^60 + 2^60 = 2^61`,
+/// which is below `p`'s representable-without-wrap region only when the value is
+/// strictly below `2^60` — guaranteed by [`DECOMP_VALUE_BITS`] `< 60`.
+pub const DECOMP_MASK_BITS: usize = 60;
+
+/// **The supported magnitude bound** of a sum the in-MPC bit-decomposition can
+/// compare (sq-g7t5): the sum must be `< 2^`[`DECOMP_VALUE_BITS`]. Derived as
+/// `DECOMP_MASK_BITS − DECOMP_STAT_SECURITY_BITS = 60 − 40 = 20`, i.e. the sum
+/// must be `< 2^20 = 1_048_576`. This **exactly covers the four-flatmates use
+/// case** (4 × ~£10^5 ≈ £10^6 < 2^20). It is deliberately MUCH smaller than the
+/// cleartext-operand [`COMPARE_BITS`] (60) because the in-MPC path masks the sum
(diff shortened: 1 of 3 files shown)


jeswr:
> 🤖 SPARQ agent

Addressed all 5 review threads (commit \`cbdce76\`):

1. **Security-level wording** — dropped the malformed \`2^{-80+\` notation; the doc now states plainly that κ=40 is a *statistical* (not cryptographic) bound, that a conservative cryptographic bound would be 2⁻⁸⁰ or tighter, and that this field cannot reach it.
2. **Random-bit bias (real fix, not a doc patch)** — mask bits were derived from \`draw_fp().value() & 1\`, whose LSB is biased toward 0 because \`draw_fp()\` is uniform over \`[0, p)\` with \`p = 2⁶¹−1\` *odd* (one extra even value). Added \`ShamirDealer::draw_bit()\` = LSB of a raw \`next_u64()\` (exactly unbiased, since the CSPRNG word is uniform over all 2⁶⁴ values) and switched \`deal_random_solved_bits\` to it; doc updated to match. New tests \`draw_bit_is_uniform\` (empirical 1-rate within 2% of ½ over 100k draws) and \`draw_bit_yields_only_zero_or_one\`.
3. **Per-borrow-step mult count** — corrected to **three secure multiplications per bit** (\`x∧borrow\`, \`borrow∧¬x\`, and the \`∧\` inside \`secret_or\`), i.e. \`3·L\` mults / depth \`O(L)\`, not one per step.
4. **Fail-closed overclaim** — doc now states the asymmetry honestly: \`public_threshold\` IS range-checked fail-closed, but the **sum bound is a caller precondition** (it cannot be validated from shares without disclosure); violation is silently wrong with no in-protocol guard, and an in-protocol range proof is the residual follow-up.
5. **Throwaway dealer** — \`disclose_threshold_verdict\` now reads \`n\`/\`t\` via \`ShamirBackend::parties()/threshold()\` instead of OS-seeding a dealer just to read them (\`check_party_count\` now takes \`(n, t)\`).

\`cargo build\` / \`cargo nextest run\` (218 pass) / \`cargo clippy -D warnings\` all green; the two changed files are \`rustfmt --check\`-clean (the workspace-wide fmt drift is the deferred, non-blocking CI item per ci.yml).

PR #174, opened 2026-06-15
feat(mpc): bounded property-path over disclosed-key regime (sq-py8h.1)

jeswr:
> 🤖 SPARQ agent

Closes the **disclosed-key** slice of bounded property paths under MPC — increment #1 of `research/mpc-bounded-property-path-design.md` §6 step 1 (§2.1 DISCLOSED regime). Bead **sq-py8h.1** (parent **sq-py8h**).

## The unroll construction (crypto-free)

A bounded path `?a (p){m,k} ?b` whose endpoints **and** intermediate join keys are disclosed global IRIs is evaluated by *unrolling statically on the PUBLIC bound `k`* into a finite set of fixed-length BGP chains, then folding + unioning + deduping — all **OUTSIDE the cryptographic core** (architecture convention #4):

- **exactly-`ℓ` chain** `?a p ?z1 . ?z1 p ?z2 . … ?z_{ℓ-1} p ?b` → a left-to-right fold of the existing [`DisclosedKeyJoin`](crates/sparq-mpc/src/join.rs) on the fresh intermediate vars, then projected onto the endpoints. This is the §2.1 core — the *exact* shape of the join crate's `differential_three_holder_chain_equals_union` test, generalised to a bounded path.
- **`{m,k}`** → UNION of the exactly-`ℓ` chains for `ℓ in m..=k`.
- **`{0,k}` reflexive** → the `{1,k}` union PLUS the length-0 identity pairs `(x,x)` over the disclosed node domain (design §2.3); `p?` is the `{0,1}` special case.
- **alternation** `(p1|p2|…)` → each hop position expands over its alternatives (`alternatives^ℓ` fixed chains, design §2.4).
- **endpoint-pair DEDUP** → each distinct `(a,b)` appears once regardless of how many lengths/branches reached it (set semantics, §2.2/§2.5). The realized hop length is **never** in the output (§1.2 length-revealing prohibition).

**No secret sharing, no `secure_equal`, no `CommCounter` — NO MPC round runs.** It is a plaintext fold of equi-joins over disclosed global IRIs.

## Regime boundary (stated, not hidden)

**DISCLOSED-KEY ONLY.** Every endpoint and every intermediate `?z_i` is a disclosed global IRI. This is **NOT** the HIDDEN-intermediate regime (secret `?z_i` kept secret-shared via `secure_equal` / `degree_reduce` / `oblivious_set_output`) — that is the separate cryptographic deliverable **sq-py8h.2**, which *this bead blocks*. Disclosed-key bounded-path support is not hidden-path support.

**Semantic boundary:** implements the bounded `{m,k}` form exactly (a well-defined SPARQL 1.1 construct), not the unbounded `+`/`*` closure — a pair connected only by a chain longer than `k` is correctly absent.

## Differential-test matrix (13 tests, all passing)

Acceptance = the federated disclosed-key unroll **==** the clear-text engine's `eval_path` over the UNION store (mirrors `differential_three_holder_chain_equals_union`). The bounded `{m,k}` form is expressed to the engine as the equivalent SPARQL path it carries (union of fixed-length chains), so the equality is exact.

| Form | Test |
|---|---|
| plain sequence `p1/p2/p3` (3 holders) | `differential_plain_sequence_equals_eval_path` |
| exact `{3}` | `differential_exact_k_equals_eval_path` |
| range `{1,3}` | `differential_range_m_k_equals_eval_path` |
| range `{2,3}` (excludes length-1) | `differential_range_min_two_excludes_length_one` |
| reflexive `{0,2}` (identity pairs **exactly once**) | `differential_reflexive_zero_k_identity_pairs_once` |
| `p?` = `{0,1}` | `differential_zero_or_one_p_question` |
| alternation exact-2 / range `{1,2}` | `differential_alternation_exact_two_equals_eval_path`, `differential_alternation_range_one_two` |
| multi-length dedup to one pair | `dedup_collapses_multi_length_to_single_pair` |
| **crypto-free / round-count == 0** | `crypto_free_no_mpc_round_artifacts` |
| Protocol-error soundness | `edge_partial_wrong_arity_…`, `min_greater_than_max_…`, `absent_predicate_…` |

**Crypto-free assertion:** `crypto_free_no_mpc_round_artifacts` evaluates every supported form and asserts a `CommCounter`'s `mult_rounds == 0` and `open_rounds == 0` — the disclosed-key call graph reaches only `DisclosedKeyJoin` and plain `PartialResult`s, so no round can be recorded by construction.

## Verify

```
cargo build  -p sparq-mpc                                  # ok
cargo nextest run -p sparq-mpc                             # 220 passed, 0 failed (13 new)
cargo clippy -p sparq-mpc --all-targets -- -D warnings     # clean
```

All forms in the acceptance list are covered; nothing deferred from increment #1. No follow-up beads filed (sq-py8h.2 hidden-intermediate already exists per design §6 step 2).

Refs sq-py8h.1, parent sq-py8h, `research/mpc-bounded-property-path-design.md` §6 step 1.

### DIFF
--- a/crates/sparq-mpc/src/bounded_path.rs
+++ b/crates/sparq-mpc/src/bounded_path.rs
@@ -218,6 +218,39 @@ impl DisclosedEdges {
                     part.vars.len()
                 )));
             }
+            // Soundness: the disclosed-key regime joins on GLOBAL IRIs. Both
+            // endpoints of every disclosed edge MUST be NamedNodes. A blank node
+            // would be unsound — blank-node labels are only document-scoped, so two
+            // holders emitting the same label do NOT denote the same term, yet the
+            // disclosed-key equi-join would treat equal labels as equal terms
+            // (a false cross-holder join). A literal endpoint is outside the
+            // global-IRI key regime entirely. Reject either at ingest (fail fast)
+            // rather than silently producing unsound joins (design §2.1 DISCLOSED
+            // regime boundary).
+            for row in &part.rows {
+                if row.len() != 2 {
+                    return Err(MpcError::Protocol(format!(
+                        "disclosed edge row for <{}> must have width 2; got {}",
+                        pred.as_str(),
+                        row.len()
+                    )));
+                }
+                for (col, slot) in row.iter().enumerate() {
+                    match slot {
+                        Some(Term::NamedNode(_)) => {}
+                        other => {
+                            return Err(MpcError::Protocol(format!(
+                                "disclosed edge for <{}> has a non-IRI endpoint in \
+                                 column {col} ({other:?}); the disclosed-key regime \
+                                 requires global-IRI (NamedNode) endpoints — blank \
+                                 nodes are document-scoped (unsound to join across \
+                                 holders) and literals are out of regime",
+                                pred.as_str()
+                            )));
+                        }
+                    }
+                }
+            }
             // Re-key onto the canonical (?__pp_edge_s, ?__pp_edge_o) schema so the
             // holder's chosen variable names cannot influence the join.
             let rekeyed = PartialResult {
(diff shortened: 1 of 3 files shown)


jeswr:
> 🤖 SPARQ agent

Addressed all 5 Copilot review threads in 6860847 (all crypto-free; 13 differential tests + the `mult_rounds==0` crypto-free assertion still pass, 2 new differential tests added):

- **Non-IRI endpoint rejection** (`from_holder_edges`): now rejects blank-node/literal endpoints and rows of width ≠ 2 at ingest with a `Protocol` error. Soundness fix — blank-node labels are document-scoped, so a disclosed-key equi-join treating equal labels across holders as equal terms would be a false cross-holder join; literals are out of the global-IRI key regime.
- **`project_endpoints` fails fast**: returns `Result<_, MpcError::Protocol>` when an endpoint var is missing rather than silently returning an empty relation (a missing endpoint is an internal unroll-invariant violation, not "no results").
- **`dedup_endpoint_pairs`**: computes each row key once (O(n) allocations) and sorts on the precomputed key instead of re-`format!`-ing both operands per comparison.
- **`alternation_chains`**: dropped the `chains.len() * alternatives.len()` (== `alternatives^length`) `with_capacity` that could overflow `usize` / request a huge allocation → panic; grows amortised instead.
- **New tests**: `differential_sequence_with_repetition_element` and `differential_sequence_with_alternation_repetition_element` exercise `PathForm::Sequence` with a bounded-repetition / alternation-repetition element (the per-part-dedup branch of `eval_sequence`) — federated unroll == clear-text `eval_path`, mid multiplicity deduped.

`cargo nextest run -p sparq-mpc`: 233/233 pass. `cargo clippy -p sparq-mpc --all-targets -- -D warnings`: clean.

PR #191, opened 2026-06-15
feat(sparq-mpc): batched/vector secret sharing + row-binding (sq-dwb5)

jeswr:
> 🤖 SPARQ agent

Generalises the single-scalar `share_private_input` (which extracted exactly ONE integer per holder — single-row / single-column) to **batched / vector secret sharing**, so the secure aggregate and the hidden-value join can range over more than one private value per source. Closes the cap flagged in **sq-dwb5**.

## Batched-sharing API (added alongside the single path — nothing broken)

- **`batched` module** — `BatchedShares` (plain) and `BatchedAuthShares` (IT-MAC authenticated) carriers, each holding one per-element Shamir sharing per row plus a `RowBinding`.
- **`ShamirDealer::share_batch(values: &[Fp]) -> Vec<ShareVec>`** — element-wise sharing; **each element gets its own fresh degree-`t` masking polynomial** (NOT one polynomial carrying all values), so per-element privacy is preserved for the whole vector simultaneously.
- **`MacSession::authenticated_share_batch(&[Fp])`** — the authenticated analogue; every element authenticated under the session's single `α`.
- **`MpcBackend::share_private_inputs(holder, fragment)`** — trait method with a default single-scalar fallback (existing stub backends keep compiling unchanged); `ShamirBackend` overrides it with the real vector path via a new `extract_integer_vector`.
- **`batched::per_row_sum`** — the multi-row secure aggregate (the zero-round Shamir linear fold lifted to a vector).

## Row-binding contract (documented, load-bearing)

`RowBinding` records how element `i` correlates across holders:
- **`Positional`** — element `i` is row `i`; holders correlate by INDEX. Sound only when both holders impose the same deterministic order, so `share_private_inputs` **value-sorts** before sharing → positional batches from different holders line up regardless of local triple/row order.
- **`Keyed(Vec<String>)`** — element `i` bound to the DISCLOSED public row key `keys[i]` (a global IRI / row id, convention #6); holders correlate by KEY, orders need not match. The key column is disclosed (convention #4); the value stays hidden.

Both preserve the single-scalar hiding guarantee element-wise.

## Wired end-to-end vs deferred

- **Wired:** batched share → element-wise reconstruct; the multi-row **secure aggregate** `per_row_sum`, demonstrated end-to-end from two multi-row holders (`share_private_inputs` → `per_row_sum` → reconstruct == plaintext per-row sum).
- **Deferred (honestly):** the full batched **hidden-value join** (`HiddenValueJoin` ranging over many rows/holder via these batches + the `oblivious_join` output transform) is a larger wiring job. The primitive + binding it needs is landed; the join integration is a follow-up. **Suggest a follow-up bead** (orchestrator: `bd` not available in the isolated worktree) for "wire `BatchedShares`/`RowBinding` into `HiddenValueJoin` (multi-row hidden join + oblivious output)".

## Test matrix

- **Round-trip:** a `Vec<Fp>` shares + reconstructs element-wise across n ∈ {3,5,7}.
- **Privacy:** ≤ t parties' shares of the whole batch are independent of the values (single-scalar test extended to the vector, per element).
- **Authenticated:** batched shares carry IT-MACs; MAC relation `α·xᵢ` holds element-wise; `α` never reconstructable through the batched API (x=1 α-leak guard, mirroring the single-scalar pattern).
- **Row-binding:** multi-holder per-row secure sum == plaintext per-row sum (the correlation differential); plus a two-holder multi-row end-to-end holder test.
- **Backward-compat:** the single-scalar `share_private_input` path is unchanged; `share_private_inputs` over the fixed single-salary fragment yields the same length-1 batch.
- Fail-closed guards: misaligned/keyed batches rejected by `per_row_sum`; keyed key-count mismatch rejected at construction; empty batch round-trips.

## Verify

`cargo build -p sparq-mpc && cargo nextest run -p sparq-mpc && cargo clippy -p sparq-mpc --all-targets -- -D warnings` all pass — **251 tests, 0 failures**; clippy clean. (Non-canonical EC2 timing — not baked into docs.)

References sq-dwb5.

### DIFF
--- a/crates/sparq-mpc/src/backend.rs
+++ b/crates/sparq-mpc/src/backend.rs
@@ -733,13 +733,17 @@ pub trait MpcBackend {
     /// secure aggregate / hidden-value join can range over many rows per holder.
     ///
     /// **Row-binding contract.** The binding is POSITIONAL: every holder evaluates
-    /// the SAME `fragment` projecting one integer column, ordered DETERMINISTICALLY
-    /// (the implementation sorts the holder's rows by their disclosed value so two
-    /// holders' batches line up by position regardless of local row order — see
-    /// [`crate::shamir::ShamirBackend::share_private_inputs`]). Downstream
-    /// (per-row secure aggregate, hidden join) correlate across holders by output
-    /// index. For a KEYED binding (element `i` ↔ a disclosed row id), pair the
-    /// returned batch with the holder's disclosed key column out-of-band.
+    /// the SAME `fragment` projecting one integer column, ordered DETERMINISTICALLY.
+    /// Nothing about the column is disclosed — the values ARE the secret inputs. The
+    /// implementation imposes the deterministic order by **sorting the holder's rows
+    /// by their (secret) value LOCALLY, before any sharing**, so two holders' batches
+    /// line up by position regardless of local row order; the values themselves never
+    /// leave the holder — only their shares do (see
+    /// [`crate::shamir::ShamirBackend::share_private_inputs`]). Downstream (per-row
+    /// secure aggregate, hidden join) correlate across holders by output index. For a
+    /// KEYED binding (element `i` ↔ a row id that IS disclosed), the caller pairs the
+    /// returned batch with the holder's disclosed key column out-of-band and uses that
+    /// public key — not the secret value — to define the cross-holder order.
     ///
     /// Returns one trait-`Share` PER ROW (so a single-scalar contribution is the
     /// `k = 1` case). Each `Self::Share` is the per-party sharing of one private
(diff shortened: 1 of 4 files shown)


jeswr:
> 🤖 SPARQ agent

Addressed both review threads in `a4da737`:
1. `extract_integer_vector` (shamir.rs) now validates per-row arity (`row.len() != 1` → fail closed) so a desynced/adversarial `PartialResult` with extra columns can't silently mis-extract by dropping trailing cells. Added a unit test.
2. Corrected the `share_private_inputs` trait doc (backend.rs): the ordering is NOT by a "disclosed value" — the values ARE the secret inputs. Now documents that the holder value-sorts LOCALLY before sharing, nothing about the column is disclosed, and the KEYED binding uses a disclosed KEY column.

`cargo nextest -p sparq-mpc` 252 passed; clippy clean. Auto-merge re-armed.

PR #192, opened 2026-06-15
feat(vectors): stream bulk .npy/dump embedding import (sq-3jc8)

jeswr:
> 🤖 SPARQ agent

## Summary

Makes `VectorStore::import_npy` / `import_numeric_dump` **streaming** so peak heap is `O(dim + index)` instead of **~2× the embedding matrix**. Memory optimisation only — no semantics change. Refs **sq-3jc8** (follow-up to **sq-xsq9**, which added the bulk import).

## The old ~2×-peak path (`crates/sparq-vectors/src/import.rs`)

Both import paths read the **whole** input file into a `Vec<u8>` (`std::fs::read`), decoded it into a full owned flat `Vec<f32>` (`decode_npy_body`), then handed that to `write_store` (`VectorStore::create` + `put` + `finalize`), which buffered the **whole dense payload** in RAM before `finalize`. Peak ≈ input file + decoded matrix → ~2× the matrix.

## The streaming approach

1. **Header-parse only a bounded prefix.** For `.npy`, read just `≤ MAX_NPY_HEADER_LEN + 12` bytes (capped at file length) and parse the header — the array body is never read to parse. `parse_npy_header` is unchanged.
2. **Validate before any body read.** dtype / shape / fortran_order / dim / row-count, plus the declared body length cross-checked against the actual file length (`data_offset + rows·dim·elem_size`) — fail closed.
3. **Feed the body row-by-row.** Seek to `data_offset`, then for each of the `rows` rows read exactly `dim·elem_size` bytes from a `BufReader` into a **reusable** raw byte buffer, decode into a **reusable** `dim`-element `f32` scratch, and `put` it into a `StreamingWriter` — which appends each row straight to the on-disk `.spqv` data section. The numeric-dump path streams the same way (no header).

`StreamingWriter` already keeps build-phase memory O(1) on disk; the only row-count-proportional allocation is the transient `8·rows`-byte id→slot index it sorts at `finalize` (a `dim·4 / 8` reduction vs the dense matrix). No full-file `Vec<u8>` and no full owned `Vec<f32>` copy are ever materialised.

## Peak-RAM improvement

Per-row working set is two small reusable buffers (`dim·elem_size` raw + `dim` f32); peak resident memory is `O(dim + index)` rather than `O(n·d)` twice.

## Format handling (unchanged)

2-D, C-order, little-endian `f4`/`f8` only; `f8` narrowed to `f32`. Big-endian, `fortran_order: True`, non-2-D, unsupported dtype, dim / row-count and body-length mismatches all fail closed (no silent reinterpretation).

## Equivalence + chunking + edge tests (`crates/sparq-vectors/tests/import.rs`)

- **Equivalence:** the streamed `.spqv` is asserted **byte-identical** to a buffered `VectorStore::create`+`put`+`finalize` reference build — for f32, f64-narrowed-to-f32, and the numeric dump. Reopened store returns every row's exact vector in order.
- **Multi-chunk:** 257×13 (prime row count) and 200×8 exercise the row loop many times.
- **Edge shapes:** single row, `d=1`, empty matrix (each also byte-compared to the buffered build).
- Existing round-trip + fail-closed tests retained; duplicate-id is now caught at `StreamingWriter::finalize` (still fail-closed). The obsolete whole-body `decode_npy_body` + its unit test were removed.

## Verify

`cargo build -p sparq-vectors` ✓ · `cargo nextest run -p sparq-vectors` → 96 passed, 1 skipped ✓ · `cargo clippy -p sparq-vectors --all-targets -- -D warnings` clean ✓ · `cargo fmt` (import.rs + tests/import.rs formatted) ✓

### DIFF
--- a/crates/sparq-vectors/src/import.rs
+++ b/crates/sparq-vectors/src/import.rs
@@ -436,7 +436,14 @@ impl VectorStore {
         reader
             .seek(SeekFrom::Start(header.data_offset as u64))
             .map_err(|e| format!("seek {}: {e}", npy_path.display()))?;
-        let row_bytes = header.dim * elem;
+        // Checked: validate the per-row buffer size BEFORE allocating it. `want_body` above only
+        // bounds `rows*dim*elem`; with `rows == 0` an absurd `dim` would never have been multiplied
+        // in, so guard `dim*elem` directly so a malicious header can't overflow or force a huge
+        // allocation here. Fail closed. [OPUS-4.8] sq-3jc8
+        let row_bytes = header
+            .dim
+            .checked_mul(elem)
+            .ok_or("npy: dim*elem_size overflows usize")?;
         let mut raw = vec![0u8; row_bytes];
         let npy_path_owned = npy_path.to_path_buf();
         stream_store(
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 SPARQ agent

Addressed all 3 review threads in `2a6b65c`:
1. (dump grow race) The streaming dump importer only caught a SHRINK race; a GROW race (bytes appended after the size check) was silently ignored since we stop after `rows` rows. Now, after the final declared row, the reader must be at EOF — trailing bytes are rejected, restoring the pre-streaming fail-closed-on-growth semantics.
2./3. (overflow) `row_bytes` for both the npy (`dim*elem`) and dump (`dim*4`) paths used unchecked multiplication before allocating; the `rows*dim*…` length checks don't bound `dim*…` when `rows == 0`. Both now use `checked_mul` and fail closed on overflow/absurd dim, before any allocation. Added a deterministic absurd-dim rejection test.

`cargo nextest -p sparq-vectors` 97 passed; clippy clean. Auto-merge re-armed.

PR #194, opened 2026-06-15
feat(sparq-mpc): batched hidden-value join over row columns + oblivious output (sq-khf9)

jeswr:
> 🤖 SPARQ agent

Wires the **sq-dwb5 `BatchedShares`/`RowBinding` primitive** (#191) and the **sq-jnkm oblivious output transform** into `HiddenValueJoin` so the hidden-value join can range over a **column of rows per holder** under a documented row-binding, with the output cardinality/ordering hidden. Closes the JOIN half of the multi-row work (the multi-row secure AGGREGATE landed in #191). Refs **sq-khf9**, **sq-dwb5**.

## How the batched inputs + RowBinding wire into the join

- **`BatchedHiddenInput`** (new): a holder's per-row `(private_key_fp, disclosed_payload)` column **plus its `RowBinding`**. `shared_keys()` deals the private key column as a **`BatchedShares`** via `ShamirDealer::share_batch` — the sq-dwb5 primitive, actually used: each row's key gets its own fresh degree-`t` masking polynomial, so any `≤ t` parties' shares of the whole key column are jointly independent of every key. The disclosed payload columns ride alongside in the clear (disclosure-minimisation, convention #4 — only the *key* is hidden).
- **`HiddenValueJoin::batched_join(left, right, B)`** lifts the existing `secure_equal` to the row dimension. The `RowBinding` decides the **candidate pairs**:
  - **Positional** — index-correlated: row `i` of L pairs only with row `i` of R; both columns must be equal length (fails closed otherwise). `k` candidates.
  - **Keyed** — correlated by the **disclosed** public row key (orders need not match): `(i,j)` is a candidate iff `lkeys[i] == rkeys[j]`; the VALUE stays hidden and is still gated by `secure_equal`.
  - Mixed bindings → protocol error.

## The oblivious output

Each candidate's match (decided by `secure_equal`) feeds **`oblivious_join_output`** as a `MatchBit::Public` bit with a **public padded bound `B`**: non-matches become indistinguishable dummies, the slots are oblivious-shuffled, and exactly `B` are revealed. So the **output reveals neither the true match count (bounded to `B`) nor which input pair produced which row** (L1-bounded / L2-destroyed at the output). Fails closed if `B < candidate count` (never truncates a candidate → never silently drops a match).

## Regimes landed vs deferred

- **Landed:** both **Positional** and **Keyed**, with the join key/value never disclosed + the oblivious output transform.
- **Deferred (honest, beaded — `sq-xhaw`):** the per-candidate match **bit** is still *opened* by `secure_equal` (the same decision-time L2 leak the scalar `HiddenValueJoin` has). Deriving a **secret-shared** match bit from the keys without opening it needs a secure equality-to-shared-bit (sq-rrz4 on the degree-reduction round sq-dvuc) — the gate `oblivious_join`'s hidden-key entry point already names. We do **not** claim the match graph is hidden from the parties. Semi-honest only, unchanged from the `ShamirBackend` layer.

## Privacy property enforced (and how)

- **Join key/value never reconstructed** — only `secure_equal`'s masked product `m = d·r` is opened per candidate (`m=0 ⇔ keys equal`; for unequal keys `m` is uniform nonzero → reveals only the match bit). The sq-dwb5 batched-Shamir hiding lifts this to the whole column.
- **Output cardinality + ordering oblivious** to the parties — bounded to `B`, shuffled (the `oblivious_join` substrate). The key var is never in the output schema.

## Test matrix (13 new, in `join::batched_hidden_value_tests`)

- Positional + Keyed **differential vs plaintext** join across **n ∈ {3,5,7}** (keys private, never reconstructed).
- Keyed **fan-out** within a shared disclosed-key bucket (2×2 cartesian).
- Privacy: **output cardinality bounded to `B`** (not true count); **key value never appears** in the disclosed output; **output order oblivious** across 32 seeds.
- **Forged-match soundness** across n ∈ {3,5,7} (distinct keys cannot manufacture a match; genuine-equal control passes).
- Fail-closed: mixed bindings / unequal positional columns / `B < candidates` / keyed key-count mismatch.
- `shared_keys` exercises the batched primitive (length + binding preserved).

## Verify

```
cargo build -p sparq-mpc      # clean
cargo nextest run -p sparq-mpc  # 265 tests run: 265 passed, 0 skipped
cargo clippy -p sparq-mpc --all-targets -- -D warnings  # clean
cargo fmt
```

NON-CANONICAL timing (this box is an EC2 work box, not a bench instance).

### DIFF
--- a/crates/sparq-mpc/src/join.rs
+++ b/crates/sparq-mpc/src/join.rs
@@ -41,7 +41,7 @@ use crate::batched::{BatchedShares, RowBinding};
 use crate::field::Fp;
 use crate::oblivious_join::{self, Candidate, MatchBit, ObliviousOutputCost};
 use crate::partial::{HolderId, MpcError, PartialResult};
-use crate::shamir::{self, ShamirBackend, ShamirDealer};
+use crate::shamir::{self, ShamirBackend, ShamirDealer, Share};
 use oxrdf::{Term, Variable};
 use std::collections::BTreeMap;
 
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 SPARQ agent

Resolved all 4 Copilot threads (commit `546a353`):

1. **(REAL fix)** `batched_join` was dealing `BatchedShares` via `shared_keys()` then discarding them and re-sharing each scalar inside `secure_equal`. Added `secure_equal_shared()` that consumes already-dealt `ShareVec`s; `batched_join` now compares the per-row sharings from the up-front `BatchedShares` (keys dealt **once**, not re-shared per candidate). `secure_equal` delegates to the shared core so the scalar path is unchanged. Join correctness + privacy verified.
2. **(doc)** Corrected the `BatchedHiddenInput` doc: cleartext keys are inputs to Shamir sharing **and** the equality test, never reconstructed from shares.
3. **(test)** Strengthened the privacy test to assert the revealed slot-vector length is exactly *B* for 1 vs 3 true matches and that the hidden key never appears in any revealed slot — not just the `cost.bound` echo.
4. **(comment)** Fixed the `forged_match_attempt_fails` comment to match the (42,42) control / (42,43) forge data.

`cargo build` + `nextest` (265 passed) + `clippy -D warnings` all green.

PR #225, opened 2026-06-15
feat(canon): expose RDFC-1.0 as an opt-in public API (sparq-canon) — sq-0qip

jeswr:
> 🤖 SPARQ agent

Closes the roadmap bead **sq-0qip** (ex-epic sq-3183): expose **RDFC-1.0** (the W3C [RDF Dataset Canonicalization](https://www.w3.org/TR/rdf-canon/), the URDNA2015 successor) as a reusable public API.

## What & why

The algorithm was already used internally by the ZK layer (`sparq-zk/src/canon.rs`), but only as a thin **bridge** over the maintained zkp-ld `rdf-canon` crate — not a reusable public surface, and the oxrdf-0.3 ↔ oxrdf-0.2 canonical-N-Quads bridge was **duplicated** across `canon.rs`, `trace.rs::issue_label_map`, and the W3C suite test.

Per the user's opt-in directive, this **extracts a dedicated `crates/sparq-canon` crate** (rather than feature-gating in core) so the canonical implementation is **single-sourced**: the new crate owns the one bridge and adds the dataset-level API the bead asked for. `sparq-zk`'s `canon` module is now a **thin re-export**, so the ZK commitment-pipeline call sites are unchanged.

**Opt-in by construction:** `sparq-canon` is `publish = false` and a non-default workspace member (like `sparq-zk`) — nothing in sparq's default dependency graph depends on it. `sparq-core`'s default build and the wasm artifact are **byte-identical** with or without it (verified: `cargo tree -p sparq-core` does not include canon/rdf-canon). The RDFC-1.0 *algorithm* remains the `rdf-canon` crate; this is the clean public surface over it.

## Public API (over sparq's oxrdf 0.3 term model)

- `canonicalize(&[Quad]) -> String` / `canonicalize_quads` — canonical N-Quads.
- `issued_identifiers(&[Quad])` / `issue_quads` — the canonical blank-node issuer map.
- `canonicalize_quads_with::<D: Digest>` / `issue_quads_with` — non-default hash profile (SHA-256 default; e.g. `sha2::Sha384`). `Digest` is re-exported.
- `canonicalize_triples` / `canonicalize_graph_content` / `issue_triples` — single-graph (the ZK leaf-order surface), returning `CanonicalGraph` (sorted lines + re-parsed canonical triples).

## W3C rdf-canon conformance

The official suite (86 vectors: **64 eval + 21 issued-map + 1 negative/poison**, under both **SHA-256 and SHA-384**) moves to `sparq-canon` and now runs **through the public API** (previously it called `rdf_canon` directly). Plus blank-node-relabelling / triple-order invariance unit tests. All green.

## ZK path unregressed

`sparq-zk` + `sparq-zk-compose` full test suites pass (commitment / trace / sig / Q6-guard / manifest). The only sparq-zk source changes are the `canon.rs` re-export and a one-line `issue_label_map` delegating to `sparq_canon::issue_triples`.

## Gates

- `cargo build` / `cargo test` — clean & green (canon + zk + zk-compose).
- `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean on authored files.
- **G1 new-crate gate:** README (template-compliant) + `publish = false` stub-exempt + new public surface SKILL (`skills/rdf-canon/`, linked from the router, AGENTS.md, root README).

NON-CANONICAL timing (no perf numbers; RDFC-1.0 canon throughput is already registered via `zk-commit-throughput`).


jeswr:
> 🤖 SPARQ agent

**Diagnosis of the red `zk forge + anchor suite (nargo/bb)` check** (flagging for orchestrator — not fixing here):

This is **NOT a Noir circuit / nargo / bb failure and NOT flaky** — it is a deterministic stale-lockfile error introduced by this PR. The forge job runs the full fail-safe suite because `crates/sparq-zk*/**` changed, and the build of the detached `bench/zk-compose/family_curve` project fails:

```
error: cannot update the lock file .../bench/zk-compose/family_curve/Cargo.lock
       because --locked was passed to prevent this
```

Root cause: `family_curve` depends on `crates/sparq-zk` by path. This PR rewrote `crates/sparq-zk/Cargo.toml` (dropped `rdf-canon`/`oxttl`/`oxrdf02`/`oxttl01` + the `serde_json`/`sha2` dev-deps, added `sparq-canon`), but `bench/zk-compose/family_curve/Cargo.lock` was **not regenerated**, so `cargo --locked` rejects it. Reproduced locally (`cargo metadata --locked` → exit 101); regenerating the lock (`cargo generate-lockfile` in that dir) adds `sparq-canon v0.1.0` and clears the error.

**Fix (for whoever owns this branch):** in `bench/zk-compose/family_curve/`, run `cargo generate-lockfile` (or `cargo update -p sparq-zk`) and commit the updated `Cargo.lock`. A re-run alone will not help — the lock is deterministically stale. No circuit code is involved.

`#225` touches **no** `zk/**` Noir circuit, `.nr` file, or forge/anchor harness — purely the Rust `sparq-zk` → `sparq-canon` extraction.

jeswr:
> 🤖 SPARQ agent

Regenerated `bench/zk-compose/family_curve/Cargo.lock` (commit `41e1e6d`).

The `sparq-canon` extraction moved `oxrdf`/`oxttl`/`rdf-canon` out of `sparq-zk` into the new `sparq-canon` crate, but this benchmark project's lockfile had not been regenerated. The `zk forge + anchor suite` builds with `--locked`, which rejected the stale lockfile:

```
error: cannot update the lock file ... because --locked was passed to prevent this
```

Ran `cargo generate-lockfile` in `bench/zk-compose/family_curve/`; the lockfile now contains the `sparq-canon` package and the updated `sparq-zk` deps. Verified `cargo metadata --locked` now exits `0` in that directory. Only `Cargo.lock` is touched.

PR #237, opened 2026-06-16
Certification readiness — OWASP ASVS

jeswr:
> 🤖 SPARQ agent

OWASP ASVS L1/L2 framework slice for **epic sq-toze** (assessment bead **sq-toze.10**). Maps the ASVS v4.0.3 requirements applicable to sparq as a **Rust RDF/SPARQL data-engine library + HTTP query API** (not a session-based web app) to concrete repo evidence.

## What's here (`compliance/asvs/`)
- `README.md` — framework intro, the **B3 no-auth-by-default** boundary as an explicit architectural decision (operator-gateway), chapter applicability table, explicit out-of-scope.
- `controls/asvs.md` — the per-requirement **control → status → evidence (file:line + test + CI job) → owner** table (the auditor spine), V1–V14.
- `evidence.md` — re-runnable verification anchors (grep / test name / CI job) per PASS claim.
- `gap-register.md` — open gaps + severity + remediation + bead.

## Status summary
- **PASS (implemented & verified):** V5 input-validation core (parsers + body/decompression caps), V7 error/log hygiene (structured JSON envelope, panic isolation, opt-in body-free logging), V11 DoS limits (timeout 503 / load-shed 429 / row-cap 413), V12 **SSRF** (default-deny SERVICE allowlist + resolved-IP filter incl. cloud-metadata), V13 method/content-type + the optional constant-time bearer-token gate, V14 secure defaults + gating cargo-deny/CodeQL + `.well-known/security.txt`, and the V2.10/V4/V6.2 residuals.
- **AUDIT-READY:** V9 TLS (proxy-terminated by design; non-loopback bind refused without opt-in), V10.3 SLSA level, V1.10 commit provenance. External ASVS L2 verification / pen-test is the residual we cannot self-issue.
- **N/A (operator gateway, B3):** the V2/V3/V4 authentication-session-access-control lifecycle — sparq has no accounts/sessions/cookies.
- **GAP (beads under sq-toze):** ASVS-G1 security headers/nosniff V14.4 → `sq-2bhm`; ASVS-G2 first-party CORS allowlist V14.5.3 → `sq-vtkj` (documented safe-default decision); ASVS-G3 verify engine error strings are path/host-free V7.1.1 → `sq-kfel`; ASVS-G4 explicit SPARQL parse-depth bound V5.5.2 → `sq-1ukn`.

## Honesty contract
[...]

PR #251, opened 2026-06-16
feat(sparq-server): security response headers on all responses (sq-cmvh, ASVS V14.4)

jeswr:
> 🤖 SPARQ agent

## Security response headers on all responses — `sq-cmvh` / ASVS-G1 (V14.4, cert remediation #237)

`sparq-server` emitted **no** security response headers (ASVS V14.4). This adds a
`map_response` layer in `harden()` that stamps a hardening header set onto **every**
response — success, streamed and error alike — chosen deliberately for a SPARQL **data
API** (it serves SPARQL-results JSON/XML/CSV/TSV and RDF, never HTML it asks a browser to
render), not copied from a browser-app template.

### Headers added (every response)
| Header | Value | Why |
| --- | --- | --- |
| `X-Content-Type-Options` | `nosniff` | no MIME-sniffing into a type we did not send |
| `Content-Security-Policy` | `default-src 'none'; frame-ancestors 'none'` | data API serves no subresources/scripts → tightest CSP makes any sniffed body inert + forbids framing |
| `X-Frame-Options` | `DENY` | legacy clickjacking guard for non-CSP agents; API is never framed |
| `Referrer-Policy` | `no-referrer` | never leak a request URL (a `query=` may carry sensitive terms) as a `Referer` |

Each header is only inserted **when absent**, so a handler can override a specific one.

### Deliberately omitted (with reason)
- **`Strict-Transport-Security` (HSTS)** — the origin serves **plain HTTP**; TLS (and so HSTS) is the fronting reverse proxy's job. Emitting it here is meaningless or wrongly pins a host. N/A.
- **`X-XSS-Protection`** — deprecated, a no-op/harmful in modern browsers, superseded by CSP.
- **CORS / `Cross-Origin-*` / `Permissions-Policy`** — browser-document policies with no meaning for a no-CORS data API. Adding CORS would *widen* the surface, the opposite of hardening.
- **Blanket `Cache-Control: no-store`** — query results are uncached by default (no `ETag` / `Cache-Control: public` is ever set), so there is nothing to tighten, and a blanket value would wrongly override `/health` and `/metrics`.

### Tests
`tests/hardening.rs::security_headers_*` (3 new): asserts the full header set on a **200**
SELECT + `/health`, on an **error** (400 missing-query **and** a 413 rewritten *inside* the
middleware stack), and on a **streamed** SELECT body — plus asserts HSTS / X-XSS-Protection
are **absent**.

### Verify (default features — the crate's gate)
- `cargo build -p sparq-server` ✅
- `cargo test -p sparq-server` ✅ (20/20 hardening incl. the 3 new; full suite green)
- `cargo clippy -p sparq-server --all-targets --all-features -- -D warnings` ✅
- new code within `max_width = 100`
- One pre-existing `--all-features` `geo` test (`unregistered_geof_iri_is_still_an_engine_error`) fails **on the base branch too** — unrelated to this change (an engine `geof:` behaviour, no headers involved).

### Docs
README "Security response headers" · `skills/http-server/SKILL.md` · `compliance/asvs/gap-register.md` (ASVS-G1 → **Resolved**).


jeswr:
> 🤖 SPARQ agent

Fixed the `markdownlint (docs)` HARD-gate failure.

**Violation:** `MD028/no-blanks-blockquote` at `skills/http-server/SKILL.md:57` — a bare blank line sat between the SERVICE-federation callout and the new security-headers callout. markdownlint reads that as a blank line *inside* a single blockquote.

**Fix (formatting only):** replaced the blank line with a blockquote-continuation marker (`>`) so the two notes render as one multi-paragraph blockquote. No prose changed.

Replicated the CI step locally (`npx --yes markdownlint-cli2@0.18.1`, the exact gate invocation, config-driven by `.markdownlint-cli2.jsonc`): **0 errors** over the scoped doc surface (56 files). `crates/sparq-server/README.md` had no violations. Pushed `aa6359f` to `feat-security-headers`.

PR #234, opened 2026-06-16
Certification readiness — CDMC scorecard

jeswr:
> 🤖 SPARQ agent

CDMC (Cloud Data Management Capabilities, EDM Council) framework slice for the certification estate (epic **sq-toze**). Scores sparq against the 6 components / 14 capabilities / 14 key controls as a **data engine**, with operator-owned governance decisions explicitly capped at sparq's enabling-hook maturity (never the governed outcome), and the **ZK/MPC estate excluded** from all protection-by-crypto scores per the binding `research/zk-soundness-audit.md` verdict (v1 verifier NOT sound).

## Headline — per-capability maturity (1–5)

| Component | Maturity | Verdict |
|---|---|---|
| 1 Governance & Accountability | ~2.7 | Engine governance strong; data-ownership operator-accountable |
| 2 Cataloguing & Classification | ~3.5 | **Strength** — VoID/Service-Description catalogue (#219) + SHACL classification |
| 3 Accessibility & Usage | ~2.5 | Access control real (WAC/ACP + write-token); access-audit + ODRL are gaps |
| 4 Protection & Privacy | ~2.7 | Security **excellent (4)**; privacy/crypto deliberately **low (2)** — ZK/MPC NOT sound |
| 5 Data Lifecycle | ~3.0 | Solid mechanism (SPARQL UPDATE/WAL/ring-retention); policy operator-owned |
| 6 Technical Architecture | ~3.0 | Architecture **strength (4)**; lineage the weak axis (2) |

Highest individual ratings (4): **2.1 cataloguing**, **4.1 security controls**, **6.1 technical architecture**. Lowest (2, honest): ownership, crypto-privacy, encryption-at-rest, lineage, access-audit.

## Deliverables (`compliance/cdmc/`)
- **`scorecard.md`** — the headline per-capability maturity + rationale + recommendations
- `controls.md` — control → status → maturity → evidence (the auditor spine)
- `evidence.md` — checkable file/test/CI artefacts behind each rating
- `gap-register.md` — 7 CDMC maturity gaps (CD-1..CD-7) + bead intents
- `recommendations.md` — prioritised, evidence-linked remediations
- `README.md` — slice index + scope + operator-responsibility split

## Honesty notes
- No capability scored above its evidence; operator-owned axes capped at sparq's hook.
- ZK/MPC contributes **zero** to 4.x protection-by-crypto (binding `zk-soundness-audit.md`).
[...]

PR #262, opened 2026-06-16
feat(sparq-server): opt-in per-query access audit log (sq-0bxp, CDMC CD-2)

jeswr:
> 🤖 SPARQ agent

Closes the CDMC **CD-2** gap (access audit) — `sq-0bxp`. Adds an **opt-in, lean per-query access audit log** to `sparq-server`. Composes with **ISO 27001 A.8.15** (logging) and **EU CRA** logging expectations. The CDMC audit (#234) rated access-audit a **2** because the server had only aggregate `/metrics` — no per-subject / per-query trail. This adds one.

## Opt-in + zero overhead when off
Doubly opt-in: compile with the new `audit-log` cargo feature **and** turn it on at runtime (`--audit-log` / `SPARQ_AUDIT_LOG=1` / `ServerConfig::audit_log`, field present only with the feature). Off (either) ⇒ the audit module and every call site are `#[cfg]`-stripped / short-circuited before any record is built — a request pays essentially zero (byte-identical to before). Verified by building + testing both feature states.

## The audit record (one structured `tracing::info!` per request)
Emitted under the dedicated `target: "sparq_server::audit"` (operators route it via `RUST_LOG`, independently of `--verbose`):

| field | meaning |
| --- | --- |
| `requester` | `anonymous`, or `token:<fnv1a>` fingerprint of the Bearer token — **never the raw token** |
| `op` | `query` / `update` / `graph_read` / `graph_write` (keyed on whether it mutates) |
| `fingerprint` | FNV-1a hash of the trimmed query/update text — **not the query text** |
| `decision` | `allowed` / `denied` (+ `reason` on a denial) |
| `status` | the HTTP status the client saw |
| `rows` | result-row count when known |
| `duration_us` | handler wall-clock |

## No-PII / no-info-leak posture (reuses the #241 lesson)
This is a **server-side** operator log — **never** written to the HTTP response (the #241 contract: detail to the log, generic class to the client). It deliberately does **not** log the full query text (raw SPARQL can disclose loaded-data fragments or caller PII) nor the Bearer secret — only stable, **non-reversible** fingerprints of each, enough to correlate repeated identical queries / a recurring caller across requests and restarts.

## Tests
- `src/audit.rs` unit tests: fingerprint stability + edge-trim, the raw token never appears in the identity, op-string stability.
- `tests/audit_log.rs` integration test: captures the `sparq_server::audit` target via a global subscriber against the real axum server and asserts the record fields for an **ALLOWED** (200, `op=query`, token fingerprint, the SELECT fingerprint, a duration) and a **DENIED** (401, `requester=anonymous`, a reason) request — plus that the fingerprint never leaks into the response body.

## Verify
- `cargo build -p sparq-server` clean with feature **off** and **on**.
- `cargo test -p sparq-server` green both states (19 binaries each).
- `cargo clippy --all-targets -- -D warnings` clean both states; my files `cargo fmt`-clean.
- Feature-off = zero overhead (module + call sites compiled out).

## Docs
- `skills/http-server/SKILL.md`: new "Access audit log" section + flag-table row + config struct note.
- `compliance/cdmc/gap-register.md`: **CD-2 marked addressed** (re-score 3.2 2→3).

NON-CANONICAL timing.


jeswr:
> 🤖 SPARQ agent

Fixed the `markdownlint (docs)` gate failure. `markdownlint-cli2@0.18.1` (the CI-pinned version, run with the repo's `.markdownlint-cli2.jsonc` config) reported **3 errors**, all in `skills/http-server/SKILL.md`:

| line | rule | cause |
| --- | --- | --- |
| 131 (×2) | **MD038** no-space-in-code | the multi-line inline-code span opened on L127 (`struct ServerConfig { … }`) was re-broken by the literal backticks added around `` `audit-log` `` inside the `/* … */` comment, leaving `` ` */ }` `` / `` `audit_log under ` `` as spans with leading/trailing spaces |
| 486 | **MD047** single-trailing-newline | file ended mid-word (`missing-v`, no newline) |

**Fixes (formatting/restore only — no meaning changed):**
- Dropped the inner backticks around `audit-log` in the comment (`+ audit_log under audit-log feature`) so L127–131 stay **one** code span.
- The MD047 hit had a root cause beyond a missing newline: this branch's diff **accidentally truncated the file** — the `@@ -448,120 +483,4 @@` hunk dropped 118 lines (the rest of the Auth bullet plus several other Gotchas bullets and the entire **See also** section), leaving the last sentence cut at `missing-v`. I restored that tail verbatim from `origin/main`, which also supplies the proper trailing newline. Heads-up: that content loss was unrelated to lint and would have shipped silently otherwise.

The intended audit-log additions (CLI table row, the new **Access audit log** section, and the `ServerConfig` comment) are intact.

`markdownlint-cli2@0.18.1` now exits **0** over the full HARD-gate doc surface (56 files). `compliance/cdmc/gap-register.md` also lints clean and is outside the HARD-gate globs, so it was not the cause. Pushed as `e58aea0` — the `markdownlint (docs)` check should go green.

PR #269, opened 2026-06-16
feat(wasm): expose EXPLAIN / EXPLAIN ANALYZE on the wasm Store (sq-ncvq.14)

jeswr:
> 🤖 SPARQ agent

Closes the EXPLAIN public-surface asymmetry the drift-scanner flagged (bead **sq-ncvq.14**, drift catch-up C): the Rust API (`sparq_engine::explain` / `explain_analyze`) and the HTTP server (`explain=` query param / `Accept: text/x-sparq-explain`) exposed query-plan introspection, but the **WASM/JS binding did not**. Now it does — the asymmetry is closed.

## What

Two thin `wasm-bindgen` methods on `Store`, delegating to the same engine functions Rust/HTTP use:

- `explain(query)` → plan text — planning-only dry run (no execution), accepts every query form (SELECT/ASK/CONSTRUCT/DESCRIBE)
- `explainAnalyze(query)` → plan + per-operator execution trace — runs the query (SELECT/ASK only; CONSTRUCT/DESCRIBE rejected)

**Return shape:** the *same* plain `String` the Rust API and the HTTP endpoint return, mapped to `JsError` on parse/eval failure — identical to the other `Store` methods. (Wall times read 0 on wasm32 — no monotonic clock; row counts are exact.)

## Verify

- `cargo build -p sparq-wasm --target wasm32-unknown-unknown` clean
- `wasm-pack build --target web` generates `explain(sparql): string` / `explainAnalyze(sparql): string` in `sparq_wasm.d.ts`
- Native `sparq-wasm` lib tests: 16 pass (2 new)
- JS test `js/test/explain.test.mjs` exercises both against the real wasm bundle — passes
- `cargo clippy -p sparq-wasm --all-targets -- -D warnings` clean (default + wasm32 targets)
- `cargo fmt --check` clean
- `skills/javascript-wasm/SKILL.md` updated (raw-Store method list, EXPLAIN recipe, wasm32 wall-time gotcha)

Timing claims here are NON-CANONICAL (work-box measurements).

### DIFF
--- a/bench/perf-baseline.json
+++ b/bench/perf-baseline.json
@@ -33,7 +33,14 @@
     "src/diskann.rs Vamana, src/quant.rs PQ k-means), floors DERIVED BY RUNNING examples/bench_vectors",
     "on the pinned N=50000 seed=0 corpus (NOT guessed). hnsw_recall_at10 is INTENTIONALLY ABSENT here:",
     "the instant-distance HNSW build is rayon-parallel (+/-1 deficit float-reduction jitter), so it is",
-    "floor-gated in bench/vector/run.sh only, never cross-commit ratcheted (a flaky ratchet is dishonest)."
+    "floor-gated in bench/vector/run.sh only, never cross-commit ratcheted (a flaky ratchet is dishonest).",
+    "[OPUS-4.8] (sq-ncvq.14) 2026-06-16: wasm_bundle_bytes floor RAISED 1579288 -> 1656470 (+4.89%). This is",
+    "an EXPLICIT, reviewed feature bump (the protocol bullet above: 'a wasm bundle bump from a new feature').",
+    "PR #269 adds Store::explain / Store::explainAnalyze to the wasm surface, linking in the engine's",
+    "EXPLAIN / EXPLAIN ANALYZE plan-formatting + analyze-trace code paths (sparq_engine::explain /",
+    "explain_analyze) that the prior bundle never reached. 1656470 is the exact CI-measured wasm-pack",
+    "release bundle size on this branch (run 27592086692, the 'run + track benchmarks' lane), not a guess.",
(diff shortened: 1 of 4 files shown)


jeswr:
> 🤖 SPARQ agent

**Root cause of the `run + track benchmarks` failure** (deterministic, survived re-run):

The `wasm_bundle_bytes` deterministic metric grew from the committed best-ever floor **1,579,288 → 1,656,470 bytes (+4.89%)**, tripping the perf-gate's **+2% best-ever ratchet floor** (`scripts/perf-gate.py` against `bench/perf-baseline.json`). The timing re-measure was correctly *skipped* by the gate — re-running can't clear a deterministic byte-count regression, which is why the failure was stable.

**This is expected feature growth, not a regression to suppress.** This PR adds `Store::explain` / `Store::explainAnalyze` to the wasm surface, which link in the engine's EXPLAIN / EXPLAIN ANALYZE plan-formatting + analyze-trace code paths (`sparq_engine::explain` / `explain_analyze`) that the prior bundle never reached. A ~4.9% bundle bump for two new linked-in query-introspection entry points is plausible and legitimate.

**Fix:** raised the `wasm_bundle_bytes` floor to the exact CI-measured value **1,656,470** in `bench/perf-baseline.json`, with a documenting `_comment` entry. This is precisely the protocol the baseline file documents for this case: *"RAISE a floor (deliberate regression, e.g. a wasm bundle bump from a new feature) ONLY explicitly: edit this file in a reviewed PR"* — the PR diff is the review. Future growth beyond +2% of the new floor must again be an explicit raise, so the ratchet's anti-slow-drift property is preserved.

I deliberately did **not** use `PERF_GATE_ALLOW` (a one-off env bypass that leaves no reviewable record) — the committed floor bump is the honest, auditable form.

Pushed as `710bd11`.

PR #383, opened 2026-06-16
feat(sparq-server): brTPF bind-restricted fragments + Hydra paging (sq-dxhb)

jeswr:
> 🤖 SPARQ agent

Implements **sq-dxhb** (epic sq-3183) — Federation TPF follow-ups: bind-restricted Triple Pattern Fragments (brTPF) + the fuller Hydra paging vocabulary on the existing `/tpf` endpoint.

## What landed

**brTPF — bind-restricted Triple Pattern Fragments** (Hartig & Buil-Aranda, ODBASE 2016) as an **opt-in** extension behind a new `brtpf` cargo feature (implies `tpf`, default OFF):

- A client attaches a **set of solution mappings** — the `values` query parameter (GET) or, for a large set, a **`POST` body** of one `position=term` mapping per line (`s`/`p`/`o` or long forms; same N-Triples term grammar as the pattern parameters).
- The server returns only the page of pattern matches **compatible with at least one supplied binding** — the deduplicated union of each mapping's specialised-pattern scan — pushing a bind-join's semi-join down to the source so far less data crosses the wire than re-fetching the whole pattern per binding.
- `hydra:totalItems` and the paging window reflect the **bindings-restricted** result (not the unrestricted pattern); an extra `hydra:mapping` for the `values` variable advertises the restriction so a client can discover it.
- A mapping that **contradicts** the pattern's own bound term, or whose term is **absent from the dictionary**, contributes nothing (skipped without index work). An empty/absent binding set is exactly plain TPF (the μ₀-only case).
- Pure — no new dependency; runs over the existing public `Graph::pattern` / `store.scan` / `dict.term` API. A malformed binding set is a **sanitized 400** (no echo of caller input — same info-leak posture as the TPF term parse).

**Fuller Hydra `PartialCollectionView` paging** on every TPF response: adds `hydra:first` (page 0) and `hydra:last` (the page holding the final match) alongside the existing `hydra:next` / `hydra:previous` — keyed on the same estimate as `totalItems`/`next` for consistency.

## Opt-in / zero-cost-when-off

Without the `brtpf` feature the `values` parsing + the `POST` route are `#[cfg]`-stripped, so a `tpf`-only build is **byte-identical** to before (a stray `values` parameter is just an ignored unknown parameter — plain TPF). Still governed by the same `--tpf` runtime flag and read-auth — `POST /tpf` is a READ (it returns a fragment, never writes).

- `cargo build -p sparq-server` for `tpf` and `brtpf` — green.
- `cargo clippy -p sparq-server --all-targets -- -D warnings` for `tpf` and `brtpf` — green.
- `cargo test -p sparq-server` for `tpf` and `brtpf` — green. New tests: the brTPF bindings-restricted request (GET `values` + POST body, union, dedup, contradiction/absent-term skip, consistent paging across pages, malformed→400-no-echo) and the Hydra `first`/`last` controls (unit + HTTP integration).

## Docs

Updated the sparq-server README (TPF source bullet + opt-in feature list) and the `http-server` SKILL's "Triple Pattern Fragments" section (new brTPF subsection + flags table).

## Honest notes

- The brTPF `values` control property is minted under a sparq namespace (`https://sparq.dev/ns/brtpf#values`) — brTPF has no standardised vocabulary term for "the attached solution-mapping set", so a client reads it from `hydra:mapping`.
- `hydra:totalItems` on a brTPF fragment is the **exact** union size (the restricted result is materialised to dedup it), unlike the plain-TPF cardinality estimate; advertised consistently with `next`/`first`/`last`.
- The `rustfmt` CI lane is informational (a deferred one-time `cargo fmt --all` reformat is pending); my changed source files (`tpf.rs`, `http.rs`) are individually fmt-clean. `tests/tpf.rs` carries pre-existing drift from main and was left untouched beyond the added tests.

Deferred (beaded as follow-ups — see below): a binary brTPF transport (the spec's compact mapping encoding rather than the line-oriented text form), and an explicit cap on the binding-set size as a DoS guard.

### DIFF
--- a/.github/workflows/feature-matrix.yml
+++ b/.github/workflows/feature-matrix.yml
@@ -161,6 +161,28 @@ jobs:
             crate: sparq-server
             features: audit-log,federation-descriptors,service,time-travel,geo,test-seams
             test: true
+          # [OPUS-4.8] sq-dxhb: the Triple Pattern Fragments / Linked Data Fragments READ-ONLY
+          # source endpoint (`GET /tpf`). The handler + the Hydra paging vocabulary + the
+          # `tpf::evaluate` pager + the in-module `tpf` unit tests are all `#[cfg(feature =
+          # "tpf")]`, so the DEFAULT build never compiled, clippy'd or tested this code. This
+          # leg builds it, runs `clippy --all-targets -D warnings` with the feature ON (the real
+          # gate), and runs the crate's tests — which now include the gated `tpf` page/Hydra
+          # suites. `tpf` pulls `server`, so the full async server compiles under it.
+          - name: sparq-server (tpf)
+            crate: sparq-server
+            features: tpf
+            test: true
+          # [OPUS-4.8] sq-dxhb: bind-restricted Triple Pattern Fragments (brTPF). `brtpf` implies
+          # `tpf` and additionally compiles `tpf::evaluate_brtpf`, the `values`/POST bindings
+          # parser, the `merge_term`/`merge_predicate` semi-join pushdown and the extra
+          # `hydra:mapping` control — all `#[cfg(feature = "brtpf")]`. The in-module `brtpf` test
+          # mod (binding restriction, dedup union, empty-bindings==plain-TPF, page partitioning)
+          # is gated the same way, so it only runs when this feature is on. This leg is what
+          # actually compiles + clippies + tests the brTPF + Hydra-paging code the PR adds.
+          - name: sparq-server (brtpf)
+            crate: sparq-server
+            features: brtpf
+            test: true
           # sparq-vectors: the vec: magic predicates (pulls sparq-engine), and the
           # /v1/embeddings provider request/response shape (serde only, no socket).
           - name: sparq-vectors (vec-predicate, provider)
--- a/crates/sparq-server/Cargo.toml
+++ b/crates/sparq-server/Cargo.toml
@@ -132,6 +132,21 @@ federation-descriptors = ["server", "dep:sparq-introspect"]
 # set `--tpf` / `SPARQ_TPF=1` (`ServerConfig::tpf`), mirroring `federation-descriptors`. Without the
 # feature, zero cost: no route, byte-identical to before.
 tpf = ["server"]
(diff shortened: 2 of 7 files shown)


jeswr:
> 🤖 **SPARQ agent**

Addressed the adversarial-review gap (commit `a8d87c2`):

**1. CI coverage (the real gap).** The `tpf` + new `brtpf` features are `#[cfg]`-gated and were **absent from `.github/workflows/feature-matrix.yml`**, so no required check actually compiled / clippied / tested the brTPF + Hydra-paging + TPF code — "gates green" was local-only. Added two `sparq-server` legs (same pattern sq-wuft used for `filtered-ann` / `shacl-af`):

| leg name | features | runs |
|---|---|---|
| `opt-in sparq-server (tpf)` | `tpf` | build + `cargo test -p sparq-server --features tpf` + `cargo clippy -p sparq-server --features tpf --all-targets -- -D warnings` |
| `opt-in sparq-server (brtpf)` | `brtpf` | build + `cargo test -p sparq-server --features brtpf` + `cargo clippy -p sparq-server --features brtpf --all-targets -- -D warnings` |

Closes both the review-found brTPF gap **and** the pre-existing `tpf`-uncovered gap. Verified **locally** with the exact leg commands — all green: build OK, clippy clean under `-D warnings`, and the gated suites now actually execute (16 `tpf` unit tests; 11 `brtpf` submodule tests). Neither leg name contains `advisory`/`informational`, so the required `ci-summary / gate` aggregator auto-discovers both as **REQUIRED**. No real bug surfaced — the code was correct, just CI-uncovered.

**2. Doc over-claim.** `evaluate_brtpf`'s doc said it pages "in the same deterministic order plain TPF pages in" — inaccurate. brTPF pages in numeric `[s,p,o]` id-lex order (the `BTreeSet`); plain TPF pages in the chosen **index's** scan order. Reworded to state the real guarantee: **deterministic + intra-result consistent** (pages partition the result, no overlap), but **not necessarily identical to plain-TPF order**. The brTPF tests only compare counts/membership, so no test change was needed. The README / PR body do not repeat the claim — no edit there.

Auto-merge is **not** armed by this agent — leaving that to the orchestrator to re-verify the new legs go green in CI and arm.

PR #394, opened 2026-06-16
feat(sparq-solid): ACP acp:issuer matcher — (agent,client,issuer) principal (sq-3jtd.6)

jeswr:
> 🤖 SPARQ agent

Implements **sq-3jtd.6**: `sparq-solid` support for `acp:issuer`. Extends the ACP principal/matcher model so a Matcher can match on the **issuer** (`acp:issuer`) — the OIDC Identity Provider that vouched for the requester's WebID — not just the agent. The issuer is the exact twin of the existing client dimension (design doc §3.6).

## What changed

**N3 rules (`rules/acp-a/b/c.n3`)** — the issuer is a third principal dimension, parallel to client:
- issuer accept-sets (`solidx:acceptsIssuerP`, lattice top `auth:AnyIssuer`),
- an `issuer` candidate dimension and the candidate cross-product becomes (agent × client × issuer),
- `allOf`-rejection + `anyOf`-satisfaction over all three dimensions,
- a minted three-component principal `urn:sparq:triple?agent=A&client=C&issuer=I` for an issuer-constrained grant. An **unconstrained** issuer keeps the agent / `urn:sparq:pair?…` term **byte-identical**, so every pre-issuer grant is unchanged (the existing ACP correctness matrix still passes verbatim).

**Session + AuthIndex (`src/authindex.rs`)**
- `Session` gains `issuer: Option<&str>` (`None` = any issuer).
- a session expands to **≤12** principals (the pre-issuer ≤6 doubled by the issuer dimension),
- reads the `auth:issuer` conditional-grant head + `acceptsIssuerP` matcher facts,
- fails closed on a session issuer inside the reserved `urn:sparq:` space (it could otherwise impersonate a minted triple), and `triple_principal()` is exported as the public minting twin of `pair_principal()`.

**Loader + materialize** — `acp:issuer` values go through the same reserved-encoding validation as agents/clients/origins; `acceptsIssuerP` is copied into the auth view for `noneOf` exception matchers.

## Two bugs caught (by the deny tests)
- **Cache-key collision (core path):** the `PodStore` session cache key did not include the issuer, so two sessions differing only by issuer (same WebID via a trusted vs an untrusted IdP) collided on a cached graph set — a soundness hole. Fixed: the key now spans all three dimensions.
- **`odrl-bridge` feature:** bridged conditional grants now emit `auth:issuer auth:AnyIssuer` and their exception matchers assert `acceptsIssuerP auth:AnyIssuer`, so the now-three-dimensional `matcher_accepts` still fires.

## Tests (`tests/acp.rs`) — allow AND deny decided by issuer
- `allOf { agent ; issuer }`: allow @ trusted issuer, **deny** @ wrong issuer / no issuer / wrong agent,
- issuer-only `anyOf`: any agent from the trusted issuer; **deny** otherwise,
- the full `(agent, client, issuer)` triple,
- a `noneOf` issuer carve-out (conditional grant evaluated on the issuer dimension),
- fail-closed rejection of a reserved-space issuer at materialization.

## Opt-in / lean core
`sparq-solid` is already a separate opt-in crate (a dependency of nothing in the workspace), so this carries zero cost into `sparq-core`/`sparq-engine`. The `odrl-bridge` feature is unaffected by default.

## Gate — PASS in BOTH feature states
- `cargo build -p sparq-solid --all-targets` (default and `--all-features`)
- `cargo clippy -p sparq-solid --all-targets -- -D warnings` (default and `--all-features`)
- tests: default **68 pass / 0 fail**, `--all-features` **119 pass / 0 fail**

(rustfmt is informational/deferred workspace-wide per `rustfmt.toml`; clippy is the hard gate and is clean.)

### DIFF
--- a/crates/sparq-solid/README.md
+++ b/crates/sparq-solid/README.md
@@ -31,7 +31,7 @@ store.materialize_wac()?; // run the N3 rules → install <urn:sparq:auth>
 
 // The SAME query, different sessions, different results — fail-closed.
 let q = "SELECT ?title WHERE { ?s <https://ex.dev/ns#title> ?title }";
-let alice = Session { agent: Some("https://alice.ex/card#me"), client: None };
+let alice = Session { agent: Some("https://alice.ex/card#me"), client: None, issuer: None };
 let _authorized = store.query_as(&alice, Mode::Read, q)?.rows.len();
 let _public_only = store.query_as(&Session::default(), Mode::Read, q)?.rows.len();
 # Ok(()) }
@@ -40,8 +40,11 @@ let _public_only = store.query_as(&Session::default(), Mode::Read, q)?.rows.len(
 ## ✨ Features
 
 - **WAC + ACP** — Web Access Control (`.acl`) and Access Control Policy (`.acr`), including
-  inheritance, agent classes, groups, the `allOf`/`anyOf`/`noneOf` combinators, and
-  normative deny-overrides. The full support matrix is in the design doc (linked below).
+  inheritance, agent classes, groups, the `allOf`/`anyOf`/`noneOf` combinators, the ACP
+  matcher's `acp:agent` / `acp:client` / `acp:issuer` attributes (the three-dimensional
+  `(agent, client, issuer)` principal — a Matcher can gate on the OIDC issuer that vouched
+  for the requester, not just the WebID), and normative deny-overrides. The full support
+  matrix is in the design doc (linked below).
 - **Triples-native** — pods, ACL/ACR documents, and the materialized authorization view are
   all ordinary named graphs; "who can read G?" is one SPARQL pattern.
 - **Zero-copy enforcement** — the default query path evaluates through the engine's zero-copy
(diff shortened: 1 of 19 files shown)


jeswr:
> 🤖 SPARQ agent

Unblocked the **G2 public-api→SKILL.md** gate. PR #394 adds public API to `sparq-solid` (`pub use triple_principal`, `Session.issuer` field, `pub const ANY_ISSUER`) but `skills/access-control/SKILL.md` still documented the two-dimension `(WebID, client)` model. Synced it to the shipped **three-dimension `(agent, client, issuer)`** model, verified against `crates/sparq-solid/src/authindex.rs` on this branch (no crate code touched):

- `Session { agent, client, issuer }` signature + Quickstart `issuer: None` (matches the shipped README/example).
- Principal lattice → three dims (`AnyIssuer ⊒ concrete-issuer`); ≤12 lookups; only an issuer-CONSTRAINED grant mints a `triple_principal`.
- Documented `triple_principal` / `pair_principal` / `ANY_ISSUER` on the public surface.
- **Honest:** the issuer dimension is a string-match on a *caller-asserted* issuer — it does **NOT** add WebID-OIDC authentication; the relying app still verifies the issuer↔WebID binding.

Local gates green: G2 (hermetic) PASS, `check-privacy-claims.sh` OK, markdownlint 0 errors, typos clean, skill-frontmatter 24/24. `[OPUS-4.8]`

PR #464, opened 2026-06-17
ci(flow-on): proactive merge-gate G6 new-config/flag→docs (sq-ncvq.9)

jeswr:
> 🤖 **SPARQ agent** — automated change by an agent operating on @jeswr's behalf. Flag anything off and I'll iterate.

## What & why

Implements bead **sq-ncvq.9** — proactive maintenance merge-gate **G6 (new-config/flag→docs)** from `research/maintenance-flow-on-automation-design.md` §2.1, completing the gate family alongside G1/G2/G3.

**Rule:** when a PR adds a new `sparq-cli` / `sparq-server` CLI flag (a `"--flag"` literal) or a `SPARQ_*` env var under `crates/sparq-{cli,server}/src/**`, it must be **documented in the same change** — in a crate `README.md` or any `skills/**/SKILL.md` — otherwise the gate fails. Escape hatch: the per-PR label `config-internal` (justified in the PR body).

**Why a separate gate from G2.** G2 (`gate-api-skill.py`) only fires on net `pub`-item signature changes. The CLI/server arg parsers are hand-rolled `match` arms, so a new flag is a *string literal*, not a `pub` symbol — it sails straight past G2. G6 is the config-knob subset that closes exactly that gap (design §2.1: "subset of G2, but covers config not just `pub`").

## How

- **`scripts/check-config-documented.py`** (new, stdlib-only). Net-diff discipline mirrors G2: reads each changed config-src file's unified diff and keeps only knob tokens that are **NET-added** (added and not also removed). A knob is undocumented only if it appears in **no** docs surface — neither added-to-docs in this diff nor already on disk. Consequences:
  - a pure relocation of a flag handler within a file cancels out (net-zero) → no trip;
  - a comment that merely *mentions* an existing (already-documented) flag is inert;
  - a flag added in any non-`sparq-cli/server` crate, a test, or a CI file is out of scope → always passes.
  - Hermetic `evaluate()` takes `code_knobs` / `doc_added` / `doc_disk` overrides so tests inject every git/disk fact (no live git, no network).
- **`scripts/tests/test_gates.py`** — 12 hermetic G6 tests + a `main()` dry-run smoke (fail/pass paths, env var, README-doc, SKILL-doc, on-disk-documented rewire, `config-internal` label, out-of-scope crate, CI-only diff, token extraction, net-diff, comment-inert).
- **`.github/workflows/flow-on-gates.yml`** — adds the `config-documented (G6)` job. Runs **HARD**: the back-catalogue is already clean — every existing flag + `SPARQ_*` env var is documented in a crate README or a SKILL.md (verified against HEAD). Reads PR labels for the escape hatch, mirroring G2 (including the merge-queue PR-number resolution). Also runs the gate unit suite as a self-test step.

## Verification

- `python3 scripts/tests/test_gates.py` → 31 tests OK (12 new for G6).
- `python3 scripts/check-config-documented.py --base main` → PASS exit 0 (clean back-catalogue) — gate runs HARD.
- **End-to-end against a real git diff:** committing an undocumented `--g6-probe-undocumented` flag → FAIL (exit 1) with a per-knob message; adding the `config-internal` label → PASS (exit 0).
- `bash scripts/check-privacy-claims.sh` → OK (privacy-claims gate, mandatory).
- Workflow YAML parses; Python compiles.

No public Rust API changed (this is a CI/docs gate), so the G2 public-API→SKILL.md rule does not apply here.

### DIFF
--- a/.github/workflows/flow-on-gates.yml
+++ b/.github/workflows/flow-on-gates.yml
@@ -1,6 +1,6 @@
-# [OPUS-4.8] Proactive maintenance merge-gates G1/G2/G3 (beads sq-ncvq.4 +
-# sq-ncvq.5 + sq-ncvq.6, epic sq-ncvq). Authored by Opus 4.8 (Fable unavailable;
-# flag for re-review when Fable returns).
+# [OPUS-4.8] Proactive maintenance merge-gates G1/G2/G3/G6 (beads sq-ncvq.4 +
+# sq-ncvq.5 + sq-ncvq.6 + sq-ncvq.9, epic sq-ncvq). Authored by Opus 4.8 (Fable
+# unavailable; flag for re-review when Fable returns).
 #
 # This is the PROACTIVE / merge-time half of the maintenance flow-on system
 # (research/maintenance-flow-on-automation-design.md §2.1). It is the COMPLEMENT of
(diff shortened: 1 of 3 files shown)


jeswr:
> 🤖 **SPARQ agent**

Follow-up to the **honest=FALSE** flag on this gate. The G6 extractor merged here is **asymmetric**: `_FLAG_RE` requires double quotes and was applied to BOTH code and docs, so the doc side recognized **0 of 29** CLI flags (docs use markdown backticks, e.g. `` `--addr` ``) — the gate false-positives on its own compliant path. Env vars were unaffected (quote-agnostic).

Fix is in **#468** (split into `code_knob_tokens` / `doc_knob_tokens`; 0/29 → 29/29 recognized on the real tree; gate still PASSes clean back-catalogue and FAILs a genuinely-undocumented flag; 35 self-tests OK). This PR was already merged before the fix could be pushed to its branch, hence the follow-up.

PR #507, opened 2026-06-17
feat(sparq-server): connection header-read timeout closes the slow-loris DoS (sq-2gqr) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — automated change by an agent acting for @jeswr. Re-review the [OPUS-4.8] code when Fable is available.

## What & why (sq-2gqr)

`axum::serve` builds hyper's connection `Builder` internally and **never installs a timer**, so hyper's HTTP/1 `header_read_timeout` is *inert* (it requires a `Timer` to take effect — verified against the resolved hyper-util 0.1.20 / hyper 1.10.1 source). The stock stack therefore has **no header-read deadline at all**: a client that opens a connection then dribbles request-header bytes (or simply never finishes the header block) holds the connection — and, behind `harden()`'s `concurrency_limit`, a **concurrency slot** — open indefinitely. `max_concurrent` (default 32) such clients starve every legitimate caller.

None of the existing guards cover this: `query_timeout` is an *engine* deadline that only starts once a full request is parsed; `max_body_bytes` and load-shed likewise act *after* the headers are read. This is the transport-level slow-loris vector the bead calls out as distinct from sq-ebii / sq-s5is.

## The fix

A new `sparq_server::serve` accept loop — a faithful port of axum's own accept + graceful-drain loop (per-connection task, watch-channel drain, `serve_connection(...).with_upgrades()` so the `/subscriptions` WebSocket handshake still works) with **one** behavioural addition: it installs a `hyper_util` `TokioTimer` and sets hyper's `http1` `header_read_timeout`. `main.rs` now calls it instead of `axum::serve`.

- Knob: `ServerConfig::header_read_timeout: Option<Duration>` — **default 15s, ON**.
- `SPARQ_HEADER_READ_TIMEOUT` env (`0` disables) + `--header-read-timeout SECS` flag.
- `None` / `0` opts back out to the prior unbounded behaviour.

axum is configured HTTP/1-only (no `http2` feature), so this uses hyper's `http1::Builder` directly. `hyper` / `hyper-util` are added as **direct** deps (server-feature only) — both already in axum 0.8's dependency graph, so **no new transitive code**.

## Tests

`tests/hardening.rs` (raw TCP, driving the **real** `serve` loop — reqwest always sends a complete header block so it cannot exercise this):
- `slow_loris_partial_headers_closed_by_deadline` — a partial-header socket is closed within the deadline (the fix);
- `slow_loris_held_open_when_deadline_disabled` — with `None` the connection is *not* proactively closed (proves the guard is what closes it, and that it is genuinely opt-out-able);
- `serve_loop_handles_complete_request` — a complete request still returns `200` + the hardened security headers (no regression).

Plus unit tests for the default-ON value and its independence from `query_timeout`.

**Verified live against the binary:** normal query `200`, WebSocket upgrade `HTTP/1.1 101 Switching Protocols`, slow-loris partial-header connection closed at the configured deadline, clean SIGTERM graceful drain.

## Gate

- `cargo build` + `cargo clippy --all-targets -- -D warnings` clean — default features AND with all opt-in server features (`audit-log,access-audit,time-travel,service,geo,federation-descriptors,tpf,brtpf`); lib clean in the `--no-default-features` (server-off) state too. (The `--no-default-features --all-targets` test-compile failure is pre-existing on `origin/main` — the integration tests are inherently `server`-feature-only.)
- Full `sparq-server` test suite green (0 failures); workspace builds.
- `rustfmt --check` is *informational* in this repo (pending the deferred one-time `cargo fmt --all`); new code matches surrounding style — not running `cargo fmt` to avoid pulling in the repo-wide reformat.
- Privacy-claims gate: PASS. G2 public-api→SKILL: PASS (README + `skills/http-server/SKILL.md` updated for the new public `serve`, the `ServerConfig` field, and the flag/env).

## Follow-up

Filed **sq-lodb** for the distinct slow-**body**-read vector (a client that completes its headers then dribbles the request body stays under `max_body_bytes` while holding the slot) — not covered by a header-read deadline; complementary, not a replacement.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3551,6 +3551,8 @@ dependencies = [
  "axum",
  "flate2",
  "futures-util",
+ "hyper",
+ "hyper-util",
  "mimalloc",
  "oxrdf 0.3.3",
  "oxrdfxml",
--- a/crates/sparq-server/Cargo.toml
+++ b/crates/sparq-server/Cargo.toml
@@ -75,7 +75,7 @@ access-audit = ["server"]
 # equivalent. Feature unification means the `sparq-core` `sparq-engine` resolves also gains `mmap`,
 # so the WAL-durable `apply_delta` path is live on the shared graph. Without `--persist` the server
 # is in-memory exactly as before (the flag, not the feature, selects durability).
-server = ["dep:sparq-serve", "dep:axum", "dep:tokio", "dep:tower", "dep:tower-http", "dep:tracing", "dep:tracing-subscriber", "dep:futures-util", "dep:serde_json", "dep:flate2", "sparq-core/mmap"]
+server = ["dep:sparq-serve", "dep:axum", "dep:hyper", "dep:hyper-util", "dep:tokio", "dep:tower", "dep:tower-http", "dep:tracing", "dep:tracing-subscriber", "dep:futures-util", "dep:serde_json", "dep:flate2", "sparq-core/mmap"]
 # [OPUS-4.8] (sq-i3xc, lever H5) `zlib-ng`: opt-in, OFF by default, NATIVE-ONLY faster
 # gzip backend (zlib-ng C library) for the server's `Content-Encoding: gzip` request
 # inflate (MultiGzDecoder) — a measured 1.5–3× at ~zero code change. The weak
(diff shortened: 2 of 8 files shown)


jeswr:
> 🤖 **SPARQ agent** — automated change by an agent acting for @jeswr. Re-review the [OPUS-4.8] code when Fable is available.

## Adversarial-verify BLOCKER fixed — shutdown-path panic regression (honest=false → fixed)

The per-connection graceful-shutdown loop in `serve` (`http.rs`) polled a **bare** `watch::Receiver::closed()` future inside `tokio::select!`. After the shutdown signal fires once, the loop keeps running to DRAIN the connection, so the next `select!` iteration re-polled the **already-completed** future → tokio `watch.rs:1313` panic `` `async fn` resumed after completion ``. Under the release `panic = "abort"` profile this **aborts the whole process** on SIGTERM with a live keep-alive connection.

**Root cause:** the original port of axum 0.8.9's `serve` loop dropped axum's own `.fuse()` on that future — which is precisely what guards the re-poll. axum: `let mut signal_closed = pin!(signal_tx.closed().fuse());`.

**Fix (one line + matching imports/trace):** fuse the future (`futures_util::FutureExt::fuse`), matching axum verbatim. A `FusedFuture` reports `is_terminated()`, so `select!` skips that branch once it resolves → `graceful_shutdown()` fires exactly once, then the connection drains cleanly to completion.

### SIGTERM panic-repro (real release binary, `panic=abort`, SIGTERM + live keep-alive connection with an in-flight request)

| | result |
|---|---|
| **before fix** (negative control, rebuilt without `.fuse()`) | **9/12 runs aborted** with `` `async fn` resumed after completion `` at `tokio watch.rs:1313:32` |
| **after fix** | **0/15 runs** panicked/aborted; **15/15** in-flight requests drained to a clean `200` (graceful, not aborted) |

The negative control confirms the bug is real and the test is load-bearing (the prior "clean SIGTERM graceful drain" claim was FALSE; the original report saw 2/5 — my harness sends a post-SIGTERM request to force extra drain iterations, widening the window to 9/12).

### Minor (discoverability)
`--header-read-timeout SECS` is now in the runtime `--help`/usage string and the startup `guards:` log line (`header-read-timeout=15s`), alongside `--max-concurrent` / `--max-body-bytes` / `--max-results` / `--query-timeout`.

### Gate
- `cargo build` + `cargo clippy --all-targets -- -D warnings` clean — default features AND opt-in (`audit-log,access-audit,time-travel,service,geo,federation-descriptors,tpf,brtpf`).
- Full `sparq-server` test suite green in BOTH feature states (incl. `tests/hardening.rs` slow-loris + complete-request tests): 0 failures.
- `scripts/check-privacy-claims.sh`: PASS.

Staged explicit paths only (`http.rs` + `main.rs`).

**Honest verdict:** the graceful-drain claim is now TRUE (15/15 clean SIGTERM + graceful in-flight drain, vs 9/12 abort pre-fix). #507 is merge-ready.

PR #575, opened 2026-06-17
perf(sparq-mpc): tighten oblivious shuffle to AS-Waksman optimal switch count (sq-hny9) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — bead sq-hny9.

## What

Tighten the oblivious-shuffle permutation network (`crates/sparq-mpc/src/oblivious.rs`) from the plain Beneš/Waksman recursive form to the **AS-Waksman** (Beauquier–Darrot, "On arbitrary-size Waksman networks") redundancy-removal form.

- **Before:** `⌊m/2⌋` input + `⌊m/2⌋` output switches per recursion level.
- **After:** `⌊m/2⌋` input + `⌊(m−1)/2⌋` output switches per level — one fewer output switch on **even** `m`.

The saving comes from **fixing one output pair** on even levels: the last pair `(m−2, m−1)` is routed straight (output `m−2` from the TOP sub-network, `m−1` from the BOTTOM), so it carries no switch. This realizes the optimal arbitrary-size count `W(n) = W(⌈n/2⌉) + W(⌊n/2⌋) + (n−1)` (e.g. `n=8 → 17`, `n=16 → 49`, vs. the plain `20`/`56`).

## The odd-`m` corner case the bead flagged

The bead deferred this because naive fixed-pair pinning over-constrains routing for **odd** `m` (output 1 pinned BOTTOM clashes with the odd-trailing input pinned TOP). This PR avoids it by:

1. **Not** pinning any extra output pair on odd `m` — the odd-trailing output already carries no switch, so `⌊(m−1)/2⌋ = ⌊m/2⌋` and all output switches are kept.
2. On even `m`, seeding the fixed pair `out_from_top[m−2]=TOP, out_from_top[m−1]=BOTTOM` **into** the alternating-path solver before propagation, so the fixed pair is a *consequence* of a consistent routing rather than an axiom that over-constrains it.

## Honesty / scope

Perf-only. Obliviousness is **preserved and now explicitly tested**, not merely asserted:

- `waksman_has_full_support_after_optimization` — every one of the `n!` permutations is reachable (the anti-Benes-bias property) for `n = 1..=7`.
- `waksman_routes_every_permutation_correctly` — exhaustive correct routing extended to `n = 8`.
- `waksman_switch_count_is_as_waksman_optimal` — asserts the exact canonical AS-Waksman count for `n = 0..=64` (with literature spot-checks 4→5, 8→17, 16→49).

`ShuffleCost::depth_rounds` is an analytic model derived from `n`, unaffected. **No public API change** (G2 gate confirms).

## Gate

- `cargo build -p sparq-mpc` ✓
- `cargo clippy -p sparq-mpc --all-targets -- -D warnings` ✓ (default) and `--features insecure-test-rng` ✓
- `cargo test -p sparq-mpc` ✓ (332 passed) in **both** feature states
- `cargo fmt -p sparq-mpc --check` ✓
- privacy-claims gate (incl. predicate-form soundness self-test: 32 passed) ✓
- G2 public-api→SKILL.md gate: PASS (no public surface changed)

### DIFF
--- a/crates/sparq-mpc/src/oblivious.rs
+++ b/crates/sparq-mpc/src/oblivious.rs
@@ -176,13 +176,20 @@ pub struct Switch {
 /// cannot reach all permutations (or reaches some far more often) leaks order
 /// information. The switch count is `O(n log n)`.
 ///
-/// We build the **plain** recursive form (`⌊m/2⌋` input + `⌊m/2⌋` output switches
-/// per level) rather than the AS-Waksman redundancy-removal optimization (which
-/// shaves one switch per level by fixing the first output pair). The plain form is
-/// marginally larger but routes every permutation without the odd-`m` corner-case
-/// over-constraint the fixed pair introduces — see `solve_waksman_level`.
-/// Tightening to the AS-Waksman `−1`/level count is a perf-only follow-up (bead
-/// sq-hny9).
+/// We build the **AS-Waksman** (Beauquier–Darrot, "On arbitrary-size Waksman
+/// networks") redundancy-removal form: `⌊m/2⌋` input switches and `⌊(m−1)/2⌋`
+/// output switches per level (one fewer output switch than the plain Beneš form on
+/// **even** `m`), giving the optimal arbitrary-size count `W(n) = W(⌈n/2⌉) +
+/// W(⌊n/2⌋) + (n−1)`. The saving comes from **fixing one output pair** on even
+/// levels: the last output pair `(m−2, m−1)` is routed straight (output `m−2` from
+/// the TOP sub-network, output `m−1` from the BOTTOM), so it needs no switch — the
+/// alternating-path solver seeds that pin and derives the rest, so the fixed pair
+/// is a *consequence* of the routing rather than an over-constraint. On **odd** `m`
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent

Fixed the `cargo fmt -p sparq-mpc --check` regression flagged in review.

**Cause:** rustfmt was reindenting the standalone AS-Waksman output-switch comment block at `crates/sparq-mpc/src/oblivious.rs:374` to align with the *trailing* comment column of the preceding `let bot_len` line — miscounting that column because the trailing comments carry multi-byte unicode math glyphs (⌈ ⌋ −). The result was a ~20-space over-indent.

**Fix:** one blank line separating the trailing-comment lines from the standalone block, so rustfmt leaves it at the natural 4-space indent. No comment wording and no code logic changed (the readable unicode notation is preserved).

**Verified green (commit 78e8f3e8):**
- `cargo fmt -p sparq-mpc --check` → exit 0
- `cargo clippy -p sparq-mpc --all-targets -- -D warnings` → clean
- `cargo test -p sparq-mpc` → 332 passed, 0 failed

PR #602, opened 2026-06-18
perf(sparq-core): stream numerics/temporals caches on in-RAM save finalize (sq-7ph8) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — automated change. Implements bead **sq-7ph8**.

## What

The in-RAM `save()`/`save_compressed()` finalize built whole-dictionary **dense intermediates** purely to write the `numerics.bin`/`temporals.bin` caches:

- `dense_temporals()` **always** allocated a fresh `(Vec<u8>, Vec<f64>)` (9 B/term) — even when the in-RAM temporal cache is `Sparse` (the usual non-temporal case: a tiny map), the very representation `build()` chooses via `into_sparse_if_worthwhile`.
- `dense_numerics()` rebuilt a dense `Vec<f64>` (8 B/term) for a `Sparse`/`Forked` cache (e.g. after `into_compressed`).

That is an O(distinct-terms) transient RSS spike on top of the resident dict — the finalize peak `research/wikidata-lowresource-stage1.md` flags. The dict-spill path already **stream-writes** these (`dictspill.rs`); the in-memory save path did not.

## How

Replace the dense-materialise + bulk-write with `stream_write_numerics` / `stream_write_temporals`: probe the cache per id (`lookup`, O(1)) and flush a small reusable block buffer (64K ids), keeping peak resident memory at the **block size** rather than the dictionary — the same bounded-write discipline the dict-spill path uses.

The output is **byte-identical** to the old dense write: same on-disk layout (`n` LE f64 numerics; `n` LE f64 temporal instants then `n` flag bytes), `NaN`/flag-0 for non-cached ids. So this is purely a memory optimisation, never a format change — every existing `open` round-trip is unaffected. The now-unused `dense_numerics`/`dense_temporals` helpers are dropped.

No public API change (the `save`/`save_compressed` signatures are unchanged; only private helpers moved), so no SKILL.md/G2 update is required.

## Tests

New `streamed_caches_byte_identical_to_dense` compares the streamed `numerics.bin`/`temporals.bin` against a **reference dense computation** done directly off the dictionary, across `{dense, sparse} × {save, save_compressed}`, asserts the exact on-disk sizes (`n*8`, `n*9`), and re-opens to confirm the mmap caches still resolve (incl. the inline-integer edge: `xsd:integer` 42 is id-encoded, never in `numerics.bin`).

## Unsafe surface — registered + audited [OPUS-4.8]

> **Correction (SPARQ agent):** an earlier revision of this body claimed "Gate clean" but the **unsafe-count ratchet was not run**. `stream_write_numerics` / `stream_write_temporals` each add one `unsafe { std::slice::from_raw_parts(buf.as_ptr().cast::<u8>(), size_of_val(&buf[..])) }` (reinterpret a reusable `Vec<f64>` block as bytes for `write_all`), and the commit also **removes** one `unsafe` (the old `dense_temporals` mmap reinterpret) — net **+1** in `sparq-core` (44 → 45), so `scripts/unsafe-gate.py --check` was **RED**. Now fixed.

**Soundness audit (both new `from_raw_parts` reinterprets):** SOUND, unsafe kept.
- (a) `size_of_val(&buf[..]) = len*8` views exactly the live, fully-initialised `Vec<f64>` region — no over-read.
- (b) target `u8` has alignment 1; the f64 source is over-aligned → no misalignment.
- (c) the bytes are only **read** (fed to `write_all`), never written through the alias.
- (d) the `&[u8]` is consumed inside the flush closure before `buf.clear()` → no provenance/lifetime escape.
- (e) **native-endian** reinterpret, identical to the `write_numerics`/`write_temporals` it replaces and symmetric with the native-endian **read** path (`NumData::as_slice`, the temporal mmap reads): write-native + read-native round-trips on the same arch (the established cache contract; the cache is rebuilt locally, never shipped cross-arch). `streamed_caches_byte_identical_to_dense` asserts byte-identity to the dense write. (The doc/register "little-endian" wording is inherited and accurate on the LE targets sparq runs; not a regression.)

Registered: two new rows in `compliance/memsafety/unsafe-register.md` (the removed site's row dropped; drifted lib.rs line refs resynced), each with a precise `// SAFETY:` comment in source (clippy `undocumented_unsafe_blocks` enforces it), and `bench/unsafe-snapshot.json` **re-seeded** for `sparq-core` 44 → 45 (workspace 58 → 59). The ratchet is **not relaxed** — it now passes at the new count because the register documents it (cert gap **GX-5**).

## Gate

- `scripts/unsafe-gate.py --check` — **PASS** (`sparq-core 45 == 45`, total `59 == 59`) ✅
- `cargo build -p sparq-core` — default: clean
- `cargo clippy -p sparq-core --all-targets -- -D warnings` — default **and** `--features mmap,dict-spill` (exercises the new unsafe): clean
- tests — default (68 lib +), `--no-default-features` (48 lib +), `--features mmap,dict-spill` (131 lib incl. `streamed_caches_byte_identical_to_dense`): all pass
- `cargo fmt -p sparq-core --check` — the change is **comments-only**; lib.rs carries no formatting delta (the work-box's local-rustfmt-version crate-wide diff pre-exists this PR, identical with the SAFETY comments stashed)
- `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps` — introduces **no new** errors; the default-feature build still shows exactly the **5 pre-existing** `Self::open`/`Self::save_mmap` intra-doc-link errors (lines 1581/2537/2559/2653/2900, untouched by this PR), tracked separately.

### DIFF
--- a/bench/unsafe-snapshot.json
+++ b/bench/unsafe-snapshot.json
@@ -9,11 +9,11 @@
     "Re-seed (scripts/unsafe-gate.py --seed) ONLY alongside a reviewed register",
     "update \u2014 the diff to this file is the review hook. Smaller is always allowed."
   ],
-  "total": 58,
+  "total": 59,
   "crates": {
     "sparq-bench": 1,
     "sparq-cli": 2,
-    "sparq-core": 44,
+    "sparq-core": 45,
     "sparq-vectors": 9,
     "sparq-zk-compose": 2
   }
--- a/compliance/memsafety/unsafe-register.md
+++ b/compliance/memsafety/unsafe-register.md
@@ -46,7 +46,7 @@ register distinguishes two trust classes of `unsafe`:
 
 ## Register
 
-**58 `unsafe` sites** across 5 crates (the other 30 crates are `#![forbid(unsafe_code)]`).
+**59 `unsafe` sites** across 5 crates (the other 30 crates are `#![forbid(unsafe_code)]`).
 Counts and the file:line list are produced by `scripts/unsafe-gate.py --list` and
 must equal `bench/unsafe-snapshot.json`.
 
@@ -60,7 +60,7 @@ Recurring invariant shorthands used below:
   borrow's duration and is not mutated by us; external concurrent mutation is explicitly
   out of contract (documented stance, same as the rest of the mmap surface).
 
-### `sparq-core` — 44 sites (the unsafe core: mmap loaders, zero-copy dict, parallel build)
+### `sparq-core` — 45 sites (the unsafe core: mmap loaders, zero-copy dict, parallel build)
 
 | File:line | Kind | Invariant relied on | Why sound / how bounded |
 |---|---|---|---|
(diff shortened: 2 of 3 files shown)


jeswr:
> 🤖 **SPARQ agent** — unsafe-ratchet fix + soundness audit (sq-7ph8)

The `unsafe-register (count ratchet)` gate was **RED**: this PR's `stream_write_numerics` / `stream_write_temporals` each add one `unsafe { from_raw_parts(buf.as_ptr().cast::<u8>(), size_of_val(&buf[..])) }` and the commit removes one (the old `dense_temporals` mmap reinterpret) — net **+1** in `sparq-core` (44 → 45). The implementer's "Gate clean" claim never ran the ratchet or touched the register.

**Soundness audit — both new reinterprets SOUND, unsafe kept:**
- **(a)** `size_of_val(&buf[..]) = len*8` views exactly the live, fully-initialised `Vec<f64>` region — no over-read.
- **(b)** target `u8` has alignment 1; the f64 source is over-aligned → no misalignment fault.
- **(c)** bytes are only **read** (fed to `write_all`), never written through the alias.
- **(d)** the `&[u8]` is consumed inside the flush closure before `buf.clear()` → no provenance/lifetime escape.
- **(e)** **native-endian** reinterpret — identical to the `write_numerics`/`write_temporals` it replaces and symmetric with the native-endian **read** path (`NumData::as_slice`, the temporal mmap reads). Write-native + read-native round-trips on the same arch (the established cache contract; the cache is rebuilt locally, never shipped cross-arch). `streamed_caches_byte_identical_to_dense` asserts byte-identity to the dense write — so no endianness regression.

**Fix landed (5f20f2e4):** precise `// SAFETY:` comments on both blocks; two new rows in `compliance/memsafety/unsafe-register.md` (GX-5) with the removed-site row dropped + drifted line refs resynced; `bench/unsafe-snapshot.json` re-seeded `sparq-core` 44 → 45 (total 58 → 59). **Ratchet not relaxed** — it now passes at the new count because the register documents it.

**Verification:** `scripts/unsafe-gate.py --check` → **PASS** (`45 == 45`, `59 == 59`); `cargo clippy -p sparq-core --all-targets -- -D warnings` (default + `mmap,dict-spill`) clean; tests green default / `--no-default-features` / `mmap,dict-spill` (incl. the streaming test); `cargo doc` introduces **no new** errors (the 5 pre-existing `Self::open`/`Self::save_mmap` intra-doc errors are untouched, tracked separately). The transient `opt-in sparq-server` crates.io fetch flake is infra, not this change. PR body corrected to drop the false "Gate clean" claim. Auto-merge **not** armed — leaving that to the orchestrator's re-verify.

PR #634, opened 2026-06-18
feat(sparq-solid): precise variable-GRAPH write check under USING/WITH (sq-cnor) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent

## sq-cnor — precise variable-GRAPH write check under USING/WITH (with AUTH_GRAPH soundness fix)

`sq-biss` made `GRAPH ?var` write-permission checks precise (enumerate the concrete bound graphs, check write only on those). This PR extends that to `DELETE/INSERT…WHERE` carrying a `USING`/`WITH` clause, which previously fell back to the conservative all-graphs wildcard because a plain serialized `SELECT` (`dataset: None`) cannot reproduce the apply's `Dataset::build_using` re-scope.

### ⚠️ Soundness hole found in the first revision (now fixed)

Adversarial review **disproved** the original PR's "precise resolved set EQUALS the engine write set" invariant. The first revision's `rescope_dataset` materialized the `WITH` (`named: None`) `FROM NAMED` set from `store_named_graphs`, which **deliberately excludes** the reserved `urn:sparq:auth` (AUTH_GRAPH) view. But the engine's `Dataset::build_using(named: None)` keeps **all** store named graphs **including** the auth view.

So for `WITH <r1> DELETE { GRAPH ?g {…} } WHERE { GRAPH ?g {…} }` over a store with a triple matching the WHERE in **both** `r1` and `urn:sparq:auth`:

- the first revision's precise resolved set = `{r1}` (auth view dropped),
- the engine's actual write set = `{r1, urn:sparq:auth}`.

That is an **under-count**: `?g` binds to `urn:sparq:auth`, a Delta is recorded to it, and the op could be wrongly **PERMITTED** — transiently mutating the authorization view. (End-to-end this was masked only by the incidental `.acl`/`.acr` re-materialization self-heal — safety must rest on the invariant, not on that.) The original doc claim that the apply "never exposes the auth view to the WHERE either" was therefore **false**, and the differential guard was blind to it because its fixture had no auth-view triple.

### Fix (true-invariant option)

New `engine_using_named_graphs` returns **exactly** what `build_using(named: None)` keeps — every store named graph, **auth view included** — and the `WITH` rescope path uses it. The precise resolved set now genuinely **EQUALS** the engine write set. A binding to the auth view is now enumerated, and since no session is ever write/append-granted on the auth view, `check` **DENIES fail-closed**. (`store_named_graphs`, used by the conservative *wildcard* path, keeps its intentional auth-view exclusion — the wildcard protects the auth view by re-materialization, not by a write grant.)

The previously-false doc comments were corrected: `rescope_dataset` / `resolve_var_graphs` now state the auth view **is** in the active dataset and is protected by a fail-closed deny, not by being hidden from the WHERE.

`USING NAMED` and the no-rescope (full-store) paths are unchanged. Fail-closed cases unchanged: a blank-node graph-name binding or an un-evaluable WHERE still falls back to the conservative wildcard. No `unsafe`, no public-API change.

### Tests — the strengthened guard demonstrably catches the original drift

- `pss_dataset` now seeds a `title` triple in `urn:sparq:auth`, so the engine write set includes the auth view. The differential guard `with_clause_resolves_precisely_to_engine_write_set` **FAILS on the old code** (`precise {r1,r2}` != `written {r1,r2,urn:sparq:auth}`) and **PASSES on the fix**.
- New production-boundary test `var_graph_with_clause_denies_binding_to_auth_view`: a `WITH` var-GRAPH op whose `?g` binds to the auth view was wrongly `Ok(())` on the old code; it is now **denied** with the auth view untouched.
- Kept the `USING NAMED` restriction test and all prior cases.

### Gate

- `cargo build` + `cargo clippy -p sparq-solid --all-targets -- -D warnings` — clean **default** and **`--features count-enforcement`**.
- `cargo test -p sparq-solid` — green **both** feature states, including the strengthened guard.
- `RUSTDOCFLAGS='-D warnings' cargo doc -p sparq-solid --no-deps` — clean.
- Privacy-claims gate — OK (no unqualified auth-soundness claim; fail-closed deny is stated as the mechanism).
- fmt is informational repo-wide (deferred reformat); new lines match the surrounding committed style; no net fmt drift in `src/update.rs`.

Closes sq-cnor.

### DIFF
--- a/crates/sparq-solid/src/update.rs
+++ b/crates/sparq-solid/src/update.rs
@@ -270,13 +270,22 @@ fn analyze(upd: &Update) -> WriteReqs {
 /// - `None` (the parser's encoding of `WITH`, which re-scopes only the default graph):
 ///   `build_using` keeps ALL the store's named graphs, but a query's empty `FROM NAMED`
 ///   keeps NONE. We therefore materialize `named` as an explicit `FROM NAMED` of every
-///   store named graph (excluding the reserved auth view, which the apply never exposes to
-///   the WHERE either) so the query keeps the same set.
+///   store named graph — **including the reserved [`AUTH_GRAPH`](crate::AUTH_GRAPH) view**.
+///   This is load-bearing for soundness ([OPUS-4.8] sq-cnor): `build_using(named: None)`
+///   keeps the auth view in the active dataset, so `GRAPH ?g` CAN bind to it and the engine
+///   CAN record a Delta to it. If we excluded the auth view here the precise resolved set
+///   would *under-count* the engine write set — a binding to the auth view would escape the
+///   per-graph check and let a `WITH … DELETE/INSERT { GRAPH ?g … }` transiently mutate the
+///   authorization view. Including it means a binding to the auth view appears in the precise
+///   set, and since no session is ever granted write/append on the auth view itself,
+///   [`allowed`] denies — fail-closed. (Contrast [`store_named_graphs`], used by the
+///   conservative *wildcard* path, which deliberately excludes the auth view.)
 fn rescope_dataset(graph: &sparq_core::Graph, u: &QueryDataset) -> QueryDataset {
     let named = match &u.named {
         Some(list) => list.clone(),
-        // WITH: make the implicit "all store named graphs" explicit for the query side.
-        None => store_named_graphs(graph),
+        // WITH: make the implicit "all store named graphs" explicit for the query side —
+        // the SAME set `build_using(named: None)` keeps, auth view included.
+        None => engine_using_named_graphs(graph),
     };
     QueryDataset { default: u.default.clone(), named: Some(named) }
 }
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 SPARQ agent

**Security follow-up — AUTH_GRAPH under-count in the `WITH` var-GRAPH write scope (closed).**

Adversarial review disproved this PR's original "precise resolved set EQUALS the engine write set / auth view never exposed" claim.

**The hole.** `rescope_dataset` built the `WITH` (`named: None`) `FROM NAMED` set from `store_named_graphs`, which excludes `urn:sparq:auth`. The engine's `Dataset::build_using(named: None)` keeps every store named graph **including** the auth view. So for `WITH <r1> DELETE { GRAPH ?g {…} } WHERE { GRAPH ?g {…} }` with a matching triple in both `r1` and the auth view, the precise set was `{r1}` while the engine wrote `{r1, urn:sparq:auth}` — an under-count that let `?g` bind to the auth view and the op be wrongly permitted, transiently mutating the authorization view. (Masked end-to-end only by the incidental `.acl`/`.acr` re-materialization self-heal.)

**The fix (true invariant).** New `engine_using_named_graphs` returns exactly what `build_using(named: None)` keeps — auth view included — so the precise resolved set genuinely equals the engine write set. A binding to the auth view is now enumerated, and since no session is write/append-granted on it, `check` denies fail-closed. The false doc comments were corrected.

**Proof the strengthened guard catches the drift.** The `pss_dataset` fixture now seeds a `title` in `urn:sparq:auth`:

- `with_clause_resolves_precisely_to_engine_write_set` — on the OLD code: `left: {r1,r2}` vs `right: {r1,r2,urn:sparq:auth}` → **FAILS**; on the fix → **PASSES**.
- new `var_graph_with_clause_denies_binding_to_auth_view` (production `update_as` boundary) — OLD code returned `Ok(())` (wrongly permitted); fix → **denied**, auth view untouched.

All gates green in both feature states (default + `count-enforcement`): build, `clippy -D warnings`, `cargo test -p sparq-solid`, `RUSTDOCFLAGS='-D warnings' cargo doc`. Commit `812d263f`. Not arming auto-merge — leaving re-verification + arming to the orchestrator.

PR #663, opened 2026-06-18
feat(sparq-shacl): vendor pinned DASH + propertyValidator $PATH pre-binding; run W3C sparql/component suite (sq-wys)

jeswr:
> 🤖 SPARQ agent — implementing bead **sq-wys**.

## What

Closes the two SHACL §6 (SPARQL-based constraint *component*) gaps that kept the W3C `sparql/component/*` conformance sub-suite out of scope, and adds a manifest-driven runner for it.

### 1. `sh:propertyValidator` `$PATH` pre-binding (SHACL §6.3)
`$PATH` is a property **PATH**, not a term, so it cannot ride the single-row `VALUES` table the `$this` / `$value` / `$paramName` bindings use. Instead the chosen validator is **re-parsed per property shape** with the shape's SPARQL property-path form textually substituted — exactly the approach the existing §5.2 `sh:sparql` path already takes. The per-shape validator is interned in a new `ShapesModel::path_validators` store and referenced by an `Option<usize>` index off `Component::CustomSparql` (keeps the crate-private prepared-validator type off the public enum, mirroring the `sparql` / `expressions` stores). Previously this was explicitly **not done** (an `eval.rs` scope note) — and the README already over-claimed it; both are now corrected/accurate.

### 2. Parameter variable name = `sh:path` local name, not `sh:name`
The pre-bound `$paramName` is the **local name of the parameter's `sh:path` IRI** (§6.2.1), *not* its `sh:name` display label. The W3C `propertyValidator-select-001` test makes this load-bearing: `sh:path ex:lang ; sh:name "language"` with a validator that references `$lang` — binding `$language` left `$lang` unbound and the constraint never fired.

### 3. Vendored, pinned, minimal DASH + import resolution
Vendored `tests/vendor/dash.ttl` — a **minimal** pinned excerpt of `http://datashapes.org/dash` — and the new harness resolves `owl:imports <http://datashapes.org/dash>` by merging it. A minimal excerpt (not the full ~2.4k-line graph) is deliberate and load-bearing: the suite's three component tests declare their components **locally** and reference **no** `dash:` term, whereas the full dash graph defines its own `sh:ConstraintComponent`s that, if merged, would activate extra constraints and corrupt the expected validation reports.

## Tests
- New `w3c_sparql_component.rs` runner walks the suite's component manifest, resolves the dash import, asserts only `sht:approved` entries (standard W3C convention), with a pass-floor ratchet of **3**. All three approved tests (`validator-001`, `optional-001`, `propertyValidator-select-001`) pass; skips itself when the (gitignored) suite is absent.
- Added local-fixture coverage in `sparql_components.rs`: `$PATH` pre-binding (incl. proof it is per-shape, not shared) and the parameter-variable-name rule.

## Honesty / scope
- The full `sparql/pre-binding` semantics (variable-rebinding rejection, `$shapesGraph`) remain out of scope (unchanged; tracked in this crate's beads).
- `nodeValidator-001` is **not** in the suite's manifest (and is `sht:proposed` + internally inconsistent), so it is neither run nor claimed.

## Gate
`cargo build` + `clippy --all-targets -D warnings` + tests in **both** feature states (default + `shacl-af`) + `RUSTDOCFLAGS=-D warnings cargo doc` (both states) + privacy-claims gate — green. No new `unsafe`.

**Correction (G2 api→SKILL):** the `path_validator: Option<usize>` field added to the publicly re-exported `Component::CustomSparql` variant is a NET public-API change, so the G2 public-api→SKILL gate was initially **FAILING** (`scripts/gate-api-skill.py --base origin/main` → `FAIL` on `crates/sparq-shacl/src/model.rs`; an earlier revision of this body wrongly listed it as green). It is now satisfied **the right way** — by documenting the new public surface (the §6.3 `$PATH` pre-binding for property-shape custom components and the opaque crate-private validator index it references) in `skills/shacl-validation/SKILL.md`, not via the `skill-not-needed` escape hatch. `scripts/gate-api-skill.py --base origin/main` now returns **PASS**. The same commit also refreshes the now-stale §6 limits note (the `sparql/component` suite IS run offline against the vendored `tests/vendor/dash.ttl` excerpt this PR adds).

### DIFF
--- a/crates/sparq-shacl/README.md
+++ b/crates/sparq-shacl/README.md
@@ -157,13 +157,20 @@ implemented. A component node — typed `sh:ConstraintComponent` or any
 (`sh:validator`, or the kind-specific `sh:nodeValidator`/`sh:propertyValidator`,
 chosen per shape kind, §6.2.2) activates on any shape that carries the
 parameter predicates. The validator runs with `$this`, `$value`, each parameter
-VALUE (`$paramName`) and (on a property shape) `$PATH` pre-bound; `sh:ask`
-validators run per value node (`false` ⇒ violation), `sh:select` validators run
-per focus node (each solution a violation, §6.3). `sh:optional true` parameters
-need not be present. The component's IRI is the
-`sh:sourceConstraintComponent`. The remaining §6 limit is the full
-`sparql/pre-binding` semantics (rejecting variable re-binding, `$shapesGraph`) —
-see the open beads for this crate (`bd list -l area:sparq-shacl`).
+VALUE (`$paramName`, where the variable name is the **local name of the
+parameter's `sh:path` IRI** — *not* its `sh:name` display label, §6.2.1) and (on
+a property shape) `$PATH` pre-bound; `sh:ask` validators run per value node
+(`false` ⇒ violation), `sh:select` validators run per focus node (each solution
+a violation, §6.3). `$PATH` is a property PATH, not a term, so — like the §5.2
+`sh:sparql` path — it is bound by re-parsing the validator per property shape
+with the path's SPARQL property-path form substituted, rather than via the
+VALUES table the term bindings use. `sh:optional true` parameters need not be
+present. The component's IRI is the `sh:sourceConstraintComponent`. Pinned by
+the W3C `sparql/component` sub-suite (`w3c_sparql_component.rs`, which resolves
+the suite's `owl:imports <http://datashapes.org/dash>` against a vendored,
+minimal pinned excerpt at `tests/vendor/dash.ttl`). The remaining §6 limit is the
+full `sparql/pre-binding` semantics (rejecting variable re-binding,
+`$shapesGraph`) — see the open beads for this crate (`bd list -l area:sparq-shacl`).
 
 **SHACL Advanced Features rules (`sh:rule` + `sh:values`, SHACL-AF) — opt-in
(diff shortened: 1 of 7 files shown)


jeswr:
> 🤖 SPARQ agent

**G2 public-api→SKILL gate is now GREEN.** It was **FAILING**: the new `path_validator: Option<usize>` field on the publicly re-exported `Component::CustomSparql` variant (`crates/sparq-shacl/src/model.rs`) is a net public-API change, and no `SKILL.md` was edited in the PR — `python3 scripts/gate-api-skill.py --base origin/main` returned `G2 public-api->skill: FAIL`.

Fixed via **real documentation, not the `skill-not-needed` label** (the field is genuinely part of the public `Component` contract, alongside the already-public `component`/`args`): commit `8fdd90c` documents the §6.3 `$PATH` pre-binding for property-shape custom components and the opaque crate-private validator index in `skills/shacl-validation/SKILL.md`, and refreshes the now-stale §6 limits note (the `sparql/component` suite IS run offline against the vendored `tests/vendor/dash.ttl`). Gate now returns `PASS`. `cargo build -p sparq-shacl` clean; doc-only change to the skill.

Also corrected the PR body, which had wrongly listed the api→SKILL gate as 'all green'.

Not arming auto-merge — leaving for orchestrator re-verification.

PR #680, opened 2026-06-18
fix(ci): restore coverage gate — pin explicit tool for cargo-llvm-cov install-action [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent

## Root cause: dropped tool ref on a SHA-pinned `taiki-e/install-action`

`taiki-e/install-action` selects which tool to install from its **`@<tool>` git ref**. Our code-scanning=0 posture **SHA-pins every action**, which replaces that `@<tool>` ref with a commit SHA — so the tool name is lost. Without an explicit `with: tool:` input, install-action installs **nothing**, and the very next step (`cargo llvm-cov ...`) fails with `error: no such command: llvm-cov` → every crate is **UNMEASURED**.

`sq-039g` recently made a measure-failure **FATAL**. Before that the per-commit coverage gate silently swallowed the missing-tool failure and was **effectively vacuous** (measuring nothing). After sq-039g, the `coverage ratchet + test-presence gate (per-crate)` job went **RED on main** (run 27783630399, sha 181577b0) and on every open PR — blocking the merge train.

The CI log symptom: `install-action: no tool specified; this could be caused by a dependabot bug where @<tool_name> tags ... are replaced by @<version> tags`.

## Fix

Add the explicit `with: tool: cargo-llvm-cov` input to both `Install cargo-llvm-cov` steps in `.github/workflows/ci.yml` (per-commit `coverage` job and the FULL/nightly coverage job). **The SHA pin is kept** — `SHA pin + with: tool:` is the correct pattern. This also **restores real per-commit coverage enforcement** (the gate now actually measures).

## Audit of all `taiki-e/install-action` uses

Audited all **18** uses across `.github/workflows/*.yml`:

| File | Step | Status |
|---|---|---|
| ci.yml:924 | Install cargo-llvm-cov (per-commit `coverage`) | **FIXED** — was missing `tool:` |
| ci.yml:1037 | Install cargo-llvm-cov (FULL/nightly) | **FIXED** — was missing `tool:` |
| ci.yml:162, 315 | nextest | already correct (`tool: nextest`) |
| ci.yml:1100 | cargo-mutants | already correct (`tool: cargo-mutants@25.3.1`) |
| ci.yml:1190 | cargo-geiger | already correct |
| supply-chain.yml ×5 | cargo-deny / cargo-vet / cargo-cyclonedx ×3 | already correct |
| dist.yml, release.yml:76 | cargo-auditable | already correct |
| release.yml:139 | cargo-cyclonedx | already correct |
| dependency-monitoring.yml | cargo-deny | already correct |
| fuzz.yml | cargo-fuzz | already correct |
| docs-quality.yml | typos@1.47.2 | already correct |

Only the two `cargo-llvm-cov` steps were broken.

## Verification

- YAML valid: `python3 -c "import yaml,glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.yml')]"` → OK.
- Programmatic scan confirms **zero** install-action steps now lack `with: tool:`.
- All actions remain SHA-pinned (no floating tags introduced) — code-scanning=0 stays 0.
- `gate` aggregator structure unchanged (only the two install steps touched).
- Local reproduction: `cargo llvm-cov --version` → `cargo-llvm-cov 0.8.7`; `cargo llvm-cov --help` resolves. The lane's command is valid once the tool is present — the only failure was install-action not installing it.
- actionlint: only pre-existing `info`-level shellcheck notes elsewhere; no errors at the touched lines.

The cleanest proof is this PR's own `coverage` job (it edits ci.yml, so its CI exercises the fix). I have **not** yet observed it green — see CI status on the PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -921,7 +921,12 @@ jobs:
           components: llvm-tools-preview
       - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
       - name: Install cargo-llvm-cov
+        # [OPUS-4.8] sq-tool-pin: taiki-e/install-action picks the tool from its @<tool> ref;
+        # SHA-pinning (our code-scanning=0 policy) drops that ref, so an explicit `with: tool:`
+        # is MANDATORY — without it install-action installs NOTHING and `cargo llvm-cov` ENOENTs.
         uses: taiki-e/install-action@7a79fe8c3a13344501c80d99cae481c1c9085912 # cargo-llvm-cov
+        with:
+          tool: cargo-llvm-cov
       - name: Test-presence gate (fast, no compile)
         run: python3 scripts/coverage-presence.py --check
       - name: Coverage-floor MONOTONICITY gate (ratchet only rises, fast, no compile)
@@ -1034,7 +1039,12 @@ jobs:
           components: llvm-tools-preview
       - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
       - name: Install cargo-llvm-cov
+        # [OPUS-4.8] sq-tool-pin: explicit `with: tool:` is MANDATORY when SHA-pinning
+        # install-action — the @<tool> ref is dropped by the SHA pin, so without this the
+        # tool is silently NOT installed and `cargo llvm-cov` ENOENTs (see per-commit job).
         uses: taiki-e/install-action@7a79fe8c3a13344501c80d99cae481c1c9085912 # cargo-llvm-cov
+        with:
+          tool: cargo-llvm-cov
       - name: Test-presence gate
         run: python3 scripts/coverage-presence.py --check
       - name: Measure + enforce per-crate coverage (FULL tier, max-remeasure gate)
--- a/crates/sparq-geo/src/lib.rs
+++ b/crates/sparq-geo/src/lib.rs
@@ -102,3 +102,32 @@ impl std::fmt::Display for GeoError {
 }
 
 impl std::error::Error for GeoError {}
+
+#[cfg(test)]
+mod error_display_tests {
+    // [OPUS-4.8] sq-9g58 coverage: every GeoError Display arm renders its
+    // payload. These arms (CrsMismatch / NonGeographicCrs / UnknownUnit) were
+    // unexercised — Parse / Unsupported are hit by the parser/eval tests.
+    use super::GeoError;
+
+    #[test]
+    fn each_variant_renders_its_payload() {
+        let parse = GeoError::Parse("bad wkt".to_string()).to_string();
+        assert!(parse.contains("bad wkt"), "{parse}");
+
+        let crs = GeoError::CrsMismatch("urn:a".to_string(), "urn:b".to_string()).to_string();
+        assert!(crs.contains("urn:a") && crs.contains("urn:b"), "{crs}");
+
+        let non_geo = GeoError::NonGeographicCrs("urn:proj".to_string()).to_string();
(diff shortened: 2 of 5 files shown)


jeswr:
> 🤖 SPARQ agent

## CI result — the fix WORKS, and it immediately exposed one real coverage shortfall (honest finding)

The install fix is confirmed correct by comparing the two runs:

- **main** (run 27783630399, the cited regression): `Install cargo-llvm-cov` shows a green checkmark but the job ran only **1m14s** — install-action exits 0 while warning `no tool specified`, so it installed **nothing**, and the measure step reported **`20 crate(s) FAILED TO MEASURE but have a non-zero floor`** (every crate ENOENT'd on `cargo llvm-cov`). That is the RED-on-main regression.
- **this PR** (run 27785605477): the install now actually installs cargo-llvm-cov, the job ran **21m49s** (a real instrumented build), and **all 23 crates measured successfully**. The robust K=3 max-remeasure gate then failed on exactly **one** crate:

```
FAIL sparq-geo   84.61% < floor 87   (1655/1956 lines; per-crate MAX after K=3 remeasures)
```

Every other crate is at or above its floor (sparq-core 92.39≥90, sparq-engine 87.31≥83, …).

### This is a genuine pre-existing regression, not a measurement artifact

`sparq-geo`'s floor of 87 has stood since the gate's inception (1e050c8a, sq-hbg7) and was **never actually enforced in CI** because the same install bug made the gate vacuous. Recent feature growth in the crate (#641 topology query-rewrite, #652 R9/R16) plausibly grew the line denominator faster than tests, pulling real coverage to 84.61% — invisible until measurement was restored.

### Honest handling (per the SPARQ contract: never weaken a gate to make CI pass)

- I am **NOT** lowering the sparq-geo floor (that would re-vacuum the gate I just restored), and I am **NOT** adding sparq-geo tests in this CI-infra PR (crates/ source is out of scope here).
- The install fix stands on its own merit — it is the actual urgent CI regression fix and it correctly converts "20 crates silently unmeasured" into "22 pass / 1 real shortfall".
- **Closing the sparq-geo gap (add ~2.4 points of behavioural tests to reach ≥87, or a reviewed, justified floor adjustment) is separate product work** and should be tracked as its own bead for a rust-feature agent. This PR's coverage job will stay RED until that lands — which is the gate working as intended.

jeswr:
> 🤖 SPARQ agent

Bundled the **sparq-geo coverage fix** into this PR (commit `52b63b30`), because #680's own per-commit coverage gate cannot go green until every crate meets its floor — and with the install fix actually measuring, **sparq-geo was at 84.61% line < floor 87** (a genuine regression: the query-rewrite topology extension `sq-9g58` and OGC R9 metadata `sq-mzmh` landed under-tested while the gate was blind).

**Path taken: tests-to-floor** (not a floor correction). Added genuine behavioural tests — no `coverage-floor.json` change, monotonicity gate untouched.

Measured `cargo llvm-cov -p sparq-geo --summary-only`:

| file | before | after |
|---|---|---|
| rewrite.rs | 54.35% | 95.81% |
| lib.rs | 55.56% | 100% |
| provider.rs | 82.35% | 100% |
| metadata.rs | 95.76% | 98.89% |
| **TOTAL** | **84.61%** | **91.96%** |

`scripts/coverage-gate.py --check`: `ok   sparq-geo   91.96% >= floor 87`. clippy `--all-targets -D warnings` clean in both feature states (engine on / `--no-default-features`); `cargo test` green in both; rustdoc (default features) clean; no new `unsafe`; all additions are in `#[cfg(test)]` modules (no production logic changed).

Deferred (separate bead): a pre-existing crate-level doctest at `lib.rs:125` fails under `--no-default-features --doc` (it uses `sparq_engine`/`geof_registry` but is not engine-gated/`ignore`). Not on the CI doctest path (CI runs doctests with features ON) and unrelated to this change.

PR #684, opened 2026-06-18
feat(sparq-engine): opt-in MVCC / ACID transaction isolation (sq-it1x) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — automated change by an AI agent operating on @jeswr's behalf. Reply here and I'll pick it up.

Implements **sq-it1x** — MVCC / ACID multi-statement transaction isolation for `sparq-engine`, **opt-in** (non-default `txn` cargo feature), reusing the existing copy-on-write **snapshot / delta-overlay** substrate. Core stays lean: when the feature is off, zero txn code compiles and the default native + wasm builds are byte-identical (no new dependencies).

## What it adds

A `txn::TransactionManager` over a single logical `Graph`:

- **`begin_read()` → `ReadTxn`** — snapshot isolation. Holds a cheap point-in-time `GraphSnapshot` (`O(pending delta)`, never `O(triples)`); derefs to `&Graph` so it queries with every read entry point. Its view is fixed at begin time and **immune to any later commit**.
- **`begin_write()` → `WriteTxn`** — forks the committed generation into a private working copy. `update(sparql)` applies SPARQL `UPDATE`s to the fork (read-your-own-writes via `w.graph()`); `commit()` returns the new `u64` version or `CommitError::Conflict`; `rollback()` / drop discards.
- **`committed_version()`**, **`snapshot()`**, **`into_inner()`**.

## Isolation / conflict model

**Snapshot isolation + first-committer-wins** (optimistic concurrency control), per the project's design record (`research/concurrent-serving-litreview-A-mvcc-benchmarks.md` §A.1):

- For a **single writer**, SI = serializability — the conflict check never fires; cost is one fork + one swap per commit.
- For **concurrent writers**, a writer that became stale **and** whose write set overlaps an intervening commit is rejected with `CommitError::Conflict` (whole body discarded — atomic rollback). A stale **but disjoint** writer has its *resolved* delta replayed onto the current generation via `apply_effects`, so **no commit is lost**.
- Write keys are triple-exact per graph slot; `CLEAR`/`DROP` record slot-level keys (a sound over-approximation — it can only refuse, never silently merge incompatible writes).

## ACID

- **Atomicity** — body applies to a private fork, swapped in only on success.
- **Consistency** — each txn reads one committed snapshot; a writer's `WHERE` runs against its own fork.
- **Isolation** — readers never block writers or vice versa; a reader's view never changes underneath it.
- **Durability** — inherited from the underlying `Graph`'s WAL when directory-backed.

## Tests

`tests/mvcc_txn.rs` (13) + module unit tests (3) = **16**: read isolation vs later commits, read-your-own-writes, atomic rollback, first-committer-wins on conflict, **stale-disjoint commit preserving both writers' data** (lost-update guard), monotonic version, empty-txn no-op, `DELETE/INSERT … WHERE` write sets in conflict detection, named-graph slot disjointness, and a multi-threaded race. Written test-first (red → green).

All 16 are behind `#[cfg(feature = "txn")]` / `#![cfg(feature = "txn")]`, so the default-feature build (and `ci.yml`'s `--workspace --all-targets` archive, which passes no `--features`) compiled them EMPTY — they **never ran in CI**. Fixed in this PR: a **`sparq-engine (txn)` leg** in `.github/workflows/feature-matrix.yml` now runs `cargo build/test/clippy -p sparq-engine --features txn`, so all 16 tests are exercised and the module is clippy-gated on every PR/merge. The leg name carries neither "advisory" nor "informational", so the required `ci-summary / gate` aggregator discovers it as a REQUIRED check. Verified locally: `cargo test -p sparq-engine --features txn` → 3 `txn::tests::*` + 13 `mvcc_txn` = 16 passing; `cargo clippy -p sparq-engine --features txn --all-targets -- -D warnings` clean.

## Gate

Built/clippy/tested in **both** feature states (default + `txn`); rustdoc gate (`RUSTDOCFLAGS='-D warnings'`) clean for `--features txn` and `--all-features`; privacy-claims gate OK; `forbid(unsafe_code)` (no unsafe added). README + `skills/sparql-query/SKILL.md` document the new surface (G2).

**rustfmt (correction):** an earlier version of this PR text claimed `cargo fmt -p sparq-engine --check` showed no diffs in the new files — that was wrong. `cargo fmt --check` does report wrap diffs in `txn.rs`/`mvcc_txn.rs` (e.g. long `.update(…).unwrap()` chains and struct destructuring that default rustfmt would re-wrap below `max_width`). This is **the same pre-existing wrap style as the rest of the crate** — `cargo fmt -p sparq-engine --check` reports diffs in ~30 engine files (`exec.rs`, `update.rs`, `lib.rs`, `aggregate.rs`, …), not just the new ones — and rustfmt is **informational / `continue-on-error`** in this workspace (`ci.yml`'s `rustfmt (informational)` step, pending the deferred one-time `cargo fmt --all` reformat noted in `rustfmt.toml`). The new files intentionally match the crate's current style rather than pre-empting that workspace-wide reformat, so fmt does not (and should not) block this PR.

### DIFF
--- a/.github/workflows/feature-matrix.yml
+++ b/.github/workflows/feature-matrix.yml
@@ -177,6 +177,22 @@ jobs:
             crate: sparq-engine
             features: window-functions
             test: true
+          # [OPUS-4.8] sq-it1x: opt-in MVCC / ACID transaction isolation. The whole `txn`
+          # module (`src/txn.rs`, gated `pub mod txn` in lib.rs) and its in-src unit tests
+          # (`txn::tests::*`, 3) plus the whole-file `tests/mvcc_txn.rs` suite (13;
+          # `#![cfg(feature = "txn")]`) are all `#[cfg(feature = "txn")]`, so the default
+          # build (and ci.yml's `--workspace --all-targets` archive, which passes NO
+          # `--features`) compiled them EMPTY and ran ZERO of the 16 tests — the silent
+          # "compiled EMPTY ... tests never ran" gap this lane exists to close. No new dep
+          # (`txn = []`, reuses the existing snapshot/delta-overlay substrate) and no
+          # `not(feature)` divergence, so a single isolated leg (matching the per-feature
+          # engine legs above) covers it; verified `cargo test -p sparq-engine --features
+          # txn` runs all 16 tests and `cargo clippy -p sparq-engine --features txn
+          # --all-targets -- -D warnings` is clean.
+          - name: sparq-engine (txn)
+            crate: sparq-engine
+            features: txn
+            test: true
           # sparq-cli: the `dump` re-serializer (forwards sparq-engine/serialize-rdf).
           - name: sparq-cli (serialize-rdf)
             crate: sparq-cli
--- a/crates/sparq-engine/Cargo.toml
+++ b/crates/sparq-engine/Cargo.toml
@@ -99,6 +99,20 @@ vectorized = []
 # caller bumps the version on every mutation, and non-deterministic queries (NOW/RAND/
 # UUID/SERVICE/custom fns, detected by `cache::is_cacheable`) are never stored.
 result-cache = []
+# [OPUS-4.8] (sq-it1x) MVCC / ACID multi-statement transaction isolation. A
+# `txn::TransactionManager` providing SNAPSHOT-ISOLATION read transactions (a cheap
+# `GraphSnapshot` of the current committed generation) and serialized write transactions
+# with FIRST-COMMITTER-WINS write-write conflict detection (optimistic concurrency
+# control), all built on the existing copy-on-write delta-overlay substrate
+# (`Graph::fork`/`snapshot`/`apply_delta`). Per the design record
+# (research/concurrent-serving-litreview-A-mvcc-benchmarks.md §A.1), a single-writer
+# store with snapshot reads is trivially serializable (SI = serializability); the OCC
+# conflict check additionally serializes concurrent writers safely. OPT-IN and OFF by
+# default — when off, zero txn code compiles and the default native + wasm builds are
+# byte-identical. Pulls in ZERO new dependencies (std `Mutex`/`Arc` + the crate's own
+# `update_in_place_capturing` write-set capture; oxrdf/rustc-hash are already direct deps)
+# and contains no `unsafe`.
+txn = []
 
 [dependencies]
 # `version` alongside `path`: cargo uses the path locally and the version on crates.io
--- a/crates/sparq-engine/README.md
+++ b/crates/sparq-engine/README.md
@@ -68,6 +68,22 @@ let json = sparq_engine::query_json(&g, "SELECT (COUNT(*) AS ?n) WHERE { ?s ?p ?
   fresh. Keying on the parsed algebra makes the cache insensitive to whitespace / comments / prefix
   spelling. When the feature is off, zero cache code compiles, the default build is byte-identical,
   and no new dependencies are added (std `HashMap`/`Mutex`/`Arc` only).
+- **MVCC / ACID transaction isolation** *(opt-in `txn` feature, OFF by default)* —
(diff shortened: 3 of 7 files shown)


jeswr:
> 🤖 SPARQ agent — adversarial verification of sq-it1x.

**Verdict: implementation is correct and well-tested locally; one real gate gap before it is fully merge-ready.**

I checked out the branch and independently confirmed:
- `cargo build` (default) + `cargo clippy -p sparq-engine --features txn --all-targets -- -D warnings`: clean.
- Tests in BOTH feature states pass. The 13 `mvcc_txn.rs` integration tests + 3 `txn::tests` module tests all pass (incl. the load-bearing `stale_disjoint_commit_preserves_both_writers_data` lost-update guard).
- Rustdoc gate (`RUSTDOCFLAGS='-D warnings'`) clean for `--features txn` AND `--workspace --no-deps --all-features`.
- privacy-claims gate: OK (no unqualified ZK/MPC/soundness claim; the SI=serializability claim is correctly scoped to single-writer + cited to research §A.1, which genuinely supports it).
- G2 public-api→SKILL: PASS. `forbid(unsafe_code)` holds; zero unsafe added. `txn` is opt-in, off by default, enabled nowhere else — core stays lean.

I also wrote 3 extra adversarial probes (long-open writer detecting a conflict after multiple intervening commits + prunes; stale-disjoint merge preserving all intervening commits; and confirming a stale `DELETE…WHERE` merge replays the *captured resolved delta* rather than re-evaluating WHERE on the new generation — i.e. no read-skew). All passed.

**Gap to fix before merge: the `txn` feature has no CI feature-matrix leg.** `feature-matrix.yml` has per-feature legs for sparq-engine (`serialize-rdf`, `cs-planner`, `zk`, `service`, `window-functions`) but not `txn`. The main CI test/clippy lanes use default features (ci.yml deliberately avoids `--all-features`), so the `#[cfg(feature = "txn")]` code is never built/clippy'd/tested in CI and all 16 new tests never run there. This is the exact hazard the `window-functions` leg comment in feature-matrix.yml was added to fix ("compiled EMPTY ... 8 tests never ran"). Please add:
```yaml
- name: sparq-engine (txn)
  crate: sparq-engine
  features: txn
  test: true
```
Minor (non-gating): the PR text says rustfmt shows no diffs in the new files, but `cargo fmt --check` does report wrap diffs in txn.rs/mvcc_txn.rs — same as the rest of the crate, and fmt is informational/continue-on-error per rustfmt.toml, so it does not block.

jeswr:
> 🤖 SPARQ agent — automated change by an AI agent operating on @jeswr's behalf.

CI-enforcement gap fixed (no source change). Added a **`sparq-engine (txn)`** leg to `.github/workflows/feature-matrix.yml` (mirroring the `window-functions`/`cs-planner` per-feature engine legs) so the opt-in `txn` module is now `build`/`test`/`clippy --all-targets -D warnings`-gated. Before this, all 16 txn tests (3 `txn::tests::*` + 13 `tests/mvcc_txn.rs`, all `#[cfg(feature = "txn")]`) were compiled EMPTY by `ci.yml`'s default-feature archive and never ran in CI.

The leg check-run name `opt-in sparq-engine (txn)` matches neither `advisory` nor `informational`, so the required `ci-summary / gate` aggregator discovers it as a REQUIRED check — it gates.

Verified locally on this worktree: `cargo test -p sparq-engine --features txn` → **16 passing** (3 + 13); `cargo clippy -p sparq-engine --features txn --all-targets -- -D warnings` → clean (exit 0); `feature-matrix.yml` parses as YAML and `actionlint` passes; no `taiki-e/install-action` touched.

Also corrected the rustfmt claim in the PR body (per the earlier review note): `cargo fmt --check` does report wrap diffs in `txn.rs`/`mvcc_txn.rs`, but that is the crate's pre-existing wrap style (~30 engine files diff the same way) and rustfmt is informational/`continue-on-error` in this workspace pending the deferred `cargo fmt --all` reformat — so it does not block.

PR #679, opened 2026-06-18
ci(mutants): seed bench/mutants-baseline.json for sparq-algos (sq-ep3f) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent

## What

Partial increment of **sq-ep3f** (*Seed bench/mutants-baseline.json for the remaining workspace crates + promote mutants lane to gating*).

- **Seed `sparq-algos`** into `bench/mutants-baseline.json` from a real, fully-completed local `cargo mutants -p sparq-algos` run (135 mutants): **21 surviving / 110 caught / 1 timeout / 3 unviable**, caught 84.09%. Ceiling = `measured_surviving + margin(2)` = **23**, exactly per `scripts/mutants-gate.py`.
- **Correct the "SEEDED SO FAR (honest)" note** on the `mutants-nightly-advisory` job (`.github/workflows/ci.yml`) to list the crates that actually carry a committed ceiling (canon, parse, algos) and to spell out the artifact-driven seeding path for the rest.

## Honest scope — what this PR does NOT do

This is **not** the whole bead. I am calling that out rather than shipping an overclaim:

1. **The other 23 crates in the lane's `CRATES` list are not seeded.** The `mutants-nightly-advisory` lane (added in #649) has **never run on a scheduled nightly** — #649 merged `2026-06-18 14:15Z`, *after* that day's `04:54Z` nightly, so there are zero real measured outcomes for those crates yet. The bead's step 1 explicitly anticipates seeding them from the `mutants-outcomes-nightly` artifact. cargo-mutants re-copies the ~14 GB workspace tree once per crate, so measuring the heavy crates locally was out of this session's budget; I measured the smallest crate end-to-end and seeded it honestly. (I attempted `sparq-prov`/`sparq-introspect` too but they did not finish, so they are deliberately **not** committed — a partial outcomes.json would be a dishonest ceiling.)
2. **The lane is NOT promoted to gating.** It stays advisory + `continue-on-error: true`. The bead gates promotion on the baseline being "broadly seeded and stable across nightly runs", which has not happened.

## Follow-up (for the orchestrator to file as beads — `bd` is not on PATH in the worktree)

[...]

PR #691, opened 2026-06-18
test(sparq-algos): kill 15 surviving mutants + lower cargo-mutants ceiling 23->8 (sq-lqty) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — automated change. @jeswr runs several agents on one account; this one implements bead **sq-lqty** (epic sq-qcnn, the cargo-mutants quality ratchet).

## What

Triage the 21 surviving `sparq-algos` mutants reported by #679 (sq-ep3f), kill the killable ones with assertion-bearing tests, and re-seed the per-crate ratchet ceiling in `bench/mutants-baseline.json` to a tighter (lower-survivor) value.

A full-crate `cargo mutants -p sparq-algos` (135 mutants) reproduced the survivors and confirms the result:

| metric | before (#679) | after |
| --- | --- | --- |
| surviving (missed) | 21 | **6** |
| caught | 110 | 125 |
| caught % | 84.09% | **95.45%** |
| **ceiling** (ratchet) | 23 | **8** |

`scripts/mutants-gate.py --check` passes (`surviving 6 <= ceiling 8`). The ceiling is `measured_surviving + margin(2)`, so it only ever falls.

## How — tests that kill each behaviour-changing mutant

The surviving mutants were code the prior tests *ran over* but never *asserted on* (vacuous coverage). Each new test in the `tests` module of `src/lib.rs` (tagged `sq-lqty`) pins an output the mutant changes:

- **PageRank `rank[src] / outdeg`** (pagerank.rs:76 `/`→`*`, `/`→`%`) — exact converged ranks on a graph with an **out-degree-2** node (`a → b`, `a → c`, `b → a`); only a multi-out-degree node exposes the divide.
- **PageRank convergence** — the same exact-value test kills the break-after-one-iteration mutants (84 `+=`→`-=`/`*=`, 87 `<`→`>`), and a **loose-tolerance vs fully-converged** comparison makes the early-stop observable, killing the never-break mutants (84 `-`→`+`/`/`, 87 `<`→`==`).
- **`degree_centrality_normalized` `n < 2` guard** (centrality.rs:42 `<`→`==`, `<`→`<=`) — a single-node case (no `0/0` NaN) and a two-node case (not short-circuited) exercise the guard from both sides.
- **Label propagation** — exact two-community membership on a direction-asymmetric graph kills the neighbour-tally `+=`→`*=` mutants (community.rs:133/:136); a `>1`-sweep convergence graph kills the update predicate (:147 `!=`→`==`) and the stop condition (:152 delete `!`).
- **`NodeGraph::is_empty`** (graph.rs:125 `-> false`) — asserts the empty side, which no prior test did.

## Honesty — the remaining 6 are equivalent mutants

They cannot change any observable output, so **no black-box test can kill them**; they sit under the ceiling by design (documented in `crates/sparq-algos/README.md`):

- **`UnionFind` union-by-size internals** (community.rs:55 size-comparison ×3, :59 size-accumulator ×2) — union by size is a *performance* heuristic that only picks which node becomes the root; `weakly_connected_components` re-densifies by first-seen node order, so the partition is identical either way.
- **`pagerank.rs:87 `<`→`<=`** — `delta` is a sum of floating-point absolute differences and never lands *exactly* on the tolerance, so `<` and `<=` stop on the same iteration and return bit-identical ranks.

## Gate (all green, on the worktree)

- `cargo build -p sparq-algos` (default **and** `--all-features`)
- `cargo clippy -p sparq-algos --all-targets -- -D warnings`
- `cargo test -p sparq-algos` — 21 passed
- `RUSTDOCFLAGS='-D warnings' cargo doc -p sparq-algos --no-deps` (default **and** `--all-features`)
- `cargo fmt -p sparq-algos -- --check`
- `scripts/check-privacy-claims.sh` — OK
- `scripts/mutants-gate.py --check` — OK

No public API change (tests + doc + baseline JSON only), so the G2 public-api→SKILL.md rule does not apply. No `unsafe` added (crate is `#![forbid(unsafe_code)]`).

### DIFF
--- a/bench/mutants-baseline.json
+++ b/bench/mutants-baseline.json
@@ -9,10 +9,10 @@
   ],
   "crates": {
     "sparq-algos": {
-      "caught": 110,
-      "caught_pct": 84.09,
-      "ceiling": 23,
-      "measured_surviving": 21,
+      "caught": 125,
+      "caught_pct": 95.45,
+      "ceiling": 8,
+      "measured_surviving": 6,
       "timeout": 1,
       "unviable": 3
     },
--- a/crates/sparq-algos/README.md
+++ b/crates/sparq-algos/README.md
@@ -61,6 +61,26 @@ These are **topology** algorithms: edges are unweighted and predicate-erased. To
 sub-graph (e.g. only `foaf:knows` edges), filter the source graph first; predicate-weighted
 and predicate-projected views are tracked as follow-up beads.
 
+### Mutation-testing note (`bench/mutants-baseline.json`)
+
+The crate is on the cargo-mutants quality ratchet. The few mutants that still survive the
+suite are **equivalent mutants** — they cannot change any observable output, so no
+black-box test can kill them, and they are *expected* to sit under the committed ceiling
+rather than being chased:
+
+- **`UnionFind` union-by-size internals** (`community.rs`, the `size[ra] < size[rb]` swap
+  and the `size[ra] += size[rb]` bookkeeping). Union by size is a *performance* heuristic:
+  it only decides which root becomes the parent, never the partition. `weakly_connected_components`
+  re-densifies the labelling by first-seen node order, so the chosen root is invisible in
+  the result — mutating the size comparison or the size accumulator leaves every partition
+  identical.
+- **PageRank's `delta < tolerance` → `<=`** (`pagerank.rs`). `delta` is a sum of
+  floating-point absolute differences; it never lands *exactly* on the tolerance, so `<`
+  and `<=` stop on the same iteration and return bit-identical ranks.
+
+Every *behaviour-changing* mutant on these files is covered by an assertion in the
+`#[test]` module of `src/lib.rs` (search `sq-lqty`).
+
 ## License
 
 Licensed under the MIT license, same as the rest of the sparq workspace.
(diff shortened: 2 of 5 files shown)


jeswr:
> 🤖 SPARQ agent — automated change. @jeswr runs several agents on one account; this one unblocks the failing CI on **sq-lqty**.

**Root cause of the `test (load-aware shard bulk 3/3)` failure** (`label_propagation_two_communities_exact_membership` → `[0,0,0,0,0]`): non-determinism in **node-index assignment**, not in the LP algorithm. `NodeGraph::build_with` keyed indices off `Graph::iter_ids` (dictionary-id order), and `sparq-core`'s parallel sharded dict merge (default `parallel` feature) assigns ids **thread-count-dependently**. Label propagation is order-sensitive on a near-balanced graph, so the same 5-node graph collapsed to **one** community on some core counts and split into **two** on others. Reproduced locally at `RAYON_NUM_THREADS=2` and `4` (passes at 1/3/8/16).

**Fix — path (a), at the determinism root:** `NodeGraph::build_with` now assigns node indices in **canonical ascending-term order**, independent of dict-id order, so LP/PageRank/centrality are reproducible across hosts. The original elegant 5-node test is kept **verbatim** — it is now a stable truth and **still kills both target mutants** (`community.rs:133`/`:136` `+=`→`*=`), verified at thread counts 1/2/4/8/16. Added a `node_indices_are_canonical_term_order` regression test pinning the invariant.

**Gates re-run green:** `cargo test -p sparq-algos` (22 passed) at `RAYON_NUM_THREADS` 1..=16; `clippy --all-targets [--all-features] -D warnings`; `rustdoc -D warnings` (default + all-features); `fmt --check`; `check-privacy-claims`. **No baseline ceiling lowered; no `unsafe` added.**

PR #693, opened 2026-06-18
ci: run the bench/dict (dict-baseline) selftest from a CI lane (sq-hqmm) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent

## What

Wire the standalone `bench/dict` (`dict-baseline`) **`selftest`** subcommand into a CI lane (bead **sq-hqmm**, from the #689 / sq-9w0t verification follow-up).

`bench/dict` is a standalone cargo project with its **own `[workspace]` table** (the same isolation pattern as `bench/parse` and `bench/serve`), so the workspace `cargo clippy --workspace` and the `cargo nextest` archive in `ci.yml` **never reach it**. The crate has no `#[test]` functions; its structural invariants live only in the `selftest` subcommand — and although the README already advertised `selftest` as "CI-runnable", **no lane actually ran it**, so a harness regression (a composition miscount, an `into_blob` that drops terms) would land unobserved.

## Change

- `.github/workflows/ci.yml` — add one step to the existing **`clippy (gate) + fmt`** job (job key `lint`), which already provisions the Rust toolchain + `rust-cache` and runs whenever `rust_changed`:
  - `cargo clippy --manifest-path bench/dict/Cargo.toml --all-targets -- -D warnings` (the workspace clippy can't reach this crate), then
  - `cargo run --manifest-path bench/dict/Cargo.toml -- selftest`.

  Both are **GATING** (the step has no `continue-on-error`). The selftest is in-process, dataset-free, and makes no wall-clock claim, so it is deterministic in CI — no quiet box needed.
- `bench/dict/README.md` — the "Selftest (CI-runnable)" section now names the lane that runs it, so the doc claim is accurate and durable.

## Why this lane

The selftest needs a Rust toolchain, so the Python-only `flow-on-gates.yml` (G1/G2/G3/G6) is the wrong home; G3 (`check-new-bench-registered.py`) is about registry/dashboard registration, not running a harness. The `lint` job is the cheapest correct fit — toolchain + warm dep cache already present, and it is a required gate via `ci-summary`.

## Honesty / scope notes

- `bench/dict` is a separate workspace (own `target/`), so `sparq-core` recompiles once more here in **debug**; the registry/git deps come warm from `rust-cache`, so the added cost is small. The CI comment states this plainly rather than claiming a free reuse.
[...]

PR #712, opened 2026-06-19
ci(supply-chain): make cargo-vet GATING leg deterministic (--frozen, no remote re-fetch)

jeswr:
> 🤖 SPARQ agent

## What

Makes the `cargo-vet (… GATING)` leg of `supply-chain.yml` **deterministic** by
verifying against the **committed** `supply-chain/imports.lock` instead of
re-fetching the remote audit aggregations on every run.

Before: `cargo vet --locked` — still re-pulls the 6 upstream `[imports.*]` audit
sets (Mozilla / Google / Bytecode-Alliance / ISRG / Embark / Zcash) and
re-aggregates them in memory each run.

After: `cargo fetch --locked` (populate the cargo cache) then
`cargo vet --locked --frozen` — `--frozen` (`Avoid the network entirely,
requiring … the cargo cache is populated …`) verifies the **checked-in** evidence
with no run-time remote re-fetch.

## Why (root cause from real logs)

During triage of PRs #693/#704 the cargo-vet leg failed with:

```
ERROR × Failed to parse toml file
  ╰─▶ missing field `user-id` for key `publisher.wasm-encoder` at line 2352
```

The **committed** `imports.lock` is valid and byte-identical between `main` and
the PR branch (its `publisher.wasm-encoder` is at line 59 and well-formed). The
error points at a *runtime-regenerated* lock (line ~2352) — i.e. cargo-vet
re-fetched a transiently malformed `publisher.wasm-encoder` block (no `user-id`)
from upstream. #704 (a no-op `taiki-e/install-action` SHA bump that touches no
`Cargo.lock`) hit this **4 consecutive runs** while a concurrent `main`
supply-chain run was **green on the identical tree**. Same cargo-vet 0.10.0,
same committed evidence — a live-network flake on a **GATING** check. Tracked as
bead `sq-miwt`.

## Honesty / scope

- The attestation is **not weakened**: a genuinely new/unaudited crate still
  FAILS (the committed `imports.lock` won't cover it); the ratchet is unchanged.
- Imports are refreshed **deliberately** (re-run `cargo vet` without `--frozen`)
  when updating the checked-in `supply-chain/` evidence — not implicitly inside
  the gate.

## Validation

- Reproduced locally (cargo-vet 0.10.2):
  `cargo fetch --locked && cargo vet --locked --frozen` on a **fresh** vet cache
  → `Vetting Succeeded`, exit 0.
- YAML valid (`yaml.safe_load`); `actionlint 1.7.12` clean.
- Gate leg name unchanged (no `advisory`/`informational` word) → `ci-summary`
  still treats it as GATING; the aggregator is intact.

Closes the recurring cargo-vet gate flake; unblocks PRs #693 (already merged
post-fix-of-a-different-leg) / #700 / #704.

### DIFF
--- a/.github/workflows/supply-chain.yml
+++ b/.github/workflows/supply-chain.yml
@@ -137,9 +137,27 @@ jobs:
         uses: taiki-e/install-action@7a79fe8c3a13344501c80d99cae481c1c9085912 # v2.81.10
         with:
           tool: cargo-vet
-      # `--locked` so vet evaluates the committed Cargo.lock, not a freshly-resolved tree.
+      # [OPUS-4.8] Populate the cargo cache so the subsequent `--frozen` vet can run
+      # fully offline (its precondition is "the cargo cache is populated or deps are
+      # vendored"). `--locked` keeps it to the committed Cargo.lock.
+      - name: Populate cargo cache (for offline vet)
+        run: cargo fetch --locked
+      # [OPUS-4.8] `--locked --frozen`: verify EVERY crate against the COMMITTED
+      # supply-chain/imports.lock + config.toml exemptions WITHOUT re-fetching the
+      # remote `[imports.*]` audit aggregations at run time. Plain `cargo vet --locked`
+      # still re-pulls those 6 upstream audit sets (Mozilla/Google/Bytecode-Alliance/
+      # ISRG/Embark/Zcash) and re-aggregates them in memory on every run; a transient
+      # bad/partial upstream response (observed: a `publisher.wasm-encoder` block fetched
+      # without its `user-id` field -> "missing field `user-id`" toml parse error) then
+      # reds this GATING leg and the whole ci-summary gate on an unrelated PR — while a
+      # concurrent run on `main` that sampled good data passes (seen on PRs #693/#704,
+      # bead sq-miwt). `--frozen` makes the attestation deterministic: it proves the
+      # ratchet against the checked-in evidence, and a genuinely-new/unaudited crate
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent — validation note

Local reproduction passes (`cargo fetch --locked && cargo vet --locked --frozen`
on a fresh vet cache → `Vetting Succeeded`, exit 0) with cargo-vet 0.10.2. CI
installs cargo-vet 0.10.0 via `taiki-e/install-action`; `--frozen` + `--locked`
are stable flags present since well before 0.10.0, so the behavior should match —
**validate on this PR's first CI run**. Auto-merge intentionally **not** armed by
me (left to the maintainer).

jeswr:
> 🤖 SPARQ agent — correction

The rationale in this PR (transient remote-imports re-fetch flake) was **wrong**.
The real cause of the cargo-vet GATING failure on PR #704 is a **cargo-vet
version** mismatch: the committed `imports.lock` uses `trusted-publisher`, parsed
only by cargo-vet **0.10.2+** (CHANGELOG #671); install-action **v2.82.0** (PR
#704's bump) installs **0.10.0**, which rejects it, whereas **v2.81.10** fell
back to cargo-binstall → 0.10.2. `--frozen` does NOT fix this (the parse fails
while LOADING the committed lock, before any network step) — proven by #704's
post-merge run still failing with `--frozen` + 0.10.0.

The `--frozen` change here is harmless (kept as determinism), but the **actual
fix** is pinning `tool: cargo-vet@0.10.2` — opened as **#713**. Apologies for the
incorrect first diagnosis.

PR #737, opened 2026-06-19
feat(sparq-introspect): export characteristic sets as SHACL node shapes (sq-bde)

jeswr:
> 🤖 SPARQ agent

Implements **sq-bde** — *Introspect: export characteristic sets as SHACL node shapes* (the last open `sparq-introspect` schema-export TODO).

## What

Adds `Introspection::to_shacl(&self) -> String`, rendering each mined characteristic set as a W3C [SHACL](https://www.w3.org/TR/shacl/) `sh:NodeShape` (N-Triples, which is valid Turtle — no serializer dependency, same as the existing `to_void` family). A characteristic set (Neumann & Moerkotte: the exact predicate set a group of subjects emits, with per-predicate multiplicity) maps almost one-to-one onto a node shape: *"an entity of this kind carries exactly these properties, this many times."*

## Mapping — and why each constraint is sound

Every constraint holds for **every** subject in the set, by the definition of a characteristic set:

- `sh:targetClass <C>` — emitted only for a class **universal** to the set (its in-set subject count equals the set's total subject count), so the shape auto-applies only where every targeted instance genuinely has this exact predicate set. A set whose subjects share no universal class (untyped / heterogeneous) yields a reusable but **target-less** shape.
- `sh:property` → `sh:PropertyShape` with `sh:path <p>` + `sh:minCount 1` per **non-type** predicate — a subject is in the set iff it emits exactly these predicates, so each occurs at least once.
- `sh:maxCount 1` **only** when `predicate_triples[i] == subjects` (avg multiplicity exactly 1). A larger sum means some subject is multi-valued → no upper bound is emitted (never an unsound constraint).
- `rdf:type` is expressed via `sh:targetClass`, so it is **not** repeated as a property shape.
- `scs:subjects` provenance triple records the support behind each shape.

The constraints are mined purely from what the data asserts, so the current graph already validates against them — a data-grounded **effective-schema floor** to edit, not an aspirational contract.

Example (alice/bob share `{type,name}`; carol adds `worksAt` twice):

```
_:shape0 a sh:NodeShape ; scs:subjects 2 ; sh:targetClass foaf:Person ; sh:property _:shape0_0 .
_:shape0_0 a sh:PropertyShape ; sh:path foaf:name ; sh:minCount 1 ; sh:maxCount 1 .
_:shape1 a sh:NodeShape ; scs:subjects 1 ; sh:targetClass foaf:Person ;
         sh:property [ sh:path :worksAt ; sh:minCount 1 ] ,        # multi-valued → no maxCount
                     [ sh:path foaf:name ; sh:minCount 1 ; sh:maxCount 1 ] .
```

## Tests & gate

- 2 new `src/lib.rs` unit tests: full multi-set mapping (universal-class targeting, no minority targets, `rdf:type` never a path, min/maxCount soundness on single- vs multi-valued predicates, `scs:subjects` provenance, valid-RDF re-parse via `oxttl`) + empty-graph edge case. One wasm32 smoke assertion in `tests/wasm.rs`.
- Gate green: `cargo build` + `clippy --all-targets -D warnings`, full crate tests (22 unit + 1 integration), `wasm32-unknown-unknown` build, both workspace rustdoc gates (`-D warnings`, default **and** `--all-features`), privacy-claims gate, G2 api→SKILL gate, no-perf-numbers + skill-frontmatter checks.
- README + `genai-retrieval` SKILL.md updated for the new public API.

### DIFF
--- a/crates/sparq-introspect/README.md
+++ b/crates/sparq-introspect/README.md
@@ -26,6 +26,7 @@ ix.to_json()                                  // pretty JSON — the machine sur
 ix.to_text_summary(2500)                      // prompt-ready digest, most-important-first, ≤ N chars
 ix.to_void("http://ex.org/dataset")           // W3C VoID description, as N-Triples (valid Turtle)
 ix.to_void_with_cs("http://ex.org/dataset")   // VoID + characteristic-set source stats (scs: ext)
+ix.to_shacl()                                 // characteristic sets → W3C SHACL node shapes
 ix.schema_summary_for(&seeds, 2500)           // retrieval-mode: schema scoped to seed IRIs
 
 // Persisted *.introspect sidecar — mine once, summarise forever without rescanning:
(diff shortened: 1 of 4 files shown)


jeswr:
> 🤖 SPARQ agent

Honesty fix pushed (`d316f108`). The crate README claimed **24 unit tests**; the real count on this branch is **22** (main has 20 `#[test]` in `src/lib.rs`, this PR adds exactly 2 SHACL node-shape tests).

Confirmed independently on the branch:
- `grep -c '#[test]' crates/sparq-introspect/src/lib.rs` → **22**
- `cargo test -p sparq-introspect` → `test result: ok. 22 passed; 0 failed` (plus 1 integration test in `tests/olympics.rs`)

Fix: `crates/sparq-introspect/README.md` line 123, `24` → `22` (test-count line only). The PR body already states the correct **22 unit + 1 integration** (the original "20→24" claim was already reconciled), so no body edit was needed. No implementation changes — the SHACL node-shape export is unchanged.

PR #841, opened 2026-06-19
test(sparq-server): feature-composition (audit-log + access-audit) + sequenced durable-failure (sq-bif.14) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — automated test-coverage work for bead **sq-bif.14** (sparq-server). Tests only; no production `src/` change.

Closes the two bounded coverage gaps the `sq-bif` assessment flagged for `sparq-server`. **`crates/sparq-server/tests/` only** — `src/http.rs` and the rest of `src/` are untouched (the brief's hot-merge-conflict scope guard).

## 1. Feature COMPOSITION — `audit-log` + `access-audit` together
Every opt-in feature was tested in isolation; no test enabled a PAIR together. New `tests/audit_composition.rs` (gated `#![cfg(all(feature = "audit-log", feature = "access-audit"))]`) builds ONE server with BOTH audit sinks compiled in and turned on at runtime. They hook the **same** request handler — one a process-global `tracing` sink, one a per-`AppState` JSON-Lines file sink — so a single request emits BOTH records.

Load-bearing invariant asserted (for an ALLOWED and a DENIED read): the two independently-derived records **agree** on operation, the *actually-enforced* decision, HTTP status, and the NON-reversible query fingerprint (both sinks share the FNV-1a-over-trimmed-text fingerprint), and the redaction boundary holds in **both** (never the raw query text or Bearer token; the fingerprint never leaks to the HTTP body). A router/init-ordering or shared-state regression would pass both single-feature legs (`tests/audit_log.rs`, `tests/access_audit.rs`) yet fail here.

## 2. SEQUENCED durable-write failure injection
The existing `test-seams` graceful-degradation tests fail a *contiguous prefix* of seals (a transient burst that clears, or a permanent jam). New test in `tests/persist.rs` drives an **interleaved** recover→fail→recover sequence through the same seal seam (`AppState::with_config_inject_durable_failure`) and asserts the fail-closed contract at **every** step: a refused write returns `503` and is never published, a recovered write commits `204`, the writer survives the whole sequence, reads keep serving the last published snapshot, and the final durable set is **exactly** the successful writes (no refused write resurrects) across a restart.

## Non-vacuity
Both tests have real assertions that fail on a regression. The sequenced test was confirmed to FAIL when the injected failures are removed (the `w1`/`w3` `503` expectations break).

## Gates — GREEN in BOTH feature states
- **default (features OFF):** `cargo clippy -p sparq-server --all-targets -- -D warnings` ✅ · `cargo test -p sparq-server` ✅ (the composition file compiles to zero tests, correctly inert)
- **ON:** `cargo clippy -p sparq-server --all-targets --features test-seams,audit-log,access-audit -- -D warnings` ✅ · `cargo test -p sparq-server --features test-seams,audit-log,access-audit` ✅

Docs updated: crate `README.md` + `skills/http-server/SKILL.md`.

## Honesty
The redaction assertions verify the documented #241 info-leak posture (query text + token never reach either sink), **not** an unqualified security guarantee — the audit trail records identities + resource IRIs **by design**. No ZK/MPC claim is involved (this crate has none). Work-box timings are non-canonical and not baked into any assertion.


jeswr:
> 🤖 SPARQ agent [OPUS-4.8]

Refreshed this branch against `origin/main` (clean `git merge origin/main`, no conflicts). It was stale-behind-main, which made the **MONOTONICITY** step of the coverage ratchet gate fail — the branch's `bench/coverage-floor.json` predated floors that merged to main (sparq-policy, sparq-algos/sparq-canon, etc.), so the monotonic check read main's newer floors as "removed". After merging main in, `bench/coverage-floor.json` is now main's verbatim (this is a tests-only branch that added no floors → union = main's floor).

Verified locally: `python3 scripts/coverage-gate.py --check-monotonic --base-ref origin/main` → **PASS (exit 0)**. The PR's deliverable is intact (`crates/sparq-server/tests/audit_composition.rs` + `persist.rs` still present in the diff vs main). CI is refreshed.

Issue #805, opened 2026-06-19
[feature]: support "pretty turtle" serialisation on website


jeswr:
With prefixes and grouping results so that `;` and `,` can be used to represent triples with shared subjects and objects.

PR #861, opened 2026-06-19
feat(sparq-engine): engine-side pretty Turtle serialiser (sq-ixc3.2, long-term home for #805) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — bead sq-ixc3.2 (parent epic sq-ixc3: cross-platform GUI).

## What this adds

An engine-side **pretty / formatted Turtle (+ TriG) serialiser** behind the existing opt-in **`serialize-rdf`** feature — the long-term engine home for the site-side `prettyTurtle` reshaper shipped in #805 (sq-gb4o, `packages/sparq-client/src/pretty-turtle.ts`). The engine now produces idiomatic, deterministic Turtle so the CLI / server / wasm can all get it for free and the site can later drop its TS formatter.

### Honest scoping note on the premise
The engine **already** had a `write_turtle` / `graph_to_turtle` (shipped under sq-678h) with `@prefix` compaction, predicate-object lists, `a` for `rdf:type`, blank nodes and RDF 1.2 triple terms. That writer is **not** "pretty" in the site's sense: it preserves the store's `iter_ids()` emission order (thread-count-dependent) and uses a fixed 4-space inline shape. This PR adds a **distinct, pretty** surface alongside it — it does not rewrite the existing one.

## API (`sparq_engine::serialize`, feature-gated)

- `PrettyOptions { indent /* default 2 spaces */, abbreviate /* default true */ }`
- `write_turtle_pretty(&[Triple], &Prefixes, &PrettyOptions) -> String`
- `write_trig_pretty(&[NamedGraph], &Prefixes, &PrettyOptions) -> String`
- `graph_to_turtle_pretty(&Graph)` / `graph_to_turtle_pretty_with(&Graph, &Prefixes, &PrettyOptions)` / `graph_to_trig_pretty(&Graph)`

### How it differs from the flat `write_turtle`
1. **Emission-order-independent**: subjects, predicates and objects are sorted by their canonical N-Triples spelling; `rdf:type` (`a`) sorts first in each block. The same triple **set** always renders to the same bytes.
2. **Output shape == the site's `prettyTurtle`**: prefix-alphabetical used-only `@prefix` header, blank-line-separated subject blocks, subject on its own line, predicate lines at one indent (`;`-joined, `.`-terminated), object lists wrapped onto continuation lines (`,\n` + two indents). The byte shape is asserted by a golden test.

Reuses the existing oxrdf term/escaping helpers — literals (datatype/lang, implicit `xsd:string` dropped), blank nodes, and RDF 1.2 triple terms `<<( s p o )>>` reproduce **losslessly**.

### CLI (low-friction exposure, also gated)
`sparq-cli dump` gains `turtle-pretty` / `trig-pretty` out-formats.

## Feature gate
Everything lives inside the already-`#[cfg(feature = "serialize-rdf")]`-gated `serialize` module — **OFF by default**, **zero new dependencies**, core stays lean.

## Round-trip + golden verification (tests, feature on)
- `pretty_golden_output_matches_site_shape` — byte-exact golden matching the site shape.
- `pretty_round_trips_all_term_kinds` — re-parses to the same triple set (literals/lang/datatype, blank nodes, triple terms, escaped IRIs).
- `pretty_ordering_is_emission_independent` — two differently-ordered inputs → identical bytes; `a` first.
- `pretty_no_abbreviate_keeps_full_iris_and_no_header`, `pretty_custom_indent`, `pretty_empty_graph_is_empty_string`, `pretty_trig_round_trips_named_graphs`.

- `cargo clippy -p sparq-engine --all-targets -- -D warnings` (default, feature OFF) ✅
- `cargo clippy -p sparq-engine --all-targets --features serialize-rdf -- -D warnings` ✅
- `cargo test -p sparq-engine` (default — no new surface compiles) ✅
- `cargo test -p sparq-engine --features serialize-rdf` (238 lib tests incl. the 7 new) ✅
- `cargo clippy/test -p sparq-cli` default + `--features serialize-rdf` ✅
- End-to-end smoke: `sparq-cli dump … turtle-pretty` then re-load → identical triple set.

## Fidelity limits (honest)
- `graph_to_turtle_pretty` abbreviates against `default_prefixes()` only — a namespace not in that set stays a full `<…>` IRI (use `graph_to_turtle_pretty_with` to pass the query's declared prefixes, as the site does). This matches the existing `graph_to_turtle` behavior.
- Blank-node labels are preserved verbatim (no canonicalisation) — round-trip is set-equality with stable labels, same as the other writers.
- The pretty writer materialises + sorts all triples (it is a formatter for human-facing output, not a streaming dump path).

## Deferred follow-ups (beaded)
- **sq-fe1s** (P2): wire pretty Turtle into the **wasm `Store.serialize`** surface + regen bindings — deferred here because wasm-pack regen + committing generated artifacts is heavy. The Rust API + CLI landed; this PR exposes no new wasm surface.
- **sq-lxej** (P3, blocked on sq-fe1s): switch the **site** over to the engine serialiser and drop the TS reshaper.
- **sq-pxdu** (P3): decide/scope **JSON-LD pretty/compaction** (explicitly out of scope here).


jeswr:
> 🤖 SPARQ agent

Diagnosed the persistent **"Inference conformance"** failure on this PR (sq-y5dz). It is **not** a problem with this PR's content and **not** stale-vs-main — this branch is byte-identical to `main` on the fetch scripts and `ci.yml`.

Root cause: the **"Fetch inference test suites (pinned)"** step downloads the OWL 2 test-case export from the Internet Archive, which reset the transfer mid-stream:
```
curl: (56) Recv failure: Connection reset by peer
```
The old `curl --retry 3` does not retry on a (56) mid-stream reset, so it red-gated every engine PR. The archive URL is reachable and the pinned sha256 still matches, so the source is fine — it's transient.

Fix is in **#864** (`scripts/fetch-inference-suites.sh`: add `--retry-all-errors` etc.). Once #864 lands on `main`, re-run the Inference conformance check here and it should pass. The pretty-Turtle serialiser deliverable on this branch is untouched by the fix.

PR #926, opened 2026-06-19
feat(sparq-vectors): persist the delta sidecar to disk (.spqd, durable add/remove/update) (sq-7e50) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — opened by an autonomous sparq agent (Opus 4.8) on @jeswr's behalf. One of several agents that may run on this account.

## What

Follow-up to **sq-pi44** (PR #880, the IN-RAM delta sidecar). The in-RAM delta was lost when the `VectorStore` handle dropped, so `compact` was the only durability path. This adds a **persisted on-disk delta sidecar** (`.spqd`) so incremental `add`/`remove`/`update` against a finalized `.spqv` base survive a process restart **without** a compact.

All strictly additive under the existing opt-in **`delta`** cargo feature (OFF by default) — **no new dependency** (reuses std + the in-tree `rustc-hash`), so the default `sparq-vectors` build is unchanged.

## On-disk format (`.spqd`, version 1, little-endian — mirrors the `.spqv` header discipline)

```text
offset 0   magic         b"SPQD"                  4 bytes
offset 4   version       u32 = 1                  4 bytes
offset 8   dim           u32                      4 bytes
offset 12  append_count  u64                      8 bytes
offset 20  tomb_count    u64                      8 bytes
offset 28  reserved      zero padding             4 bytes
offset 32  fingerprint   base graph fingerprint   24 bytes   (the sq-32i5 generation tie)
offset 56  appends       [append_count × (id:u32, dim × f32)]
           tombstones    [tomb_count × (id:u32)]
```

The header carries the **base graph fingerprint**, so a sidecar written against generation N is validated/rejected against a base of generation M ≠ N on open (routed through the existing `apply_delta` generation tie). The in-RAM `VectorDelta` is reused verbatim as the in-memory representation.

## API (all `delta`-gated)

- `VectorStore::save_delta` / `save_delta_to` — **crash-durable**: write to a sibling `.spqd-tmp`, `fsync` (`sync_all`), then atomic `rename` — the same write-tmp-then-rename discipline as `StreamingWriter`. An unstarted delta persists an empty, generation-bound sidecar.
- `VectorStore::open_with_delta` / `open_with_delta_at` — open base mmap + replay the persisted delta (through `apply_delta`'s generation tie + a `dim` check). No sidecar ⇒ the bare base.
- `VectorStore::sibling_delta_path` / `has_persisted_delta`; `SPQD_MAGIC` / `SPQD_VERSION`.
- `VectorDelta::{to_bytes,from_bytes}` (pub(crate)) — `from_bytes` is **fail-closed**: bad magic/version, zero dim, an **exact-length partial-write/truncation guard** (checked arithmetic against the header counts — short OR long rejected), non-finite components, and duplicate/conflicting-id rejection.

## Tests (`tests/delta_persist.rs` + serde unit tests in `delta.rs`)

- **Result-equivalence (REQUIRED):** `open_with_delta(base + persisted-delta)` == the in-RAM `(base + delta)` view == `compact()` (sorted effective sets + `len` + per-id `get` + a search agree).
- **Crash / partial-write:** a truncated mid-body / torn-header / trailing-garbage / bad-magic sidecar is detected and **rejected, not UB**.
- **Base-fingerprint-mismatch:** a gen-A sidecar replayed onto a gen-B base errors (generation mismatch); the same sidecar replays cleanly onto its own generation.
- Plus: dim-mismatch rejection, atomic-replace (no stray tmp), empty-delta round-trip, and serde unit tests (None-generation round-trip, byte-determinism regardless of map order, sub-header rejection).

## Gates — green in BOTH feature states

| gate | default (feature OFF) | `--features delta` |
|---|---|---|
| `cargo clippy -p sparq-vectors --all-targets -- -D warnings` | ✅ | ✅ |
| `cargo test -p sparq-vectors` | ✅ | ✅ (7 new integration + 7 new serde unit) |

README kept under the ≤120-line template cap (passes `check-readme-template.py`, 0 deviations); typos gate clean. No hard-coded performance numbers.

Closes **sq-7e50**.

### DIFF
--- a/crates/sparq-vectors/README.md
+++ b/crates/sparq-vectors/README.md
@@ -75,9 +75,10 @@ let _neighbours = nearest_term_exact(&store, &graph, &some_term, 10);
   and reopen **that** (`Graph::open`, frozen id order); **never** re-parse the source RDF
   (pinned in `tests/staleness_contract.rs`).
 - **Incremental add / remove / update (opt-in `delta`)** — a `.spqv` is build-once-immutable; the
-  `delta` feature adds an **in-RAM delta sidecar** (`add`/`remove`/`update` → append map + tombstones)
-  that `get`/`iter`/search transparently union (honouring tombstones), generation-tied to the base
-  (sq-32i5). `compact` folds it into a from-scratch-equivalent base. **In-RAM only** (lost on drop; `compact` is the durability path); a persisted delta is a tracked follow-up. No new dep. See rustdoc/SKILL.
+  `delta` feature adds a **delta sidecar** (`add`/`remove`/`update` → append map + tombstones) that
+  `get`/`iter`/search transparently union (honouring tombstones), generation-tied to the base
+  (sq-32i5); `compact` folds it into a from-scratch-equivalent base. `save_delta`/`open_with_delta`
+  persist+replay it to a crash-durable `.spqd` (tmp+fsync+rename, partial-write & mismatched-base detected) so mutations survive a restart without a compact. No new dep. See rustdoc/SKILL.
 - **Verbalization, quantization, hybrid fusion** — `verbalize` / `embed_entities` render
   per-entity text (multilingual, char-budgeted, single named graph at a time);
   `ScalarQuantizer` (4×) and `ProductQuantizer` for large stores; `fuse_rrf` /
--- a/crates/sparq-vectors/src/delta.rs
+++ b/crates/sparq-vectors/src/delta.rs
@@ -85,7 +85,7 @@
 //! `.spqd-tmp`, `fsync`s it, then atomically `rename`s it over the final path — the same
 //! write-tmp-then-rename + `sync_all` discipline as [`StreamingWriter`](crate::store::StreamingWriter):
 //! a crash mid-write leaves the previous `.spqd` (or none) intact, never a half-written sidecar at
-//! the live path. **Partial-write detection.** [`from_bytes`](VectorDelta::from_bytes) recomputes
+//! the live path. **Partial-write detection.** `from_bytes` (the crate-internal decoder) recomputes
 //! the EXACT expected length from the header (`56 + append_count·(4 + dim·4) + tomb_count·4`) and
 //! rejects any file that is shorter OR longer — a truncated (or trailing-garbage) sidecar is a
 //! descriptive `Err`, never an out-of-bounds read or a silently-short replay.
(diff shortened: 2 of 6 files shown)


jeswr:
> 🤖 **SPARQ agent** — gate-unblock for `clippy (gate) + fmt (non-blocking)`

**Root cause (not clippy — rustdoc).** The check whose display name is `clippy (gate) + fmt (non-blocking)` failed in ~1m, well before a full workspace clippy would even finish. The workspace clippy step (`cargo clippy --workspace --all-targets -- -D warnings`) was actually **clean**; the failure was the **rustdoc** step bundled into the same `lint` job:

```
cargo doc --workspace --no-deps --all-features   (RUSTDOCFLAGS="-D warnings")
```

It hit two intra-doc-link errors in the new `delta`-gated code — surfaced **only** under `--all-features` (the default build compiles no delta code, which is why per-crate / default clippy passed locally). So it is feature-related, just on the rustdoc surface rather than a clippy lint:

1. `crates/sparq-vectors/src/delta.rs:88` — module doc linked ``[`from_bytes`](VectorDelta::from_bytes)`` but `from_bytes` is `pub(crate)` → `rustdoc::private-intra-doc-links`.
2. `crates/sparq-vectors/src/store.rs:777` — ``[`save_delta_to`]`` was unqualified and did not resolve to the associated method → `rustdoc::broken-intra-doc-links`.

**Minimal fix (doc-text only, 2 lines, no API/behaviour change).**
- delta.rs: render the private decoder as plain inline code ``` `from_bytes` (the crate-internal decoder) ``` — matching how the rest of that module already refers to the `pub(crate)` `from_bytes`/`to_bytes`.
- store.rs: qualify the link as ``[`save_delta_to`](Self::save_delta_to)`` — matching the `Self::`-qualified links used throughout the same doc block.

Feature design untouched; no `#![allow]` added.

**Gates re-run green (this work box):**
- `cargo doc --workspace --no-deps --all-features` (`RUSTDOCFLAGS=-D warnings`) → exit 0 (the failing step)
- `cargo doc --workspace --no-deps` (default features, `-D warnings`) → exit 0
- `cargo clippy --workspace --all-targets -- -D warnings` → clean
- `cargo clippy -p sparq-vectors --all-targets --features delta -- -D warnings` → clean
- `cargo test -p sparq-vectors --features delta` → all green (incl. `delta` + the new `delta_persist` suites)

Pushed `dd7668af` onto the existing branch (fast-forward from `8d588026`). `fmt` shows pre-existing whole-crate drift from a rustfmt version skew that predates `8d588026` and is unrelated to this PR — left untouched (fmt is non-blocking, and reformatting 270+ unrelated lines would be churn). Auto-merge left OFF for the orchestrator to arm after verification.

Issue #903, opened 2026-06-19
Port the federated adaptive re-planner to the LOCAL evaluator


jeswr:
> 🤖 **SPARQ agent** (one of several @jeswr runs on this account).
>
> Deep-research feature proposal for your review — one of 17 surfaced by a parallel research sweep (competitive parity · W3C/emerging specs · query-engine · ops · DX · reasoning/privacy). **No work has started; this is for your decision.** The honesty caveats in the body are load-bearing (ZK/MPC are NOT externally audited; work-box benchmarks are non-canonical; first-party authz is plaintext access control, not a privacy guarantee; the RDF-1.2 canon extension would be non-standard).

## Summary
Port the existing federated divergence-triggered re-planner + hysteresis (`crates/sparq-fedplan/src/adaptive.rs`) to the **local** evaluator: after each materialised join, compare actual vs estimated cardinality, and if a remaining-BGP estimate was off by > k x, re-order the **unexecuted** patterns. Opt-in feature mirroring the fed gate.

## Why / use-case
This is the cheapest robustness lever sparq has and it is currently sitting unused on the wrong side of the local/federated boundary. Because the local evaluator **fully materialises** each intermediate, the TRUE cardinality of every completed stage is known **for free** — the expensive part of adaptive query processing (getting actuals) is already paid. `research/optimization-techniques.md` explicitly recommends this as 'near-zero-cost' and it is not implemented for the local path. It directly attacks the BENCHMARKS.md 'gap widens with scale' symptom by making GOO mis-orders self-correcting.

## Fit with sparq (+ what partially exists today, honestly)
Strong — it is **mostly a port** of the existing fed `ReplanPolicy` + hysteresis to consume row counts the local evaluator already produces. No new cost model, no new operators. Verified: adaptive re-plan exists ONLY in sparq-fedplan (adaptive.rs); the local path in exec.rs commits to one GOO/DPccp order and runs to completion (no replan/divergence machinery in exec.rs for the local join loop).

## Prior art (engines/specs + refs)
- ReOpt mid-query re-optimization; order-robustness arXiv:2502.15181
- Adaptive query processing survey (Deshpande/Ives/Raman)
- sparq's own `sparq-fedplan/src/adaptive.rs`

## Sketch / options
- Extract the divergence/hysteresis policy from fedplan into a shared helper so both paths use one implementation.
- Trigger only between materialised stages on the remaining BGP; cap re-orders to avoid thrash (hysteresis already designed in the fed version).
[...]

PR #948, opened 2026-06-20
feat(sparq-engine): port adaptive divergence-triggered re-planner to the LOCAL evaluator (sq-p6p6) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — opening this PR for bead **sq-p6p6** (#903). Auto-merge intentionally OFF; the orchestrator/maintainer decides arming.

## What this implements

Ports `sparq-fedplan`'s `adaptive-replan` (`ReplanPolicy` divergence + hysteresis, `AdaptiveExecutor`) into the **local** greedy BGP join loop in `crates/sparq-engine/src/exec.rs::eval_bgp_binary`, behind the new opt-in **`adaptive-replan-local`** cargo feature (OFF by default).

The local path fully **materialises** every intermediate join, so the TRUE running cardinality — and the observed per-variable distinct-value count — is known for free at each stage boundary. When it diverges from the planner's independence-model estimate past the shared `ReplanPolicy::divergence_factor` (default 4×), the not-yet-joined patterns are re-picked with the observed selectivities substituted in (re-using the same `goo_pick` cost model — **no new cost model, no new operators**), adopting the corrected order only when it wins by the `improvement_margin` hysteresis band (anti-thrash).

It is a **port**, not a redesign.

## Result correctness — the gate (PASSES)

The output multiset is **identical to the static plan regardless of re-order** — a BGP answer is the order-independent natural join of its per-pattern multisets, and the adaptive path only re-orders the not-yet-joined suffix (the materialised prefix is carried across unchanged, and merge/hash/bind joins are all multiset-preserving).

Proved by the ported `replan_result_equals_static` oracle in `crates/sparq-engine/src/adaptive.rs`, which is **non-vacuous**: the fixture asserts the divergence trigger fires AND the corrected order is actually adopted (a real switch), then asserts the adaptive result multiset equals the static one (the static baseline obtained from the SAME feature-on binary via a bench/oracle thread-local toggle). Plus a no-divergence-is-inert test and an across-shapes equivalence test.

## Shared-module decision (CONDITIONAL — see below)

The divergence/hysteresis policy is factored into a new pure crate **`sparq-replan-policy`** (`publish = false`, `forbid(unsafe_code)`, zero deps, inlined `f64` arithmetic) so the local + federated re-planners cannot diverge on the rule. It is a **verbatim port** of fedplan's `diverges`/adopt logic.

I deliberately did **NOT** rewire `sparq-fedplan` onto the shared crate: its own richer `ReplanPolicy` (with the extra latency/EWMA knobs) is left untouched, so the federated path is **provably unaffected** (its 34 `adaptive-replan` tests stay green). The shared crate is consumed by the **local** engine only. Because the shared functions are verbatim ports, the two paths still cannot **semantically** diverge — committing the shared abstraction without touching fedplan's tested path is the low-risk way to satisfy "fed + local can't diverge" while honouring "keep per-path if the seam would regress either path."

## Benchmark + HONESTY — work-box-INDICATIVE, NON-CANONICAL

> These numbers are from this EC2 work box, which is **NON-CANONICAL**. They are **NOT** baked into any doc or test. **Do NOT arm this PR on a perf claim.**

A focused in-process micro-benchmark (`crates/sparq-engine/tests/adaptive_replan_bench.rs`, run with `--ignored --nocapture`) times the SAME query via the SAME `eval_bgp_binary`, adaptive-on vs forced-static, on a mis-estimated multi-pattern fixture (where a switch fires), a collapse fixture, and a correctly-estimated control:

- **Mis-estimated fan-out (1 switch):** ~0.93–0.96× — adaptive is **slightly SLOWER**.
- **Collapse (0 switches):** ~0.99× — the static planner is already optimal; no trigger.
- **Correctly-estimated control (inert):** ~1.00× — **no degradation** when the re-planner doesn't fire.

**Honest finding:** the port is correctness-complete, but it shows **no perf win** on this work box, and a small regression when it fires. Root cause: the LOCAL planner's per-pattern cardinality estimates are **exact** (index-range counts), so it is already well-ordered; the only divergence the reorder can exploit (the running-ndv cap mis-estimate) flips two **multiplicative** arms, which does not reduce total materialised work — it only adds the re-plan overhead. (This matches the bead's own SCOPE/NO-EFFECT note.)

**Per the bead's HARD benchmark gate, the perf gate is UNMET.** The shared-module no-degradation result IS satisfied (control ~1.0×, and fedplan is untouched), but the perf-win premise is not. The orchestrator/maintainer should decide whether to land a correct, opt-in, **perf-neutral** port (off by default, zero overhead when off, harmless when on) and settle the perf question on a canonical host / via a redesign, or defer. I am NOT claiming a speedup.

- `cargo build` + `cargo clippy --all-targets -D warnings` + `cargo test`:
  - **default (feature OFF):** engine 199 tests pass; static planner is byte-identical (zero adaptive code compiled in; `sparq-replan-policy` not pulled in).
  - **feature ON (`adaptive-replan-local`):** engine 209 tests pass (incl. the oracle).
  - `sparq-replan-policy`: 4 tests pass. `sparq-fedplan --features adaptive-replan`: 34 tests pass (unchanged).
- README kept ≤120 lines; `skills/sparql-query/SKILL.md` updated. Both document the HONEST status.


jeswr:
> 🤖 **SPARQ agent** (one of several @jeswr runs on this account).

**Closing as a measured negative result — not landing.** The port is correctness-complete (the `replan_result_equals_static` oracle is non-vacuous: it fires a real divergence-triggered switch then asserts multiset-equality to the static plan; opt-in, default byte-identical, both feature states green). **But the benchmark gate is unmet:** mis-estimated fan-out ~0.93–0.96× (slight regression), inert control ~1.00×, no win anywhere — all **work-box-indicative / NON-canonical**, not baked anywhere.

**Why it structurally can't win on the local path:** the local planner's per-pattern cardinality estimates are **exact** (index-range counts), so the greedy GOO order is already well-chosen. The only divergence a runtime reorder can exploit flips two *multiplicative* join arms, which does not reduce total materialised work. This is the opposite of the federated case (where estimates are genuinely uncertain) — so the fedplan port doesn't transfer a win.

Your #903 greenlight was conditioned on "make sure this does improve performance in the main module"; it doesn't, so per that condition I'm not merging a maintained opt-in feature + a new `sparq-replan-policy` crate for zero benefit. The work + this negative finding are preserved in the closed PR + bead **sq-p6p6**.

**If a real local win is wanted**, the candidate is a *pruning-aware* reorder (push observed cardinality into the bind-join eligibility threshold) — NOT a divergence-triggered arm-swap. Beaded as a fresh redesign. Reopen this if you'd prefer to land the perf-neutral opt-in anyway.

PR #1239, opened 2026-06-23
test(sparq-wasm): correctness coverage (sq-bif)

jeswr:
> 🤖 SPARQ agent (Opus 4.8). Advancing the P1 umbrella **sq-bif** (workspace correctness-test coverage sweep) for crate **`sparq-wasm`** — the published `@jeswr/sparq` browser/WASM bundle. Test-only; the umbrella stays OPEN.

## What this adds

`sparq-wasm` was already exceptionally well-covered: every public `Store` method is exercised natively in `tests/exported_api.rs` and end-to-end in a real wasm32 runtime in `tests/web.rs`, plus per-module `src` tests. So rather than pad, this fills the **4 genuine remaining gaps** in the EXISTING native target `tests/exported_api.rs` — no new cargo feature / `cfg`-gate, no wasm-bindgen harness change:

1. **`SolutionCursor` partitioning with `batch_size > 1` over MULTIPLE batches** incl. a trailing **partial** batch (5 rows / batch 2 → `[2, 2, 1]`). The prior cursor tests only used `batch_size 1` (uniform single-row steps) or an oversized single batch, so the multi-row `end`/`pos` advancement and the ordered, no-gap/no-dup reassembly were unpinned.
2. **The EXACT-MULTIPLE boundary** (4 rows / batch 2 → two full batches, **no** spurious trailing empty batch) — the precise case the exhaustion guard `pos == total && total != 0` is written for; a naive `pos < total` loop would emit a phantom empty batch.
3. **`DESCRIBE` through the exported `Store::queryQuads`** — the other graph-valued form (prior quad tests only covered `CONSTRUCT`). Asserts the Concise Bounded Description reaches the binding and an **inbound** subject is NOT pulled in.
4. **`queryQuadsChunks` at a NON-multiple boundary** (5 triples / batch 2 → `[2, 2, 1]`) with byte-for-byte reassembly — the existing test only used an even `batch_size`-1 split.

## Why these catch real regressions

Each cursor/chunk test is **mutation-validated**: it was confirmed to FAIL under a targeted source mutation and pass after revert —

| Mutation in `src/lib.rs` | Caught by |
|---|---|
| `end` off-by-one (`+ batch_size + 1`) | exact-multiple + partial-trailing |
| drop the `pos == total && total != 0` guard | exact-multiple (phantom empty batch) |
| quad-chunk batch `+ 1` | partial-tail |

These assert specific row counts, ordered values, and CBD direction — not "no panic".

## Gates (both feature states, run in-worktree)

- `cargo build` / `cargo test -p sparq-wasm` — default (features OFF): lib 19, `exported_api` 22 → **26**.
- `cargo test -p sparq-wasm --all-features` — lib 47, `exported_api` 26 → **28** (2 extra are the feature-gated shacl tests).
- `cargo clippy -p sparq-wasm --all-targets -- -D warnings` and `--all-features` — clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc -p sparq-wasm --no-deps --all-features` — clean.
- `rustfmt --check` clean on the changed file (the workspace's deferred-reformat diffs are only in untouched files; I did not reformat them).

Test-only, **159 insertions**, no public-API / README / `SKILL.md` change. Coverage not regressed — net-new assertions on previously-untested partition + CBD branches.

Auto-merge intentionally **OFF**.

### DIFF
--- a/crates/sparq-wasm/tests/exported_api.rs
+++ b/crates/sparq-wasm/tests/exported_api.rs
@@ -252,6 +252,101 @@ fn exported_query_cursor_empty() {
     );
 }
 
+/// [OPUS-4.8] sq-bif: cursor PARTITIONING with a `batch_size > 1` over MULTIPLE batches —
+/// the case the existing tests never reach (they use only batch_size 1, where every step
+/// advances by exactly one row, or an oversized single batch). A 5-row result with
+/// batch_size 2 must split into batches of `[2, 2, 1]` (the trailing PARTIAL batch via
+/// `min(pos + batch_size, total)`), the per-batch row counts must be exact, and the rows —
+/// in `ORDER BY` order — must reassemble with NO drop, duplication, or reorder. This pins
+/// the `end`/`pos` advancement against an off-by-one and drives the `pos == total`
+/// exhaustion branch from a step larger than one (the existing batch_size-1 test reaches
+/// it one row at a time; this reaches it after a partial trailing step).
+#[test]
+fn exported_query_cursor_partial_trailing_batch() {
+    let data = "@prefix ex: <http://ex/> .\n\
+        ex:a ex:n \"1\" . ex:b ex:n \"2\" . ex:c ex:n \"3\" . ex:d ex:n \"4\" . ex:e ex:n \"5\" .";
+    let store = Store::load(data, "turtle").unwrap();
+    let q = "PREFIX ex: <http://ex/> SELECT ?n WHERE { ?s ex:n ?n } ORDER BY ?n";
+    let mut cur = store.query_cursor(q, 2).unwrap();
+    assert_eq!(cur.row_count(), 5);
+    assert_eq!(cur.batch_size(), 2);
+
+    let b0 = cur.next().expect("batch 0");
+    let b1 = cur.next().expect("batch 1");
+    let b2 = cur.next().expect("batch 2 (partial)");
+    assert!(
+        cur.next().is_none(),
+        "exhausted after the trailing partial batch"
+    );
+
+    // Per-batch row counts: [2, 2, 1] — the partition is exact (the final batch is the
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent

Addressed all five review threads in 5cb90f17 (tests-only; all 26 `exported_api` tests pass, `clippy -D warnings` clean):

1. **SELECT cursor `value` substring matching (line ~309)** — valid fragility. Extraction now anchors on the `?n` binding key via a new `select_binding_values(batch, "n")` helper rather than scanning every `"value":"`, so a future second projected variable can't be conflated. I deliberately did NOT add `serde_json`: this crate's SPARQL-JSON is produced by its own hand-rolled, dependency-free serializer with a fixed key order (`src/serialize.rs` — "dependency-free … no serde_json"), so a dev-only JSON dep purely for this assertion would be disproportionate, and the key-anchored scan covers the only realistic fragility. Justification documented on the helper.

2. **DESCRIBE CBD negative assertion (line ~418)** — valid, tightened. Now asserts the SPECIFIC inbound triple is absent (`<http://ex/alice> <http://ex/knows> <http://ex/bob>`) instead of any mention of `<http://ex/alice>`, so the test reflects the real "no inbound edge" invariant and would still pass if alice legitimately appeared as the object of an outgoing triple.

3-5. **CONSTRUCT whole-vs-chunked byte-for-byte (lines ~433/440/455)** — valid determinism concern. CONSTRUCT/DESCRIBE graphs have set semantics and the engine makes no triple-emission-order guarantee (`construct::instantiate` is "first-production order", driven by `eval_select` row order), so comparing two separate executions byte-for-byte over-specifies the contract. Both comparisons now compare the SORTED non-empty line set (new `sorted_nt_lines` helper), pinning the load-bearing invariant (same triple set) while the chunk-count / chunk-size assertions continue to enforce the partition.

PR #1240, opened 2026-06-23
test(sparq-nlq): correctness coverage (sq-bif)

jeswr:
> 🤖 SPARQ agent [OPUS-4.8]

Advances P1 bead **sq-bif** (workspace correctness-coverage sweep) for crate **`sparq-nlq`** — the opt-in natural-language-to-SPARQL surface.

`sparq-nlq` was already exceptionally well-covered (the loop, the dictionary constraint, entity/relation linking, the exec-accuracy harness, provenance/citations/qualification, and the endpoint client all have real, end-to-end suites). Rather than pad, I found the GENUINE remaining gaps — documented-but-untested branches, asymmetric error/Result paths, and degenerate-config invariants — and added MEANINGFUL tests that each assert a SPECIFIC behaviour and would catch a real regression. All additions go into EXISTING test targets / EXISTING features; **no new cargo feature or cfg-gate** (no new feature-matrix fragment).

## Gaps filled (each pins a previously-unexercised branch)

- **`cite.rs`** (`citations`): **multi-source render** — a subject with two `prov:wasDerivedFrom` sources cites BOTH, with **document-wide continuing numbering** (the documented "a Finding derived from two sources cites both" branch of the per-source loop, which no single-source fixture reached); empty-records render (vacuous resolution `1.0`); marker-leads-footnote.
- **`provenance.rs`** (`citations`): `provenance_for` over a **2-source subject** (the OPTIONAL cross-product deduped into two distinct `SourceLink`s; single-valued quality axes read once, not duplicated per row); `lookup_query` embeds the subject in **every** axis + determinism.
- **`lib.rs`**: `extract_sparql` **empty/unclosed** fenced block → `None` (the `!q.is_empty()` and missing-`body_end` guards); case-insensitive bare-keyword gating; **`RecordingLlm` does NOT record failed calls** (documented invariant); `ReplayLlm` empty-fixture miss + malformed-fixture error path.
- **`eval.rs`**: `F1::score` **asymmetric one-side-empty** (predicted-empty vs non-empty gold, and vice versa → F1 `0`, not exact — the `p==0`/`g==0` ratio guards); `AnswerSet` **multiset** duplicate-row preservation + the greedy intersection merge over duplicates.
- **`qualify.rs`** (`citations`): `band_for` **inverted-thresholds collapse-Medium** (the documented degenerate-config invariant); abstention floor is **strict `<`** (confidence `== floor` does NOT abstain — locks against an accidental `<=`).
- **`endpoint.rs`** (`nlq-endpoint`): the `EndpointConfig` builder chain (`with_max_tokens`/`with_timeout`/`with_api_key`); `request_url` trailing-slash normalisation; `non_empty_env` unset-vs-empty.

- `cargo build` / `cargo test` — green **default** AND **`--all-features`** (also verified **`--features citations`** alone and **`--features nlq-endpoint`** alone). 118 individual tests pass under `--all-features`.
- `cargo clippy -p sparq-nlq --all-targets -- -D warnings` — clean (default + `--all-features`).
- Workspace rustdoc gate: `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — clean.
- `rustfmt --check` clean on every changed file.
- Test-only (392 insertions; no public API / README / SKILL change). Coverage not regressed — net-new assertions on previously-untested branches.

Auto-merge intentionally **OFF**. Umbrella **sq-bif** left **OPEN**.

### DIFF
--- a/crates/sparq-nlq/src/cite.rs
+++ b/crates/sparq-nlq/src/cite.rs
@@ -302,4 +302,82 @@ ex:bare a pkg:Finding ;
         assert!(block.contains("no source recorded"));
         assert!(block.contains("<http://ex/bare>"));
     }
+
+    /// A subject derived from TWO sources must cite BOTH, with the numbering continuing
+    /// across the two citations (not reset per subject) — the `render` contract: "one
+    /// `Citation` per distinct source link" / "a Finding derived from two sources cites
+    /// both". This is the multi-source branch of the `for link in &rec.sources` loop, which
+    /// no single-source fixture exercises. [OPUS-4.8]
+    #[test]
+    fn subject_with_two_sources_cites_both_with_continuing_numbers() {
+        let g = Graph::load_str(
+            r#"@prefix pkg:  <https://sparq.dev/ns/pkg#> .
+               @prefix prov: <http://www.w3.org/ns/prov#> .
+               @prefix ex:   <http://ex/> .
+               ex:multi a pkg:Finding ;
+                 prov:wasDerivedFrom ex:src-a , ex:src-b ;
+                 pkg:confidence "0.8" .
+               ex:src-a a pkg:Source .
+               ex:src-b a pkg:Source ."#,
+            "turtle",
+        )
+        .unwrap();
+        // Bind ONLY the multi-source finding so the citation count is exactly its sources.
+        let result = query_with_budget(
+            &g,
+            "PREFIX pkg: <https://sparq.dev/ns/pkg#> \
+             SELECT ?f WHERE { VALUES ?f { <http://ex/multi> } ?f a pkg:Finding }",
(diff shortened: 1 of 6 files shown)


jeswr:
> 🤖 SPARQ agent

Valid concern — fixed in bfc11ffa. The unit test now takes a `static ENV_LOCK: Mutex<()>` guard before mutating env, mirroring the established pattern in `tests/endpoint_stub.rs`. As you note, `std::env::{set_var,remove_var}` mutate process-global state and race any concurrent `getenv`/`setenv` from another thread in the same binary (UB even though the probe key is unique to this test), and cargo runs this binary's unit tests on a thread pool. The lock serializes the mutation against any other env-mutating test in the binary.

PR #1251, opened 2026-06-24
feat(fedplan): opt-in predicate-selectivity non-star join cardinality (sq-jsuzr) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — implementing bead **sq-jsuzr** (grounded improvement survey **§C2**, `research/codebase-improvement-opportunities-2026-06-23.md`).

## What

The non-star independence fallback in `sparq-fedplan`'s `independence_estimate` (plan.rs) approximates the join-key number-of-distinct-values (ndv) by the larger of the two joined leaf cardinalities — `out = l*r/max(|L|,|R|)` — and **ignores** the `distinct_subjects`/`distinct_objects` statistics every `SourceDescriptor` already carries (mined from served VoID property partitions).

This adds an **opt-in** `PlanOptions::use_predicate_selectivity` knob. When on, a connected **non-star** join folds the candidate's *actual* per-predicate distinct count into the join-key ndv:

- join variable in the candidate's **subject** position ⇒ `Σ distinct_subjects` over the candidate's retained sources;
- join variable in its **object** position ⇒ `Σ distinct_objects` (a multi-source union sums per source, mirroring the leaf-cardinality union);
- estimate becomes `out = |L|·|R| / Σ distinct(join-key)`, bounded below by 1 and above by the Cartesian product (monotone, exactly like the max-leaf path).

When the relevant distinct count is **absent** (VoID leaves it 0 when unknown), or the join key is in neither position, it falls back to the identical max-leaf heuristic — it never fabricates an ndv. **Star joins keep using the characteristic-set estimate regardless of the flag.**

## Correctness contract (load-bearing)

This is a **cost-estimate-only** knob. It changes the planner's cardinality estimate — and hence its join order and bind-vs-hash decision — but the **result multiset is byte-identical regardless of the knob**: BGP join is commutative/associative, so any complete order over the same patterns answers identically. The flag is **OFF by default**, so the prior estimate is preserved exactly for A/B comparison.

Tests added (drive the real `plan_bgp` / `cost_join` path):
- `predicate_selectivity_off_is_unchanged_independence_estimate` — OFF (default and explicit `false`) == the prior independence estimate, same tree/order/cost.
- `predicate_selectivity_on_uses_distinct_subjects_for_subject_join` / `..._distinct_objects_for_object_join` — ON reads the right per-predicate distinct count for the join-var position.
- `predicate_selectivity_falls_back_to_max_leaf_when_stats_absent` — absent distinct counts ⇒ ON == OFF.
- `knob_is_multiset_neutral_even_when_order_flips` — both plans permute the same pattern set `{0,1,2}` even when the estimate reorders the plan (the planner-layer result-equivalence guarantee).
- `star_join_unaffected_by_predicate_selectivity_knob` — a covered star is untouched by the knob.

## Opt-in / honesty stance

- **Opt-in / lean:** the new surface is a runtime `PlanOptions` field defaulting to `false`. The capability is also opt-in at the cargo-feature level — the whole crate is behind the `fedplan` feature (OFF by default); `sparq-core`/`sparq-engine` never depend on it. No new cargo feature is introduced (a single boolean knob within the existing feature), so the feature-matrix golden and `.github/feature-matrix.d/sparq-fedplan.yml` are untouched; the existing `adaptive-replan` leg already runs every `plan.rs` test under `fedplan`.
- **Honesty:** no performance number or speedup is asserted anywhere. Whether the tighter, skew-aware estimate yields faster plans is a **measurable hypothesis** to confirm on the canonical perf host (sq-0g6g), not on this work box and not in this PR.

## Gates (all green, both feature states)

- `cargo build`/`clippy --all-targets -D warnings`/`test` for `sparq-fedplan`: default (fedplan OFF), `--features fedplan`, and `--features adaptive-replan` — all green; downstream `sparq-fedplan-mpc` (fedplan-mpc) green.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean.
- terminology / privacy-claims / typos — clean. Feature-matrix assemble test — 7 passed.

Docs: `skills/federated-planning/SKILL.md` documents the knob, its semantics, the fall-back, and the honesty caveat. The crate README is an unchanged internal-stub (publish=false), so the `readme-template` gate is unaffected.

Files: `crates/sparq-fedplan/src/plan.rs`, `skills/federated-planning/SKILL.md`.

### DIFF
--- a/crates/sparq-fedplan/src/plan.rs
+++ b/crates/sparq-fedplan/src/plan.rs
@@ -165,6 +165,22 @@ pub struct PlanOptions {
     /// front is the latency/memory risk, so the streaming operator is preferred. Set to
     /// `f64::INFINITY` to always use plain hash. Default `100_000` rows.
     pub stream_threshold: f64,
+    /// [OPUS-4.8] sq-jsuzr (survey C2). Opt-in: use the candidate's per-predicate
+    /// **distinct-subject / distinct-object** statistics (already carried by every
+    /// [`crate::SourceDescriptor`]) for the join-key ndv in the **non-star** join estimate,
+    /// instead of the coarse `max(|L|, |R|)` leaf-cardinality approximation. When `true` and
+    /// the candidate's retained sources carry the relevant distinct count for the join
+    /// variable's position, the non-star (independence-fallback) output cardinality is scaled
+    /// by a weighted per-predicate selectivity; when those stats are absent it falls back to
+    /// the same max-leaf heuristic as when this is `false`.
+    ///
+    /// This is a **cost-estimate-only** knob: it changes the planner's cardinality estimate
+    /// (and thus its join order / bind-vs-hash decision), but the planned join's *result
+    /// multiset is unchanged* — BGP join is commutative/associative, so any order over the
+    /// same patterns yields the same answers. Default `false` (preserves the prior estimate
+    /// exactly, for A/B comparison). Star joins still use the characteristic-set estimate
+    /// regardless of this flag.
+    pub use_predicate_selectivity: bool,
 }
 
 impl Default for PlanOptions {
@@ -172,6 +188,7 @@ impl Default for PlanOptions {
         PlanOptions {
             request_cost: 1.0,
             stream_threshold: 100_000.0,
+            use_predicate_selectivity: false,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 **SPARQ agent** — adversarial pre-arm verification of this PR (bead `sq-jsuzr`, survey §C2, `sparq-fedplan`).

Independently checked out the PR head in a fresh worktree and ran all four hard gates; all pass cleanly.

**1. Correctness preserved (load-bearing) — PASS.** The knob is provably cost-estimate-only: in `cost_join`, the `out` cardinality is consumed only for upward propagation + the bind/hash/streaming algo+cost choice; it never touches source selection or pattern inclusion (those are fixed upstream in `select_sources`, independent of `PlanOptions`). So it can only reorder joins / swap a physical join algorithm — both result-multiset-preserving (BGP join is commutative/associative). The equivalence tests confirm this: `predicate_selectivity_off_is_unchanged_independence_estimate` asserts OFF == the prior estimate byte-for-byte (same join order, total_cost, and cardinality), and `knob_is_multiset_neutral_even_when_order_flips` asserts the ON plan is a complete permutation of the same pattern set `{0,1,2}` even when the order flips. Both green.

**2. Honesty — PASS.** No baked/asserted perf number in the diff or PR body. The only "faster" mention is the explicit caveat that it is *"a measurable hypothesis to confirm on the canonical perf host (sq-0g6g), not a claim made here."* `no-perf-numbers (docs)` is green.

**3. Opt-in / lean — PASS.** Runtime `PlanOptions::use_predicate_selectivity` bool, default `false`. Whole crate stays behind `fedplan` (default `[]`, `publish=false`, `forbid(unsafe_code)`); `mod plan` is `#[cfg(feature="fedplan")]` so default builds compile the new code out entirely. No new cargo feature, no `.github/`/feature-matrix-fragment/golden changes; `sparq-core`/`sparq-engine` never depend on it. Absent VoID stats ⇒ falls back to the identical max-leaf estimate (never fabricates an ndv).

**4. Gates green both states — PASS.** Independently ran `cargo clippy --all-targets -D warnings` in default + `--features fedplan` + `--features adaptive-replan` (all clean) and `cargo test`: default (crate compiles out), `fedplan` (75 pass, incl. the 6 new tests), `adaptive-replan` (109 pass). Downstream `sparq-fedplan-mpc` clippy clean. PR CI has zero failures (gate aggregator legitimately IN_PROGRESS).

Arming with `--squash --auto`.

PR #1250, opened 2026-06-23
feat(serve): in-process embedding seam — embed facade + GenerationRing front door (sq-xa15c, #1248 items 1+2) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — implementing bead **sq-xa15c** for the PSS agent's in-process embedding ask in **#1248** (items 1 + 2), so `solid-server-rs` can embed SPARQ **in-process** instead of over the SPARQL-1.1-over-HTTP transport.

### Item 1 — the embedding facade
A new `sparq_serve::embed` module: a small, **documented** tier-1 facade over the existing engine entry points, matching the shapes PSS listed so an `EmbeddedSparqClient` binds cleanly:

- `embed::query_json(&Graph, sparql) -> Result<String>` (SPARQL-JSON) + `embed::query(..) -> QueryResult` + `query_json_with_budget`
- `embed::ask(&Graph, sparql) -> Result<bool>`
- `embed::update_in_place(&mut Graph, sparql)` (+ `update_in_place_atomic` for all-or-nothing on one graph)
- `embed::apply_delta_nquads(&mut Graph, inserts, deletes)` (quad-level, per-graph; blank nodes by label)
- existence / metadata probes: `embed::exists`, `embed::named_graph_exists`, `embed::metadata -> Metadata { triples, named_graphs }`

These are **thin re-exports / one-line wrappers** — no reimplementation, no new behaviour. Binding to this one module (instead of reaching into `sparq-core`/`sparq-engine` directly) buys the embedder a single place the maintainer intends to keep stable.

### Item 2 — the runtime-agnostic `GenerationRing` core
**The extraction was already done in a prior wave**: the runtime-agnostic core was factored out of `sparq-server`'s axum/tokio glue into `sparq-serve` as `pub GenerationRing<S>`; `sparq-server` already re-uses `sparq_serve::GenerationRing` and holds **no private copy** (verified: there is no `struct GenerationRing` anywhere in `sparq-server`). So there is already ONE implementation and zero semantic divergence.

This PR therefore (a) makes that core the **documented embedding front door** by re-exporting `GenerationRing` / `Generation` / `GraphApplier` / `Writer` (+ `RingConfig`, `WriterConfig`, `PodId`, `TimeTravelConfig`) through `embed`, and (b) adds the **missing regression test** the brief requires — see below.

## Additive / semver-pending stance
[...]

PR #1252, opened 2026-06-24
feat(core): opt-in per-block Bloom filters on high-NDV permutation columns (sq-wihld) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** implementing bead **sq-wihld** (grounded survey **§A1**, `research/codebase-improvement-opportunities-2026-06-23.md`).

## What

Per-block **Bloom filters** on the high-NDV (subject/object) leading column of a block-compressed permutation (`crates/sparq-core/src/compress.rs`).

A `CompressedPerm`'s sparse directory `(first-triple, byte-offset)` already gives an *implicit min/max zone map* that prunes **range** scans. But for an **equality-bound leading column** — a point/prefix lookup on a constant subject/object — whose id falls inside several overlapping blocks' `[min,max]` spans (the common case on a high-NDV column), nothing skips a block that cannot contain it. This adds a tiny per-block Bloom bitset over each block's distinct leading ids and probes it in `range` before `decode_block_at`, skipping any block the filter proves cannot hold the constant.

The maintainer's own research (`research/data-structures.md` §A2 / §3 #1) flags this exact gap and says **"prototype this first."**

## Opt-in / lean stance

- New cargo feature **`block-bloom`**, **OFF by default**. Pure-`std` bitset — **no new dependency**; the lean default / wasm build pays nothing.
- Built **only for high-NDV leading columns** past a density gate (average distinct leading ids per block), so dense/predicate-leading columns never pay filter bytes that would never skip.

## Correctness contract (load-bearing)

- **Feature OFF → byte-identical.** The `bloom` field and the skip code are entirely `#[cfg(feature = "block-bloom")]`-gated; the default path is the original code verbatim.
- **Feature ON → results identical to OFF.** Zero false **negatives** by construction (a present leading id always probes "maybe present", so no matching row is ever skipped). A false **positive** costs one wasted block decode whose rows the existing range trim discards — the `range` output is unchanged.
- **On-disk format unchanged.** The filter is **in-RAM / build-time only** — it is *never* written to the `SPQCPRM1` format, so a perm built with the feature on persists byte-identically to one built with it off, and a memory-mapped perm carries no filter (the out-of-core path is untouched).

## Tests (real path, both feature states)

- Existing store-level plan-agreement test `compressed_scans_match_raw` runs through the Bloom skip path with the feature ON and confirms **identical rows + estimates** vs the raw store across every bound/unbound pattern shape — the feature-on-vs-off equivalence proof at the query-facing scan API.
- New `compress::tests::bloom_range_equals_binary_search_and_skips_blocks` asserts `range` over equality-bound leading keys equals a raw binary search (the feature-OFF oracle) over **present + absent** keys, **and** that the filter actually skips candidate blocks (proves the optimisation is engaged, not a silent no-op).
- `bloom_inert_for_non_equality_shapes` (full scan + multi-key range bypass the filter) and `bloom_decode_all_roundtrips`.
- Gates run GREEN in both states: `cargo build` / `clippy --all-targets -D warnings` / `test` in **default (OFF)** and with **`--features block-bloom`**, plus `--all-features` clippy, `block-bloom,mmap,dict-spill` test (mmap on-disk path), and `RUSTDOCFLAGS=-D warnings cargo doc --workspace --no-deps --all-features`. `readme-template`, `typos`, `privacy-claims`, `terminology` clean.

## Honesty

Per the repo's empirical-honesty rule, **no performance number / speedup is asserted** anywhere. The payoff is a measurable **hypothesis** — block-skip rate on selective equality-bound patterns and end-to-end latency — to be confirmed on the **canonical perf host (`sq-0g6g`)**, not on this non-canonical work box.

## CI

Adds the gating leg `opt-in sparq-core (block-bloom)` to `.github/feature-matrix.d/sparq-core.yml` and the shared golden `scripts/tests/feature-matrix-legnames.golden.txt` (assemble test green).

Auto-merge intentionally **OFF** — the verify stage arms it.

### DIFF
--- a/.github/feature-matrix.d/sparq-core.yml
+++ b/.github/feature-matrix.d/sparq-core.yml
@@ -31,3 +31,11 @@
   crate: "sparq-core"
   features: "jsonld"
   test: true
+
+# [OPUS-4.8] sq-wihld (survey §A1): opt-in per-block Bloom filters on the high-NDV
+# leading column of a block-compressed permutation. OFF by default (lean core); this
+# leg runs the bloom equivalence + block-skip tests (compress::tests::bloom_*).
+- name: "sparq-core (block-bloom)"
+  crate: "sparq-core"
+  features: "block-bloom"
+  test: true
--- a/crates/sparq-core/Cargo.toml
+++ b/crates/sparq-core/Cargo.toml
@@ -32,6 +32,16 @@ parallel = ["dep:rayon", "dep:memchr"]
 compact-index = []
 # Out-of-core querying: persist the indexes and memory-map them (native only).
 mmap = ["dep:memmap2"]
+# [OPUS-4.8] sq-wihld (survey §A1): OPT-IN per-block Bloom filters on the high-NDV
+# (subject/object) leading column of a BLOCK-COMPRESSED permutation. The directory's
+# first-triple already gives an implicit min/max zone map that prunes RANGE scans, but for
+# an EQUALITY-BOUND leading column whose id overlaps several blocks' [min,max] spans, only a
+# Bloom filter can skip the blocks that cannot contain it. OFF by default so the lean default
+# / wasm build pays no filter bytes; it is a pure-std bitset (NO new dependency). Zero false
+# negatives by construction, so results are identical to the feature-off path — only fewer
+# blocks decoded. The filter is in-RAM only and never serialised, so the on-disk SPQCPRM1
+# format is unchanged. See `compress::block_bloom`.
+block-bloom = []
 # [OPUS-4.8] sq-dvyi: JSON-LD ingest for the in-memory loaders (`load_str` /
 # `load_dataset` / `load_reader`). OPT-IN and OFF by default so the DEFAULT lean wasm
 # bundle (which the perf-gate tracks via `cargo build -p sparq-wasm`, no features) does
--- a/crates/sparq-core/README.md
+++ b/crates/sparq-core/README.md
@@ -38,7 +38,9 @@ assert_eq!(count, 1);
   `insert_triple(s, p, o)` / `remove_triple(s, p, o)` a single triple from `oxrdf` terms, or apply
   a whole batch with `apply_delta` — in place, with an optional write-ahead log.
 - **Out-of-core store** — query datasets larger than RAM from a memory-mapped on-disk store,
-  with optional block compression and near-zero resident heap.
+  with optional block compression and near-zero resident heap. The opt-in `block-bloom` feature
+  adds per-block Bloom filters on high-NDV columns to skip blocks on equality-bound point lookups
+  (result-equivalent, never serialised).
 - **Named graphs & RDF 1.2** — full quad storage and
   [triple terms](https://www.w3.org/TR/rdf12-concepts/).
 - **Thread-safe sharing** — `Graph` is `Send + Sync`, so one store serves many server threads;
--- a/crates/sparq-core/src/compress.rs
+++ b/crates/sparq-core/src/compress.rs
@@ -98,6 +98,150 @@ impl Blocks {
     }
 }
 
+/// [OPUS-4.8] sq-wihld (survey §A1) — OPT-IN per-block Bloom filters on the leading
+/// (column-0) ids of a block-compressed permutation. The directory's first-triple already
+/// gives an implicit min/max zone map that prunes RANGE scans, but for an EQUALITY-BOUND
+/// leading column (a point/prefix lookup, `lo[0] == hi[0]`) whose id falls inside several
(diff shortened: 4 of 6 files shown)


jeswr:
> 🤖 **SPARQ agent** — adversarial pre-arm verification (bead **sq-wihld**, survey §A1, `sparq-core`).

Verified all four hard gates locally on the PR head (non-canonical box → deterministic pass/fail + correctness only, no perf number baked):

**1. Correctness preserved.**
- *Feature OFF → byte-identical:* every change in `compress.rs` is `#[cfg(feature = "block-bloom")]`-gated (or a doc comment); the default `range`/`encode`/`open` path is the original code. Default `clippy -D warnings` + `cargo test -p sparq-core` pass.
- *Feature ON == OFF:* the load-bearing equivalence test `compress::tests::bloom_range_equals_binary_search_and_skips_blocks` asserts `range` over equality-bound leading keys equals a **raw binary-search oracle** (the feature-OFF reference) across **present + absent** keys, **and** that `total_skipped > 0` (the skip path is genuinely engaged, not a silent no-op). The store-level plan-agreement test `store::tests::compressed_scans_match_raw` passes **with the feature ON** — identical rows **and** estimates vs the raw store across every bound/unbound pattern shape (bound-subject routes through the Bloom skip). Zero false negatives by construction (whole-block skip preserves global sort order; any block holding the key probed "maybe present").

**2. Honesty.** No asserted perf/speedup number in the diff or PR body; payoff is framed as a hypothesis to confirm on the canonical host (`sq-0g6g`). The `~0.25 B/triple` figure is a derivable *space* cost, not a speed claim.

**3. Opt-in / lean.** `block-bloom = []`, OFF by default, **no new dependency** (pure-std bitset); on-disk `SPQCPRM1` format unchanged (filter is in-RAM/build-time only — a mapped perm carries `None`, verified by `compressed_save_open_roundtrip` under `block-bloom,mmap,dict-spill`). New CI leg present in **both** the per-crate fragment and the shared golden; `test_feature_matrix_assemble.py` passes locally.

**4. Gates green both states.** `clippy --all-targets -D warnings` + `cargo test` pass in **default (OFF)**, in **`--features block-bloom`** (95 tests), and in **`block-bloom,mmap,dict-spill`** (168 tests). The per-crate CI leg `opt-in sparq-core (block-bloom)` is SUCCESS.

The one red CI leg — `MSRV check` — is a **transient registry network flake** (`error: failed to get quinn ... OpenSSL SSL_ERROR_SYSCALL`, failing during dependency download before any compile), **not caused by this PR** (which adds no dependency); it is re-runnable once the in-progress run completes. Arming with `--auto` so the merge waits for legitimate CI completion.

jeswr:
> 🤖 SPARQ agent

**MSRV check fix — root cause was a transient crates.io download, not a >1.88 API.**

The failing `MSRV check (Rust 1.88, declared floor)` was NOT caused by the Bloom code using a post-1.88 std API. The job died in its **registry-update / dependency-fetch** phase, before compiling any source:

```
error: failed to get `quinn` as a dependency of package `reqwest v0.12.28`
  Caused by: download of qu/in/quinn failed
  Caused by: curl failed
  Caused by: [55] Failed sending data to the peer (OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 0)
```

This is the same transient `curl failed` mode the conformance job already self-heals (`sq-shacl-flake`); the MSRV step has no retry, so a one-off network hiccup leaves it red until re-run.

**Verified locally on the pinned 1.88 toolchain** (`rustc 1.88.0`), reproducing the exact CI command:

- `cargo +1.88 check --workspace --exclude sparq-py --exclude sparq-hdt` → exit 0
- `cargo +1.88 check -p sparq-core` (default) and `--features block-bloom` → exit 0

The new `block_bloom` module is pure-std (FNV-1a hash, a `Vec<u64>` bitset, slice `chunks`) with no post-1.88 surface. **No source change was required** — the fix is re-running the transient job, after which the `ci-summary` gate re-aggregates.

**Coverage was not at risk.** Every Bloom line is `#[cfg(feature = "block-bloom")]`-gated and the per-crate coverage gate measures sparq-core with `--features mmap,dict-spill`, so the Bloom module is compiled out of the measured build and cannot drop the floor (unlike the always-on additions that bit #1250). Local re-measure: sparq-core **93.06% ≥ 90% floor** (`coverage-gate.py --check-robust` → 1 ok / 0 fail).

Gates re-run green in-worktree: `clippy --all-targets -D warnings` and `cargo test -p sparq-core` in BOTH states (default + `block-bloom`), and `cargo doc --workspace --no-deps --all-features` with `RUSTDOCFLAGS=-D warnings`. Auto-merge (squash) left armed. `memmap2` unchanged at 0.9.11.

PR #1257, opened 2026-06-24
feat(engine): SPARQL 1.2 MULTIPLICITY() aggregate extension (sq-v411r, survey §B2)

jeswr:
> 🤖 **SPARQ agent** — implementing bead **sq-v411r** (survey **§B2**) in `sparq-engine`.

## What

Exposes the SPARQL 1.2 algebra's `multiplicity(μ|Ω)` device — which replaced the informal `card[Ω](μ)` in §18.4 BGP matching and the Set-Function definitions — as a zero-argument extension builtin **`MULTIPLICITY()`**, valid inside an aggregate's argument expression:

```sparql
SELECT (SUM(?x * MULTIPLICITY()) AS ?w) WHERE { SELECT ?x WHERE { ?s ex:v ?x } }
```

When the WHERE-clause solution bag is `{10,10,10,20,20}`, `SUM(?x * MULTIPLICITY())` folds the **distinct** members `{10(×3), 20(×2)}` = `10·3 + 20·2 = 70`, which equals plain `SUM(?x)` over the bag. That identity is the whole point of the device.

## Honesty / conformance stance (load-bearing)

The published **W3C SPARQL 1.2 Query rec and the editor's draft define `multiplicity` ONLY as an algebra/semantics notation — there is NO callable `multiplicity()` builtin in the spec, and NO conformance test for it.** I verified this against the pinned W3C suite (`w3c/rdf-tests @ f25dbc0`): there is no `multiplicity` test directory, and `sparql11/aggregates` (41/0), `sparql11/grouping`, `sparql11/subquery`, and every `sparql12/*` lane stay **100%** with this change.

So this is a **clearly-labelled sparq vendor EXTENSION** implementing the 1.2 algebra device — **not** a 1.2-conformance feature. The survey §B2 wording ("SPARQL 1.2 adds the standard `multiplicity()` function") conflated the algebra notation with a user-callable builtin. The boundary is documented in the README, the `sparql-query` SKILL, `vendor/spargebra/SPARQ-PATCHES.md` §10, and the rustdoc; a P3 follow-up bead corrects the survey record.

**SPARQL 1.2 `multiplicity` conformance suite cases now passing: none exist — there is no W3C conformance suite for a `multiplicity()` builtin.** The change is verified to leave the existing aggregate/grouping/subquery conformance lanes unchanged (no regression).

## Approach (ADDITIVE, side-effect-free)

- **Parser** (vendored spargebra): one `BuiltInCall` alternative `MULTIPLICITY() NIL()` (the zero-arg shape of `NOW()`/`RAND()`), gated behind `sparql-12`. To keep the shared `Function` enum **byte-compatible with downstream crates that match it exhaustively** (`sparopt`/`spareval` via `oxigraph`, pulled in by `sparq-bench` and so compiled under `--workspace --all-features`), it emits a reserved-IRI `Function::Custom(<urn:sparq:fn:multiplicity>)` — **no new enum variant**, no `Display` change.
- **Engine** (`exec.rs`): a thread-local `Cell<Option<u64>>` holds the current group member's bag cardinality; the aggregate evaluator sets it per member before `eval_expr`, and the `F::Custom` dispatch reads it (checked **before** the cast/registry so the reserved IRI can never be shadowed). When the argument calls `MULTIPLICITY()` the evaluator folds the **distinct** solutions weighted by cardinality (Set-Function algebra), so a value occurring `k` times contributes `k·x`, not `k²·x`. Outside an aggregate → expression error → unbound.
- **Zero cost on the common path**: the distinct-collapse + thread-local write are gated on an AST walk (`expr_uses_multiplicity`); an aggregate that never references `MULTIPLICITY()` is byte-for-byte unchanged and folds the full bag exactly as before.

## Tests (real query path — `crate::query`)

`SUM(?x*MULTIPLICITY())==SUM(?x)`; `Σ MULTIPLICITY()==COUNT(*)`; per-group cardinality; all-unique → multiplicity 1; out-of-aggregate → unbound; no-multiplicity aggregate folds the full bag; direct `group_multiplicities` gating; + a custom-aggregate (`window-functions`) variant in `aggregate.rs`.

## Gates (GREEN in both feature states)

- `cargo build` / `clippy --all-targets -D warnings` / `cargo test` — default (feature OFF) **and** `--features window-functions` (ON). Engine lib: **221 + 8** tests pass.
- Workspace rustdoc gate: `cargo doc --workspace --no-deps --all-features` with `RUSTDOCFLAGS="-D warnings"` — clean (this is what surfaced the exhaustive-match hazard that drove the reserved-IRI design).
- `readme-template --enforce`: **0 deviations**, engine README at the 120-line cap.
- MSRV **1.88**: `cargo +1.88 check -p sparq-engine` and `--workspace --exclude sparq-py --exclude sparq-hdt` — clean. `memmap2` stays 0.9.11; no new dependency.
- terminology / privacy-claims / markdown-typos — clean. `sparq-wasm` builds for `wasm32` (the builtin is pure, no native-only deps).

Feature flags exercised: **default** (`parallel,regex,digest`) and **`window-functions`**. Auto-merge intentionally **OFF**.

Closes sq-v411r.


jeswr:
> 🤖 **SPARQ agent** — adversarial pre-arm verification of **sq-v411r** (survey §B2, `MULTIPLICITY()`).

**Verdict: NOT arming — one HARD gate is red (`run + track benchmarks` → `wasm_bundle_bytes`).** Routing to @jeswr for the baseline decision. Everything else passes; the semantics are correct.

### What passed (verified independently, at PR head `442b5193`)

- **CORRECT (group multiplicity, not the cs.rs estimate).** I authored a *fresh* multiset with a different distribution than the PR's tests — bag `{5×4, 7×1, 9×2}` — and confirmed end-to-end via the real query path:
  - `SUM(?x * MULTIPLICITY())` = **45** = plain `SUM(?x)` over the bag (5·4 + 7·1 + 9·2). The Set-Function identity holds.
  - per-group `MAX(MULTIPLICITY())` grouped by `?x` = **4, 1, 2** for x=5,7,9 — the right per-group bag cardinality.
  - `SUM(MULTIPLICITY())` = **7** = `COUNT(*)` = bag size.
  The collapse-to-distinct-then-weight logic in `group_multiplicities` is the load-bearing piece (a value occurring `k` times contributes `k·x`, not `k²·x`), and it's correct. This is **not** the characteristic-set `multiplicity` in `cs.rs` (that file is untouched; the two usages are orthogonal — CS is a planner cardinality estimate). All 7 `multiplicity_builtin` tests + the `window-functions` custom-aggregate test pass.
- **ADDITIVE.** The only deleted lines are the SUM/AVG member loop, refactored into a `match group_multiplicities {…}` whose `None` arm is byte-identical to the old loop. An aggregate that never calls `MULTIPLICITY()` takes `None` and behaves exactly as before (confirmed by `aggregate_without_multiplicity_folds_full_bag`). Parser change is a single `Function::Custom("urn:sparq:fn:multiplicity")` arm — no `Function` enum variant, downstream exhaustive matchers untouched. `cs.rs` not in the diff.
- **MSRV / clippy / tests.** `cargo +1.88 check -p sparq-engine` ✅; `cargo clippy -p sparq-engine --all-targets -D warnings` clean ✅; `cargo test -p sparq-engine` (default + `window-functions`) green ✅.
- **COVERAGE.** `COVERAGE_CRATES=sparq-engine scripts/coverage.sh` = **88.12%** vs floor **83**; `coverage-gate.py --check-robust` → `1 ok / 0 fail`. ✅
- **HONESTY.** No fabricated perf numbers. The PR body/README/SKILL/PATCHES accurately label this a **non-standard vendor extension** — I verified there is **no** W3C `multiplicity` conformance directory/`.rq`/manifest in the pinned suite (`rdf-tests @ f25dbc0`), matching the stated "no conformance test exists" claim.

### Why I'm not arming — `wasm_bundle_bytes` HARD gate

```
wasm_bundle_bytes: floor=1656470  cur=1692910  -> REGRESSION +2.20% > +2% threshold [DETERMINISTIC]
```

This is **deterministic, not a flake** (re-running won't clear it). I attributed the delta by building `sparq-wasm --release --target wasm32-unknown-unknown` at both points:

| build | bytes | vs floor |
|---|---|---|
| merge-base `4d8e035a` (PR base, **passed** CI) | 1,687,485 | **+1.87%** |
| PR head `442b5193` | 1,693,484 | **+2.20%** |
| **PR's own contribution** | **+5,999 (~6KB, +0.36%)** | |

The PR's ~6KB growth is legitimate and small (new parser arm + eval dispatch; the 332-line `exec.rs` diff is mostly `#[cfg(test)]`, excluded from release wasm). **The trip is dominated by pre-existing drift**: main is already at **+1.84%** against a stale floor (1,656,470 set long ago; merge-base + current main `dc36748c` both measure 1,686,910 / +1.84% and pass only because they're under +2%). The PR's modest, justified increase crossed the line.

**This needs a maintainer judgment call, not an agent override.** The gate header's own remedy is to **RAISE the floor in a reviewed `bench/perf-baseline.json` bump** (or `PERF_GATE_ALLOW='wasm_bundle_bytes'`). I won't unilaterally bump the perf baseline inside an arming pass — that's exactly the deliberate, reviewed decision the ratchet exists to force. @jeswr, please either bump the floor to absorb this ~6KB (recommended; the code is correct and additive) or advise. Auto-merge left **OFF**.

*Note: `parse_ns_per_byte` +1.55% is within the timing noise-band (advisory, non-blocking) — not a concern.*

jeswr:
> 🤖 **SPARQ agent** — unblocked by the maintainer-approved wasm-baseline re-base (#1286, floor→1686907, +2% guard retained): the `run + track benchmarks` wasm gate now passes against the current-main floor, and the coverage ratchet is green. `MULTIPLICITY()` shipped honestly as a clearly-labelled SPARQL-1.2 *vendor extension* (no claim it is standard). All required checks green/pending with zero failures — arming.

jeswr:
> 🤖 **SPARQ agent**

## Unblocked: `coverage ratchet (per-crate)` was a measurement artifact, not a real regression

**Blocker (re-derived from live state):** the only failing real check was `coverage ratchet + test-presence gate (per-crate)` → `gate` fail; everything else (wasm/benchmarks already green after #1286, clippy, conformance, supply-chain) was green. The gate reported `sparq-engine 65.09% < floor 83` and held at ~60–65% across all 3 robust re-measures.

**Root cause — a poisoned coverage DENOMINATOR, not under-coverage.** The coverage job's restored `target/` (Swatinem/rust-cache) carried `sparq-engine` objects compiled with extra **opt-in** features (`zk`/`yannakakis`/`service`/`window-functions`/`semijoin-bitmap`/`cs-planner`/…) baked in. `cargo llvm-cov --package sparq-engine` runs *default* features but **reused those instrumented objects**, so llvm-cov counted ~3.7k feature-gated regions the default-feature tests never exercise as "uncovered" — inflating `lines_total` while the actual covered lines were fine.

Evidence (all on one box, same tooling):

| build | lines_pct | covered / total | exec.rs lines |
|---|---|---|---|
| `main` (clean, CI) | 88.09% | 9101 / 10331 | 6448 |
| this branch (clean worktree) | **88.12%** | 9233 / 10478 | 6595 (+147 = the MULTIPLICITY code) |
| this branch (CI, cache-poisoned) | 65.09% | 9233 / **14184** | ~11501 |

The **covered count is identical (9233)** across clean and poisoned runs — only the denominator was poisoned. So the branch's true coverage is ~88%, comfortably above the floor; nothing in the diff under-covers.

**Fix (root cause, not a workaround):** `cargo llvm-cov clean --workspace` at the start of the per-commit measurement loop in `scripts/coverage.sh`, so every per-crate build instruments only the regions its own requested feature set compiles — independent of any restored cache. This mirrors the existing `measure_merged` path (which already cleans) and does **not** mask a regression: it re-compiles and re-measures the *same* tests. **The floor is unchanged (still 83); auto-merge left enabled.**

**Verified locally** — on the contaminated worktree that measured 57%, the fixed `scripts/coverage.sh` now measures:

```
sparq-engine  lines= 88.31%  (9311/10543)
ok  sparq-engine  88.31% >= floor 83
VERDICT: PASS
```

**Also added** (the direct-tests ask + extra margin), in `crates/sparq-engine/src/exec.rs`'s `mod multiplicity_builtin`:
- `expr_uses_multiplicity_walks_every_arm` — exercises **every** arm of the `MULTIPLICITY()` detection walk (the gate that selects the distinct-collapsing fold), including the `EXISTS` non-descent contract.
- `multiplicity_context_guard_nests_and_restores` — the thread-local RAII `Guard`'s nest/restore re-entrancy.

`MULTIPLICITY()` stays a clearly-labelled **non-standard vendor extension** (no W3C conformance claim) — unchanged.

**Local gates re-run green:** `cargo test -p sparq-engine` (9 multiplicity tests incl. the 2 new), `cargo clippy --all-targets -D warnings`, `cargo +1.88 check --all-targets`, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features`, and the coverage gate (`coverage-gate.py --check-robust` → PASS).

Pushed `366e3c69` to this branch to re-trigger a fresh CI cycle (the prior `gate` concluded "fail" and needed a fresh run, not just a sub-job re-run). Once the coverage job re-runs clean, the PR is mergeable and auto-merge lands it.

Issue #820, opened 2026-06-19
GUI: unintrusive stats


jeswr:
It would be nice to have stats for things like current disk usage, loading time remaining for ingests of large files etc. without them being displayed in an overly intrusive way.

jeswr:
> 🤖 **SPARQ agent** — autonomous agent on @jeswr's account (Claude Opus 4.8).

Triaged against `origin/main` (HEAD `dc36748c`). The **unintrusive stats surface already exists** and is part-way to this ask: `gui/app/src/components/workbench/status-bar.tsx` is a 24px (`h-6`) footer that already shows, unobtrusively — last-query latency (`performance.now`), row count, store size in quads, target, loader mode (native vs in-tab), and the persistence backend.

The two **specific** stats you named are still gaps:
- **Current disk usage** of the persistent store — not surfaced (needs a backend size probe on the Tauri fs path).
- **Loading-time-remaining / progress + ETA** for large-file ingests — not surfaced (needs a progress signal from the loader).

I've beaded those two additions as **`sq-su1oe` (P2)**, scoped to extend the existing status bar in the same non-intrusive style rather than adding a separate panel. Leaving this open until they land.

— 🤖 SPARQ agent

jeswr:
> 🤖 **SPARQ agent**

First pass at this in the **website /try REPL** lane: **#1260** (auto-merge left OFF — this is your design call). It adds a muted, toggleable one-liner under the REPL header showing triple count, in-memory store footprint, and the most-recent ingest's source size + *measured* load time. Two honesty notes: there's no server "disk" on the static WASM site so I surface heap-resident size as "~N in memory" (not relabelled as disk usage), and the WASM load is opaque so I show real elapsed load time rather than a fabricated ETA.

The **literal** disk-usage + ingest-ETA you describe fit the **Tauri desktop GUI** status bar (they need an fs probe + a loader progress signal that don't exist in the static site) — I filed `sq-a1ibz` for that and left this issue open for it. Please steer the selection/placement/visibility on #1260 to taste.

jeswr:
> 🤖 **SPARQ agent** — @jeswr's agent for `jeswr/sparq`. Status update since the 2026-06-24 pass.

**Desktop GUI — both asks are now live on `origin/main`:**

[...]

PR #1270, opened 2026-06-27
feat(gui): OS-reported disk-usage probe for the status bar gauge (sq-cno90)

jeswr:
> 🤖 **SPARQ agent**

Closes the **#820** follow-up bead **sq-cno90**: tighten the GUI status-bar disk gauge from the snapshot-bytes ESTIMATE to the precise **OS-reported** on-disk size — in the desktop (Tauri) shell only, falling back to the honest estimate on the web target.

## What this implements
- **`gui/src-tauri/src/disk.rs`** — a minimal `disk_usage` Tauri command. It resolves `$APPLOCALDATA/workspaces` (the same base dir the granted `fs` capability + the `@sparq/client` workspace store key off) and returns its **real byte total** via `dir_size_bytes`: a recursive `stat()` walk that sums each regular file's `metadata.len()`. It follows **no symlink out of the tree** (least-privilege — cannot escape / loop), skips an unreadable entry rather than panicking, and reports a missing dir as `bytes: 0, exists: false` (an honest fresh-install answer).
- **Least-privilege capability** (`capabilities/default.json`) — documents that `disk_usage` is an app-defined command invocable by `core:default` and adds **NO new fs-plugin surface**: it reads via the engine's own `std::fs`, confined in Rust to the `workspaces/` subtree. (App-defined commands are not permission-gated in Tauri 2; adding a bogus `fs:` grant would misrepresent that it uses the fs plugin — it does not.)
- **Frontend wiring** — `tauri-ipc.ts` `nativeDiskUsage()` feature-detects the Tauri API; the engine context exposes `diskBytes` (OS-reported, `null` on web) + `refreshDiskUsage()` (probed on mount and after each persisted import). The status bar **PREFERS** the OS figure in the desktop shell (`disk N`) and **FALLS BACK** to the estimate on the web target (`≈disk N`) — clearly labelled which of the two it is, never fabricated.
- The existing #820 redesign styling (conic disk ring, ingest meter) is kept.

## Honesty
The gauge goes from estimate → OS-reported **in the desktop shell only**; on the hosted web target there is no native FS, so it stays the snapshot-bytes estimate (shown as `≈disk`). Either way it is a real measured value or the labelled estimate — never a fabricated number. No performance figure is asserted.

## A note on the brief's premise
The brief described the gauge as already wired to `snapshotBytes`; on `origin/main` the #820 redesign wires it to `storeBytes` (the snapshot-bytes estimate, computed via `snapshotBytes(store)`). Same intent — this PR is the precise-probe follow-up that bead described.

## Gates (all green)
Run in **BOTH** feature states — default (feature OFF) **and** `--features hdt`:
- `cargo build` · `cargo clippy --all-targets -- -D warnings` · `cargo test` (`gui/src-tauri`)
  - 5 **direct** unit tests for the byte-summing walk: nested-dir sum, single-file real length, empty dir, missing path → 0, **symlink-not-followed** (Unix).
- MSRV `cargo +1.88 check --all-targets` (both states).
- `gui/app`: `npm run typecheck` · `npm run lint` (no ESLint warnings/errors) · `npm run build:tauri` (static export) — all green.

`gui/src-tauri` is a standalone crate (own `[workspace]`, excluded from the main workspace gate), validated in `.github/workflows/gui.yml` which installs the webview system libs.

Cites **sq-cno90** + **#820**.

### DIFF
--- a/gui/README.md
+++ b/gui/README.md
@@ -78,8 +78,12 @@ gui/
 - **Top bar** (`h-10`): a LOCAL⇄ENDPOINT target switch, the store size, a ⌘K hint, a theme toggle,
   and an engine status LED.
 - **IDE tab strip** + a full-bleed work area (default = Query).
-- **Status bar** (`h-6`): the **measured `performance.now()` latency** of the last run, the row
-  count, the target, and the persistence backend.
+- **Status bar** (`h-7`): the **measured `performance.now()` latency** of the last run, the row
+  count, the target, the persistence backend, the live store size, an unintrusive ingest meter
+  (`sq-vw3ax`, #820), and a **disk gauge**. The disk gauge shows the **OS-reported** on-disk size
+  of the app-data `workspaces/` dir in the desktop shell (a recursive native `stat()` via the
+  `disk_usage` command, `sq-cno90`) and falls back to the **snapshot-bytes estimate** on the web
+  target — always labelled which of the two it is (`disk` vs `≈disk`), never a fabricated figure.
 
 ## The direct native engine link (desktop)
 
@@ -88,7 +92,10 @@ gui/
 (`load` / `query` / `query_quads` / `update_in_place` / `explain` / `explain_analyze` / `count` /
 `ask` / `store_size`) but backed by the full native store. Its `#[cfg(test)]` tests exercise the
 engine calls directly. The foundation frontend runs the in-tab WASM engine in BOTH targets for
-**query** (the honest, working-today path).
+**query** (the honest, working-today path). `src-tauri/src/disk.rs` adds the `disk_usage` command
+(`sq-cno90`): a recursive `stat()` of `$APPLOCALDATA/workspaces` that returns its real on-disk byte
+total for the status bar's disk gauge (least-privilege — confined in Rust to that subtree, follows
+no symlink out of it; the byte-summing walk is unit-tested directly).
 
 For **ingest**, the Import drawer (`sq-ixc3.13`) calls the engine's **native loader** IPC
 (`load_path` / `load_text`) when running in the desktop shell: a disk file — including
--- a/gui/app/src/components/workbench/import-drawer.tsx
+++ b/gui/app/src/components/workbench/import-drawer.tsx
@@ -97,7 +97,7 @@ function ImportDrawer({
   open: boolean;
   onOpenChange: (open: boolean) => void;
 }) {
-  const { importRdf, snapshotStore } = useEngine();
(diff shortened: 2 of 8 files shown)


jeswr:
> 🤖 **SPARQ agent**

**Verified + armed for auto-merge (squash).** Checked out the PR head (`c8f0d08`) in a fresh worktree and independently re-ran the gates on Rust 1.88.

**Correctness (sq-cno90 spec):**
- `disk_usage` returns a **real recursive byte total** — `dir_size_bytes` is a `std::fs` `symlink_metadata` walk summing each regular file's `metadata.len()` (apparent size). Confirmed by running the **5 direct unit tests** through the real function: nested-dir sum, single-file real length, empty dir → 0, missing path → 0, and **symlink-not-followed** (the link's 9999-byte target is excluded; only the 50-byte real file counts) — least-privilege escape/loop guard holds.
- **Web fallback path intact**: `nativeDiskUsage()` returns `null` off-Tauri (`tauriInvoke()` is null) → `diskBytes` stays `null` → status bar shows `≈disk <estimate>` (snapshot bytes). Desktop prefers the OS figure as `disk <N>`. Labels distinguish the two honestly; never fabricated.
- `refreshDiskUsage()` is re-probed on mount and after each persisted import (import-drawer).

**No regression (both feature states):**
- `cargo +1.88 build` (default) ✓
- `cargo clippy --all-targets -- -D warnings` clean — **default** and **`--features hdt`** ✓
- Full test suite green: **13 tests** (5 disk + 8 engine), both default and `--features hdt` ✓
- Frontend: `npm run typecheck` ✓, `npm run lint` (no ESLint warnings/errors) ✓, `npm run build:tauri` static export ✓

**Honesty:** no fabricated perf numbers in the diff; web-vs-desktop divergence documented in PR body + README + code comments; no Cargo manifest/lock/deny.toml changes. The PR honestly flags that `origin/main` wires the gauge to `storeBytes` (not `snapshotBytes` as the brief said) — same intent.

**Remaining red legs are not caused by this PR:**
- `cargo-deny` — transient crates.io-index fetch flake (`HTTP2 framing` / `connection reset`); **no dependency changes** in this PR; re-run triggered.
- `GUI build + clippy (win-x64, advisory)` — pre-existing: `icons/icon.ico` is absent from the repo on `origin/main` (only `.png` icons exist); PR touches no icons.
- `GUI tauri-driver e2e (Linux, advisory)` — pre-existing WebDriver capability-negotiation harness issue; PR touches no `gui/e2e/`.

Auto-merge (squash) enabled; will land once the required `gate` clears.

PR #1254, opened 2026-06-24
feat(sparq-engine): exact-bitmap semi-join reducer on dense u32 ids (sq-gr8mb, §A3) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** implementing bead **sq-gr8mb** (grounded improvement survey **§A3** — `research/codebase-improvement-opportunities-2026-06-23.md`), crate `sparq-engine`.

## What this adds

An **opt-in exact-bitmap semi-join reducer** (CIDR'26 *"Not Yannakakis"*; ranked the single best technique-architecture fit in `research/optimization-techniques.md` §1.1/§2(a) — *ADOPT NOW*).

`crates/sparq-engine/src/exec.rs` runs binary BGP joins with bind/merge but had **no semi-join prefilter**. After `eval_bgp_binary` materialises a join's left side, it now builds a membership filter (`semijoin::KeyFilter`) over **one** connecting variable's distinct ids:

- a flat **bitmap** over the dense `u32` ids — the dictionary assigns dense ids, so the bitmap is a perfect-hash, **zero-false-positive**, branch-free membership test;
- an exact **hash set** when the keys are sparse + huge (a flat bitmap would waste memory).

`scan_to_bindings` takes an optional pre-filter and **drops a row whose join key is absent before it enters the join** — a row that would fail the join anyway.

## Correctness contract (load-bearing)

This is an **OPT-IN performance feature**, `semijoin-bitmap`, **OFF by default**.

- **Feature OFF → byte-identical.** Every change on the default path is `#[cfg(feature = "semijoin-bitmap")]`-gated; the cfg'd `scan_to_bindings` parameter and its call arguments vanish, the per-row `build_row` gains no check. The default native + wasm builds are byte-identical (no new deps; no `unsafe`).
- **Feature ON → identical RESULTS to OFF.** The filter is membership-**exact** (no false positives), so it removes **only** rows the downstream join would have dropped anyway — the answer is unchanged, only fewer rows are scanned. A row survives a multi-variable join only if it matches on **every** shared variable, so filtering on one of them is a valid necessary condition.
- Proven by `tests/semijoin_differential.rs`, which is **intentionally NOT feature-gated** so it runs in **both** states; each query's result is compared against an independent brute-force reference, so the on-vs-off equivalence is a cross-build identity. Representative star / snowflake / chain / selective / disjoint / repeated-variable shapes. The existing plan-agreement / differential suites still pass in both states.

## Opt-in / honesty stance

[...]

PR #1272, opened 2026-06-27
feat(sparq-engine): Yannakakis bottom-up full-semijoin prepass for acyclic BGPs (sq-5zf8i, §A4) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent**

Implements bead **sq-5zf8i**: an opt-in Yannakakis **bottom-up full-semijoin prepass** for ACYCLIC BGPs in `crates/sparq-engine` (per `research/codebase-improvement-opportunities-2026-06-23.md` **§A4** + `research/optimization-techniques.md` §1.1/§2(a2)). Built on top of the merged A3 exact-bitmap reducer (sq-gr8mb / #1254).

### What it does
For an acyclic BGP, BEFORE the main join, reduce every relation against its neighbours in the join tree so no "dangling" tuple (one that joins with nothing downstream) is ever materialised by the join — directly attacking the materialising evaluator's #1 cost (intermediate-result blow-up).

- Materialise each pattern's relation once, build a join tree over the relations' join-variable sets (`semijoin::build_join_tree`), then sweep it **bottom-up (leaf→root)** and **top-down (root→leaf)**, semijoining each relation against its neighbours. The reduced relations are joined by the existing generic conjunctive `join_bindings`.
- **Reuses the A3 machinery** as the brief requires: the per-edge membership test is the *same* exact `semijoin::KeyFilter` (bitmap on dense ids / hash set on sparse+huge); routing is the *same* `bgp_is_cyclic` / `bgp_uses_binary` GYO test the executor already has — **acyclic ⇒ semijoin-reduce then join; cyclic ⇒ existing LFTJ, unchanged**. The `yannakakis` feature **implies** `semijoin-bitmap`.
- **Cost-gated** (pure-overhead guard): skipped when the estimated max relation is below a threshold, so a tiny BGP transparently uses the ordinary binary plan.

### Correctness (load-bearing)
A semijoin is a pure **FILTER** — it removes only rows the final join would itself drop — so the RESULT is **identical** to the feature-off binary plan for any join order. OPT-IN and **OFF by default**: with the feature off none of the prepass compiles and the executor path is byte-identical (every new branch is `#[cfg(feature = "yannakakis")]`-gated; the default conjunctive fast-path calls `eval_bgp_binary` exactly as before).

Proven by `tests/yannakakis_differential.rs` (NOT feature-gated → runs in BOTH states; each query checked against an independent brute-force reference, so `on == off` cross-build) over **chain / star / snowflake / cyclic / empty / no-reduction** shapes incl. a pushed-down-FILTER case, plus direct in-crate unit tests for the engaged prepass (above the cost-gate, dangling tuples dropped) and the cost-gate fallback. The empty / single-pattern / unsatisfiable / triple-term / empty-default-view / armed-zk-recorder cases all defer to `eval_bgp_binary`.

### Honesty
**No performance number is baked anywhere.** The intermediate-row-count and end-to-end reduction on chain/snowflake workloads is a measurable **HYPOTHESIS** for the canonical perf host (bead `sq-0g6g`), not this work box.

A note on the brief's premise: A3 (#1254) landed `KeyFilter` in `crates/sparq-engine/src/semijoin.rs` (not a separate "membership module"); this PR extends that same file with the pure join-tree topology and adds the prepass driver to `exec.rs` — exactly as the brief intended, just spelled against the actual merged layout.

### Feature flag
`yannakakis` (OFF by default), `= ["semijoin-bitmap"]`. Zero new dependencies; no `unsafe`.

### Gates run (both states)
- `cargo build` / `cargo clippy --all-targets -- -D warnings` / `cargo test` — GREEN in **default** (275 passed / 0 failed) **and** `--features yannakakis` (288 passed / 0 failed).
- `cargo clippy --all-features --all-targets -- -D warnings` — clean.
- MSRV `cargo +1.88 check` — clean in both states.
- `python3 scripts/check-readme-template.py --enforce` — 0 deviations (README 120 lines).
- feature-matrix golden test (`scripts/tests/test_feature_matrix_assemble.py`) — passes (new `opt-in sparq-engine (yannakakis)` leg).
- `cargo build --workspace` — clean.

Regression caught + fixed during development: the empty-BGP (`{}` → unit row) under an Empty default-graph view; the prepass now defers `patterns.len() < 2` to `eval_bgp_binary`, restoring identical behaviour (re-verified by `tests/dataset_view.rs`).

Closes sq-5zf8i. Auto-merge intentionally OFF — the verify stage arms.


jeswr:
> 🤖 **SPARQ agent** — verification (sq-5zf8i, engine-yannakakis)

Verified the PR head in a fresh worktree (cargo + the 1.88 MSRV toolchain). **ARMED** (`--squash --auto`).

**Correctness (HARD — the answer-preserving claim):**
- **Feature-OFF byte-identical:** every new symbol is `#[cfg(feature = "yannakakis")]`-gated and the `not(feature)` arm of the two executor seams is the verbatim `eval_bgp_binary(...)` call — no non-gated logic change. Default build / `clippy -D warnings` / test suite all green.
- **Feature-ON == OFF across BGP shapes:** the non-feature-gated `tests/yannakakis_differential.rs` runs in BOTH states against an *independent* brute-force reference and passes both ways (transitive on==off). I confirmed the engaged multi-pattern prepass is itself correct by temporarily forcing the cost-gate to 1 — all 8 shapes (chain-3 / star-3 / snowflake-3 / filtered-2 / disjoint-2 / no-reduction-3 / cyclic→LFTJ) still matched brute-force exactly. The per-column `KeyFilter` AND-membership semijoin is **sound** (it never drops a row with a real join partner; the final join removes any over-kept rows), which is all Yannakakis answer-preservation requires.

**Gap I found + fixed (pushed as 21bb76e3):** the shipped "engaging" differential datasets all sat *below* the `YANNAKAKIS_MIN_REL=4096` cost-gate (chain max_est=3200, star/snowflake=1500), so the feature-on leg silently exercised only the **binary fallback** — the engaged multi-pattern path was untested end-to-end (only the 2-pattern in-src unit test engaged) and the docstrings ("above the cost-gate so the prepass engages") were inaccurate. I sized the engaging datasets above the gate (chain len=2200→4400; `social_dataset(4200)`), kept small-star/disjoint below it for the fallback, fixed the count assertion + docstrings, and re-verified via a gate probe that each now reports `engages=true` and still matches brute-force in both states.

**No regression:** `cargo test -p sparq-engine` (default) and `--features yannakakis` (227 lib + all integration suites) both green, 0 failed; `cargo clippy -p sparq-engine --all-features --all-targets -- -D warnings` clean; `cargo +1.88 check` clean in both states.

**New gating CI leg + golden:** the `opt-in sparq-engine (yannakakis)` feature-matrix leg is wired and gates (not advisory); `scripts/tests/test_feature_matrix_assemble.py` and the `feature-matrix-legnames.golden.txt` golden both updated and passing.

**Honesty:** no baked perf/accuracy numbers (the payoff is documented as a hypothesis for the perf host sq-0g6g); README template-compliant (120 lines). The one inaccuracy (engagement docstrings) is now corrected.

Auto-merge ON; will land once CI on 21bb76e3 is green.

jeswr:
> 🤖 **SPARQ agent**

Fixed the failing `opt-in sparq-engine (semijoin-bitmap)` clippy leg.

**One-line cause.** The Yannakakis prepass topology in `crates/sparq-engine/src/semijoin.rs` — `TreeNode`, its `is_root` method, and `build_join_tree` (plus their 5 join-tree unit tests) — is used ONLY by the `yannakakis` path (`eval_bgp_yannakakis` in `exec.rs`) but was left ungated. Built with only the sibling `semijoin-bitmap` feature (no `yannakakis`), those items are dead code, so `clippy -D warnings` (`-D dead_code`) failed: `TreeNode is never constructed` / `is_root is never used` / `build_join_tree is never used`.

**Fix.** Gated each Yannakakis-only item behind a real `#[cfg(feature = "yannakakis")]` (not a blanket `#[allow(dead_code)]`, so the items genuinely don't exist in non-yannakakis builds — the opt-in stays lean). `KeyFilter` and its merge/membership machinery stay **ungated** (shared with `semijoin-bitmap`). No runtime behaviour change in either feature. This matches the intent already documented in `.github/feature-matrix.d/sparq-engine.yml`, which states `build_join_tree` + `TreeNode` + their 5 tests are all meant to be `#[cfg(feature = "yannakakis")]`.

**Verified locally — clippy `--all-targets -- -D warnings` matrix:**

| feature state | result |
|---|---|
| default (neither) | ✅ pass |
| `semijoin-bitmap` (the failing leg) | ✅ pass |
| `yannakakis` | ✅ pass |
| `semijoin-bitmap yannakakis` | ✅ pass |

`cargo test -p sparq-engine --features yannakakis` → **227 passed, 0 failed**, and the 5 join-tree tests (`chain_join_tree_is_a_single_spine`, `star_join_tree_roots_at_the_hub`, `disconnected_bgp_yields_one_root_per_component`, `single_relation_is_its_own_root`, `snowflake_join_tree_well_formed`) + the 2 exec engaged/cost-gate tests + the `yannakakis_differential` suite all still compile and run (not silently cfg'd out). Rustdoc all-features gate (`RUSTDOCFLAGS="-D warnings" cargo doc -p sparq-engine --no-deps --all-features`) clean.

Pushed `ebc6197a`. Auto-merge left enabled.

PR #1271, opened 2026-06-27
feat(shacl): SHACL 1.2 targets + SPARQL node-expression value/target/severity (sq-rnkdh) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — implementing bead **sq-rnkdh** (epic sq-waf9o): SHACL 1.2 TARGETS + SPARQL constraint extras in `crates/sparq-shacl`.

## What this implements

All on the **always-on** SHACL Core / SHACL-SPARQL path — **no feature flag**, green in BOTH default (feature OFF) and `shacl-af` ON. These are SHACL-1.2 core/sparql suite features (not SHACL-AF rules), so they reuse the always-compiled `sh:sparql` `$this` pre-binding rather than the `shacl-af`-gated node-expression algebra.

- **`sh:targetWhere [ <inline shape> ]`** (`Target::Where`) — focus nodes are every data-graph node that CONFORMS to the inline (object) shape (conformance via the validator's own memoised `conforms`).
- **`sh:shape`** — a DATA-graph triple `?n sh:shape ?S` makes `?n` a focus node of shape `?S` (the data-driven dual of `sh:targetNode`; discovered per data graph in `validate_graph`, not as a per-shape `Target`).
- **`sh:ShapeClass`** — "a class that is also a node shape" (1.2): discovered as a shape root and given an implicit class target, exactly like `rdfs:Class`.
- **SPARQL-valued targets & value nodes** (`PreparedSelectExpr`) — `sh:targetNode [ sh:select … ]` (`Target::Sparql`) computes focus nodes from the first result variable; on a property shape `sh:values [ sh:select … ]` / `[ sh:sparqlExpr "EXPR" ]` COMPUTES the value nodes (`$this` = focus) instead of traversing `sh:path` (the reported `sh:resultPath` stays the shape path). `sh:sparqlExpr` is derived as `SELECT ((EXPR) AS ?result) WHERE {}`.
- **Constraint-level `sh:severity`** on a `sh:SPARQLConstraint` overrides the shape's default severity for its results.

## Tests made to pass via the REAL `validate()`

Vendored `shacl12-test-suite` entries (pinned suite `b6e73695`):

- core: `targets/targetWhere-001`, `targets/shape-001`, `targets/targetClassImplicit-002`
- sparql: `targets/targetNode-select-001`, `node/sparql-001`, `property/property-select-001`, `property/property-sparqlExpr-001`

## 1.2-draft semantics

[...]

PR #1275, opened 2026-06-27
feat(shacl): SHACL-SPARQL 1.2 pre-binding propagation + sh:sourceConstraint + node-expr harness fidelity (sq-mue75)

jeswr:
> 🤖 **SPARQ agent**

Implements **sq-mue75** (epic sq-waf9o): SHACL-SPARQL 1.2 pre-binding propagation, `sh:sourceConstraint`, and node-expression harness fidelity in `crates/sparq-shacl/`.

## PRIMARY — pre-binding VALUES propagation (`src/sparql.rs::push_values_down`)
SHACL-SPARQL pre-binds the focus node (`$this`/`$value`/`$PATH`) by injecting a `VALUES` block that must reach the leaf graph patterns. The old descent followed a single chain; it now propagates into **every scope the variable can reach**:
- **UNION** — into BOTH branches independently.
- **sibling Join** — into BOTH children (a `{ FILTER(bound($this)) }` sibling over an empty inner now sees the binding).
- **sub-SELECT `Project`** — through the projection when the variable is explicitly projected; a `SELECT *` (empty projection list) re-scopes it, so the VALUES is joined ABOVE (the spec-rejection case). `GRAPH` blocks are descended too.

Makes these vendored W3C `shacl12-test-suite/tests/sparql/pre-binding` entries PASS via the **real** `validate()`:
- `pre-binding-002` (UNION), `pre-binding-005` (sibling Join/BGP+FILTER), `pre-binding-007` (nested SELECT projecting `$this`).

## SECONDARY — `sh:sourceConstraint` (`src/eval.rs`, `report.rs`, `model.rs`)
`sh:sparql` results now carry `sh:sourceConstraint` = the `sh:SPARQLConstraint` node (distinct from `sh:sourceShape`), per SHACL §5.2.2, emitted in the Turtle report. Core constraints and §6 SPARQL-based **components** carry `None` (the spec omits it there). The shacl12 SPARQL harness now verifies the field.

## SECONDARY — node-expr harness fidelity (`tests/w3c_node_expr.rs`)
The runner re-implemented the node-expression algebra inline and compared order-blind. It now drives the **real** crate path via `eval_node_expression` with **order-aware** (sequence) comparison, falling back to multiset only for order-insensitive set operators. To support this the crate parser learned the SHACL 1.2 `shnex:` spellings of the shared algebra (`pathValues`/`nodes`/`filterShape`/`intersection`/`union`) and `eval_node_expression` now registers inline filter/function shapes.

Driving real code surfaced and **FIXED** a genuine crate bug: a decimal `sum` rendered `42` rather than the canonical `42.0`^^xsd:decimal (`func::decimal_literal`). One honest divergence remains — the harness `sht:scope-*` var-injection entry (`var-bound`), which the crate's evaluation has no counterpart for — recorded as **xfail** (documented, not faked); see the deferred follow-up below.

## FLOORS (re-measured, both feature states)
- `w3c_sparql_shacl12.rs`: `BASELINE_PASS` 14→17, `BASELINE_GAP` 10→7 (the 7 `sht:Failure` rejection entries remain).
- `w3c_node_expr.rs`: real-eval path = **62 pass + 1 xfail** (the previous 63 inline-pass count preserved as 62 + 1 xfail); `BASELINE_PASS` = 62, `XFAIL_VAR_SCOPE` = 1.
- `w3c_core_full_shacl12.rs`: floors unchanged (129 default / 130 `shacl-af`).
- Two-sided ratchet kept (pass ≥ floor AND gap ≤ ceiling); the DEFAULT floor equals the default achieved count; no all-pass assertion.

## Gates (GREEN in BOTH default AND `--features shacl-af`)
- `cargo test -p sparq-shacl` — green both states (0 failures).
- `cargo clippy -p sparq-shacl --all-targets -- -D warnings` — both states.
- `cargo +1.88 check -p sparq-shacl` — both states.
- `python3 scripts/check-readme-template.py --enforce` — 0 deviations (README ≤120 lines).
- Direct unit tests added for the new arms (union / sibling-join / projecting sub-SELECT / SELECT*-above / GRAPH) and for `sh:sourceConstraint` presence/absence.
- No regression: the merged core-value + targets tests and the full existing suite still pass.

## Deferred (follow-up bead)
- **Caller-supplied node-expression variable scope** — a `shnex:var "<name>"` resolving a harness `sht:scope-<name>` binding (W3C `node-expr/shnex/var-bound`); the crate binds only `"focusNode"`. Kept honest as an xfail with a clear note in the harness.

Closes sq-mue75.

### DIFF
--- a/crates/sparq-shacl/README.md
+++ b/crates/sparq-shacl/README.md
@@ -49,9 +49,8 @@ if report.conforms_violations_only() { /* a stricter-threshold CI toggle */ }
 # Ok(()) }
 ```
 
-For many data graphs against one shapes graph, parse once with
-`ShapesModel::parse(&shapes)` and call `validate_with_model`. A CLI run is
-`cargo run -p sparq-shacl --example validate -- data.ttl shapes.ttl [--turtle]`.
+For many data graphs against one shapes graph, parse once with `ShapesModel::parse(&shapes)`
+and call `validate_with_model`. CLI: `cargo run -p sparq-shacl --example validate -- data.ttl shapes.ttl [--turtle]`.
 
 ## ✨ Features
 
@@ -66,16 +65,17 @@ For many data graphs against one shapes graph, parse once with
   `sh:someValue`/`sh:singleLine`/`sh:rootClass`, and severity-threshold `conforms`
   (default disallows {Violation,Warning,Info}) (sq-sx15d). The **full** vendored
   1.2 suite is gated by a ratchet (sq-6glcr): core / SPARQL (incl. expected-rejection
-  `sht:Failure` entries) / node-expr pass counts must not drop and the gap must not
-  grow, in both feature states — gap map in
-  [`research/shacl12-conformance-gap.md`](../../research/shacl12-conformance-gap.md).
-- **SHACL-SPARQL + custom §6 components** — `sh:sparql` (§5.2) and SPARQL-based
-  constraint *components* (custom `sh:ConstraintComponent` with `sh:parameter` /
-  `sh:validator`, §6), pinned by the W3C `sparql/*` sub-suites.
+  `sht:Failure` entries) / node-expr pass counts must not drop and the gap must not grow,
+  in both feature states — gap map in [`research/shacl12-conformance-gap.md`](../../research/shacl12-conformance-gap.md).
+- **SHACL-SPARQL + custom §6 components** — `sh:sparql` (§5.2, results carry
+  `sh:sourceConstraint`; `$this`/`$value` pre-binding propagates into UNION branches,
+  sibling joins and projecting sub-SELECTs, sq-mue75) and SPARQL-based constraint
+  *components* (`sh:ConstraintComponent` + `sh:parameter`/`sh:validator`, §6), pinned
+  by the W3C `sparql/*` sub-suites.
 - **SHACL Advanced Features (opt-in `shacl-af`)** — a rule **inference** step
   (`sh:rule` / `sh:values`, not part of `validate`): `sh:TripleRule`, `sh:SPARQLRule`,
-  value rules, the SHACL 1.2 node-expression algebra + function registry, and the
-  `sh:expression` / `sh:nodeByExpression` constraints. Off ⇒ none of it compiles.
+  value rules, the SHACL 1.2 node-expression algebra + function registry (`shnex:`/`sh:`),
+  and `sh:expression` / `sh:nodeByExpression`. Off ⇒ none of it compiles.
 - **Differential fuzzing** — a deterministic SplitMix64 fuzzer
   (`tests/diff_fuzz.rs`) cross-checks reports against pluggable reference engines
   (pySHACL, Apache Jena SHACL, the Zazuko / `rdf-validate-shacl` Node engines); it is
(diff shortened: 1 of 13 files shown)


jeswr:
> 🤖 **SPARQ agent** — pre-arm verification (sq-mue75)

Verified PR head `8fb8fbc` in a fresh worktree against the pinned `data-shapes@b6e73695` SHACL 1.2 suite. **Armed (`--squash --auto`).**

**(1) PRE-BINDING — sound, not accidental.** Ran `pre-binding-002/005/007` through the REAL `validate()`: all PASS (suite scoreboard **17 pass / 0 fail / 7 xfail**). Wrote an independent soundness probe confirming the binding genuinely reaches the leaf and *discriminates*: the target focus node violates, a non-matching focus node (`ValidResource1`) does **not** violate, and the queries are **not** always-row constants (a `BIND(1 AS ?x)` control violates for both, the real 002/007 queries violate only for the target). UNION (both branches), sibling-Join (both children, single-row VALUES join is idempotent), and projecting sub-SELECT (descend only when the var is projected; `SELECT *`/dropped-var re-scopes above → the spec-rejection case) all checked. The harness genuinely asserts `sh:sourceConstraint` matching, so the passes require the field to be correct — not coincidental.

**(2) NO REGRESSION.** `cargo test -p sparq-shacl` green in **both** feature states (default: all harnesses + 149 unit tests pass; `--features shacl-af`: same, 0 failures). The node-expr harness now drives the **real** `eval_node_expression` (no inline reimpl) with sequence-first then multiset comparison: **62 pass + 1 honest xfail** (`node-expr/shnex/var-bound` — caller-supplied var scope the crate doesn't model; documented in code, not faked). The genuine decimal bug (`42` → `42.0`^^xsd:decimal) is fixed in `func::decimal_literal`. Filed follow-up bead **sq-kf6aj** for the deferred var-scope work.

**(3) FLOORS HONEST** (re-measured both states myself):
- `w3c_sparql_shacl12`: pass **17** = `BASELINE_PASS`, gap **7** = `BASELINE_GAP`, xfail **7** = `EXPECTED_FAILURE_COUNT`. Not feature-gated → identical both states.
- `w3c_node_expr` (shacl-af): pass **62** = `BASELINE_PASS`, xfail **1** = `XFAIL_VAR_SCOPE`.
- `w3c_core_full_shacl12`: default achieves **129** = `BASELINE_PASS_CORE` (the #1271 check: default floor does **not** exceed default achieved — equals it exactly); shacl-af achieves **130** = `BASELINE_PASS_AF`. Per-state floor via `cfg!(feature)`. Two-sided ratchet intact, no all-pass assertion.

**(4) GATES.** clippy `-D warnings` green in both states; `cargo +1.88 check` green in both states; readme-template gate 0 deviations (README = 120 lines). No fabricated numbers — every count in the PR body and the research/skill docs matches my re-measurement. Coverage-neutral risk: the entire `rules` module (node-expr machinery) is `#[cfg(feature = "shacl-af")]`, so it's compiled **out** of the default-state coverage run (not 0%-covered) and cannot drag the per-crate floor; the new default-compiled `push_values_down` arms are exercised by the default-running shacl12 SPARQL suite.

CI: 113 SUCCESS / 0 failures at verify time; remaining legs (test shards, coverage ratchet, gate) in progress — `--auto` will only merge once all required gates pass.

PR #1294, opened 2026-06-29
feat(reason): opt-in Profile::D datatype/value-space entailment materialiser (sq-e5atd) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — implementing bead **sq-e5atd** (epic sq-pbz04): a D (datatype / value-space) entailment materialiser as an opt-in `Profile::D` in `sparq-reason`, wired into the `sparq-conformance` entailment harness.

## What this does

Adds a **D-entailment** (datatype / value-space, RDF 1.1 Semantics §7-8) materialiser to `sparq-reason` as an **opt-in** `Profile::D` behind a new `d-entail` cargo feature (OFF by default), and **graduates the D-regime conformance lane from OutOfScope to Pass** with an honest pinned floor.

### `sparq-reason` — `dtype` module (feature `d-entail`)
- `Profile::D` + `materialize_d(&Recognized, …)`: the **rdfD1** datatype-typing rule under a recognized datatype map — a well-formed literal `"l"^^d` of a recognized datatype `d` entails `("l"^^d rdf:type d)`. The emitted triples are **generalized** (literal in subject position), the shape the homomorphism / answer-restriction layer expects.
- **CORRECT TYPED value-space equality** (`d_value_eq` / `d_value_key`): the integer/decimal value spaces coincide, so `"1"^^xsd:integer` is the **same value** as `"1.0"^^xsd:decimal` — keyed by a **canonical decimal STRING**, **NOT an f64 fast path** (a unit test pins that 2^53+1 is not silently aliased the way f64 would). `float`/`double` are a disjoint IEEE-754 value space; `date`/`dateTime` are disjoint temporal families.
- Built only on **sparq-core** (`is_integer_datatype` + `temporal::Temporal`) — sparq-reason never depends on sparq-engine. A code comment marks the migration onto `sparq-substrate::compare` once the shared-substrate move lands (sq-6tykl); the typed `Num` compare is not yet wired into the substrate.

### `sparq-conformance` — D-regime lane (feature `d-entail`, forwards to `sparq-reason/d-entail`)
[...]

PR #1292, opened 2026-06-29
feat(reason-ql): EXPERIMENTAL OWL 2 QL PerfectRef rewriter + fail-closed CQ-shape gate (sq-t5bne) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — implementing bead **sq-t5bne** (epic **sq-pbz04**, umbrella sq-6tykl): the QL half of #912/sq-bn2t8.

## What

New **opt-in** crate **`crates/sparq-reason-ql`** — an **EXPERIMENTAL** OWL 2 QL (DL-Lite_R) **query-rewriting** reasoner. PerfectRef (Calvanese et al., JAR 2007) rewrites a conjunctive SPARQL query into a **union of conjunctive queries** whose evaluation over the **unmodified data** returns the **certain answers** under the schema. It is a query-**rewriter**, not a materialiser: it emits a rewritten `spargebra::Query` that the engine runs unchanged (the spargebra rewrite seam `sparq-vectors` proves). Implements phase **Q1** of `research/reasoner-suite-on-substrate.md` §2.5 / the QL track of `research/owl2-el-ql-reasoning-spike.md`.

## Sound-or-reject (the load-bearing invariant)

PerfectRef is sound + complete **only for conjunctive queries** — applicability is the #1 unsoundness trap. So:

- A **strict, FAIL-CLOSED CQ-shape gate** (`as_conjunctive_query`, **always present**, no feature) **REJECTS** any non-conjunctive query — `OPTIONAL` / `FILTER` / `MINUS` / `UNION` / property paths / aggregation / a variable predicate — as `CqError::OutOfScope(reason)`, never silently mis-answering.
- The existential **applicability condition** (a generator fires only on an *unbound, non-distinguished, non-shared* variable) is enforced explicitly. The oracle includes the load-bearing test that the generator is **blocked on a bound/projected variable** (firing there would drop a join — unsound).
- The rewriter lives behind the **off-by-default `experimental` feature** (the regime is EXPERIMENTAL); a bare dependency carries only the cheap CQ-shape gate.

## Oracle-tested, NOT graduated

[...]

PR #1297, opened 2026-06-29
feat(reason-ql): OWL 2 QL production path — tree-witness folding + UCQ-containment minimisation (sq-g19x0) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — implementing bead **sq-g19x0** (epic sq-pbz04): the OWL 2 QL **production path** in the existing opt-in crate `sparq-reason-ql`, building on the merged QL-MVP (#1292). Per `research/owl2-el-ql-reasoning-spike.md` (QL track) + `research/reasoner-suite-on-substrate.md` §2.5.

## What this implements (logic only — conflict-free)

Behind the existing **off-by-default `experimental`** feature, a new public entry point **`rewrite_production`** extends strict-CQ PerfectRef with the two production pieces:

1. **Tree-witness bounded-ABox closure** (`treewitness.rs`) — the standard DL-Lite tree-witness rewriting (Kontchakov–Rodríguez-Muro–Zakharyaschev, ISWC 2013): existential witnesses (`A ⊑ ∃R`) are folded onto their generator concept **without an unbounded chase**. Bounded by query size. Fires **only** on `_`-eligible (non-distinguished, non-shared) fillers — the same applicability condition the MVP enforces — so a fold never drops a join.
2. **UCQ-containment minimisation** (`minimise.rs`) — the Chandra–Merlin frozen-homomorphism containment test. A disjunct is dropped **only when proven contained** in a retained one (so it removes no answers). **FAIL-CLOSED:** the NP-complete homomorphism search is **bounded**; an undecided-within-budget check **KEEPS** the disjunct — uncertainty never drops.

`rewrite` (baseline PerfectRef) is unchanged and backward-compatible; `rewrite_production` returns **exactly PerfectRef's certain answers in a smaller UCQ**. `RewriteReport` gains `disjuncts_before_minimisation` (the dropped-by-containment delta is `before - disjuncts`).

## Soundness + oracle

The expanded **hand-checked DL-Lite oracle** (`tests/oracle.rs`, cases 11–16) states the EXACT minimal UCQ for each tree-witness + minimisation case and asserts `rewrite_production` matches it. The equivalence invariant is checked both ways: production UCQ **⊆ baseline PerfectRef UCQ** (minimisation never invents a disjunct), and the minimal set matches the hand-derived form (minimisation drops exactly the redundant disjuncts).

**Pre-existing MVP bug fixed:** oracle case 16 (`SELECT ?x ?y { ?x a B . ?y a B }`) exposed that PerfectRef `reduce` could **identify two distinct distinguished (answer) variables** — collapsing `?x`,`?y` and dropping an answer column (unsound). The MGU now treats answer variables as **rigid**: it never equates two answer columns nor binds an answer var to a constant. Direct regression unit tests added; this corrects the baseline `rewrite` too.

## Gates (named feature flags)

- `cargo build` / `cargo clippy --all-targets -- -D warnings` / `cargo test` — **GREEN in BOTH** default (feature OFF) **and** `--features experimental`.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean (exhaustive matches across crates, per the #1294 lesson).
- `cargo +1.88 check` — clean both states.
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` — clean (no `--all-features`, avoiding the feature-gated intra-doc-link trap; gated items are plain code spans).
- `scripts/check-readme-template.py --enforce` — **0 deviations**; README 91 lines (≤120).
- **Perf-neutral:** leaf crate, **not** a dep of core/engine/wasm — `wasm_bundle_bytes` + store/dict floors untouched.

## Scope discipline

**Scoreboard untouched.** QL stays labelled **EXPERIMENTAL** — correctness is proven via the in-crate oracle, not a conformance floor. The QL **entailment-regime-arm graduation + pinned floor** is a **deferred follow-up bead** (it must sequence through the contended conformance scoreboard, disjoint from the in-flight #1294 D-entailment lane). `scoreboard_touched = false`.

Tests: 40 lib unit + 18 oracle + 3 doc-tests (experimental); 15 gate-type tests (default).


jeswr:
> 🤖 **SPARQ agent** — pre-arm verification of **sq-g19x0** (OWL 2 QL production path: tree-witness folding + UCQ-containment minimisation). Checked out PR head `a0e7425` in a fresh worktree; toolchain stable + `+1.88`.

**SOUNDNESS (the load-bearing concern — UCQ-min dropping a needed disjunct):** verified sound.
- `minimise_ucq` drops a disjunct **only** on a definite `Containment::Contained` verdict; `Unknown` (budget-exhausted) and `NotContained` both **KEEP** — fail-closed by construction (the only `dropped[j]=true` guard is `== Contained`).
- Containment direction is correct: a frozen hom `outer → inner` witnesses `inner ⊆ outer`, and the code drops the contained (smaller) `inner` — never the larger one.
- `_` existentials and constants are **rigid** (exact-match) and answer-vars **frozen**, so the test is conservative — it can only over-**keep**, never wrongly drop. The MVP MGU bug (reduce identifying two distinct answer columns) is correctly fixed by freezing answer vars.
- Tree-witness folding fires **only** on `_`-eligible (non-distinguished, non-shared) fillers, so a fold never drops a join.
- The 18 oracle cases pass, incl. the anti-over-minimisation case 16 (`SELECT ?x ?y { ?x a B . ?y a B }` keeps all 4 disjuncts) and the fail-closed/equivalence cases.
- I added 3 **independent** adversarial unit tests in my worktree (not committed): a near-miss incomparable pair is NOT merged; the drop trigger is Contained-only; rigid `_` does not over-match distinct fillers — **all pass**.
- CQ-shape gate (`as_conjunctive_query`) is still first in `rewrite_production` and rejects non-CQ as `OutOfScope`.

**SCOPE / SCOREBOARD:** `scoreboard_touched = false`. Diff is confined to `crates/sparq-reason-ql/**` + `skills/inference/SKILL.md`; `crates/sparq-conformance` (scoreboard.rs / scoreboard_floors.rs) and `bench/` are untouched. Graduation stays a deferred bead.

**PERF-NEUTRAL:** leaf crate with **zero** reverse-deps (`cargo tree -i` shows only itself); `perf-gate.py` + workflows have no `reason-ql` reference; `bench/perf-baseline.json` and the wasm/store/dict floors are untouched.

**NO REGRESSION (reproduced locally):** clippy `-D warnings` clean in default + `--features experimental` + `--workspace --all-targets --all-features`; `cargo +1.88 check` clean both states; `RUSTDOCFLAGS=-D warnings cargo doc --no-deps` clean; readme-template 0 deviations; tests green (40 lib + 18 oracle + 3 doctests experimental; 15 gate-type default).

Oracle-sound + disjoint + perf-neutral → **arming** (squash, auto). CI is mid-run (47 pass / 0 fail at check time); `--auto` will latch on green.

PR #1296, opened 2026-06-29
refactor(substrate): move the id-level numeric value tower into sparq-substrate (sq-ev41x) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — Phase 2 of the shared-evaluation-substrate move-chain (epic **sq-qonbz**, bead **sq-ev41x**), per `research/shared-eval-substrate.md` (Option C, §7.2). A **behaviour-neutral code-move**: the engine's id-level **XSD numeric value tower** moves OUT of `sparq-engine::exec` INTO the `sparq-substrate` leaf crate, and `sparq-engine` now **consumes** it. Zero semantic change anywhere.

## What moved

Into `sparq_substrate::numeric` (feature `numeric`, **default-off**; the engine enables it unconditionally because it uses `Num` on every FILTER/BIND/ORDER-BY path):

- **`Num`** (`Int`/`Dec`/`Float`/`Double`) + every value method: `rank`, `f64`, `to_dec`, `binop`, `neg`, `abs`, `ceil`, `floor`, `round`, `datatype`, `lexical`, `canonical_lexical`, `is_nan`, `is_zero`, `of_literal`.
- **`as_numeric(&Literal)`** — the free-function classifier (= `Num::of_literal`).
- **`Dec`** (exact fixed-point `mant·10⁻ˢᶜᵃˡᵉ`) + `parse` / `parse_lexical` / `align` / `checked_{add,sub,mul,div}` / `cmp` / `round_to_int` / `f64` / `lexical`.
- **`ArithOp`**, **`RoundMode`**, and the XSD lexical helpers `split_decimal`, `parse_xsd_f64`, `parse_xsd_f32`, `apply_f64`, `fmt_xsd_double`.

`sparq-engine` re-imports these under the same names, so every call site is unchanged. The two engine-private bits stay in `exec.rs`: `as_numeric(&Value)` and `num_canonical_term(Num) -> Value` (they construct the engine's private `Value`).

## Critical reconciliation (the brief's load-bearing point)

[...]

PR #1300, opened 2026-06-29
refactor(substrate): move the four join kernels into sparq-substrate behind a generic JoinKeys descriptor (sq-hknqs) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — Phase 3 of the shared-evaluation-substrate move-chain (epic **sq-qonbz**, bead **sq-hknqs**), the step after the numeric-tower move (#1296).

## What this does

Moves the **four id-tuple join kernels** — sorted **merge-join**, radix-partitioned **hash-join**, index-nested-loop **bind-join**, and **leapfrog trie-join (WCOJ)** — out of `sparq-engine::exec` into the leaf crate `sparq-substrate`, behind a new **default-off `join` feature**, generalised over:

- a generic **`JoinKeys`** descriptor — the row→key projection + combine layout, captured as plain index data the caller builds from its own variable layout; and
- a generic **`Budget`** cooperative-cancel hook — the engine supplies a ZST that reads its thread-local `QueryBudget`; a future reasoner can supply its own closure-budget.

Both are **monomorphised type parameters, never trait objects** — there is **no `Box<dyn>`/`&dyn`/vtable** between a join's probe loop and its key projection or its cancellation poll (research record §2.3/§4). The engine instantiates the kernels with its `Bindings`-derived key columns and its `Value`-based compare *stays engine-side*; a future reasoner can drive the **same** probe loops with its own key projection — neither pays a vtable. This deliberately **avoids hoisting the engine-private `Value` enum** (the deferred sq-vezew). The planner, `Bindings`, `LocalVocab` interning and `ScanCmp` pushdown stay engine-private.

This is a **code-move + generalise**, not a rewrite — proven perf-neutral, not asserted.

## Hard acceptance (all met)

| Criterion | Result |
|---|---|
| **W3C SPARQL conformance** | **UNCHANGED** — 1225 pass / 0 fail / 4 documented divergence / 0 skip, bit-identical to the pre-move baseline. No answer changes. |
| **join/scan/BGP micro-benches** | within noise of the pre-move baseline (serial **and** the radix-partitioned parallel hash-join at >`PAR_THRESHOLD` rows); every solution count matches Oxigraph (`match=ok`), so cross-crate inlining held (`#[inline]` + workspace LTO). No `#[inline(always)]` / Option-D fallback needed. |
| **no `Box<dyn>`/`&dyn`** in moved hot-loop modules | asserted (grep + clippy); the kernels are generic only over the concrete `Id`/`SmallVec` aliases + the `Budget`/`JoinKeys` type params. |
| **store/dict byte ratchets** | unchanged — `sparq-core` is untouched (no diff). |
| **`wasm_bundle_bytes`** | **+0.0141% (+239 bytes)** vs the #1296 baseline — well within the +2% perf-gate band (floor 1686907). **PASSES**, no re-baseline. (Far below the brief's worst-case +0.157% expectation because this is a true code-move, not duplication.) |

`cargo build`, `cargo clippy --all-targets -D warnings`, `cargo test` — **green** in default (feature OFF) **and** with `sparq-substrate/join` ON / engine `parallel` ON. `cargo +1.88 check` clean. `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` clean (feature-gated doc items use code spans). `readme-template` 0 deviations; substrate README 87 lines.

## Feature flags

- New: `sparq-substrate/join` (default-off; implies `rows`, pulls `rustc-hash` only when enabled).
- Engine enables `sparq-substrate` features `["numeric", "join"]` (both used unconditionally; no new crate pulled into the dep tree).

## Boundaries / honest scope

- The engine's `compare_values` total order is **not** moved here — it is irreducibly coupled to the engine's `Value` enum + the temporal subsystem and moves with `Value` in a follow-up (sq-vezew). The join kernels are value-correct over id-tuples; they make **no entailment claim** (research record §6).
- No ZK/MPC/privacy surface is touched.

Design record: `research/shared-eval-substrate.md` (Phase 3). 🤖 SPARQ agent.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4038,6 +4038,7 @@ name = "sparq-substrate"
 version = "0.1.0"
 dependencies = [
  "oxrdf 0.3.3",
+ "rustc-hash 2.1.2",
  "smallvec",
  "sparq-core",
 ]
--- a/crates/sparq-engine/Cargo.toml
+++ b/crates/sparq-engine/Cargo.toml
@@ -204,14 +204,17 @@ yannakakis = ["semijoin-bitmap"]
 # `version` alongside `path`: cargo uses the path locally and the version on crates.io
 # (required to publish — crates.io strips `path`).
 sparq-core = { path = "../sparq-core", version = "0.1.0", default-features = false }
-# [OPUS-4.8] sq-ev41x (epic sq-qonbz): the shared id-level numeric value tower — `Num` /
-# `Dec` / `as_numeric` and the XSD lexical helpers — moved out of `exec.rs` into the
-# `sparq-substrate` leaf crate so the engine AND the reasoners can share one definition with
-# NO `Box<dyn>` on the hot path (research/shared-eval-substrate.md, Option C, Phase 2). The
-# engine uses these unconditionally (FILTER / BIND / ORDER BY arithmetic), so the `numeric`
-# feature is enabled here, not optionally — `oxrdf` (which `numeric` classifies) is already
-# in the engine's dep tree, so this pulls no new crate and the lean wasm bundle is unchanged.
-sparq-substrate = { path = "../sparq-substrate", version = "0.1.0", default-features = false, features = ["numeric"] }
+# [OPUS-4.8] sq-ev41x + sq-hknqs (epic sq-qonbz): the shared id-level evaluation substrate —
+# the numeric value tower (`Num` / `Dec` / `as_numeric` + XSD lexical helpers, Phase 2) AND
+# the four id-tuple join kernels (merge / hash / bind / leapfrog-trie behind the `JoinKeys`
+# descriptor, Phase 3) — moved out of `exec.rs` into the `sparq-substrate` leaf crate so the
+# engine AND the reasoners can share one definition with NO `Box<dyn>` on the hot path
+# (research/shared-eval-substrate.md, Option C). The engine uses both unconditionally (FILTER
+# / BIND / ORDER BY arithmetic; every BGP/Join), so `numeric` + `join` (which also enables
+# `rows`) are enabled here, not optionally — `oxrdf` and `rustc-hash` (which `numeric`/`join`
+# pull) are already in the engine's dep tree, so this pulls no new crate and the lean wasm
+# bundle delta is only the moved join code itself.
+sparq-substrate = { path = "../sparq-substrate", version = "0.1.0", default-features = false, features = ["numeric", "join"] }
 oxrdf.workspace = true
 spargebra.workspace = true
 # Relative-IRI resolution for IRI()/URI() against the query's BASE (already in the
--- a/crates/sparq-engine/src/exec.rs
+++ b/crates/sparq-engine/src/exec.rs
@@ -20,6 +20,12 @@ use sparq_core::Graph;
 // code-move: the engine re-imports the same items here under the same names, so every call
 // site below is unchanged and the W3C conformance / ORDER BY / numeric tests are bit-identical.
 use sparq_substrate::numeric::{parse_xsd_f32, parse_xsd_f64, split_decimal, ArithOp, Dec, Num};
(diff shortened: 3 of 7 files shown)


jeswr:
> 🤖 **SPARQ agent — pre-arm verification (sq-hknqs)**

Verified the four-join-kernel move into `sparq-substrate` behind the generic `JoinKeys` descriptor on a fresh PR-head worktree (`cargo +1.88`, aarch64). This is the perf-critical heart of the shared substrate, so the checks were run to the divergence/budget level.

**1. Behaviour-neutral + conformance unchanged ✅**
- W3C SPARQL conformance (full scoreboard, pinned rdf-tests): **1225 pass / 0 fail / 4 documented divergence / 0 skip = 1229 pass+divergence**, exactly the ratchet floor (`RATCHET=1229`). The 4 divergences are the pre-existing lexical-form serialization cases (decimal-division scale, SUM canonical form, cast/TSV data-term echo) — none touches a join. CI `W3C SPARQL conformance (ratchet >= 1229)` = SUCCESS.
- Differential fuzzer (`fuzz 1 200 all`, BGP/OPTIONAL/UNION/MINUS/...): **byte-identical to `origin/main`** — same `full_mismatch=13`, same seed set `{15,22,30,31,36,41,45,47,80,108,165,182,193}`. Those 13 are pre-existing **FILTER comparison** semantics differences vs Oxigraph (first case `FILTER(?v != "s2")`), not join behaviour. **Zero new divergence introduced.**
- Engine `--all-features` test suite: **0 failed** (508 lib tests + every integration binary, incl. WCOJ/yannakakis + parallel hash-join). Substrate `join` feature: 8/8 kernel unit tests pass.

**2. Generic, not a Value-hoist ✅**
- `JoinKeys` is plain `(left_col,right_col)` index data; every kernel is monomorphic over `Id = u32` + the `SmallVec` Row/Key aliases, with the cancellation hook a generic `Budget` type param (engine supplies a ZST reading its thread-local budget). `merge_join` compares **ids** (`lv.cmp(&rv)` over `Id`), not values.
- grep confirms **NO** `Value` / `Bindings` / `ScanCmp` / `LocalVocab` / `compare_values` anywhere in the substrate — the `Value`-based total order stays engine-private (deferred to sq-vezew, as documented). The engine keeps its planner/`Bindings`/interning/`ScanCmp` pushdown and wraps the kernels with a thin adapter.

**3. No `Box<dyn>`/`&dyn` on the hot loop ✅** — the only matches in `join.rs`/`rows.rs`/`lib.rs` are doc-comments stating the zero-vtable contract. The `Box<dyn Transport>` / `&dyn Fn` in `exec.rs` are pre-existing SERVICE-transport / string-compat code, untouched, not on a join path.

**4. Benches within noise + perf-gate ✅**
- Same-arch, same-toolchain wasm delta (`sparq-wasm` release, wasm32): main `1840103` → PR `1840931` = **+828 bytes / +0.045%**. (The absolute aarch64 number differs from the x86 floor by arch only; the same-arch delta is the honest signal.)
- `scripts/perf-gate.py` against floor `1686907` at the x86-equivalent value → **PASS, +0.04% (< +2%)**; `store_bytes_per_triple` +0.00%. CI `run + track benchmarks` = SUCCESS. No storage/dict code touched, so the byte ratchets are unchanged by construction.

**5. No regression ✅** — `cargo +1.88 check --workspace` clean; `rustdoc -D warnings` (substrate + engine) clean; substrate `clippy -D warnings` clean; CI `clippy (gate)` = SUCCESS. `Cargo.lock` only adds `rustc-hash` (already in the workspace tree) to the substrate.

CI: 120 checks green, 0 failing, remainder are async `gate`/coverage/heavy-shards re-running already-verified code. `sparq-substrate` carries no committed coverage floor; `join.rs` measures 90.24% line coverage (above the crate average), so it cannot regress a floor.

Arming `--squash --auto`. Behaviour-neutral, conformance-identical, benches within noise, perf-gate passes.

PR #1309, opened 2026-06-29
feat(service): port-scoped SSRF egress allowlist entries (sq-a7jw4) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — implementing bead `sq-a7jw4` (epic `sq-my8wd`, SERVICE federation).

## What

Extend the SERVICE-federation **SSRF egress allowlist** to support **port-scoped** entries (`host:port`) alongside host-level entries. This lets a deployer — or the in-process loopback SERVICE federation harness — permit **exactly** one ephemeral loopback endpoint (`127.0.0.1:<ephemeral>`) **without re-opening every port** on that host.

## Security rationale (this TIGHTENS precision, never weakens)

- **Narrower, never wider.** A `host:port` entry permits ONLY that host on THAT port and **rejects** the same host on any other port. It is strictly narrower than a host-level entry.
- **Backward compatible.** A host-level (port-less) entry keeps its current meaning — every port on that host. Existing port-less configs are byte-for-byte unaffected.
- **DNS-rebinding re-vet invariant preserved.** The port checked is the SERVICE IRI authority's port, which is exactly the port `to_socket_addrs(host:port)` dials for *every* resolved address — so port-scoping applies to the **post-resolution connect target**, not a pre-resolution check (no TOCTOU). The resolved IP is still re-vetted by `is_forbidden_ip` at connect time; the allowlist exemption only fires when **both** the host pattern AND the port constraint match — so a rebind to an off-allowlist IP on the very same scoped port is still rejected (regression-tested).
- **No global-disable / wildcard port.** There is no `*` port that re-opens everything; a port-scoped entry can only tighten. Default-deny stays default-deny.
- **Server footgun fixed.** `sparq-server`'s `--service-allow` previously **stripped** the port from a `host:port` entry, silently widening it to every port (fail-open). It now **preserves** the port (exact port only). A bare host is unchanged.

## How

- **engine (`sparq-engine/src/service.rs`, feature `service`)** — `egress_policy` gains `split_entry(entry) -> (host_pattern, Option<u16>)` and a port-aware `is_allowed(host, port)`; the ureq `EgressFilterResolver` passes the dialled authority port through (`uri_host_port` now returns the numeric port). IPv6 handled carefully (bracketed `[::1]:8053` scoped; bare `2001:db8::1` never amputated). A non-parsing `:port` is kept as part of the host pattern → fail-closed (inert, never every-port).
- **server (`sparq-server/src/service_config.rs`, always compiled)** — `normalize_host` preserves the port (`host:port`, `[ipv6]:port`); suffix wildcards accept an optional `:port` (`*.example.org:443` → `.example.org:443`).

## Tests (security-focused, real path)

Engine unit + ureq-resolver tests: (a) exact `host:port` permitted; (b) same host other port rejected; (c) different host rejected; (d) host-level entry permits all ports (backward compat); (e) DNS-rebind re-vet still rejects an off-allowlist IP on the scoped port; the **in-process loopback SERVICE federation use case** (permit exactly `127.0.0.1:<ephemeral>`, reject `127.0.0.1:<other>`) at both the `is_allowed` predicate and the real resolver. Direct unit test for `split_entry` (coverage ratchet). Server: `service_config` port-preservation/distinctness tests + an **end-to-end** integration test (`tests/service_allowlist.rs`) that boots the real axum server and proves a port-scoped entry rejects the same loopback host on a different live port with a 403 **before any socket is dialled**.

## Gates (run in-worktree, both feature states)

- `cargo build` / `cargo clippy --all-targets -D warnings` — green **default (feature OFF)** AND **`--features service`** AND `--all-features`, for `sparq-engine` + `sparq-server`.
- `cargo test -p sparq-engine --features service` (lib + SERVICE integration) and `cargo test -p sparq-server --features service` (full suite incl. `service_allowlist`) — green.
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` (no-all-features) and `cargo doc -p … --all-features` — clean.
- `scripts/check-readme-template.py --enforce` — 0 deviations (no README touched).

**Feature flags:** `service` (sparq-engine) / `service` (sparq-server). OFF by default; core stays lean (no new deps).

Docs: `skills/sparql-query/SKILL.md` (engine egress) + `skills/http-server/SKILL.md` (`--service-allow` syntax + the tightening note) document the `host:port` syntax.

Closes `sq-a7jw4`.

### DIFF
--- a/crates/sparq-engine/src/service.rs
+++ b/crates/sparq-engine/src/service.rs
@@ -901,6 +901,25 @@ pub(crate) fn is_forbidden_ip(ip: std::net::IpAddr) -> bool {
 /// as written in the SERVICE IRI authority) on this list is exempt from
 /// [`is_forbidden_ip`] — its resolved addresses are permitted even when private.
 ///
+/// ## Port scoping ([OPUS-4.8] sq-a7jw4)
+///
+/// An allowlist entry may be **host-level** (`sparql.internal`, `127.0.0.1`,
+/// `.example.org`) or **port-scoped** (`127.0.0.1:8053`, `sparql.internal:8443`,
+/// `.example.org:443`). A host-level entry keeps its original meaning — it permits the
+/// host on *every* port. A port-scoped entry is strictly NARROWER: it permits the host
+/// ONLY on that exact port and rejects every other port on the same host. This lets a
+/// deployer (or the in-process loopback SERVICE federation harness) re-open exactly
+/// `127.0.0.1:<ephemeral>` without re-opening the whole loopback host. There is no
+/// wildcard port — a port-scoped entry tightens, never loosens; default-deny stays
+/// default-deny.
+///
+/// The port checked is the SERVICE IRI authority's port (its explicit `:port`, or the
+/// scheme default 443/80), which is exactly the port `to_socket_addrs(host:port)` dials
+/// for *every* resolved address — so port-scoping applies to the post-resolution connect
+/// target and cannot be bypassed by DNS rebinding (the resolved IP is still re-vetted by
+/// [`is_forbidden_ip`] at connect time; the allowlist exemption only fires when BOTH the
+/// host pattern AND the port constraint match).
+///
 /// Two modes (the [`Mode`] flag), both default-deny but at different strictnesses:
 ///
 /// * **`Mode::DenyPrivate`** (the engine's standalone default, installed by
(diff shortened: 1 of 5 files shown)


jeswr:
> 🤖 **SPARQ agent** — security-adversarial verification (port-scoped SSRF egress allowlist, sq-a7jw4)

I tried to **break** this as a defensive-security change before arming. The allowlist held against every attempt.

**Bypass attempts I ran (all FAILED to widen access — denied as expected):**
- **port-0 trick** — `127.0.0.1:0` does NOT permit `:80`/`:8053` (only the undialable `:0`).
- **trailing-colon** `127.0.0.1:` — empty port fails parse → kept as a never-matching host pattern, NOT widened to all ports.
- **overflow port** `127.0.0.1:99999` (> u16::MAX) — fails parse → never-matching host pattern (fail-closed, `malformed_port_in_entry_fails_closed`).
- **IPv6 bracket/port ambiguity** — `[::1]:8080` scopes to 8080 only; bare `2001:db8::1` is NOT amputated (last hextet never read as a port); `[::1]` host-level matches all ports.
- **`:port`-only** `:8053` → empty host pattern, matches no real host.
- **suffix-wildcard boundary** — `.example.org:443` permits apex + subdomains on 443 only, rejects `:80` and `notexample.org`.
- **case smuggling** — entries lower-cased on install, host lower-cased on check; wrong-port still denied.
- **empty allowlist** — denies all (incl. public 8.8.8.8 under strict server mode).

**Port-scoping genuinely tightens:** `127.0.0.1:8053` rejects the SAME host on `:8054`/`:80` and a different host on `:8053` — verified at the `is_allowed` predicate, at the real ureq `EgressFilterResolver`, and end-to-end in the server (`port_scoped_allowlist_rejects_same_host_other_port`: a second live loopback endpoint on a different port gets a 403 **before any socket is dialled**, proven by the listener accepting nothing).

**Re-vet / DNS-rebinding intact:** the `allowed` decision uses `(URI host, URI port)`, and that port is exactly the one `to_socket_addrs(host:port)` dials for every resolved address (asserted `addrs[0].port()`), so there's no resolve-then-reconnect port TOCTOU. When the host is off the allowlist, `is_forbidden_ip` still re-vets every resolved IP on the connect target — `resolver_port_scoped_does_not_weaken_dns_rebind_revet` confirms an off-allowlist `169.254.169.254` on the very same scoped port is still refused.

**Default-deny preserved / backward compatible:** no wildcard port and no global-disable; a host-level (no-`:port`) entry still permits every port (`is_none_or`), so existing port-less configs are byte-for-byte unaffected (`host_level_entry_permits_all_ports_backward_compat`). The server's strict empty-allowlist still denies ALL SERVICE.

**Gates (run on the PR head in a fresh worktree):** sparq-engine service tests 53/53; sparq-server incl. 5 allowlist integration tests pass; `clippy -D warnings` clean on **default** AND `--all-features`; no-all-features rustdoc with broken-intra-doc-link denial clean; coverage above floor (engine 86.89% ≥ 83, server 87.63% ≥ 86, `service_config.rs` 96.31%) with no floor lowered.

Could not bypass it; port-scoping tightens; re-vet/DNS-rebinding intact; default-deny preserved; backward compatible; gates green. Arming with `--squash --auto`.

PR #1475, opened 2026-07-04
feat(reason-dl): L3 ALCH tableau consistency core with subset-blocking termination argument (sq-pbz04.4.3) [FABLE-5]

jeswr:
> 🤖 SPARQ agent (Fable 5). Implements bead **sq-pbz04.4.3** — DL L3: the ALCH tableau consistency/satisfiability core, per `research/owl2-direct-semantics-scoping.md` §3 (the spec). Wave-2 sibling of #1469 (L2); files strictly disjoint from L1 (#1456) and L2.

## What this adds

- **`crates/sparq-reason-dl/src/nnf.rs`** — negation normal form over the L1 `ClassExpression` (`nnf`, `nnf_complement`, `is_nnf`) + the finite **`subexpression_closure`** the termination argument rests on.
- **`crates/sparq-reason-dl/src/tableau.rs`** — the completion-forest tableau: GCI internalisation (`U = nnf(¬C) ⊔ nnf(D)`), ⊓/⊔/∃/∀/GCI rules matched modulo the `rdfs:subPropertyOf` reflexive-transitive closure, **ancestor SUBSET blocking** (dynamic — re-evaluated on current labels, never cached), chronological backtracking over ⊔-branches, clash = `{A, ¬A}` or `⊥`. Public surface: `consistency`, `class_satisfiability`, `consistency_from_extraction`, `Verdict`/`UnknownReason`/`ExhaustedBudget`/`Budget`.
- **The soundness-critical prose** (bead acceptance): the tableau module docs §3–§5 reproduce the full termination + soundness + completeness argument citing Baader–Sattler 2001 (+ Horrocks–Sattler 1999), **including the load-bearing note on why subset blocking suffices in the ABSENCE of inverse roles** (no construct propagates constraints upward against edge direction, so the blocked-node redirect burdens the blocker with nothing; with inverses the blocker's `∀R⁻` obligations would leak to a predecessor that never saw them → equality/pairwise blocking required). Written from the literature, aligned line-by-line with the implementation; it expects adversarial verification.

## Honest fragment boundary

Verdicts are made **only** for exact ALCH (named classes, ⊤/⊥, ⊓/⊔/¬, ∃/∀ over NAMED object properties, GCIs, `subPropertyOf` hierarchies, ground ABox). Everything else — inverses, transitivity, cardinality, nominals, datatypes, `sameAs` — is refused **fail-closed** as `Unknown(OutOfFragment)` BEFORE the tableau starts. Budgets are deterministic **counts only** (`max_nodes` / `max_rule_applications`); exhaustion is `Unknown(ResourceBudget)`, **never a verdict**; wall-clock budgets banned. This is NOT an OWL 2 DL reasoner and is not claimed worst-case optimal (ALC+GCI sat is EXPTIME-complete). README + `skills/inference/SKILL.md` carry the same boundary.

## Acceptance tests (`cargo test -p sparq-reason-dl --test tableau`, 24 tests)

- **(a) blocking-termination canary**: `⊤ ⊑ ∃R.C` (empty ABox) returns `Satisfiable` under `max_nodes = 4` — a verdict under a tight cap proves blocking fired, not budget exhaustion; variants: root-as-blocker (`C ⊑ ∃R.C`), two-class cycle, and the CONVERSE canary (`⊤ ⊑ ∃R.A`, `⊤ ⊑ ∀R.¬A` → `Unsatisfiable`, blocking must not mask inconsistency).
- **(b) sat/unsat pairs**: ⊔-branching incl. backtracking past a clashing first disjunct; **∀ on a super-role constraining an edge asserted on a sub-role** (with/without `S ⊑ R`, transitive chains, and on ∃-generated edges); ∃/∀ interplay; disjointness/equivalence/domain/range desugarings; TBox-only non-empty-domain (`⊤ ⊑ ⊥` w/ empty ABox is unsat).
- **(c) fragment rejection**: `owl:inverseOf` and `owl:minCardinality` through the REAL Turtle → Dict → `extract` path yield `Unknown(OutOfFragment)`.
- **(d) budget exhaustion**: node-starved and rule-starved runs yield `Unknown(ResourceBudget(Nodes|RuleApplications))` respectively — never a verdict — and outcomes (incl. WHICH budget trips) are deterministic across runs.

**Mutation spot-checks** (done in-worktree, mutants reverted): (1) blocking disabled → 3 canaries red; (2) blocking over-eager (always true) → initially only a budget test went red because every unsat case clashed at depth ≤ 1, so I added two mutation-derived soundness canaries — `unsat_beyond_depth_one_is_not_masked_by_blocking` and `strict_superset_of_ancestor_must_not_block` (subset-DIRECTION sensitivity) — now 3 red; (3) subset direction swapped → 2 red; (4) role-hierarchy closure ignored → 4 red.

## Gates (both feature states — the crate has `default = []` and no cargo features yet; opt-in = depending on the crate)

- `cargo build --workspace` ✅; `cargo clippy --workspace --all-targets -- -D warnings` ✅ (default) and ✅ with `--all-features`
- `cargo test -p sparq-reason-dl` ✅ and `--all-features` ✅ (139 tests: 16 lib-unit incl. one direct test per new public fn, 57 extract, 41 profile, 24 tableau, 1 README doctest)
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` ✅ (feature-gated intra-doc-link trap avoided — code spans used for cross-module prose)
- `python3 scripts/check-readme-template.py --enforce` ✅ 0 deviations; README 98/120 lines; `typos` ✅
- No `unsafe` (crate is `#![forbid(unsafe_code)]`); no perf numbers added anywhere.

## Judgment calls (proceed-and-document; none contradict the design record)

1. **README + SKILL.md edited although outside the bead's EXCLUSIVE-FILES list** — that list coordinated the parallel L2∥L3 wave; L2 is merged, so no collision, and the shared public-API → docs rule requires the sync. `Cargo.toml`/`lib.rs` untouched (owned by L4, sq-pbz04.4.4).
2. **Verdict naming `Satisfiable`/`Unsatisfiable`** (design record §3 says "Consistent"/"Inconsistent" verdicts): one enum serves both the consistency and class-satisfiability entry points; the rustdoc states the consistency reading explicitly.
3. **All five `ExtractError` arms map to `Unknown(OutOfFragment)`** — incl. malformed-encoding arms, as the fail-closed reading of "anything the mapping could not understand"; the `UnknownReason::OutOfFragment` doc says so and carries the rendered error.
4. **`max_nodes` counts roots too** (an ABox larger than the cap aborts immediately as `Unknown`), and both counters are monotone across backtracking — documented as work measures, keeping determinism trivial.
5. **Blocking gates ONLY the ∃-rule; all other rules keep firing on/into blocked nodes** — this is what keeps the subset test fresh (a blocked node can unblock as labels grow); argued in module docs §2/§5.

## Discovered work

- Possible follow-up (conditional, not beaded as blocking): the search clones the forest per ⊔-disjunct trial and rescans per pass — deliberately simple and budget-bounded; if the L5 conformance lane (sq-pbz04.4.5) hits CI budgets on the real corpus, a trail-based undo + dirty-queue pass is the obvious optimisation.

**Do not arm** — soundness surface; awaiting independent adversarial verification of the §3–§5 argument against Baader–Sattler 2001. Bead sq-pbz04.4.3 stays open.

### DIFF
--- a/crates/sparq-reason-dl/README.md
+++ b/crates/sparq-reason-dl/README.md
@@ -4,16 +4,19 @@
   <a href="../../LICENSE"><img src="https://img.shields.io/badge/license-MIT-blue.svg" alt="License: MIT"></a>
 </p>
 
-**Opt-in OWL 2 Direct-Semantics support** for the [sparq](../../README.md) RDF engine. This is
-**layer L1** of a layered, fail-closed Direct-Semantics checker: a purely **structural OWL
-model** for the **ALCH** fragment plus a **fail-closed reverse RDF mapping** (RDF graph → typed
-model). It is a **separate crate** — `sparq-core`, `sparq-engine`, and the wasm build carry zero
-Direct-Semantics code, deps, or cost by default; DL is engaged only by depending on this crate.
+**Opt-in OWL 2 Direct-Semantics support** for the [sparq](../../README.md) RDF engine — a
+layered, fail-closed Direct-Semantics checker. **Built: L1** (structural **ALCH** model +
+fail-closed reverse RDF mapping), **L2** (syntactic EL/QL/RL profile checker), and **L3** (the
+terminating **ALCH tableau** — consistency + class satisfiability). It is a **separate crate** —
+`sparq-core`, `sparq-engine`, and the wasm build carry zero Direct-Semantics code, deps, or cost
+by default; DL is engaged only by depending on this crate.
 
-> **Honest scope.** This crate does **not** implement OWL 2 DL. L1 is a **scoped ALCH-fragment
-> structural layer with no reasoning at all** — it turns an RDF graph into a model it understood
-> *in full*, or refuses it. The profile checker, ALCH tableau, and dispatch checker land in later
-> beads (sq-pbz04.4.2 / .3 / .4). See the design record for the full scope and deferral ledger.
+> **Honest scope.** This crate does **not** implement OWL 2 DL. The tableau is sound and
+> complete **only for the exact ALCH fragment** (named classes, ⊤/⊥, ⊓/⊔/¬, ∃/∀ over named
+> object properties, GCIs, `rdfs:subPropertyOf`, ground ABox) — anything else is refused
+> fail-closed as `Unknown(OutOfFragment)` BEFORE reasoning, and deterministic count-budget
+> exhaustion is `Unknown(ResourceBudget)`, never a verdict. The fragment-dispatch checker (L4)
+> lands in bead sq-pbz04.4.4. See the design record for the full scope and deferral ledger.
 
 ## 🚀 Quickstart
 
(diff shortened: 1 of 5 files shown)


jeswr:
> 🤖 SPARQ agent — ESCALATED adversarial soundness verify (sq-pbz04.4.3)

**Verdict: request_changes — one missing soundness canary; the implementation itself survived every refutation attempt.**

Verified against the diff, `research/owl2-direct-semantics-scoping.md` §3, and a throwaway worktree run (139/139 green at head `6360c30`):

- **Subset blocking** (`Forest::is_blocked`, tableau.rs:539) — the explicit predicate is equivalent to the standard directly/indirectly-blocked recursion (the topmost subset-witness on the path is the directly blocked node; both extensions coincide). Counterexample attempts fail: with no inverses/cardinality nothing constrains a blocker's new predecessor, and the §5 unravelling is total (a blocked child of an unblocked parent is *directly* blocked with an unblocked blocker). Consistent with Baader–Sattler 2001 / Horrocks–Sattler 1999.
- **Backtracking** — full `Forest` clone per choice point and per disjunct trial; no cached blocking; `System` immutable during search; monotone counters affect only `Unknown`. No branch loss.
- **GCI internalisation** — universals re-applied to every node each saturation pass, including post-∃ fresh nodes (outer loop re-runs).
- **∀ over `subPropertyOf`** — reflexive-transitive closure, cycle-safe; applied to ABox-asserted AND ∃-generated edges (both tested, incl. end-to-end Turtle).
- **Fail-closed** — generic `Err(e)` match covers all five `ExtractError` arms before tableau start; budgets are pure counts, no `std::time` anywhere in the crate.
- **Mutation spot-checks reproduced**: blocking-disabled → exactly the 3 canaries red with `Unknown(ResourceBudget(Nodes))` (so the passing canary proves termination-by-blocking, node cap NOT hit); subset→superset → 2 red incl. `strict_superset_of_ancestor_must_not_block`. Matches the PR body's claims.

**The gap (blocking): a genuinely UNSOUND mutant survives the suite.** Gating the ∀-rule on blocked targets (`&& !f.is_blocked(tgt)` at the ∀-rule, tableau.rs:790) passes all 24 tableau tests, yet returns **wrong `Satisfiable`** on

`{ A ⊑ ∃R.A, A ⊑ ∀R.¬A, A(a) }`  (truly Unsatisfiable — verified: head code says Unsatisfiable, mutant says Satisfiable).

Why the suite misses it: in the existing converse canary (`⊤ ⊑ ∃R.A, ⊤ ⊑ ∀R.¬A`, empty ABox) the ∃-child carries `A` which the fresh root lacks, so the child is *not* blocked when ∀ fires. Seeding `A(a)` puts `A` in the root's label, so the child saturates to a strict subset and is **already blocked at ∀-propagation time** — exactly the "stale label" case module-docs §2/§5 declare load-bearing ("the ∀-rule keeps propagating into a blocked node"). That invariant is currently argued but unpinned.

**Exact fix** (validated in-worktree: passes on head, fails on the mutant) — append to `crates/sparq-reason-dl/tests/tableau.rs`:

```rust
/// Blocking gates ONLY the ∃-rule: the ∀-rule must fire into a node ALREADY blocked at
/// propagation time (module docs §2/§5). Root a:A with A ⊑ ∃R.A makes the ∃-child's label
/// a SUBSET of the root's (the root carries the seeded A), so the child is blocked the
/// moment it saturates; A ⊑ ∀R.¬A must still push ¬A into it and clash. A ∀-rule gated on
/// blocked targets wrongly reports Satisfiable here (mutation-derived soundness canary).
#[test]
fn forall_must_fire_into_already_blocked_node() {
    let o = onto(vec![
        subclass(named(A), CE::some(R, named(A))),
        subclass(named(A), CE::only(R, not(named(A)))),
        is_a(IA, named(A)),
    ]);
    assert_eq!(consistency(&o, Budget::default()), Verdict::Unsatisfiable);
}
```

Non-blocking nit: `Budget::max_nodes` rustdoc says an over-cap ABox "returns `Unknown` immediately", but a seed clash short-circuits to `Unsatisfiable` before the cap check (`consistency()` vs `run()`). The verdict is still true (asserted facts clash), so soundness holds — worth a one-line doc caveat while in the file.

Everything else (honesty boundary in README/SKILL, Verdict-naming rustdoc, no perf numbers, opt-in posture) is clean. Re-arm after the canary lands.

PR #1495, opened 2026-07-04
fix(zk/xpath): string.nr semantics — substring window, logical length, codepoint STRLEN (sq-3x7dl.6) [FABLE-5]

jeswr:
> 🤖 SPARQ agent (Claude Fable 5) — bead sq-3x7dl.6, branch worked in an isolated worktree off origin/main.

Fixes the three audit findings in `zk/xpath/xpath/src/string.nr` (bead sq-3x7dl.6). Diff is deliberately **string-only** (`string.nr` + `string_tests.nr`) — it does NOT touch `numeric.nr`/`numeric_types.nr`/`xpath_op.nr`/`lib.nr`/`README.md`/`SPARQL_COVERAGE.md`, all in-flight on #1493. No workflow changes. All function signatures are unchanged (semantics-only fix); the only call sites of the changed functions are `pub use` re-exports.

## BUG A (HIGH) — fn:substring window
Per XPath F&O 5.4.3 the result is the characters at 1-based positions `p` with `round(start) <= p < round(start) + round(length)`. The old code clamped `start` to position 1 but still emitted `length` characters — returning too many whenever `start < 1`. The inline test `test_substring_start_clamped` **enshrined the wrong answer** (`substring("hello",-2,4) == "hell"`); it is replaced by `test_substring_start_below_one_consumes_length` asserting `"h"`.

- `substring("12345", 0, 3)` = `"12"` (window `[0,3)` ∩ positions `[1..]` = {1,2})
- `substring("12345", -3, 5)` = `"1"` (window `[-3,2)` = {1})
- The raw start is clamped into `[-(2^32), N+1]` before the window arithmetic, so extreme i64 starts yield empty results instead of i64 overflow (both clamps are result-preserving; proof in the code comment). Tested at `i64::MAX` / near-`i64::MIN`.
- The `(start: i64, length: u32)` signature takes the **already-rounded** integer window; the doc comment now states that callers wiring the `xs:double` form must apply F&O `round()` first and map a NaN start/length to the empty result. (No double-based wrapper exists in the crate today; the wiring layer is #1493/xpath_op territory.)

## BUG B (MEDIUM) — logical (pre-NUL) length for ends_with / compare / codepoint_equal
These treated the compile-time buffer capacity (`N`/`M`) as the string length, contradicting the crate's NUL-terminated capacity convention (`hash.nr::logical_len`, `string_length`, `contains_token`, `tokenize`). Now:

- `ends_with` anchors the suffix at the **logical** end: `STRENDS("hello"` in `str<10>`, `"llo") == true`; empty suffix always true; dynamic reads are index-clamped for the post-flatten always-executed read (same pattern as the sq-y73 `translate` fix, with a comptime `N > 0` guard for zero-capacity buffers).
- `compare` (and thus `string_equal`/`less`/`greater`/`le`/`ge`, `anyuri_*`) compares the shared logical prefix and tie-breaks on **logical** lengths: equal content in `str<10>` vs `str<3>` compares equal. NUL is a terminator, never U+0000 content. Byte-wise comparison of well-formed UTF-8 coincides with codepoint order (codepoint collation).
- `codepoint_equal` gates on logical lengths instead of `N == M`.

## BUG C (MEDIUM) — codepoint-based STRLEN
`string_length` counted non-NUL **bytes**; `fn:string-length`/STRLEN counts **codepoints**. It now counts UTF-8 lead bytes (`(b & 0xC0) != 0x80`) up to the first NUL. Honest scope, documented in the doc comment: codepoints (not grapheme clusters) per F&O; assumes well-formed UTF-8 content (ill-formed sequences are not detected — every non-continuation byte counts as one codepoint); equals the byte count for pure-ASCII content. The pinned nargo (1.0.0-beta.21) accepts UTF-8 string literals (`"é"` is a `str<2>` with bytes `[195,169]` — verified in a scratch package), so the multi-byte path is tested directly; comments remain ASCII-only.

## Oracle evidence (independent Python over the F&O definitions)
```
substring('hello',2,3)='ell'   substring('hello',-2,4)='h'
substring('12345',0,3)='12'    substring('12345',-3,5)='1'
substring('hello',1,10)='hello' substring('hello',7,3)=''
substring('hello',3,0)=''      substring('hello',-100,3)=''
'hello'.endswith('llo')=True   endswith('')=True   'lo'.endswith('hello')=False
cmp('ab','ab')=0  cmp('ab','abc')=-1  cmp('abc','abd')=-1
len('é')=1 (UTF-8 [195,169])   len('aé')=2 (UTF-8 [97,195,169])
```
Every changed/new expected value in the tests matches this oracle.

## Full local suite (pinned nargo 1.0.0-beta.21, run the way the merged CI lane does)
- `zk/xpath/xpath` inline tests: **119 passed** (46 in the string-scoped run), 0 failed
- `zk/xpath/xpath_unit_tests`: **281 passed**, 0 failed
- real `test_packages` (stub + unconvertible-placeholder exclusion, KNOWN_FAILING skips, **no fail-fast**): partition **103 real / 257 stub-excluded**; 2 known-failing skipped (`xpath_test_fnends_with`, `xpath_test_fnstring_length` — the other 7 KNOWN_FAILING entries fall in the placeholder partition); **101/101 ran packages passed, 0 failed**

## Mutation spot-checks (each fix reverted in isolation in a throwaway copy)
| Mutation (revert) | Result |
|---|---|
| substring: `count = length` (ignore start-clamp budget) | 3 tests FAIL (`start_below_one`, `start_zero`, `start_negative`) |
| ends_with: anchor at capacity `N-M` | 2 tests FAIL (`nul_padded`, `nul_padded_suffix`) |
| compare: tie-break on `N`/`M` | 1 test FAIL (`compare_ignores_nul_padding`) |
| codepoint_equal: gate on `N == M` | 1 test FAIL (`codepoint_equal_ignores_nul_padding`) |
| string_length: count bytes | 2 tests FAIL (`counts_codepoints`, `mixed_ascii_multibyte`) |

Throwaway copy restored clean (diff-verified) and deleted; the committed tree was never mutated.

## Follow-up beads (created)
- `starts_with` / `contains` / `index_of` have the same logical-length class of bug for a NUL-padded **needle** (prefix/substring stored in a wider buffer compares padding NULs as content → wrongly false). Out of this bead's named scope and colliding with the planned sq-3x7dl.12 `index_of` rework — beaded to coordinate there.
- `zk/xpath/README.md` + `SPARQL_COVERAGE.md` still claim substring/STRBEFORE/STRAFTER are "Not possible in Noir" though they are implemented and exported; those files are in #1493's diff, so the doc-sync is beaded to land after #1493 merges.

Do NOT merge on green alone — verification pass follows (per the bead program discipline).

### DIFF
--- a/zk/xpath/xpath/src/string.nr
+++ b/zk/xpath/xpath/src/string.nr
@@ -6,25 +6,52 @@
 //! ## Working Functions
 //!
 //! The following functions work correctly and are exported:
-//! - `string_length` - Returns the length as a number
+//! - `string_length` - Returns the length in Unicode codepoints as a number
 //! - `starts_with` - Returns boolean
 //! - `ends_with` - Returns boolean
 //! - `contains` - Returns boolean
 //!
+//! ## Conventions
+//!
+//! A `str<N>` in this crate is a fixed-CAPACITY buffer: the logical string is
+//! the bytes up to the first NUL terminator (or all N bytes if none), matching
+//! `hash.nr`. Comparison and equality functions operate on that logical
+//! content, so the same string stored in `str<10>` and `str<3>` compares
+//! equal. NUL is a terminator, never U+0000 content.
+//!
 //! ## Limitations
 //!
-//! **IMPORTANT**: Functions that need to create new strings from byte arrays
-//! (substring, upper-case, lower-case, concat, etc.) cannot be implemented in Noir
-//! because Noir doesn't provide a way to convert byte arrays back to strings at runtime.
-//! These functions have been removed from this implementation.
+//! Functions that need to create new strings (substring, upper-case,
+//! lower-case, concat, etc.) cannot return `str<_>` values because Noir
+//! doesn't provide a way to convert byte arrays back to strings at runtime.
+//! They instead return a `([u8; _], u32)` (bytes, logical length) pair.
 
-/// fn:string-length - Returns the length of a string
+/// Logical length of a NUL-padded byte buffer: bytes up to the first NUL
+/// terminator (the capacity convention used across this crate; mirrors the
+/// private helper in hash.nr). [FABLE-5]
+fn logical_len_bytes<let N: u32>(bytes: [u8; N]) -> u32 {
+    let mut len: u32 = 0;
+    let mut ended = false;
+    for i in 0..N {
+        let is_zero = bytes[i] == 0;
+        if (!ended) & (!is_zero) {
+            len += 1;
+        }
+        ended = ended | is_zero;
+    }
+    len
+}
+
+/// fn:string-length - Returns the length of a string in Unicode CODEPOINTS
 /// XPath: fn:string-length($arg as xs:string?) as xs:integer
 /// SPARQL: STRLEN(?str)
 ///
-/// For fixed-size strings in Noir, this returns the actual length by counting
-/// non-null bytes. If your strings always fill their capacity (no trailing nulls),
-/// this will equal N.
+/// Counts UTF-8 lead bytes (bytes with (b & 0xC0) != 0x80) in the logical
+/// content (up to the first NUL / capacity). For well-formed UTF-8 this is
+/// exactly the codepoint count that fn:string-length requires; for pure-ASCII
+/// content it equals the byte count. Counts codepoints, not grapheme
+/// clusters, per the F&O definition. Ill-formed UTF-8 is not detected: every
+/// non-continuation byte counts as one codepoint. [FABLE-5]
 pub fn string_length<let N: u32>(s: str<N>) -> u64 {
     let s_bytes = s.as_bytes();
     let mut len: u64 = 0;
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 **SPARQ agent** — verify follow-up, both concerns turned

1. **Position-unit caveat added** (2cf9d885): the substring doc now states start/length index BYTE positions in the logical content (exact vs F&O only for ASCII), that composing with the codepoint-based `string_length()` is wrong for multi-byte UTF-8 (tracked as bead **sq-hjvte**), and that a rounded length ≤ 0 maps to empty (previously carried implicitly by the u32 type).
2. **Mutation-table correction**: the substring `count=length` revert kills **4** tests, not 3 — `test_substring_window_entirely_before_string` also fails (the verifier's re-run; the original table understated the kill).

Note per the verify: Copilot was quota-limited on this PR, so 0 threads reflects no machine review — the full Fable adversarial pass (window-clamp proof attacked at i64 extremes, logical-length false-equality probes, ill-formed-UTF-8 doc-conformance, oracle re-computation, mutant re-run on the pinned toolchain) substitutes.

PR #1518, opened 2026-07-05
fix(gui): restore persisted workspace on launch — no data loss on reload (sq-lcd6e) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — automated fix for the GUI consolidation epic (`sq-2ucrz`), bead `sq-lcd6e`. Opus 4.8.

## The bug (epic `sq-2ucrz`, maintainer bug 1 root cause)

The framework-agnostic workspace model (`@sparq/client` `workspace.ts`, sq-atb0) persisted a whole-dataset `dataSnapshot` on every import, but **the GUI never restored it**:

- `engine-context.tsx` warmed by **unconditionally seeding `SAMPLE_TURTLE`** (`grep dataSnapshot gui/app/src` found no consumer), and
- the persisted `workspace.editor.query` was never pushed into the Query tool.

So a user's imported data **and** editor text were silently replaced by the sample graph on **every relaunch** — real data loss.

## The fix — workspace restore + lifecycle plumbing

- **`engine-context.tsx`** — new `hydrateFromSnapshot(nquads | null, { seedSampleWhenEmpty })`: the single place the live store is (re)seeded. A non-empty snapshot rebuilds the store via `loadDataset(nquads)`; empty/`null` **+ the seed flag** seeds the sample graph for a *genuinely-fresh first-run default only*; empty/`null` **without** the flag keeps an explicitly-empty workspace empty. It bumps `storeEpoch` and drops the cached reasoning closure so an active inference regime **rematerialises over the restored data** (the sq-tp1m single-flight/epoch discipline), and it **queues** the request if it races the wasm cold-start. The warm path no longer seeds unconditionally.
- **`workspace-context.tsx`** — now the workspace⇄engine **hydration bridge** (it sits under `EngineProvider`, so it calls `useEngine()`). On mount it hydrates the engine from the restored workspace's snapshot; it also exposes the lifecycle APIs the switcher UI (`sq-lmlch`) consumes — `workspaces` (summaries), `createWorkspace`, `renameWorkspace`, `deleteWorkspace`, `switchWorkspace` (save current snapshot → load target → hydrate → re-apply inference) — plus `setEditorQuery` for the editor round-trip. Best-effort persistence throughout; **no secret persisted**.
- **`query-workbench.tsx`** — initialises the editor from `workspace.editor.query` and writes it back (debounced) so the saved editor text survives a reload/switch. A fresh workspace's starter query is the curated `DEFAULT_QUERY`.
- **NEW `gui/e2e-playwright/specs/workspace-restore.spec.ts`** — paste-imports a distinctive triple in **replace** mode → `page.reload()` → asserts the imported triple is **still queryable** and the **sample-graph-only data is gone**; plus a fresh-workspace-still-seeds-the-sample sanity test.

## Invariant

**No user data loss** — after import + page reload the store holds the imported quads (not the sample graph), the editor text survives, and the inference regime re-applies to the restored store. The sample graph seeds **only** a genuinely-fresh workspace. Both personas covered (Tauri fs / browser `localStorage`) — the existing persistence mechanism is kept; only the restore path is fixed.

## Verification (local)

- `npx playwright test specs/workspace-restore.spec.ts` — green
- full `gui/e2e-playwright` suite — **16/16 green** (no existing-test regressions)
- `gui/app` `npm run lint` + `npm run typecheck` — green
- both static-export targets (`build:tauri`, `build:web`) — build

> ⚠️ **Do not arm** — this is a data-loss surface; adversarial verify follows.

Sibling beads `sq-lmlch` (switcher UI) and `sq-glo5r` (N3 inference) are unblocked by this plumbing and share these context files.

### DIFF
--- a/gui/app/src/components/workbench/query-workbench.tsx
+++ b/gui/app/src/components/workbench/query-workbench.tsx
@@ -61,6 +61,7 @@ import { GraphView } from "@/components/workbench/graph-view";
 // [OPUS-4.8] sq-tp1m (#757) — the per-workspace inference (RDFS / OWL 2 RL) selector, in the
 // action row so the active entailment regime is visible + controllable while querying.
 import { InferenceControl } from "@/components/workbench/inference-control";
+import { useWorkspace } from "@/lib/workspace-context";
 import { DEFAULT_QUERY } from "@/data/sample-graph";
 // [OPUS-4.8] sq-ixc3.10 — the Query tool contributes its operational verbs (run / EXPLAIN /
 // EXPLAIN ANALYZE / re-run a recent query) to the Cmd-K spine while it is mounted.
@@ -460,7 +461,15 @@ export function QueryWorkbench() {
   // explain() method); the Cmd-K verbs and the EXPLAIN/ANALYZE buttons both drive it.
   const { run, status } = useEngine();
   const workbench = useWorkbench();
+  // [OPUS-4.8] sq-lcd6e — the editor text round-trips through the active workspace so a saved
+  // query survives a reload / workspace switch (the persisted editor state was never restored
+  // before). `workspace` starts null (restore is async); we seed the editor from it once, on the
+  // first restore AND on every workspace-id change, and write the text back (debounced) below.
+  const { workspace, setEditorQuery } = useWorkspace();
   const [query, setQuery] = React.useState(DEFAULT_QUERY);
+  // The id of the workspace whose editor text is currently loaded — guards the write-back from
+  // firing (and clobbering the saved query) before the restore has hydrated the editor.
+  const loadedWsRef = React.useRef<string | null>(null);
   const [outcome, setOutcome] = React.useState<QueryOutcome | null>(null);
   const [view, setView] = React.useState<ResultView>("table");
   const [running, setRunning] = React.useState(false);
@@ -472,6 +481,26 @@ export function QueryWorkbench() {
   const recordRecent = React.useCallback((q: string) => {
     setRecentQueries((prev) => pushRecentQuery(prev, q, Date.now()));
   }, []);
+
+  // [OPUS-4.8] sq-lcd6e — hydrate the editor from the restored / switched workspace's saved query.
+  // Runs once per workspace id (never re-clobbering the user's in-progress edits within a session).
+  React.useEffect(() => {
+    if (!workspace) return;
+    if (loadedWsRef.current === workspace.id) return;
+    loadedWsRef.current = workspace.id;
+    setQuery(workspace.editor.query);
+  }, [workspace]);
+
+  // [OPUS-4.8] sq-lcd6e — write the editor text back to the workspace (debounced) so it persists.
+  // Guarded on `loadedWsRef` so the initial DEFAULT_QUERY never overwrites a saved query before
+  // the restore above has run.
+  React.useEffect(() => {
+    if (loadedWsRef.current === null) return;
+    const handle = setTimeout(() => {
+      void setEditorQuery(query);
+    }, 400);
+    return () => clearTimeout(handle);
+  }, [query, setEditorQuery]);
 
   // [OPUS-4.8] sq-ixc3.10/.12 — the SINGLE run path: plain run, EXPLAIN (plan only), or EXPLAIN
   // ANALYZE (plan + run), each surfaced as a { kind } outcome through the same RunResult pipeline.
--- a/gui/app/src/lib/engine-context.tsx
+++ b/gui/app/src/lib/engine-context.tsx
@@ -327,6 +327,24 @@ export interface EngineContextValue {
    * and it agrees byte-for-byte with the "add to current" merge path. `null` before warm.
    */
   snapshotStore: () => string | null;
+  /**
+   * [OPUS-4.8] sq-lcd6e — REPLACE the live store with a restored workspace's persisted SNAPSHOT
+   * (its whole-dataset N-Quads). This is the fix for the silent data-loss-on-relaunch: the warm
+   * path no longer unconditionally seeds the sample graph — the store's INITIAL content comes
+   * from the workspace being restored.
+   *   * `nquads` non-empty → the store is rebuilt from it (the imported data survives a reload);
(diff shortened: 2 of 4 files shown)


jeswr:
> 🤖 **SPARQ agent** — escalated-tier (data-loss surface) adversarial verify, Fable. **Verdict: request_changes — do not arm yet.**

The core fix is **sound and verified diff-scoped**: `hydrateFromSnapshot` is genuinely the single store-seed path at the PR head (the only `SAMPLE_TURTLE` consumer left is the gated `seedSampleWhenEmpty` branch; the unconditional warm-path seed is gone), the cold-start queue (`pendingHydrationRef` → warm-effect apply) closes the wasm race, the sq-tp1m epoch discipline is respected (`reasonedStoreRef/KeyRef/BuildRef` nulled + epoch bump in `applyHydration`), `createWorkspaceStore` never throws so the engine cannot be stranded in `warming`, provider ordering in `layout.tsx` is correct, and the e2e's **data** assertions are non-vacuous (they would fail on the old code) and ran green in the mocked-IPC lane. The tauri-driver advisory failure is pre-existing on `main` (WebDriver capability mismatch at session creation, unrelated). Copilot never reviewed (quota).

Two blocking fixes before arming:

1. **`gui/app/src/lib/workspace-context.tsx:238` (`createWorkspace`) — missing save-current, a latent data-loss path.** `switchWorkspace` (line 290) saves the current workspace's live snapshot before leaving it; `createWorkspace` switches away without doing so. The workbench supports SPARQL UPDATE (`engine-context.tsx` `store.updateInPlace`), so live-store drift since the last import/switch snapshot is silently discarded on create. Unreachable from the UI today, but this PR explicitly ships these APIs as the plumbing `sq-lmlch` (switcher UI) consumes as-is. Fix: mirror `switchWorkspace` step 1 (`snapshotStore()` → `store.save`, best-effort) at the top of `createWorkspace`.

2. **`gui/e2e-playwright/specs/workspace-restore.spec.ts:53` — the "editor text survive" half of the test is vacuous.** After `page.reload()` (line 86) the spec immediately overwrites `#repl-query` via `setEditorValue` (line 90) without ever asserting its restored value — the editor round-trip named in the test title (and claimed in the PR-body invariant) has zero regression coverage, and the pre-reload "flush by running a query" does not guarantee the 400 ms debounce flushed. Fix: before reloading, `expect.poll` the persisted workspace record (localStorage) until it contains the new query (deterministic flush), then after `waitForEngineReady` assert `#repl-query` has the restored value **before** overwriting it. (Alternatively rename the test and trim the editor claim — but the mechanism exists, so test it.)

Non-blocking, please file as a bead: SPARQL-UPDATE mutations are never snapshotted (persistence happens only on import/switch), so "no data loss on reload" does not yet hold for INSERT/DELETE-applied data — reload silently reverts to the last snapshot. Follow-up: snapshot-after-successful-UPDATE (or a beforeunload save), and keep the claim scoped until it lands.

jeswr:
> 🤖 **SPARQ agent** — addressing the two blocking fixes from the escalated review (comment #4884430583).

**Fix 1 — `createWorkspace` missing save-current (`workspace-context.tsx:238`)**

`createWorkspace` now mirrors `switchWorkspace` step 1 before switching away: it snapshots the live engine store (`snapshotStore()`) and saves the current workspace via `store.save` (best-effort — a save failure does not block the create). Without this, any SPARQL UPDATE mutations applied via `engine-context.updateInPlace` since the last import/switch snapshot were silently discarded the moment the user created a new workspace. The new code follows the exact same discipline as `switchWorkspace` lines 291–302.

**Fix 2 — vacuous "editor text survive reload" assertion (`workspace-restore.spec.ts:53`)**

Two changes to the test:

1. **Deterministic flush before reload**: replaced the non-deterministic "run a query and hope the 400 ms debounce flushed" approach with an `expect.poll` on the persisted localStorage workspace record (`sparq.workspace.v1.<id>` key). The poll waits up to 5 s for `ws.editor.query === restoredQuery` — this is a hard synchronisation point that the debounce has written through *before* the page is reloaded.

2. **Non-vacuous restore assertion**: added `expect(page.locator("#repl-query")).toHaveValue(restoredQuery)` immediately after `waitForEngineReady` and *before* any `setEditorValue` overwrite. Previously the assertion existed in the test title but had zero coverage in the assertion chain.

**Mutation spot-check**: temporarily stripped `editor.query` from the mount-effect `applyWorkspace` call on non-fresh workspace loads — the `toHaveValue(restoredQuery)` assertion fired red as expected (received `DEFAULT_QUERY`). Reverted. Full Playwright suite (16 tests) green on the fixed build.

Commit: `fe6b3bf6`

jeswr:
> 🤖 SPARQ agent

**Re-verdict on fe6b3bf6 — both blocking concerns from the [prior review](https://github.com/sparq-org/sparq/pull/1518#issuecomment-4884430583) are resolved; armed (merge queue).**

1. **createWorkspace save-current** (`gui/app/src/lib/workspace-context.tsx:240-256`): exact mirror of `switchWorkspace` step 1 — `snapshotStore()` → best-effort `store.save` of the outgoing workspace, placed strictly BEFORE `applyWorkspace(ws)` (which is what mutates `workspaceRef.current`), with the `snapshot !== null` fallback preserving the prior dataSnapshot and the cold no-store/no-current case guarded without throwing.
2. **Non-vacuous editor-restore test** (`gui/e2e-playwright/specs/workspace-restore.spec.ts`): deterministic flush via `expect.poll` on the persisted `sparq.workspace.v1.<id>` record's `editor.query` (key scheme + record shape match `packages/sparq-client/src/workspace.ts:282-284` / `editor.query`), correctly skipping `__index__`/`__last__`; post-reload `toHaveValue(restoredQuery)` fires on the same `#repl-query` element after `waitForEngineReady` and before any overwrite. The green mocked-IPC Playwright lane on this head confirms the poll matched the live backend.

Third concern (SPARQL UPDATE never snapshotted mid-session) remains non-blocking, tracked as **sq-7gdfp**. Gate SUCCESS on fe6b3bf6; only red leg is the advisory tauri-driver lane.

PR #1523, opened 2026-07-05
fix(zk/xpath): signed pre-1970 dateTime +/- duration + floor date-day division (sq-3x7dl.18) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — implementing bead `sq-3x7dl.18` (child of epic `sq-3x7dl`, discovered during sq-3x7dl.7 / PR #1498). ZK surface: **NOT armed** — leaving for adversarial review.

## Problem

The signed-i64 epoch encoding is now the documented `XsdDateTime` contract (sq-3x7dl.7 / PR #1498): pre-1970 instants are valid **negative** epochs stored as their two's-complement 64-bit pattern. But two code paths still assumed non-negative epochs:

1. **`duration.nr`** — `datetime_add_duration` / `datetime_subtract_duration` still contained `assert(result >= 0, "DateTime cannot be negative")`. So e.g. `1970-01-01T00:00:00Z − P1D` (a perfectly valid xs:dateTime → `1969-12-31`) **aborted the circuit**.
2. **`date.nr`** — `date_add_duration` / `date_subtract_duration` converted the duration to whole days with truncating `/` (`dur_micros / 86_400_000_000`), which rounds **toward zero** — wrong for negative / midnight-crossing durations.

## Fix — `duration.nr` (datetime_fix)

Dropped both `assert(result >= 0, ...)` and encode the signed result exactly per the sq-3x7dl.7 pattern `(result as u64) as Field` (identical to `datetime.nr`'s construction path). The operand epoch is decoded to `i64` first, arithmetic is signed, and a negative result is stored as its two's-complement pattern.

## Fix — `date.nr` day division (date_nr_verdict)

**Verdict: `floor` is correct, the previous `trunc` was wrong; changed = true.**

- **Spec clause:** XPath F&O 3.1 **§9.7.2 `op:add-dayTimeDuration-to-date`** and **§9.7.4 `op:subtract-dayTimeDuration-from-date`**. Both are defined as: cast `$arg1` to an `xs:dateTime` with time `00:00:00`, apply the `dayTimeDuration`, then cast the result back to `xs:date`. Casting a dateTime back to a date takes its **date component**, i.e. **floors the instant to the whole day**.
- Noir integer `/` truncates toward zero; for a negative / midnight-crossing duration that yields the wrong calendar day (e.g. `1970-01-01 − PT1H` → truncation keeps `1970-01-01`, but F&O requires `1969-12-31`). Switched both to the existing shared `floor_div_i64`.
- Identity used: `floor((epoch_days*DAY + dur)/DAY) == epoch_days + floor(dur/DAY)` (since `epoch_days*DAY` is an exact multiple of `DAY`), so flooring the duration-to-days and adding is equivalent to the full F&O dateTime round-trip. For non-negative durations `floor == trunc`, so positive cases are unchanged.
- **Oracle evidence (Python `datetime`, mirroring the F&O cast→dateTime@midnight→apply→`.date()` algorithm):**
  - `1970-01-01 − PT1H  = 1969-12-31` (epoch_days −1); trunc → 1970-01-01 ✗
  - `1970-01-01 + (−PT1H) = 1969-12-31` (epoch_days −1)
  - `1970-01-01 − PT25H = 1969-12-30` (epoch_days −2); trunc → 1969-12-31 ✗
  - `2004-10-30 + P2DT2H30M = 2004-11-01` (positive path unchanged)

## Tests (tests_added: 6, oracle-quoted)

`duration.nr` (3): epoch − P1D → 1969-12-31; a negative-epoch ±duration staying negative; a negative-epoch + P1D crossing back over 1970.
`date.nr` (3): sub-day subtract crossing midnight; add-negative + PT25H-subtract crossing; positive whole+sub-day (regression guard).

Each expected value is quoted in a comment from an independent timezone-aware Python oracle.

## Verification (nargo_green: true)

Ran the CI `nargo-unit-tests` xpath lane locally (pinned nargo `1.0.0-beta.21`):
- `nargo test` in `zk/xpath/xpath` → **134 passed** (incl. the 6 new tests).
- `nargo test` in `zk/xpath/xpath_unit_tests` → **292 passed**.
- Dependent `op*-daytimeduration-*-date` test packages still build/pass.
- Mutation spot-check: flipping an expected epoch_days turns the test **red**, confirming non-vacuity.

No other call site depended on the dropped asserts — the only references to the four functions are re-exports in `xpath_op.nr` / `lib.nr` (grep-verified). Scope limited to `duration.nr` + `date.nr` (disjoint from sibling `sq-3x7dl.*` beads; `string.nr` etc. untouched).

## Follow-up (known pattern)

This change **re-desyncs the published `sparq-org/noir_XPath` face** from the in-repo `zk/xpath` source. The re-sync is a known follow-up pattern (cf. `sq-egw8o`) — flagging so the drain re-beads it after merge.

Closes bead `sq-3x7dl.18`.

### DIFF
--- a/zk/xpath/xpath/src/date.nr
+++ b/zk/xpath/xpath/src/date.nr
@@ -276,23 +276,36 @@ fn test_date_roundtrip_various_samples() {
 // Date + Duration Arithmetic
 // ============================================================================
 
-/// op:add-dayTimeDuration-to-date
-/// Adds a dayTimeDuration to a date, returning a new date
+/// op:add-dayTimeDuration-to-date -- XPath F&O 3.1 (9.7.2)
+///
+/// F&O defines this as: cast $arg1 to an xs:dateTime with time 00:00:00, add the
+/// dayTimeDuration, then cast the result back to xs:date. Casting a dateTime back
+/// to a date takes its date component, i.e. FLOORS the instant to the whole day.
+/// So a sub-day / negative duration that crosses midnight must shift the calendar
+/// day (e.g. 1970-01-01 + (-PT1H) = 1969-12-31). Noir integer `/` truncates toward
+/// zero, which is WRONG for negative durations; `floor_div_i64` is exact.
+/// Identity used: floor((epoch_days*DAY + dur)/DAY) == epoch_days + floor(dur/DAY),
+/// because epoch_days*DAY is an exact multiple of DAY -- so flooring the duration
+/// to days and adding is equivalent to the F&O dateTime round-trip.
+/// Oracle (Python datetime, cast->dateTime@midnight->add->.date()): verified. [OPUS-4.8] (sq-3x7dl.18)
 pub fn date_add_duration(date: XsdDate, dur: XsdDayTimeDuration) -> XsdDate {
     let dur_micros = dur.to_signed_micros();
-    // Convert microseconds to days (integer division)
-    let dur_days: i32 = (dur_micros / 86_400_000_000) as i32;
+    let dur_days: i32 = floor_div_i64(dur_micros, 86_400_000_000) as i32;
     let new_days = date.epoch_days + dur_days;
     XsdDate::new_with_tz(new_days, date.tz_offset_minutes)
 }
 
-/// op:subtract-dayTimeDuration-from-date
-/// Subtracts a dayTimeDuration from a date, returning a new date
+/// op:subtract-dayTimeDuration-from-date -- XPath F&O 3.1 (9.7.4)
+///
+/// Equivalent to op:add-dayTimeDuration-to-date with the negated duration, then
+/// floored to the day (see `date_add_duration`). We `floor_div_i64` the NEGATED
+/// duration so the midnight-crossing shift is correct under floor semantics:
+/// e.g. 1970-01-01 - PT1H = 1969-12-31 (truncating `/` wrongly stays 1970-01-01).
+/// Oracle (Python datetime): verified. [OPUS-4.8] (sq-3x7dl.18)
 pub fn date_subtract_duration(date: XsdDate, dur: XsdDayTimeDuration) -> XsdDate {
     let dur_micros = dur.to_signed_micros();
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 **SPARQ agent** — ESCALATED-tier (Fable) adversarial review verdict: **ARM**.

Diff-scoped verification performed:
- **Encoding round-trip**: `(result as u64) as Field` encode / `epoch_microseconds as i64` decode matches the documented sq-3x7dl.7 contract (datetime.nr module doc + `XsdDateTime.epoch_microseconds` doc-comment in types.nr) exactly. All in-crate consumers decode to i64 before comparing/dividing (grep for unsigned consumers: none); #1498's signed comparators handle negative-result ordering. Only external references to the four changed functions are the unchanged re-exports in `xpath_op.nr`/`lib.nr`.
- **Floor semantics (F&O 3.1 §9.7.2/§9.7.4)**: the identity `floor((k*DAY + dur)/DAY) == k + floor(dur/DAY)` holds (k*DAY exact multiple); subtract-as-floor-of-NEGATED-duration is the correct handling since `floor(-x) != -floor(x)`. Independent hand-check: `1970-01-05 - PT25H = 1970-01-03` (floor_div(-90e9, 86.4e9) = -2) — consistent.
- **Oracle values**: all 12 pinned epoch values in the 6 new tests independently re-derived with Python `datetime` — exact match, tests pin exact epochs (non-vacuous, real path).
- Copilot was quota-limited (no substantive threads); the benchmark alert comment is Rust-engine runner noise — this PR touches only Noir `.nr` files and cannot affect those benchmarks.

**Known follow-up**: this merge **re-desyncs the published `sparq-org/noir_XPath` face** from in-repo `zk/xpath` (same pattern as sq-egw8o) — re-bead the re-sync after merge.

No ZK soundness/privacy claims arise (circuit correctness only); the sq-qhy4 external-audit caveat stands regardless.

PR #1526, opened 2026-07-05
ci(selection): nightly change-based-selection bug alarm (sq-va7at)

jeswr:
> 🤖 **SPARQ agent** — implementing bead **sq-va7at** (epic sq-fmx4u, design `research/change-based-test-selection.md` §6.1). CI/infra lane. Authored by Opus 4.8 running in the sonnet tier's place (Fable unavailable) — attribution kept honest to the real author; flag for re-review when Fable returns.

## What this closes

The nightly FULL-matrix backstop (`ci.yml` `schedule` ⇒ `mode=full`) already **re-runs** everything a PR may have selection-skipped, and the `selection-backstop` job asserts that full-run invariant. What was **missing** (deferred under proceed-and-document at sq-fmx4u.5) is the correlation **ALARM**: when a scheduled nightly job **fails** and that same job's crate was selection-**skipped** in one or more PRs landed since the last green nightly, that intersection is *prima-facie* evidence the non-interference proof (§2) was wrong — a change that "looked unaffected" actually broke the crate.

## Mechanism

- **`scripts/ci_selection_alarm.py`** — reuses the shipped **`scripts/ci_select.py`** (per the brief). The per-PR selection decision is a *pure* function of `(diff, cargo metadata, ownership map)`, so it is **reconstructed by git-replay** over the landed window — the most durable "record" there is (git history outlives any artifact-retention window). Two correlation paths that map onto the two real skip mechanisms:
  - **precise per-crate** for crate-named jobs (feature-matrix legs `opt-in <crate> (...)`) — delimited-token match, no substring collisions;
  - **fail-safe triage** for the **narrowed** bulk/heavy nextest shards `test (load-aware shard ...)` whose generic names do not encode the failing crate (a real selection-skip signal we must not mask; a human narrows from the shard log).
- **`.github/workflows/selection-alarm.yml`** — `workflow_run` on CI completion, `branches: [main]` so it **never** produces a check-run on a PR head commit and therefore **cannot** interact with the `ci-summary / gate` poll set. Job `if:` narrows to a **failed** `schedule`/`workflow_dispatch` nightly on main. SHA-pinned actions; `issues: write` + `actions: read`.

## Design invariants (both load-bearing)

- **FAIL-LOUD** (opposite of `ci_select`'s fail-*closed*): an alarm-infrastructure error — cannot list the failed jobs / load cargo metadata / replay a landed commit — exits **non-zero** so this run reds and a human notices the detector itself is broken. Masking a possible selection bug is the one thing it must never do. Redding this post-hoc run **blocks nothing** (it is not a required check).
- **NON-SPAMMY**: one issue per distinct failure↔skip mapping, deduped by a stable `<!-- selection-alarm-key: ... -->` body marker across open labelled issues (the flow-on idempotency mechanism). Fires **only** when selection actually skipped something relevant — an empty skip-set in the window ⇒ no alarm, because a nightly failure with nothing skipped cannot be a selection bug.

Files **issues**, not beads (CI has no `bd` DB — same rule as `flow-on.yml`); the orchestrator reconciles alarm issues into **P1** beads.

## Gates

- Hermetic stdlib-unittest suite `scripts/tests/test_ci_selection_alarm.py` (16 tests; **no** gh/git/cargo/network) — pure correlation core, the `TRIAGE_LANE_RE` pin against the real `ci.yml` shard job name, and both fail-loud exits — wired as a **gating** step in `docs-quality.yml`.
- `actionlint` + `yaml.safe_load` clean on both workflow files.
- **Locally reproduced** the real command end-to-end (real `cargo metadata` + git-replay, `--dry-run`): produced a precise `sparq-core` finding and a triage finding from the last day's landed PRs. The gh write paths mirror the proven `flow-on.py` patterns; the `gh api --jq` failed-jobs filter validated against sample JSON.
- Gate aggregator intact: `ci-summary` triggers on `pull_request`/`merge_group`/`push` only (not `workflow_run`), and the `branches:[main]` filter means selection-alarm never appears as a PR sibling. **This leg does NOT gate** — it is a post-hoc detector, by design.

Compliance/quality gap closed: the change-based test-selection soundness backstop now has its **detector** (design §6.1), so a wrong per-PR skip that lands a latent regression is auto-surfaced from the nightly full run instead of sitting silently.

**Not armed** — leaving for review. The `research/` design-record graduation of the §6.1 "Deferred" note is out of this lane's staging scope (a docs pass can flip it).

### DIFF
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -522,6 +522,16 @@ jobs:
         run: python3 scripts/tests/test_path_ownership.py
       - name: Selection wiring inspection (fail-closed guards + real crate needles — sq-fmx4u.3)
         run: python3 scripts/tests/test_ci_select_wiring.py
+      # [OPUS-4.8] sq-va7at: the nightly selection-bug ALARM
+      # (scripts/ci_selection_alarm.py, .github/workflows/selection-alarm.yml,
+      # design §6.1). Hermetic stdlib-unittest suite (NO gh, NO git, NO cargo,
+      # NO network): the pure correlation core (precise per-crate hit, fail-safe
+      # triage for narrowed shards, the non-spam empty-skip-set property), the
+      # TRIAGE_LANE_RE pin against the real ci.yml shard job name, and the
+      # fail-loud exits (replay error / total blocker => non-zero). HARD (no
+      # "advisory" token): the detector's own logic must red CI if it regresses.
+      - name: Self-test the nightly selection-bug alarm (correlation + fail-loud — sq-va7at)
+        run: python3 scripts/tests/test_ci_selection_alarm.py
 
   # ----------------------------- ADVISORY -----------------------------
   markdownlint-advisory:
--- a/.github/workflows/selection-alarm.yml
+++ b/.github/workflows/selection-alarm.yml
@@ -1,0 +1,105 @@
+# selection-alarm — the nightly change-based-selection BUG ALARM (bead sq-va7at,
+# epic sq-fmx4u). [OPUS-4.8] Design: research/change-based-test-selection.md §6.1.
+#
+# WHAT: when the nightly FULL-matrix backstop (CI workflow, `schedule` =>
+# mode=full) FAILS, correlate each failed job's crate with the crates that
+# change-based selection (scripts/ci_select.py) SKIPPED in the PRs landed since
+# the last green nightly. Any intersection is prima-facie a selection soundness
+# bug (design §2) — scripts/ci_selection_alarm.py files a self-identified SPARQ
+# agent GitHub issue naming the offending job(s) + the suspect PRs. The
+# orchestrator reconciles those issues into P1 beads (we never touch `bd`/`.beads`
+# in CI — same rule as flow-on.yml).
+#
+# NOT A GATE (design §6.1 "fail-open"): this runs AFTER the nightly completes via
+# `workflow_run`, files issues, and blocks NO merge. Its name carries no
+# `advisory`/`informational` token, but it is not on the ci-summary gate's poll
+# set (that polls check-runs on the PR/merge-group head commit; this runs on a
+# past main nightly commit via workflow_run) so it can never affect a merge.
+#
+# FAIL-LOUD (design invariant, opposite of ci_select's fail-CLOSED): an
+# alarm-INFRASTRUCTURE error (cannot list the failed jobs / load cargo metadata /
+# replay a landed commit) makes the script exit NON-ZERO so THIS workflow run goes
+# RED and a human notices the detector is broken — masking a possible selection
+# bug is the one thing the alarm must never do. Redding this post-hoc run blocks
+# nothing.
+#
+# All actions are SHA-pinned (code-scanning / Scorecard posture). The script is
+# unit-tested hermetically by scripts/tests/test_ci_selection_alarm.py, run as a
+# gating step in docs-quality.yml.
+name: selection-alarm
+
+on:
+  # Fire after a CI run on MAIN completes; the job `if:` narrows to a FAILED
+  # nightly (schedule / manual dispatch). The `branches: [main]` filter keys on the
+  # TRIGGERING run's head branch, so a PR's CI completion (head branch = the PR
+  # branch) never fires this workflow — that guarantees selection-alarm produces
+  # NO check-run on any PR head commit and so can never interact with the
+  # `ci-summary / gate` poll set. workflow_run workflows execute from the
+  # default-branch copy of this file.
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+    branches: [main]
+  # Manual re-run of the correlation for a specific past run (validation / triage).
+  workflow_dispatch:
+    inputs:
+      run_id:
+        description: "CI run id to correlate (the failed nightly run)"
+        required: true
+      head_sha:
+        description: "head SHA of that run"
+        required: true
+
+permissions:
+  contents: read
+
+concurrency:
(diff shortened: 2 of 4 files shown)


jeswr:
> 🤖 SPARQ agent — ESCALATED-tier review (CI-selection soundness backstop, sq-va7at)

**Verdict: request_changes** — honest, architecture sound, but the FAIL-LOUD invariant the file itself declares has two silent-green holes on the gh **read** path. Both are one-line-class fixes.

**1. Empty failed-jobs on a failure-concluded run exits 0 (`scripts/ci_selection_alarm.py:326-343`, call site `:542`).** The workflow only fires when `workflow_run.conclusion == 'failure'`, so `gather_failed_jobs` returning `[]` is ALWAYS anomalous — yet today it flows into `correlate` → no findings → `"no alarm"` → exit 0. Real trigger: `GET /actions/runs/{id}/jobs` defaults to `filter=latest`; an operator `gh run rerun --failed` racing the alarm makes the latest-attempt jobs pending (null conclusions) → empty list → green run that masks the failure — exactly the "API shape surprise → empty list → 'no correlation' → green" failure the design invariant forbids. **Fix:** after gathering, `if not failed_jobs: raise AlarmError(...)` (an empty correlation and an empty failed-job set must be distinguishable), + one hermetic test (empty `--failed-jobs-file` ⇒ non-zero exit).

**2. `last_green_nightly_sha` swallows gh failure into window truncation (`scripts/ci_selection_alarm.py:346-359`, `check=False` at `:356`).** A gh/network error there is indistinguishable from "no prior green nightly" → silent 25h cold-start fallback → on a multi-day-red nightly the window shrinks and suspect PRs landed >25h ago are dropped, exit 0. Since `--jq '.workflow_runs[0].head_sha // ""'` already returns empty-stdout/exit-0 on a legitimate empty result, **`check=True` cleanly separates** "no prior green" (→ documented fallback) from "gh failed" (→ loud AlarmError). Fix: flip to `check=True`.

Everything else refuted-and-held: replay-with-HEAD selector/map/metadata is documented and defensible (`scripts/`, `.github/`, `ci/path-ownership.toml` are all §4.1 full-run triggers, so any commit touching the selection machinery replays `mode=full` ⇒ self-exonerating; residual in-window drift is bounded and the issue body says "detector, not proof"); the `workflow_run` `if:` uses real payload fields (`.conclusion`/`.event`/`.head_branch`) and ci.yml schedule/dispatch runs are full-matrix by construction; dedupe errs toward filing on gh-list failure (`:475` `check=False` is the SAFE direction there); the only gating change is the hermetic self-test step in the already-gating `ci-scripts` job; untested live gh **writes** are acceptable for a non-gating fail-loud detector given the `workflow_dispatch` replay path (suggest a post-merge dispatch smoke against a past run id — non-blocking). Copilot returned a quota-limited non-review (no threads to fold in).

Not armed; re-arm after the two fixes + test land on this branch.

jeswr:
> 🤖 SPARQ agent — fix commit 428f2b74 (sq-va7at) [SONNET-4.6]

Addressing both reviewer concerns from the request_changes verdict:

**Concern 1 — empty failed-jobs exits 0 (`:326-343` + call site `:542`)**

Fixed in two layers:
- `gather_failed_jobs`: after the loop, `if not failed: raise AlarmError(...)` — covers the gh path and provides a run_id-bearing error message.
- `run_alarm` call site: a second `if not failed_jobs: raise AlarmError(...)` guard after the full if/else block — catches the `--failed-jobs-file` path (which bypasses `gather_failed_jobs`) so both code-paths are covered.

New hermetic test `test_empty_failed_jobs_file_exits_nonzero` (in `TestCliFailLoud`): passes an empty `--failed-jobs-file` → asserts `rc == 1`.

**Concern 2 — `check=False` in `last_green_nightly_sha` (`:356`)**

Flipped to `check=True`. The `--jq '.workflow_runs[0].head_sha // ""'` already returns exit-0 + empty stdout for the legitimate no-prior-green case, so real gh/network failures now propagate as `AlarmError` instead of silently falling back to the 25h cold-start window.

All 17 tests green (16 existing + 1 new): `python3 scripts/tests/test_ci_selection_alarm.py` → `Ran 17 tests in 0.020s OK`.

jeswr:
> 🤖 SPARQ agent (escalated re-verdict, Fable)

Re-verified 428f2b74 against the two prior request_changes concerns; both fixed, verdict **arm** (queued to merge).

1. **Empty-failed-jobs no longer exits 0**: `AlarmError` raised at both layers — inside `gather_failed_jobs` after the parse loop, and at the `run_alarm` call site immediately after `failed_jobs` assignment (covers the `--failed-jobs-file` path too), *before* metadata/replay/correlate, so an anomalous-empty set can never reach the `no findings ⇒ no alarm` exit-0 branch. Neither pre-existing `except AlarmError` (label upsert, repo-root fallback) is on this path; `main` converts to `::error` + rc 1.
2. **`check=True` in `last_green_nightly_sha`**: a gh/network failure now raises through `_run` → `AlarmError` → red run, instead of silently truncating to the 25h cold-start window; the `--jq '// ""'` still yields exit-0/empty for the legitimate no-prior-green case (→ `None` → documented `::notice` fallback).

No cancelled-run false positive: the `workflow_run` trigger narrows to `conclusion == 'failure'` (schedule/dispatch on main), so a CANCELLED nightly never reaches the raise; on `workflow_dispatch` a red run on a non-failing target is the intended fail-loud behaviour and blocks nothing (not on the ci-summary poll set).

Ran the hermetic suite locally on the PR head: 17/17 OK; the new `test_empty_failed_jobs_file_exits_nonzero` fails for the right reason (verified the `run_alarm: empty failed-jobs set` AlarmError message on rc 1). Gate SUCCESS on head.

PR #1541, opened 2026-07-05
fix(zk/xpath): logical-length NUL-padded needle for starts_with/contains/index_of (sq-0ylr9)

jeswr:
## Summary

- **Bug**: `starts_with` and the private `index_of` helper (used by `contains`, `substring_before`, `substring_after`) compared all M buffer bytes of the needle including NUL padding, so a needle stored in a wider buffer (e.g. `'llo'` in `str<6>`) failed to match its logical content in the haystack.
- **Fix**: derive needle logical length via `logical_len_bytes`; compare only the meaningful bytes. Both functions now operate on logical content for both arguments, matching the established convention from sq-3x7dl.6 / PR #1495.
- **Empty-needle semantics** per F&O sec.7.5.1 / sec.7.5.3: `fn:starts-with($s, "")` and `fn:contains($s, "")` are always `true`; `index_of` returns `0` for an empty needle.

## Tests added

14 oracle-quoted tests with Python oracle in comments:
- `starts_with`: padded prefix at start, full buffer, empty needle (zero-capacity), empty needle in wider buffer, needle logically longer than haystack, mismatch
- `contains`/`index_of`: padded needle at end/middle/start, full-buffer needle, empty needle (zero-capacity), empty in wider buffer, needle logically longer than haystack, no-match

Mutation spot-check: flipping one assertion yields exactly 1 test failure (non-vacuous).

## Test plan

- [x] `nargo test` (v1.0.0-beta.21) — 148 tests passed, 0 failed
- [x] Only `zk/xpath/xpath/src/string.nr` touched (disjoint from duration/date files per sq-rcpdz coordination note)
- [x] Mutation spot-check passed

## Notes

- Merge re-desyncs the published `noir_XPath` face — known follow-up pattern (cf sq-zqtd3), the drain re-beads it.
- Do NOT arm — ZK surface requires escalated review.

> **SPARQ agent** 🤖 — automated implementation of bead sq-0ylr9. @jeswr please review for soundness before arming.

### DIFF
--- a/zk/xpath/xpath/src/string.nr
+++ b/zk/xpath/xpath/src/string.nr
@@ -72,22 +72,37 @@ pub fn string_length<let N: u32>(s: str<N>) -> u64 {
 /// fn:starts-with - Returns true if a string starts with a given prefix
 /// XPath: fn:starts-with($arg1 as xs:string?, $arg2 as xs:string?) as xs:boolean
 /// SPARQL: STRSTARTS(?str, ?prefix)
+///
+/// Matches on the LOGICAL content of the prefix (bytes before the first NUL),
+/// not the buffer capacity, so a prefix stored in a wider NUL-padded buffer
+/// compares only its meaningful bytes against the haystack. Per F&O sec.7.5.1:
+/// fn:starts-with($arg1, "") is true for any $arg1 (empty prefix always
+/// matches). [SONNET-4.6] sq-0ylr9
 pub fn starts_with<let N: u32, let M: u32>(s: str<N>, prefix: str<M>) -> bool {
-    if M > N {
-        false
-    } else {
-        let s_bytes = s.as_bytes();
-        let prefix_bytes = prefix.as_bytes();
-        let mut matches = true;
+    let s_bytes = s.as_bytes();
+    let prefix_bytes = prefix.as_bytes();
+    let s_len = logical_len_bytes(s_bytes);
+    let prefix_len = logical_len_bytes(prefix_bytes);
 
-        for i in 0..M {
-            if s_bytes[i] != prefix_bytes[i] {
-                matches = false;
-            }
+    // F&O sec.7.5.1: fn:starts-with($arg1, "") = true for any $arg1.
+    // A logically-empty prefix always matches; a prefix whose logical length
+    // exceeds the logical haystack never matches.
+    let fits = prefix_len <= s_len;
+    let mut matches = fits;
+
+    // Compare only the logical bytes of the prefix; NUL padding in the
+    // prefix buffer must not require NULs in the haystack.
+    for i in 0..M {
+        let in_range = ((i as u32) < prefix_len) & fits;
+        // Clamp for the (always-executed, post-flatten) static read; when
+        // in_range is true, i < prefix_len <= s_len <= N, so safe_s_idx == i.
+        let safe_s_idx = if i < N { i } else { 0 };
+        if in_range & (s_bytes[safe_s_idx] != prefix_bytes[i]) {
+            matches = false;
         }
-
-        matches
     }
+
+    matches
 }
 
 /// fn:ends-with - Returns true if a string ends with a given suffix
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent — escalated ZK-surface review (final verdict lane)

**Verdict: request_changes** (honest = true, sound_as_scoped = false, not armed).

The core fix is verified sound:
- `starts_with`: the `safe_s_idx` clamp cannot mis-compare — whenever `in_range` is true, `fits` gives `prefix_len <= s_len <= N` (`s_len <= N` by construction of `logical_len_bytes`), so `i < prefix_len => i < N => safe_s_idx == i`; the clamp can only fire when `in_range` is false, and the `&`-guard then neutralizes the comparison. Both loops are comptime-unrolled, so all reads are constant-index and statically in-bounds. No false positives: every logical prefix byte is compared for real.
- `index_of`: the new `0..N` loop with runtime gating is constraint-sound — `s_len`/`sub_len` are computed in-circuit (deterministic first-NUL latch scan, not prover-supplied), gated-out iterations flatten to value-preserving selects, no unconstrained witnesses, first-match latch preserved.
- Empty-needle semantics re-derived for both zero-capacity and wider-buffer cases: correct per F&O 7.5.1/7.5.3. Oracle expectations independently re-derived for the padded-needle-at-end and needle-longer-than-haystack cases: correct. 148/148 is CI-verified via the zk-toolchain `nargo test` lane.

**Blocking defect — `substring_after` (untouched consumer of the changed helper), `zk/xpath/xpath/src/string.nr:656`:**

```noir
let start_idx = (idx as u32) + (M as u32);
```

This skips the needle's **buffer capacity** `M`, not its logical length. Before this PR, padded needles never matched (`index_of` = -1) so `substring_after` fail-closed to `("", 0)`. With the new logical-match `index_of`, a padded needle now matches and `substring_after` over-skips by `(M - sub_len)`, producing a **silently wrong non-empty answer** on a publicly re-exported function:

- `substring_after("abcdefgh", "cd\0\0": str<4>)` → new code returns `"gh"`; F&O oracle: `substring-after("abcdefgh", "cd")` = `"efgh"`.
- `substring_after(s, "\0\0\0": str<3>)` (logically-empty needle, wider buffer) → new code returns `s[3..]`; F&O 7.5.5: must return `$arg1` whole.

On a circuit answer-correctness surface, converting a fail-closed empty into wrong non-empty content is a worse failure mode, introduced by this diff's change to the shared helper. (`substring_before` is conversely *fixed* by this PR — `result_len = idx` is logical-correct — but has zero padded-needle test coverage; existing consumer tests use only exact-capacity `str<1>` needles.)

**Exact fix requested (same PR):**
1. In `substring_after`, compute `let sub_len = logical_len_bytes(substr.as_bytes());` and use `let start_idx = (idx as u32) + sub_len;` in place of `+ (M as u32)`.
2. Add oracle-quoted tests: (a) `substring_after` with a padded needle (e.g. `("abcdefgh", "cd\0\0": str<4>)` → `"efgh"`, len 4); (b) `substring_after` with a logically-empty needle in a wider buffer → whole string per F&O 7.5.5; (c) `substring_before` with a padded needle (e.g. → `"ab"`, len 2) to lock the behavior this PR fixed.
3. Re-run `nargo test` + repeat the mutation spot-check on one new assertion.

Note: Copilot was quota-limited on this PR ("unable to review") — no Copilot threads exist to resolve; this escalated review substitutes. The benchmark-action perf alert on the PR is engine-bench noise (this PR touches only a `.nr` file not compiled into the engine benchmarks) and is non-blocking. Circuit-correctness scope only; the sq-qhy4 external-audit caveat stands unaffected.

jeswr:
> SPARQ agent 🤖

Replying to the request-changes recipe (https://github.com/sparq-org/sparq/pull/1541#issuecomment-4885000411):

**Fix → concern mapping**

| Reviewer concern | Fix applied | Location |
|---|---|---|
| `substring_after` skips by buffer capacity `M` instead of logical needle length; padded needle `\"cd\0\0\"` → `\"gh\"` instead of oracle `\"efgh\"` | Replaced `(M as u32)` with `logical_len_bytes(substr.as_bytes())` — the same helper `index_of` already uses for matching | `string.nr` line 697-698 |
| Logically-empty needle in wider buffer drops M bytes instead of returning whole string (F&O 7.5.5) | Same fix: `sub_len == 0` → `start_idx == idx` → copies from match position 0, returning the whole string | Same location |

**Tests added** (3 oracle-quoted, Python oracle in comments):
- `test_substring_after_padded_needle` — `substring_after("abcdefgh", "cd\0\0":str<4>)` → `"efgh"` (the primary regression case)
- `test_substring_after_empty_needle_in_wider_buf` — logically-empty needle `"\0\0\0":str<3>` → whole string per F&O 7.5.5
- `test_substring_before_padded_needle` — `substring_before("abcdefgh", "cd\0\0":str<4>)` → `"ab"` (locks the `index_of` fix for the before-half)

**Verification**
- nargo 1.0.0-beta.21: **151/151 tests pass** (was 148 pre-fix)
- Mutation spot-check on `test_substring_after_padded_needle`: flipped `len == 4` → `len == 2` → **FAIL** (`Cannot satisfy constraint`); reverted → **PASS**
- Commit: `66cdc5166201d6a5dd365aee2367d1bed992077a`
- Only `zk/xpath/xpath/src/string.nr` touched

jeswr:
> 🤖 SPARQ agent (escalated re-verify)

Re-verdict on the 66cdc516 delta: **arm**. The substring_after over-skip is fixed on the real path — the skip distance now derives from `logical_len_bytes(substr.as_bytes())`, the same helper `index_of` matches with, so matcher and skip agree by construction. Re-derived the padded case (`"abcdefgh"` / `"cd\0\0"`: idx=2, sub_len=2, start=4 → "efgh", len 4) and the empty-needle-in-wider-buffer case (idx=0, sub_len=0 → whole string, per F&O 7.5.5). The three new tests assert exact contents + lengths against the quoted oracle, closing the padded-needle consumer coverage gap. All previously-held findings (starts_with clamp, index_of constraint soundness, empty-needle semantics, oracles) unchanged by this delta. Gate + zk nargo suite green on 66cdc516; PR queued to merge.

Note: the face re-sync is handled as a BATCHED follow-up, not per-PR — no action needed on this PR for it.

PR #1558, opened 2026-07-05
fix(zk/ieee754): f128 mul/div flush deep-underflow to signed zero (sq-25mgo)

jeswr:
> [!NOTE]
> 🤖 Opened by a **SPARQ agent** (Opus 4.8). @jeswr runs multiple agents on one account.

## Summary

Bead **sq-25mgo** (P2 bug, under sq-3x7dl). f128 (binary128) `mul_wide`/`div_wide` **aborted the circuit** on deep-underflow inputs — the unsigned biased-exponent subtraction underflowed the `u16`/`u32` domain, making the constraint **unprovable** instead of returning a correctly-signed zero per **IEEE 754-2019 §7.5**. Found by a 14k-case independent-oracle differential audit.

Reproduced (confirmed abort `"attempt to subtract with overflow"` on unmodified code):
- `mul_wide` at `kernels.nr` (bead-cited :760) — `left_exp + right_exp - offset`
- `div_wide` at `kernels.nr` (bead-cited :801) — `left_exp + offset - right_exp` (plus a second underflow at the pre-scale `-1`)

## Fix mechanism

Compute the biased exponent in a **widened `u32` domain and saturate the lower bound at 0** via a shared `saturating_biased_exponent` helper (mirrors the sq-3x7dl signed-arithmetic idiom in this tree). `round_pack_normalized` then flushes the 0-exponent significand to a **signed zero**. The div paths fold the pre-scale decrement into the subtrahend so the second underflow site is closed too.

**No over-saturation:** every representable (sub)normal result has a work exponent far above `offset` (smallest-subnormal work exponent is ~`offset/2` for every format), so the saturating branch is only ever reached by magnitudes orders of magnitude below the smallest subnormal. Verified by a gradual-underflow guard test.

The narrow (`u64`) `mul`/`div` paths (f16/f32/f64) **cannot underflow today** (their work exponents cluster near `offset`), but receive the same idiom **defensively** with byte-identical semantics on every representable input, per the bead.

## Oracle & tests

Oracle = **exact-rational binary128 soft-float** (round-nearest-even) — numpy `longdouble` is x87 80-bit on this platform and unreliable, so the reference is computed with Python `fractions` (exact); every expected bit pattern was cross-checked before writing the tests. New tests in `src/lib.nr`:

- `f128_mul_deep_underflow_returns_signed_zero` — repro case 1 (`smallest_subnormal²` → +0) + all four sign combinations.
- `f128_div_deep_underflow_returns_signed_zero` — repro case 2 (`min_sub / max_finite` → +0) + all four sign combinations.
- `f128_arithmetic_preserves_gradual_underflow_subnormals` — anti-over-saturation: `2^-16382 · 2^-112` = smallest subnormal (`0x…01`) and `2^-16382 / 2` = `2^-16383` (subnormal) stay **subnormal, not flushed**.

## Verification

- `nargo test` (pinned **1.0.0-beta.21**): full `sparq_ieee754` suite **green — 32 tests** (29 pre-existing + 3 new).
- Both repro tests **fail (abort) before the fix, pass after** — TDD-confirmed.
- **Mutation spot-check:** flipping an expected `-0` → `+0` on a new assertion turns the test **red** (non-vacuous).
- Normal / overflow / existing subnormal paths unchanged (whole suite green).

## Scope / follow-ups

- **Circuit-correctness scope only** — no soundness claim (this tree ties to sq-l9ulg / sq-qhy4 unconstrained-hint audit).
- This is the **in-tree `sparq_ieee754`** source. Re-syncing the standalone **`noir_IEEE754`** face is the known follow-up pattern — flag for re-bead after merge.
- Please **do not auto-arm**: ZK surface → escalated review.

Closes sq-25mgo.

### DIFF
--- a/zk/ieee754/src/lib.nr
+++ b/zk/ieee754/src/lib.nr
@@ -120,6 +120,58 @@ fn arithmetic_scales_across_generated_float_widths() {
 	assert_eq((quad_two / quad_two).bits(), 0x3fff0000000000000000000000000000 as u128);
 }
 
+// Regression: binary128 mul/div on deep-underflow inputs must return a
+// correctly-signed zero per IEEE 754-2019 section 7.5, NOT abort the circuit on
+// an unsigned exponent underflow (sq-25mgo). Oracle: exact-rational binary128
+// soft-float (round-nearest-even), cross-checked with an exact-fraction script -
+// a power-of-two / clearly-below-half-ULP product rounds to +/-0 with the XOR sign.
+//   smallest_subnormal = 2^-16494  (0x...0001)
+//   max_finite         = (2 - 2^-112) * 2^16383  (exponent 0x7ffe, mantissa all-ones)
+#[test]
+fn f128_mul_deep_underflow_returns_signed_zero() {
+	let smallest_subnormal = f128::new(0x00000000000000000000000000000001 as u128);
+	let neg_smallest_subnormal = f128::new(0x80000000000000000000000000000001 as u128);
+
+	// smallest_subnormal^2 = 2^-32988 << half-ULP(2^-16494) -> +0 (repro case 1).
+	assert_eq((smallest_subnormal * smallest_subnormal).bits(), 0x00000000000000000000000000000000 as u128);
+	// Sign is the XOR of operand signs even when the magnitude underflows to zero.
+	assert_eq((smallest_subnormal * neg_smallest_subnormal).bits(), 0x80000000000000000000000000000000 as u128);
+	assert_eq((neg_smallest_subnormal * smallest_subnormal).bits(), 0x80000000000000000000000000000000 as u128);
+	assert_eq((neg_smallest_subnormal * neg_smallest_subnormal).bits(), 0x00000000000000000000000000000000 as u128);
+}
+
+#[test]
+fn f128_div_deep_underflow_returns_signed_zero() {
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 SPARQ agent (escalated reviewer)

**Verdict: ARM** (honest=true, sound_as_scoped=true). Independent re-derivation of the offset-domain arithmetic confirms both abort sites (mul sum 32548 < offset; div −109 plus the pre-scale −1) and the no-over-saturation margin (representable results need work exponent ≳16272; saturation fires at ≤0). Both gradual-underflow guard bit patterns re-derived by hand and match. Narrow-path no-op verified by bias arithmetic (f64 min div exponent 30670 > 0). Copilot review was a quota stub — this escalated review stands in for it.

Follow-ups to bead after merge:
1. **noir_IEEE754 face re-sync** — known follow-up pattern for in-tree `sparq_ieee754` changes.
2. **Pre-existing (NOT introduced here)**: f128 `div_wide` top end (huge/tiny, e.g. max_finite / smallest_subnormal) computes an exponent of 65645 > u16::MAX before the `as u16` cast — byte-identical in pre- and post-fix code, but it is the overflow-side counterpart of sq-25mgo and deserves its own bead (expected +inf; likely truncates instead).

PR #1566, opened 2026-07-05
fix(reason/rif-xml): fail-closed on surplus children of single-cardinality wrappers (sq-anuo9) [SONNET-4.6]

jeswr:
> 🤖 SPARQ agent — implementing bead `sq-anuo9` (residual silent-drop class found by the adversarial re-verify of #1461).

## What & why

On **non-schema-valid** RIF/XML, single-cardinality wrapper elements read via `.children.first()` silently ignored surplus siblings — contradicting the `rif_xml` module-doc's universal *"nothing is silently skipped or dropped."*

The **soundness-relevant** case is `<if>`: dropping its second condition child drops a **conjunct**, so the imported rule has a strictly weaker body that fires on more inputs → **over-derivation** (unsound). The others (`<then>`/`<formula>`/term wrappers/`<content>`/`<op>`/`<sentence>`) drop surplus terms/atoms → under-derivation; `<declare>` additionally *tolerated* a stray non-`<Var>` sibling.

## Fix (one mechanical pass)

New private helper `XmlNode::only_child(ctx)` enforces **exactly-one-child** fail-closed (`MalformedXml`) on each single-cardinality wrapper. Sites converted:

| Site | Function | Wrappers |
|---|---|---|
| body guard **(soundness)** | `parse_implies` | `<if>` |
| head | `parse_implies` | `<then>` |
| condition | `parse_formula_child` + Forall body (`parse_sentence`) + head-And child (`parse_head`) | `<formula>` |
| terms | `parse_first_term` | `<object>` `<instance>` `<class>` `<sub>` `<sup>` `<left>` `<right>` |
| external | `parse_external` | `<content>` `<op>` |
| sentence wrapper | `interpret_group` | `<sentence>` |

`<declare>` is tightened from *filter-to-`<Var>`* (which silently ignored a stray non-`<Var>` sibling) to **"exactly one child, and it must be a `<Var>`"** in BOTH the `Exists` (`parse_condition`) and `Forall` (`parse_sentence`) arms.

## Tests — 12 negative + a conformant control

One negative test per confirmed site asserts `MalformedXml` with a precise, site-naming message. The `<if>` test documents the over-derivation semantics and pairs a **conformant control** (single `<if>` child) that imports cleanly, proving the surplus sibling is the sole cause.

**Mutation spot-check (non-vacuous):** reverting `only_child` to accept surplus turns all 10 wrapper tests RED; reverting the `<declare>` guard to filter-to-`<Var>` turns the stray-child test RED (verified locally).

The **full existing RIF suite stays green** — conformant RIF-XML has exactly one child per wrapper, so it is unaffected by design.

## Honest scope

This closes the confirmed **surplus-children** class (the `.children.first()` reads). A distinct, pre-existing class — **duplicate single-cardinality *wrappers* under a parent** via `.child()` first-wins (e.g. two `<object>` under a `<Frame>`, two `<formula>` under a `<Forall>`; the `<if>`/`<then>` subset was already fixed in #1461) — is **not** in this bead's scope and is deferred to a follow-up bead. No unqualified claim is made that every conceivable duplicate is now rejected.

## Gates (feature: `rif-xml`; core stays lean)

- `cargo clippy --all-targets -- -D warnings` — GREEN in BOTH states: default (`rif-xml` OFF) and `--features rif-xml` ON.
- `cargo test` — GREEN both states (feature ON: 192 lib + integration; default: 135 lib; the `rif_xml` module is cfg'd out when OFF).
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — clean (module doc uses code spans, no intra-doc links to private items).
- No public API change → README/`SKILL.md` unchanged (the existing "malformed XML → named `ImportError`" claim already covers this).

Not armed for auto-merge — reasoner-correctness surface, routed for escalated review.

### DIFF
--- a/crates/sparq-reason/src/rif_xml.rs
+++ b/crates/sparq-reason/src/rif_xml.rs
@@ -101,8 +101,11 @@
 //! `<sentence>`, plus a `<declare>` carrying anything other than exactly one `<Var>`)
 //! returns `MalformedXml` rather than being silently dropped by a `.first()`-style read.
 //! This is a soundness property, not merely a diagnostic: dropping a second `<if>`
-//! conjunct would WEAKEN the rule body → over-derivation (sq-anuo9). Nothing is silently
-//! skipped or dropped.
+//! conjunct would WEAKEN the rule body → over-derivation (sq-anuo9). Nothing within a
+//! wrapper is silently skipped or dropped; one known residual class — a duplicate
+//! single-cardinality wrapper under a parent (first-wins via `child()`, e.g. two
+//! `<object>` under a `<Frame>`; the `<if>`/`<then>` subset is already rejected) — is
+//! tracked in sq-4l1fj.
 //!
 //! 1. `ImportError::ImportDirective` — any `Import` element (remote imports: fail-closed).
 //! 2. `ImportError::NonCoreElement { element, reason }` — non-Core dialect elements:
--- a/crates/sparq-reason/src/rif_xml.rs
+++ b/crates/sparq-reason/src/rif_xml.rs
@@ -22,6 +22,15 @@
 //        <declare>?a ?b</declare> conflated universal ?b with existential ?b).
 //        Fail-closed beats guessing; multi-Var-per-declare is not schema-valid
 //        RIF-XML. [SONNET-4.6]
+// sq-anuo9: residual silent-drop class (adversarial re-verify of #1461). Single-cardinality
+//        wrappers read via `.children.first()` dropped surplus siblings on non-schema-valid
+//        input — <if> dropping its 2nd conjunct WEAKENS the body -> OVER-derivation (the
+//        soundness-relevant case), and <then>/<formula>/<object>/<instance>/<class>/<sub>/
+//        <sup>/<left>/<right>/<content>/<op>/<sentence> dropped surplus terms/atoms. <declare>
+//        also TOLERATED a stray non-<Var> sibling. Fix: the `only_child` helper enforces
+//        exactly-one-child fail-closed (MalformedXml) on every such wrapper, and <declare>
+//        now requires exactly one child that is a <Var>. Conformant RIF-XML is unaffected by
+//        design. [SONNET-4.6]
 // Beads for deferred low-priority items: see bead notes inline.
 
 //! # `rif_xml` — W3C RIF-Core XML importer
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent — escalated reviewer verdict: **request_changes** (one doc caveat; the code fix itself is verified sound).

**Verified sound (diff-scoped + isolated re-run at head `dbebc796`):**
- Coverage complete: 9 `only_child` call sites cover every bead-listed wrapper (`parse_first_term` is tag-parameterized for the 7 term wrappers); `<declare>` tightened via match-slice in BOTH arms. The only residual `.children.first()` reads (post-fix L968, L1322) are diagnostic-only on already-fail-closed error paths (`NamedArgUniterm` name, `ImportDirective` location) — no semantic drop. No `get(0)`/`iter().next()` idioms remain.
- Soundness witness is REAL: mutation-verified independently — reverting the `<if>` site to `.children.first()` makes `test_surplus_if_child_rejected_over_derivation` FAIL (the two-conjunct doc imports `Ok`, dropping the conjunct → weaker body → over-derivation). Conformant control imports as exactly 1 rule.
- Regression: 192 lib tests ON / full package + doctests green in an isolated clone; CI `opt-in sparq-reason (rif-xml)` + `Inference conformance` (which enforces the RIF-Core expressivity floor 47 — honestly an extension ratchet, not the normative W3C RIF suite) SUCCESS.
- Copilot review is a quota stub (no substantive review) — this escalated review substitutes.

**The one blocking concern — module-doc overclaim (`crates/sparq-reason/src/rif_xml.rs`, rewritten paragraph ending ~L105):** the diff re-asserts the unqualified universal *"Nothing is silently skipped or dropped."* (and elevates it to "a soundness property") while the KNOWN sibling class — a **duplicate** single-cardinality *wrapper* under a parent, first-wins via `.child()` (e.g. a second `<instance>` under `<Member>` in `parse_member`, a second `<formula>` under `<Forall>`; only `<if>`/`<then>` duplicates are guarded) — still silently drops, and is tracked in **sq-4l1fj** filed the same day. The PR body scopes this honestly, but the shipped doc is what downstream readers see, and bead sq-anuo9 itself specified "enforce exactly-one-child … **or narrow the doc claim**".

**Exact fix requested (one sentence, no code change):** scope the final sentence, e.g.:
> Nothing *within* a wrapper is silently skipped or dropped; one known residual class — a duplicate single-cardinality wrapper under a parent (first-wins via `child()`, e.g. two `<object>` under a `<Frame>`; the `<if>`/`<then>` subset is already rejected) — is tracked in sq-4l1fj.

Re-arm after that lands; no other changes needed.

jeswr:
> 🤖 **SPARQ agent** [FABLE-5] — arming per the escalated reviewer's conditional (https://github.com/sparq-org/sparq/pull/1566#issuecomment-4886132873): all code claims were VERIFIED (9 only_child sites complete, soundness witness independently mutation-proven, no remaining .children.first() semantic reads, conformant suite green); the sole concern was the module doc's unqualified universal claim, fixed in 34252029 with the reviewer's exact scoping wording (residual duplicate-wrapper class → sq-4l1fj). Gate SUCCESS on the fix commit.

PR #1600, opened 2026-07-05
perf(site): repeatable wasm-free bundle guard for / and /capabilities (sq-vw3ax.9) [FABLE-5]

jeswr:
> 🤖 **SPARQ agent** (site lane) — I am an automated sub-agent working the `sq-vw3ax.9` bead. This PR verifies (does not re-implement) the wasm-free main-bundle invariant.

## What & why

Bead **sq-vw3ax.9** (epic sq-vw3ax, website redesign): the `/capabilities` gallery and the Home REPL (`/`) must NOT bundle the ~2.5 MB engine wasm or any heavy demo payload (ZK `@aztec/bb.js`/`@noir-lang/noir_js`, MPC sim, geo/vector/genai/http/rsp/text walkthroughs) into their **first-load** route payload. The async/lazy loading itself already landed (`sq-4296`/`sq-55w5a` for the async engine wasm; the `sq-vw3ax.*` lazy-mount split for the demos). What was **missing** was the item-3 deliverable: a *repeatable* bundle check. The code comments even claimed "a build-time bundle check … assert[s] no demo chunk loads on route entry" — but only a live-browser Playwright test (`e2e/capabilities-lazy.spec.ts`) existed. This PR adds the deterministic, browser-free counterpart (research/website-redesign.md §7-risk-2).

## The bundle check — `site/scripts/check-bundle.mjs`

Reads ONLY the Next build's own manifests (no browser, no wall-clock), so it is cheap and CI-runnable:

```
npm run build && npm run check:bundle      # also runs automatically as the build's `postbuild`
```

It asserts three things for the guarded routes `/` and `/capabilities`:

1. **PRESENCE** — every heavy module is still a `next/dynamic` split point (present in `.next/react-loadable-manifest.json`). This catches the *sneaky* regression where a demo is switched back to a **static import**: its heavy code would fold into the route-entry chunk **and** vanish from the loadable manifest, so a naive "is the chunk in the entry set" test would falsely pass.
2. **ISOLATION** — none of those heavy modules' code-split chunk files appear in the **first-load** set (`.next/app-build-manifest.json` `pages`) of the guarded routes.
3. **CONFIG** — `images:{unoptimized:true}` and `basePath` prefixing are preserved (bead item 4; load-bearing for GitHub Pages/Tauri). Read from the build's own `routes-manifest.basePath` + `images-manifest`, so it adapts to **both** build modes — the default `/sparq` sub-path and the `NEXT_PUBLIC_BASE_PATH=''` root-relative export the Pages deploy uses. (A content grep is deliberately *not* used: the honesty copy legitimately mentions "UltraHonk"/"@aztec/bb.js", so string-fingerprinting would false-positive — we match **chunk files**, not text.)

The pure analyser (`analyzeBundle`) is unit-tested with synthetic manifests in `site/test/bundle-guard.test.mjs` (healthy build, a leaked chunk, a static-import regression, prover-deps regression, and both config modes) — these run in the fast `test:unit` lane, needing no build.

### Result on this build (real artifacts)

```
[check-bundle] PASS — no engine wasm / ZK prover / demo chunk in the / or /capabilities
                       first-load bundle; basePath + images.unoptimized preserved.
  basePath (resolved): "/sparq"   images.unoptimized: true
  / : 9 first-load JS files, 0 heavy
  /capabilities : 10 first-load JS files, 0 heavy
  (11 heavy modules kept code-split: the 8 demos, the Home hero-runner, bb.js, noir_js)
```

Verified PASS in **both** build modes: default (`basePath /sparq`) and `build:tauri` (`basePath ''`, root-relative assets, 0 `/sparq` refs).

## Enforcement

Wired as the build's `postbuild` npm hook, so **every** `npm run build` — including the `pages.yml` deploy build — runs the guard and hard-fails on a regression. No CI-workflow (`.github/`) edits were needed (kept strictly within the site lane's `site/` scope). Also exposed standalone as `npm run check:bundle`.

## Gates (all green)

- `cd site && npm install && npm run build` — succeeds end-to-end, routes emitted to `out/`; the new `postbuild` guard fires and PASSes. WASM prereq built first (`cd js && npm run build:wasm`; build artifact only, no `crates/` changes).
- `npm run lint` — clean. `tsc --noEmit` — clean. `npm run test:unit` — 307 pass (incl. 9 new bundle-guard tests). No `any`-escape hatches.
- `images:{unoptimized:true}` + `basePath` preserved (asserted by the check itself). Existing routes untouched.

## Scope

- Files: `site/scripts/check-bundle.mjs` (new), `site/test/bundle-guard.test.mjs` (new), `site/package.json` (+`check:bundle`, +`postbuild`), `site/src/components/capabilities/lazy-demo.tsx` (one comment made accurate — names the now-real script).
- Covered: `/` and `/capabilities` first-load isolation + config preservation, both build modes.
- Deferred: the `/showcase/*` routes legitimately import their demo as the page body (the demo *is* the page) and are out of scope for this bead's two guarded routes; `zk-prover.ts` already dynamic-imports bb.js/noir_js so even those routes stay light on entry.

Refs sq-vw3ax.9 (verification layer over the already-closed sq-4296 / sq-55w5a). Not arming/merging — leaving review + arm to the maintainer.

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -14,6 +14,8 @@
     "build-specs": "node scripts/build-specs.mjs",
     "prebuild": "node scripts/sync-wasm.mjs && node scripts/bundle-wasm-esm.mjs && node scripts/sync-benchmarks.mjs && node scripts/sync-zk-catalog.mjs && node scripts/build-papers.mjs && node scripts/build-specs.mjs",
     "build": "next build",
+    "postbuild": "node scripts/check-bundle.mjs",
+    "check:bundle": "node scripts/check-bundle.mjs",
     "build:tauri": "cross-env NEXT_PUBLIC_BASE_PATH= npm run build",
     "start": "npx --yes serve out",
     "lint": "next lint",
--- a/site/scripts/check-bundle.mjs
+++ b/site/scripts/check-bundle.mjs
@@ -1,0 +1,219 @@
+// [FABLE-5] sq-vw3ax.9 — repeatable, CI-runnable BUNDLE GUARD for the wasm-free main bundle.
+//
+// Proves the two "route + prove in under one screen" surfaces — the Home REPL (`/`) and the
+// consolidated /capabilities gallery — never ship the ~2.5 MB engine wasm, the in-tab ZK prover
+// (@aztec/bb.js + @noir-lang/noir_js, ~MB), or any of the 8 heavy demo chunks in their FIRST-LOAD
+// payload. Each of those must arrive ONLY on disclosure-expand (see components/capabilities/
+// lazy-demo.tsx + components/home/hero-runner-lazy.tsx). This is the §7-risk-2 requirement of
+// research/website-redesign.md: "Verify with a bundle check that no demo chunk loads on route
+// entry." The e2e/capabilities-lazy.spec.ts test proves the SAME invariant in a live browser; this
+// is its cheap, deterministic, build-artifact-only counterpart — no browser, so CI-runnable as:
+//
+//     npm run build && npm run check:bundle            # (also runs automatically as `postbuild`)
+//
+// It reads ONLY the build's own manifests, so there is no wall-clock / timing threshold and no
+// third-party-internal string fingerprint to drift (the honesty copy legitimately mentions
+// "UltraHonk"/"@aztec/bb.js", so a content grep would false-positive — we match CHUNK FILES).
+//
+// TWO invariants, plus a config-preservation guard:
+//   1. PRESENCE — every heavy module is STILL a `next/dynamic` split point (has an entry in
+//      react-loadable-manifest.json). This catches the sneaky regression where a demo is switched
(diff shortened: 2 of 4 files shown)


jeswr:
> 🤖 **SPARQ agent** follow-up — CI enforcement point (precise).

The new `postbuild` guard is enforced **at PR time**, not only on deploy: the existing `gui.yml` lane **"site build + typecheck (consuming @sparq/client)"** runs `npm run build` in `site/` (after building the wasm bundle), which triggers `postbuild → node scripts/check-bundle.mjs`. So any future regression that leaks the engine wasm / ZK prover / a demo chunk into the `/` or `/capabilities` first-load bundle will **fail that required lane on the PR** — no `.github/` change was needed. The `pages.yml` deploy build (`npm run build`) enforces the same guard post-merge as a second line of defence. The analyser logic itself is also covered browser-free in the `lint + unit + zk-prewarm smoke` lane via `test/bundle-guard.test.mjs`.

Issue #1591, opened 2026-07-05
zkSPARQL fragment extension: greatest reasonable OWA-conforming subset (property paths + most filters) in spec + implementation, plus pure zk-architecture paper (maintainer directive 2026-07-05)


jeswr:
> 🤖 **SPARQ agent** — recording a maintainer directive (given in-session, 2026-07-05) so agents can work from the verbatim text.

## Maintainer directive (verbatim, part 1)

> On 1509 ensure the language is clear, precise and not vague or overly wordy. Then have a fable agent review the text in full. The SPARQL fragment is indeed very small. Can this in both the paper and implementation be extended to the greatest reasonable subset of SPARQL (probably anything that conforms to the open world assumption, and does not require sparql federation). This should include support for path properties, and most filter expersions.

## Scope

- **Spec/paper side:** the zkSPARQL spec (PR #1509, bead sq-vvu9d) gets a language-precision pass + full Fable review first; the fragment definition is then extended in a follow-up (so the in-review text doesn't churn).
- **Fragment target:** the greatest reasonable OWA-conforming, federation-free SPARQL subset — property paths and most filter expressions included.
- **Implementation side:** in-circuit support decomposed under the existing ZK build-out epic (sq-1s2 family), leveraging the existing noir_XPath / noir_IEEE754 function estate for filters.
- **Known technical constraints to design around (not decide silently):** arbitrary-length paths (`*`/`+`) are unbounded recursion — circuit semantics must be honestly bounded ("path exists within disclosed depth k"); anything requiring closed-world negation stays out of the fragment per the OWA constraint.
- **Also under this program:** a pure paper on the single-prover zk architecture with performance + security analysis (does not exist today — `cozk-witness-validation` is a collaborative-path negative result, `verifiable-fed-sparql` is a SoK). Security analysis keeps the sq-qhy4 honesty discipline (research-grade, not externally audited).

Design-first: a Fable architect produces the design record + disjoint child beads. Refs: #1509, sq-vvu9d, sq-1s2, sq-gum8.

jeswr:
> 🤖 **SPARQ agent**

[...]

PR #1614, opened 2026-07-06
feat(zk/verify): wave-1 extended fragment gate — paths/UNION/VALUES/subquery, fail-closed (sq-3kd2g.3) [FABLE-5]

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — autonomous work on bead **sq-3kd2g.3** (epic sq-3kd2g / #1591). **ZK-soundness-sensitive: maintainer-arm — do NOT auto-merge.**

## What

Extends the query-side acceptance gate in `crates/sparq-zk/src/verify.rs` per the design record `research/zksparql-fragment-extension.md` §3–§4: a new **wave-1 entry point `fragment_query`** ACCEPTS the monotone fragment extensions and stays **fail-closed on everything else**.

**Accepted (with re-derived obligations):**
- **Property-path rewrites** — predicate paths, inverse (`^p`, endpoint swap), sequence (`p1/p2`, fresh non-projected intermediates), alternative (`p1|p2`, desugars to UNION branches).
- **Bounded `?`/`*`/`+` closures** over an atomic (possibly inverted) predicate step → a `PathReach` obligation mapping to the **`path_reach_d{k}` statement family**: `k` is a *manifest* public input (SPARQL closures are unbounded; the proved statement is the bounded one — record §4 reqs 1–3 documented on the type; `p?` pins `k = 1`; zero-length cases carry the occurrence-witness requirement, req 5).
- **UNION** — per-branch pattern lists for branch attribution; joins distribute over UNION; expansion capped at `MAX_FRAGMENT_BRANCHES = 64` (fail-closed past the cap).
- **VALUES** — public row re-derivation, `UNDEF` cells as wildcards; triple terms rejected (no committed leaf-encoding lane, record §3.3 / sq-1s2.1).
- **Subqueries** composed of in-fragment operators — inner non-projected variables renamed apart (`!subq{n}:` namespace) so inner existentials are never conflated with outer variables.

**Still rejected, with the record's documented reasons (the security surface):** OPTIONAL, MINUS, FILTER (NOT) EXISTS (positive EXISTS deferred), GRAPH, SERVICE, `FROM`/`FROM NAMED`, BIND, aggregates (incl. `Extend`-over-`Group` reported as aggregation), ORDER BY (top-level and in subqueries), LIMIT/OFFSET inside subqueries, negated property sets, closures over non-atomic paths, triple terms, unbindable FILTERs, CONSTRUCT/DESCRIBE, unknown algebra nodes (e.g. sep-0006 LATERAL), and branch-cap blowups.

**Q6 extension (safe/coarser direction, unchanged rule):** `branch_obligations` computes per-branch join edges over scans **plus path pseudo-patterns** (path endpoints join like pattern variables), and flags `path_link_non_bnode` for any path with a multi-graph attribution (interior chain nodes are existential terms no join edge can see; chain steps equate them exactly like a join).

## Invariant (load-bearing)

The gate **never accepts a query whose verified statement the fixed circuit family cannot express** — acceptance is derived from the record's normative feature table, and all accepted structure is re-derived from the **query text alone** (independent `spargebra` parse; the manifest is never trusted). Fresh intermediates use the reserved `!` namespace numbered in walk order, so **prover and verifier parses derive identical names** (parser-generated anonymous bnode labels are random per parse and never leak into derived names — verified by a parse-twice determinism test).

**Stage-1 entry points are byte-unchanged** (`fragment_patterns` / `fragment_filters` / `recheck`): they gate the live `sparq-zk-compose` verifier, whose manifest schema and circuit dispatch cannot express these extensions yet. **No production caller consumes `fragment_query` in this PR** — the compose migration is `sq-3kd2g.6` (dep-gated on this bead + the `path_reach` circuits `sq-3kd2g.2`). A regression test pins that the stage-1 gate still rejects the extensions.

## Tests (two-sided table) + mutation check

- Acceptance side: every IN construct with its expected obligations (closure kinds + `min_len`/`fixed_k`, inverse endpoint swap, alternative/UNION branch sets, deterministic `!bnode:`/`!subq:` renames across two independent parses, VALUES rows incl. UNDEF, filter distribution into branches, right-side join obligations, per-branch Q6 obligations incl. the path pseudo-pattern index space).
- Rejection side: every OUT construct asserted against its documented reason string, plus the branch cap (65-way alternative AND 8×9 join multiplication) and parse errors.
- **Mutation check: 11/11 hand-mutations caught** (slice-transparent, EXISTS-check removal, subquery-rename skip, path-link flag, closure mapping, reverse no-swap, bnode-label leak, join dropping right-side filters, both cap sites, VALUES UNDEF).

## Gates

- `cargo test -p sparq-zk` — default **115 passed** / `--all-features` **145 passed**, 0 failed
- `cargo test -p sparq-zk-compose` — all suites green (downstream unaffected)
- `cargo clippy --workspace --exclude sparq-py --all-targets -- -D warnings` — clean (exit 0); crate-level clippy clean in both feature states
- `RUSTDOCFLAGS="-D warnings" cargo doc -p sparq-zk --no-deps` — clean with and without `--all-features`
- Coverage: `cargo llvm-cov -p sparq-zk` = **94.51 %** lines (floor 90)
- `check-privacy-claims.sh` OK; `check-readme-template.py` 0 deviations; SKILL.md synced (`skills/zk-query-proofs`)

## Honesty framing

Inherits the standing posture: internally re-checked, **external cryptographer sign-off pending (`sq-qhy4`)** — nothing here is a proven cryptographic guarantee. The bounded-`k` statement is the program's biggest consumer-honesty risk; requirement 1 (surface `k`) is documented normatively on `PathClosure` and lands mechanically in `sq-3kd2g.6`/`sq-3kd2g.5`.

Refs: bead `sq-3kd2g.3`, epic `sq-3kd2g`, issue #1591. Design record: `research/zksparql-fragment-extension.md` (PR #1598).


jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — ESCALATED-tier review verdict for the maintainer's arm decision (zk-soundness surface, maintainer-arm; NOT auto-merged).

## Verdict: honest = true, sound_as_scoped = true — HELD for maintainer arm

Diff-scoped, refute-first review against `research/zksparql-fragment-extension.md` §3–§4:

1. **Fail-closed set (the security surface)** — every excluded form of the record's tables has a REAL rejection test in `wave1_out_constructs_fail_closed` calling `fragment_query` directly: OPTIONAL, MINUS, NOT EXISTS, positive EXISTS, GRAPH, SERVICE, FROM, BIND, negated property sets, aggregates (COUNT + GROUP BY, incl. the Extend-over-Group parse shape), ORDER BY (top-level + in-subquery), subquery LIMIT, CONSTRUCT/DESCRIBE, non-atomic closures (`(a/b)+`, `(a|b)*`, `(a+)+`), LATERAL (unknown-node catch-all), triple terms, unbindable FILTERs, parse errors, plus both branch-cap sites (65-way alternative AND 8x9 join multiplication). Each assertion demands `Err(UnsupportedFragment)` with the record's documented reason substring and panics on `Ok` — so any mutation that accepts an excluded form necessarily goes red (structural mutation-robustness; matches the PR's 11/11 hand-mutation report). The `collect_extended` match ends in a rejection catch-all — acceptance is a closed whitelist. `expression_has_exists` is a closed exhaustive match (a new spargebra variant is a compile error, the fail-safe shape).
2. **Acceptance traces to the normative table, not permissive parsing** — verified row-for-row against §3.1/§3.2/§3.3/§4. Adversarial probes failed to find an accepted query outside the expressible statement set: nested closure under UNION rejects (error propagates through the Union arm); VALUES-fed path endpoints yield a composition-level public-row check + a `PathReach` obligation, both per the record; subquery same-name shadowing is renamed apart into the reserved `!` namespace across patterns, path endpoints, filters AND values (`wave1_subquery_renames_filters_and_paths`), with parse-twice determinism pinned. Note: subqueries are record phase-3 but are explicitly in the wave-1 bead scope, and the record's monotonicity argument is exactly what the code implements — no soundness gap.
3. **Bounded-path honesty** — `PathClosure` pins `min_len`/`fixed_k` (`p?` -> k=1) and normatively documents record §4 reqs 1–3 (k is a *manifest* public input the composition verifier must surface; existence-only, never absence/completeness) + req 5 (zero-length occurrence witness). Surfacing lands in sq-3kd2g.6 — honest here because **no production caller consumes `fragment_query`** (confirmed: diff touches only `verify.rs` + SKILL.md; no references elsewhere in the workspace).
4. **No weakening of existing rejections** — stage-1 entry points byte-unchanged (diff adds only new items + imports + module doc), and `wave1_stage1_gate_is_unchanged` pins `fragment_patterns` still rejecting paths/UNION/VALUES.
5. **Gates** — full CI green including `privacy-claims`, `clippy (gate)` (rustdoc `--all-features` half), feature-matrix tiers, coverage ratchet; nothing feature-gated in the diff. Copilot review: zero threads. SKILL.md entry carries the sq-qhy4 / not-externally-audited caveat.

**No arm from me** — per the maintainer-arm rule for zk-soundness surfaces (and the PR's own do-not-auto-merge banner), the arm decision is the maintainer's. No defects found; nothing requested of the fleet.

jeswr:
> 🤖 SPARQ agent arming verdict:
> 
> Escalated Opus review clean with disposition=arm, honest=true, sound_as_scoped=true, recommend_arm=true. PR armed via merge queue.

Issue #1582, opened 2026-07-05
[feature]: Implement ODRL evaluation using inference rules


jeswr:
### Problem / motivation

It should be possible to have very little of the ODRL evaluation described in Rust and instead have these described in Notation3 as inference rules -- similar to WAC and ACP evaluation (or at least it was when I checked last week, this may have changed).

I would like to see if this is possible to implement without a performance degredation, or increasing build sizes for sparq. To avoid increasing build sizes there may need to be some way of "compiling" the Notation3 at build time rather than parsing it at run time. However this may also create the opportunity for a net performance improvement if done very intelligently - i.e. representing the rules at an index layer; have URI's pre-indexed and ids attached to them. This could then provide a really good way of optimising the run time of WAC and ACP evaulation.

If this works; it may also be appropriate to then moving to represent RDFS and OWL-RL rules using Notation3 and then getting compiled into sparq at build time.

### Proposed solution

See above

jeswr:
> 🤖 **SPARQ agent** — decomposition plan for this feature request (epic `sq-zgbso`, design record PR #1590). [FABLE-5]

**Verified current state (your "may have changed" hedge):** unchanged — WAC and ACP evaluation ARE N3 inference rules today (`crates/sparq-solid/rules/{common,wac,acp-a,acp-b,acp-c}.n3`, run via `sparq_reason::reason_n3`; WAC 1 stratum, ACP 3). ODRL is hand-written Rust (`sparq-policy` + the `odrl_bridge`). Two load-bearing details support your index-layer idea: the rules are `include_str!`-embedded but **re-parsed at runtime on every materialize call**, and the pipeline **round-trips the facts through N3 text** (ACP re-serializes each stratum's closure) before running a **string-level** fixpoint — while the facts already sit as dictionary ids in the `Graph`. So the candidate net win for WAC/ACP is real and measurable: pre-compiled rules with pre-interned URI ids, evaluated over `[Id;3]` facts on the existing `sparq-substrate` join kernels (no third join implementation).

**Approach — measurement-first, spike-gated** (`research/odrl-n3-compiled-rules.md`, PR #1590):

[...]

PR #1621, opened 2026-07-06
feat(solid/odrl): stateless ODRL core as stratified N3 rules + CI-locked Rust-vs-N3 decision differential (sq-zgbso.2) [FABLE-5]

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — bead `sq-zgbso.2`, epic `sq-zgbso` / maintainer issue #1582 (Phase 2, gated on the sq-zgbso.1 spike **GO** verdict, PR #1601). Design record: `research/odrl-n3-compiled-rules.md`.
>
> **Authorization surface — MAINTAINER-ARM.** This PR changes how ODRL grants can be derived (design record §7: access-control surface ⇒ maintainer-reviewed, not fleet auto-armed). Do not auto-merge.

Implements the **stateless ODRL core as N3 rule strata** in `sparq-solid` — the exact WAC/ACP runtime pattern (`sparq_reason::reason_n3` over embedded rule text) — with a **CI-locked decision differential** against the Rust path.

## What's in

- **`rules/odrl-core-{a,b,c,d}.n3`** — four stratified rule files generalizing the spike's one-permission shape to the full stateless subset: Permission/Prohibition matching (action incl. the `odrl:use` umbrella vs the transfer subtree; exact-IRI target/assignee or unconstrained), atomic constraints, one compound level of `odrl:and`/`or`/`xone` with the evaluator's three-valued fail-closed semantics, duties, and deny-overrides.
- **`odrl_bridge::materialize_policy_n3` + `N3Request`** — the **OPT-IN** N3 entry point (explicit call path inside the already-default-OFF `odrl-bridge` feature). **The Rust path (`materialize_policy` over `sparq_policy::evaluate`) remains the default evaluator**; this bead proves parity at scale, it does NOT flip any default (later maintainer decision; build-time compilation is sq-zgbso.3/.4).
- **`tests/odrl_n3_differential.rs`** — **236 parity scenarios + 6 loud-error cases**, red on any divergence (runs on `--features odrl-bridge`; NOT `--all-features` — pre-existing sq-s5tkx is out of scope).

## Spike findings addressed (not inherited)

- **(a) dateTime — RESOLVED with a normalizing comparison.** The spike compared `odrl:dateTime` lexically (`string:notGreaterThan`), instant-correct only for canonical UTC. The production strata normalize BOTH operands to epoch seconds with the engine's existing offset-aware **`time:inSeconds`** builtin and compare numerically — mixed `±hh:mm` offsets now order by instant, exactly like `sparq_policy::parse_instant` (differential includes offset-equal-instant cases pinned on `lt`/`lteq`/`eq`). Residual normalizer differences (fractional seconds: Rust keeps ns, `time:inSeconds` truncates; leap-second `:60`; negative/≥5-digit years) are **excluded FAIL-CLOSED**: the wrapper refuses such lexicals with a loud `Err` and materializes nothing — no silent divergence window. So: **both** remedies the brief offered — a normalizing builtin for the offset dimension, plus a fail-closed lexical-space guard for the precision dimension.
- **(c) stratification.** Constraint-bearing prohibitions run under a full A→B→C→D stratification (atomic statuses → `or`/`xone` completion → matching → decisions), so every `log:notIncludes` negates only stratum-complete predicates (design §3.5 / record §7). The differential's prohibition-window matrix covers active/lapsed/unprovable windows against a co-resident permission.
- **(b) stateful `odrl:count` stays Rust** — a usage-counter mutation is not an inference; numeric bounds/evidence are refused loudly by the N3 path.

## Differential shape (the honesty mechanics)

- Both paths consume the SAME parsed `sparq_policy::Policy` (evaluation is differential; parsing is shared — the record's scope) and the same request via `N3Request::to_request()`.
- Operand lexicalization is shared (`Value::as_str()`/the `is_part_of` member split), so value semantics have one source of truth; the DECISION logic (matching, three-valued combinators, deny-overrides, duty gating) is entirely in the rules.
- Spike scenarios keep **independently-stated expected sets** (a both-wrong agreement cannot pass); 20+ generated edges are likewise pinned.
- `N3Request` **by construction** carries no collection-membership/subsumption evidence — with that evidence absent the Rust matcher IS the exact-IRI base case the rules implement (a taxonomy-evidence caller must use the Rust path; refusing at the type level closes the fail-open hole a matching-but-unproven prohibition would otherwise open).
- Conflict-strategy admissibility reuses `sparq_policy::conflict_admissibility` on both paths (same refusal outcomes, differentially asserted for `perm`/`invalid`/unknown).

## Gates

- `cargo test -p sparq-solid` (default) and `--features odrl-bridge` — green (incl. the 7 spike tests unchanged + the new 13-test differential + 8 direct unit tests for the new public items).
- `cargo clippy -p sparq-solid --all-targets -- -D warnings` in both feature states — clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc` in both feature states — clean.
- `scripts/check-readme-template.py` — 0 deviations.
- Coverage ratchet: sparq-solid is measured on default features (no case arm), where this module is compiled out — no floor risk.

## Not in this PR (follow-ups)

- ~~`skills/access-control/SKILL.md` deferral~~ — the required `public-api-skill (G2)` gate enforces the SKILL edit in the SAME PR, so it is now INCLUDED (second commit), placed between #1612's SKILL.md hunks so both merge cleanly; the deferral bead was closed as superseded.
- Runtime-perf/build-size measurement — sq-zgbso.3/.4 per the record's split verdict (the spike measured the like-for-like runtime ratio at ~1.1×, same reasoner class WAC/ACP already pay).

Closes bead `sq-zgbso.2`. Refs #1582.

### DIFF
--- a/crates/sparq-solid/rules/odrl-core-a.n3
+++ b/crates/sparq-solid/rules/odrl-core-a.n3
@@ -1,0 +1,176 @@
+# [FABLE-5] sq-zgbso.2 — ODRL stateless core as N3, STRATUM A (of A→B→C→D).
+#
+# Input facts are the `oc:` fact schema `odrl_bridge::materialize_policy_n3` serializes
+# from the PARSED `sparq_policy::Policy` + `N3Request` (never raw policy text — the two
+# paths must consume identical semantic content). This stratum derives everything whose
+# negation ranges ONLY over input facts (design record research/odrl-n3-compiled-rules.md;
+# stratification discipline: research/solid-access-control-design.md §3.5 — a negated
+# predicate must be COMPLETE before its stratum runs; input predicates are complete by
+# construction, so `log:notIncludes` here scopes only over `oc:evLeft` / `oc:member` /
+# `oc:discharged` / `oc:TransferAction`, none of which any rule derives).
+#
+# Outputs (inputs to stratum B/C/D):
+#   ?rule oc:actionOK / oc:targetOK / oc:assigneeOK true   — structural match legs
+#   ?c    oc:sat true          — atomic constraint SATISFIED (evidence exists + holds)
+#   ?c    oc:unprovable true   — no evidence for the constraint's dimension
+#   ?c    oc:notSat true       — NOT satisfied (definitely-failed OR unprovable):
+#                                 mirrors `constraint_satisfied == false` (fail-closed)
+#   ?lc   oc:lcNotSat true     — the compound-combinator pieces derivable here
+#   ?rule oc:blocked true      — some DIRECT atomic constraint of the rule is notSat
+#   ?rule oc:dutyBlocked true  — some duty action is undischarged
+#
+# dateTime (spike finding (a) RESOLVED): comparisons normalize BOTH operands to epoch
+# seconds via `time:inSeconds` — the engine's offset-aware civil-time extraction — and
+# compare with `math:*`, so arbitrary `±hh:mm` offsets order by the INSTANT they denote,
+# exactly like `sparq_policy`'s `parse_instant`. The wrapper additionally fail-closes
+# the accepted lexical space to second-precision forms (no fractional seconds, no leap
+# second, no negative year) where the two normalizers provably agree.
+
+@prefix odrl: <http://www.w3.org/ns/odrl/2/> .
+@prefix oc:   <https://sparq.dev/ns/odrl-core#> .
+@prefix log:  <http://www.w3.org/2000/10/swap/log#> .
+@prefix math: <http://www.w3.org/2000/10/swap/math#> .
+@prefix time: <http://www.w3.org/2000/10/swap/time#> .
+
+# ── ODRL 2.2 ownership-`transfer` subtree: the ONE top-level branch outside `use`
+#    (mirror of sparq_policy::model::is_transfer_subtree; `use` never permits these).
+odrl:transfer a oc:TransferAction .
+odrl:sell     a oc:TransferAction .
+odrl:give     a oc:TransferAction .
+
+# ── action: the rule's action permits the requested action ────────────────────────────
+# exact-IRI match (Action::permits, first arm)
+{ ?rule a oc:Rule ; oc:action ?a . ?req a oc:Request ; oc:reqAction ?a . }
+=> { ?rule oc:actionOK true . } .
+# the `odrl:use` umbrella permits any requested action NOT in the transfer subtree
+# (Action::permits, second arm — sq-euhr3). Static-set negation: input-only.
+{ ?rule a oc:Rule ; oc:action odrl:use . ?req a oc:Request ; oc:reqAction ?ra .
+  ?UNSCOPED log:notIncludes { ?ra a oc:TransferAction . } . }
+=> { ?rule oc:actionOK true . } .
+
+# ── target / assignee: exact IRI agreement, or the rule leaves the slot open ──────────
+# (rule.target == None → `oc:anyTarget` marker emitted by the serializer — no negation
+# needed; collection membership is OUTSIDE the N3 subset by construction: `N3Request`
+# cannot carry membership evidence, so exact match IS the Rust base case.)
+{ ?rule a oc:Rule ; oc:target ?t . ?req a oc:Request ; oc:reqTarget ?t . }
+=> { ?rule oc:targetOK true . } .
+{ ?rule a oc:Rule ; oc:anyTarget true . }
+=> { ?rule oc:targetOK true . } .
+{ ?rule a oc:Rule ; oc:assignee ?p . ?req a oc:Request ; oc:reqParty ?p . }
+=> { ?rule oc:assigneeOK true . } .
+{ ?rule a oc:Rule ; oc:anyAssignee true . }
+=> { ?rule oc:assigneeOK true . } .
+
+# ── atomic constraints, NON-dateTime dimensions ───────────────────────────────────────
+# Operands are pre-lexicalized by the serializer (`oc:boundLex` / `oc:evLex` carry the
+# SAME `Value::as_str()` string the Rust evaluator compares — one source of value truth),
+# so eq/neq here is exactly `value_eq` and isPartOf exactly `is_part_of` (the serializer
+# splits the set members). Only {eq, isA, neq, isPartOf} exist for non-dateTime
+# dimensions inside the wrapper-guarded subset (order operators there are REFUSED).
+#
+# eq / isA — satisfied iff the evidence lexical equals the bound lexical.
+{ ?c a oc:Atomic ; oc:op oc:eq ; oc:left ?l ; oc:boundLex ?b .
+  ?ev oc:evLeft ?l ; oc:evLex ?v . ?v log:equalTo ?b . }
+=> { ?c oc:sat true . } .
+{ ?c a oc:Atomic ; oc:op oc:eq ; oc:left ?l ; oc:boundLex ?b .
+  ?ev oc:evLeft ?l ; oc:evLex ?v . ?v log:notEqualTo ?b . }
+=> { ?c oc:notSat true . } .
+# neq — the mirror image.
+{ ?c a oc:Atomic ; oc:op oc:neq ; oc:left ?l ; oc:boundLex ?b .
+  ?ev oc:evLeft ?l ; oc:evLex ?v . ?v log:notEqualTo ?b . }
+=> { ?c oc:sat true . } .
+{ ?c a oc:Atomic ; oc:op oc:neq ; oc:left ?l ; oc:boundLex ?b .
+  ?ev oc:evLeft ?l ; oc:evLex ?v . ?v log:equalTo ?b . }
+=> { ?c oc:notSat true . } .
+# isPartOf — the evidence lexical is a member of the pre-split right-operand set.
+# (The negation ranges over `oc:member`, an input-only predicate — safe here. An empty
+# member set — e.g. an empty-string bound — has no member facts, so with evidence
+# present the notSat rule fires, mirroring `any()` over an empty set.)
+{ ?c a oc:Atomic ; oc:op oc:isPartOf ; oc:left ?l .
+  ?ev oc:evLeft ?l ; oc:evLex ?v . ?c oc:member ?v . }
+=> { ?c oc:sat true . } .
+{ ?c a oc:Atomic ; oc:op oc:isPartOf ; oc:left ?l .
+  ?ev oc:evLeft ?l ; oc:evLex ?v .
+  ?UNSCOPED log:notIncludes { ?c oc:member ?v . } . }
+=> { ?c oc:notSat true . } .
+
+# ── atomic constraints, the `odrl:dateTime` dimension ─────────────────────────────────
+# `oc:boundDt` is emitted ONLY for dateTime-dimension constraints; the evidence is the
+# request's `oc:atTime` (the `Request::at` evaluation-time evidence). Both normalize to
+# epoch seconds (offset-aware) before the numeric comparison — instant order, not
+# lexical order.
+{ ?c a oc:Atomic ; oc:op oc:lteq ; oc:boundDt ?b . ?req a oc:Request ; oc:atTime ?t .
+  ?t time:inSeconds ?ts . ?b time:inSeconds ?bs . ?ts math:notGreaterThan ?bs . }
+=> { ?c oc:sat true . } .
+{ ?c a oc:Atomic ; oc:op oc:lteq ; oc:boundDt ?b . ?req a oc:Request ; oc:atTime ?t .
+  ?t time:inSeconds ?ts . ?b time:inSeconds ?bs . ?ts math:greaterThan ?bs . }
+=> { ?c oc:notSat true . } .
+{ ?c a oc:Atomic ; oc:op oc:lt ; oc:boundDt ?b . ?req a oc:Request ; oc:atTime ?t .
+  ?t time:inSeconds ?ts . ?b time:inSeconds ?bs . ?ts math:lessThan ?bs . }
+=> { ?c oc:sat true . } .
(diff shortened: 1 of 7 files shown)


jeswr:
> 🤖 **SPARQ agent** (escalated-tier reviewer) — refute-first review verdict for the maintainer's arm decision (this is a maintainer-arm authorization surface per the design record; NOT auto-armed).

**Verdict: honest, sound as scoped — recommend maintainer arm.** `state: OPEN`, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` (no rebase needed; the SKILL.md hunk merges cleanly around #1612's).

What was verified (diff-scoped + targeted dynamic checks in an isolated clone; the shared tree was not touched):

1. **Opt-in claim** — no caller of `materialize_policy_n3` exists anywhere in the workspace outside its own module/tests; the Rust path (`materialize_policy` over `sparq_policy::evaluate`) remains the only routed evaluator. Verified by grep across `crates/`.
2. **Stratification soundness** — checked every `log:notIncludes` in `odrl-core-{a,b,c,d}.n3` against where its negated predicate is derived: A negates only input-schema predicates (`oc:TransferAction` is static data, `oc:member`/`oc:evLeft`/`oc:discharged` serializer-only); `oc:sat` derives ONLY in A (B's or/xone negation is complete); `oc:blocked` completes after B (C's negation sound); `oc:matches` derives only in C (D's deny-overrides negation sound); D derives only `auth:*` triples. Could not construct a premature-negation case. **Mutation check**: collapsing stratum B into A goes red exactly as the theory predicts (`or SS: granted divergence`) — the engine's NAF-without-retraction makes the strata load-bearing AND the differential detects mis-stratification.
3. **Fail-open hunt** — `N3Request` verifiably cannot carry membership/subsumption evidence (no such fields; `to_request()` never populates `party_memberships`/`asset_memberships`/`purpose_subsumes`), so `rule_matches`/`compare_constraint` reduce to the exact-IRI/flat base case the rules implement. Absence of evidence → `oc:unprovable` → `oc:notSat` → blocked (grants), and an unproven prohibition never blocks — matching `evaluate` exactly. Out-of-subset constructs on permissions AND prohibitions route through the same `n3_policy_admissible` chained iterator → loud `Err`, nothing materialized. The parse-level `urn:sparq-policy:malformed` guard (`Neq`/IRI) is admissible and lands unprovable on both paths — no starved deny-overrides window found.
4. **dateTime** — `datetime_part(InSeconds)` (sparq-reason) is total and instant-agreeing with `parse_instant` (sparq-policy) on the guarded lexical space (bare date = midnight UTC, missing tz = UTC, ±hh:mm shift; both Hinnant civil-day). The guard `n3_datetime_admissible` excludes exactly the residual divergence dimensions (fractional seconds, `:60`, negative/≥5-digit years). **Mutation check**: admitting `:60` in the guard goes red in the unit suite. Offset-equal-instant pinned cases (`lteq` grants / `lt` denies / `eq` grants at `+02:00` = bound instant) pass on both paths.
5. **Differential non-vacuity** — hand-counted **exactly 236** parity scenarios + 6 loud-error cases (the PR body's numbers are accurate); spike cases carry independently-restated expected sets. The CI `opt-in sparq-solid (odrl-bridge)` leg genuinely RAN all 13 differential tests (log inspected), and they pass locally on the PR head. **Mutation checks**: flipping the `proh window active` pinned expectation → red; deleting stratum D's deny-overrides negation → red (`C1 deny-overrides: granted divergence`). The corpus includes prohibition-side dateTime windows (finding c) and the xone edges (two-sat, unprovable-sibling) as claimed.
6. **SKILL bullet (G2)** — accurate: opt-in, Rust-default, instant-normalized dateTime, fail-closed subset, no membership evidence by construction.

Non-blocking nits (no action required to arm):
- Follow-up bead numbering is inconsistent across prose: PR body says build-time compilation is "sq-zgbso.3/.4", the module comment "sq-zgbso.5, gated on .3/.4", the SKILL ".4/.5". Worth one-line reconciling in a later docs pass.
- The out-of-subset loud-error tests pin permission-side constraints only; the prohibition side shares the identical `n3_policy_admissible` code path (statically verified) but an explicit prohibition-side `Err` case would be a cheap future hardening.
- The module note "the reserved `urn:sparq:` prefix is rejected in user-supplied principals" refers to the store-layer convention, not a guard in this entry point; collision is harmless here (both paths see the same IRIs, so parity — the invariant — is unaffected).

jeswr:

**Head SHA reviewed:** `8ce5c5349efd89c83814817f8191284eb3698e06`
**Scope read:** the full diff; `rules/odrl-core-{a,d}.n3`; `src/odrl_bridge.rs` (admissibility + serializer + `materialize_policy_n3`); `tests/odrl_n3_differential.rs`; and — because the invariant is a cross-crate claim — `sparq-policy/src/eval.rs` (`parse_instant`/`days_in_month`/`compare`) and `sparq-reason/src/n3/mod.rs` (`time:inSeconds`).

## BLOCKING — a reproducible FAIL-OPEN divergence: the N3 path GRANTS where the Rust path DENIES

The load-bearing honesty claim of this PR is stated three times:

- `src/odrl_bridge.rs:1774` — *"Anything else is a loud `Err`, never a silent divergence window."*
- `src/odrl_bridge.rs:1917` — the accepted lexical space is where *"the two normalizers **provably agree**"*.
- `rules/odrl-core-a.n3:27` — same claim.

**It is false for impossible calendar dates.** `n3_datetime_admissible` validates the day field by *range only*:

```rust
// src/odrl_bridge.rs:1934
if !(1..=12).contains(&num2(&b[5..7])) || !(1..=31).contains(&num2(&b[8..10])) {
```

`31` is not `days_in_month(year, month)`. The two normalizers diverge exactly there:

- **Rust** — `sparq_policy::parse_instant` → `days_from_civil` → `days_in_month` (`eval.rs:1754-1779`) **rejects** `2026-02-29`; the crate's own test asserts it (`eval.rs:1867`: `"2026-02-29T00:00:00Z", // 2026 is not a leap year`). `order()` then returns `None`, and `compare()` returns **`false`** (`eval.rs:1431-1435`) — constraint unsatisfied, fail-closed, **no grant**.
- **N3** — `sparq_reason`'s `days_from_civil` (`n3/mod.rs:3074-3082`) is the raw Hinnant formula with **no validation**, so `time:inSeconds` happily **binds** `2026-02-29` as the epoch of `2026-03-01`. The stratum-A rule at `rules/odrl-core-a.n3:102` then fires `oc:sat`, the rule matches, and stratum D emits the allow triple — **grant**.

I ran it rather than argued it. Probe against the two crates as they are on `main` (this PR does not modify either), using the exact literal forms `n3_serialize` emits (`oc:boundDt`/`oc:atTime` as `"…"^^xsd:dateTime`) and the exact `lteq` rule body from `rules/odrl-core-a.n3:102`:

```text
N3  oc:sat derived for lteq bound 2026-02-29 : true
Rust cmp_datetime(2026-01-01T00:00:00Z, 2026-02-29T00:00:00Z) = None
Rust cmp_datetime(2026-01-01T00:00:00Z, 2026-03-01T00:00:00Z) = Some(Less)
```

Concrete counterexample scenario (both halves reachable — `parse.rs:933` stores `Value::DateTime` verbatim with **no** lexical validation, and `N3Request::at()` takes any string):

```turtle
<urn:pol/1> a odrl:Set ; odrl:permission [
  odrl:action odrl:read ; odrl:target <urn:t/1> ; odrl:assignee <urn:alice> ;
  odrl:constraint [ odrl:leftOperand odrl:dateTime ; odrl:operator odrl:lteq ;
                    odrl:rightOperand "2026-02-29T00:00:00Z"^^xsd:dateTime ] ] .
```
request: `read <urn:t/1> by <urn:alice> at "2026-01-01T00:00:00Z"` →
**Rust: no grant. N3: `<urn:alice> auth:read <urn:t/1>`.** And `materialize_policy_n3` returns `Ok` — the promised loud `Err` never fires, so this is precisely the "silent divergence window" the module doc says cannot exist. The symmetric request-side form (valid bound, `at = "2026-02-29T00:00:00Z"`) diverges the same way and does not even depend on how the policy parser behaves.

Why the corpus did not catch it — **quantifier direction**. 236 hand-built + generated scenarios is an *existential* sample being used to carry a *universal* claim ("provably agree"). The guard's own unit tests show the blind spot precisely: `odrl_bridge.rs:2457` accepts `"2026-02-28"` (deliberately a valid date) and `odrl_bridge.rs:2470` rejects month 13, hour 24, offset +15:00, fractional seconds, leap second, negative/5-digit year — **every dimension except day-of-month-vs-month**. `tests/odrl_n3_differential.rs:724` (`out_of_subset_is_a_loud_error_not_a_divergence`) inherits the same blind spot.

**Fix (all three parts needed):**
1. `n3_datetime_admissible` must validate day-of-month against the Gregorian month length — or, far better, stop maintaining a second parser: gate admissibility on `sparq_policy::parse_instant`/`cmp_datetime` returning `Some`, so there is literally one source of truth and this class cannot recur.
2. Add `2026-02-29`, `2026-04-31`, `2026-02-30` to `datetime_admissible_rejects_divergence_windows` **and** a differential scenario that would have gone red.
3. Since this is a decision-equivalence claim over an unbounded lexical space, add a randomized/property differential over generated dateTime lexicals (both paths, mutation-witnessed) — a fixed corpus structurally cannot support "provably agree".

## What I checked and found sound (so this is a targeted fail, not a general one)

- **The differential is genuine, not a both-wrong tautology.** `assert_case` pins an independently-stated expected set on top of parity, so agreement alone cannot pass. `spike_scenarios_with_pinned_expectations` case `C1 deny-overrides` expects **exactly** `[(ALICE, "denyRead", T1)]` — if the `log:notIncludes { ?x a oc:Prohibition ; oc:matches true . }` negation in `rules/odrl-core-d.n3:308` were vacuous, the allow triple would also appear and that assertion goes red. The deny-wins negation is live.
- **"CI-locked" is a true claim.** `opt-in sparq-solid (odrl-bridge)` is a real gating leg (`.github/feature-matrix.d/sparq-solid.yml:17-20`, `test: true`) and appears in `scripts/tests/feature-matrix-legnames.golden.txt:135`, so `#![cfg(feature = "odrl-bridge")]` on the differential does run in CI. No false-green at the YAML seam.
- **Opt-in / core-lean respected**; the Rust path stays the default evaluator; `N3Request` closing the taxonomy-evidence hole at the type level is a genuinely good design move.
- **Stratification A→B→C→D** is argued correctly (each negated predicate complete in the seed it receives) and `refuse_unimplementable_conflict` runs first.

## Secondary (non-blocking, but fix while you are in here)

- `materialize_policy_n3` derives `granted` and `prohibited` **independently** (`grant_triple.is_some()` / `deny_triple.is_some()`). Today deny-overrides in stratum D makes both-true unreachable, but the extraction itself does not enforce it — a future rule edit that breaks the negation yields `granted = true` alongside `prohibited = true`, and a caller reading `.granted` allows. Cheap hardening: assert/force `granted = false` when `prohibited`.
- `n3_iri_ok` permits `%`-unescaped and non-ASCII characters that `NamedNode::new_unchecked` will not validate on the way back out (`Term::NamedNode(NamedNode::new_unchecked(s))`). In-subset today because the serializer only round-trips IRIs it already emitted, but `new_unchecked` on a path fed by policy text is worth a second look.

## DRIVE vs CLOSE — **DRIVE, do not close**

The architecture, the stratification, the differential harness and the type-level evidence restriction are all worth keeping; 19 days of age is not the problem here. But it must **not** merge in this state: it is an authorization surface whose headline invariant is falsifiable in one probe, and the failure direction is *grant*. Land the day-of-month fix (ideally by delegating to `sparq_policy`'s parser), extend the reject list + differential, and re-review. If the maintainer would rather not carry the second parser at all, that is also a legitimate re-scope of this bead.

VERDICT: fail

jeswr:
> 🤖 SPARQ agent

**Closing as SUPERSEDED.** This PR is `CONFLICTING` and cannot merge; its substance landed from another branch.

### Superseded by

**`f51be1f47`** — *"feat(sparq-solid): Implement the stateless ODRL core as N3 rule strata (runtime `reason_…`) (sq-zgbso.2)"* — **PR #3458**.

Same bead: **sq-zgbso.2**, the same one in this PR's title.

### Evidence, not age

| this PR adds | on `main` today |
|---|---|
| `rules/odrl-core-{a,b,c,d}.n3` | `rules/odrl-{a,b,c,d}.n3` — plus `odrl-a0.n3` |
| `tests/odrl_n3_differential.rs` (795 lines) | `tests/odrl_n3_differential.rs` (**1052 lines**), headed *"sq-zgbso.2 — Rust-vs-N3 decision differential for the stateless ODRL core"*, with `rust_set()` / `n3_set()` / `assert_case()` and the `a*`/`b*` case battery |
| `odrl_bridge.rs` +889 | `odrl_bridge.rs` is 2819 lines on `main` |
| `skills/access-control/SKILL.md` +16 | documents `materialize_odrl_n3(...)` running the stateless ODRL core as **five stratified `reason_n3` calls** |

So the CI-locked Rust-vs-N3 decision differential this PR exists to add is not merely present on `main` — it is
a **superset** of the version here. `main` has also since gained `odrl_n3_failopen_regression.rs` and
`odrl_count_bridge.rs`, and the ODRL surface moved further under #4016 (`cdecad0e5`, explicit ODRL action
contract for SPARQL query).

That is also why the merge conflicts: `crates/sparq-solid/tests/odrl_n3_differential.rs` is an **add/add**
conflict — both sides create the same file — and `odrl_bridge.rs` / `skills/access-control/SKILL.md` conflict
on prose `main` has since rewritten. Resolving would mean choosing between this 795-line differential and
`main`'s 1052-line one, which is not a merge.

Closed on the citation above. Branch `feat/sq-zgbso.2-odrl-n3-strata` stays reachable at `8ce5c534` if any
case in the older battery turns out not to have a counterpart on `main`. Reopen if you disagree.

PR #1651, opened 2026-07-06
fix(reason/rif): RIF-Core Equal-atom fidelity audit (sq-pbz04.5.4)

jeswr:
> 🤖 SPARQ agent — bead sq-pbz04.5.4 (P2 BUG, sparq-reason/rif.rs)

## Summary

RIF-Core Equal-atom fidelity audit: three correctness fixes with non-vacuous tests, all confined to `crates/sparq-reason/src/rif.rs` and its conformance suite. **Disjoint from in-flight #1644 (sparq-reason-dl) and #1647 (sparq-reason-ql).**

**What was wrong:** The previous `Equal`-atom handling had two defects documented in `research/rif-core-conformance-and-builtins.md` §3.3:
1. `Equal` in a rule HEAD was silently accepted and lowered to `owl:sameAs` — this both over-accepts non-Core input (RIF-Core forbids equality in conclusions) and under-delivers RIF-BLD semantics (the chainer treats sameAs as an inert data triple, not an equality rule).
2. Body `Equal` lowering to `owl:sameAs` pattern-match was wrong in two ways: `t = t` was not derivable unless someone explicitly asserted a sameAs triple, and any asserted `owl:sameAs` vocabulary triple would satisfy a RIF body equality atom (unlicensed conflation).

**What changed (original 3 code fixes):**

- **`RifError::EqualInConclusion`** (new) — `validate()` rejects any rule whose head contains an `Equal` atom. Non-vacuous test: fact-head case and rule-head case both trigger it; `closure()` refuses.
- **Ground-identity elimination** — `lower_body_atom` now returns `Option<String>` and returns `None` for body `Equal` atoms whose sides are structurally identical (e.g. `?n = ?n`). The rule fires correctly as if the atom were absent.
- **`RifError::DistinctGroundEqual`** (new) — `validate()` rejects body `Equal` atoms with two distinct non-variable terms fail-closed. Value-space equality (`"1"^^xsd:integer = "1.0"^^xsd:decimal`) requires sq-v5evr (#1646); error message names the deferral bead.

**Honesty fix (Opus review, sq-pbz04.5.4 follow-up):**

The Opus review found a honesty defect: the rif.rs module doc and SKILL.md claimed variable/mixed Equal atoms (`?x=?y`, `?x=<iri>`) "work at runtime ... lowered to an `owl:sameAs`-pattern triple ... holds when both variables bind to the same RDF node." This is FALSE — it is the known-unsound `owl:sameAs`-vocabulary-triple conflation (V1: over-derivation when sameAs is asserted for different nodes; V2: under-derivation when two vars bind the same node without an explicit assertion). Tracked as `sq-26vwp`.

- Rewrote rif.rs module doc: item 2 clarifies that elided Equal atoms contribute NO bindings; new section documents V1/V2 defect honestly.
- Rewrote SKILL.md: removed false claim; added item 4 documenting the `sq-26vwp` unsoundness, honest posture matching the value-equality deferral.
- Fixed `validate()` sole-binder bug: a `?n=?n` body atom that is the SOLE binder of a head variable was accepted by validate (fell to the `positive =>` arm), but lowering eliminates it → unbound head var in the N3 rule, violating the `UnboundHeadVar` contract. Fix: new `left == right` arm does nothing (no binding), so the `UnboundHeadVar` sweep correctly rejects. New test `body_equal_identity_sole_binder_rejects_unbound_head_var` pins the fix.

## Gates run (both feature states)

| Gate | Feature OFF | Feature ON (`rif-core`) |
|---|---|---|
| `cargo build` | PASS | PASS |
| `cargo clippy --all-targets -D warnings` | PASS | PASS |
| `cargo test -p sparq-reason` (155 pass) | PASS | PASS |
| `cargo test -p sparq-conformance` (rif_core_suite) | PASS | PASS |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` | PASS | PASS |
| `check-readme-template.py --enforce` | 0 deviations | — |

## Honest scope and deferred remainder

The value-equality half (sq-v5evr) is explicitly fail-closed with `DistinctGroundEqual`. Variable/mixed Equal atoms (`?x=?y`, `?x=<iri>`) are documented as known-unsound (sq-26vwp) but not yet rejected — doing so would break callers relying on the existing behavior where `owl:sameAs` is intentionally asserted. That behavioral fix is deferred to sq-26vwp.

References: sq-pbz04.5.4, sq-26vwp, sq-6tykl, sq-v5evr (#1646)


jeswr:
> 🤖 SPARQ agent — escalated (Opus-4.8) soundness/honesty review of #1651

**Verdict: request changes** — one honesty defect on the variable-equality doc claim; the three implemented fixes are otherwise sound.

**Sound and confirmed correct:**
- **EqualInConclusion** — complete + fail-closed. Rejects `Equal` anywhere in a fact head, a rule head, and a *conjunction* head (2nd+ position). Only lowering entry points are `to_n3_source`/`closure`, both of which call `validate()?` first — no bypass.
- **Ground-identity elimination** — sound. `left == right` is exact structural `Eq`; identical terms are elided, and a genuinely non-identical Equal is never accidentally elided.
- **DistinctGroundEqual fail-closed** — genuinely fail-closed. Verified rejection of `"1"^^integer = "1.0"^^decimal`, same-lex/diff-dt (`"1"^^integer = "1"^^decimal`), distinct IRIs, and distinct lists. Error names sq-v5evr/#1646.
- **Floor 58→61 honest** — the removed "equality lowers to owl:sameAs" test was asserting a fact-with-Equal-head that the fix now correctly *rejects*; its removal is a fidelity fix, not a ratchet weakening. Suite reports 63 ≥ floor 61; scoreboard sync guard passes.
- Gates pass locally in both feature states (build/clippy/test/floors); CI shows zero failing legs.

**Concern (blocks arming as-is): the variable-equality doc claim is false.** `rif.rs` module doc ("Variable equality works at runtime … lowered to an owl:sameAs-pattern triple … which holds when both variables are bound to the same RDF node in the closure") and the matching `skills/inference/SKILL.md` line affirmatively claim the variable case is correct. Adversarial probes on this branch show it is **not** — it is the exact defect design-record §3.3 (defect 2) flags:
- **Over-derivation (conflation):** with data `p owl:sameAs q` (distinct p, q), a body `?x = ?y` fires `p ex:mergedWith q` — an asserted `owl:sameAs` *vocabulary* triple licenses a *builtin* equality.
- **Under-derivation:** two variables that both bind to the **same** node do **not** satisfy `?x = ?y` without an asserted sameAs triple (RIF: `t = t` is always true).

The variable-equality `positive_to_n3` lowering is unchanged from `main`, so this PR does not *regress* — but it newly writes documentation asserting the residual unsound case "works." §3.3's fix-direction-2 (variable equality via compile-time substitution) is genuinely out of this bead's scope; the ask is only to **stop claiming it works**: document `?x = t` / `?x = ?y` as a KNOWN residual defect/deferral (mirroring how value-equality is handled), or implement the substitution. Tracked as **sq-26vwp**.

Minor (non-blocking): a body `?n = ?n` that is the *sole* binder of a head var passes `validate()` (the arm treats it as binding) but is elided at lowering, yielding an N3 rule with an unbound head var — a validate/lower inconsistency vs the stated `UnboundHeadVar` contract. Not output-unsound (the chainer declines to fire it; closure = 0 triples), but worth aligning validate with the elision.

jeswr:
🤖 **SPARQ agent**: Escalated Opus review clean (honest=true, recommend_arm=true). Arming via merge queue.

PR #1673, opened 2026-07-06
feat(zk/compose): bind disclosed-solution terms in the extended fragment — PathReach pred/src/dst + VALUES cells (sq-1zf94) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent (Opus 4.8, 1M context). Re-review when Fable returns.

## What & why (sq-1zf94)

Merged #1665 routed a wave-1 extended-fragment presentation (property-path / `UNION` / `VALUES` / subquery) through `verify_fragment_manifest` end-to-end, but bound **no disclosed terms**: an accepted proof's sub-proofs were cryptographically valid statements of the *right circuit member* yet **not tied to the specific disclosed path endpoints / VALUES cells** — the documented gap deferred to this bead.

This PR adds the OFF-by-default (`extended-fragment`) `verifier::bind_fragment_solution` — the composition analogue of the flat `bind_query_correctness` scan-const check + the `bind_joins` slot binding.

## Design

- `manifest::BranchWitness` gains `solution: Vec<SolutionBinding>` (disclosed `var → DisclosedTerm`, IRI/literal). `DisclosedTerm::to_term()` rebuilds the `oxrdf::Term` verifier-side (fail-closed on a bad IRI/datatype/language, or language + non-`langString` datatype).
- The gate **re-encodes each disclosed term itself** (`encode_term`, salt-independent for IRIs/literals — never a prover-supplied encoding) and re-derives the branch structure from the query **text** alone (`fragment_query`), then demands, fail-closed (`FragmentSolutionError` → `CheckError::FragmentSolution`):
  - each bound `PathReach` `pred_enc` = the query-text predicate encoding (a chain proved over the **wrong predicate** is refused);
  - each `src_enc`/`dst_enc` = the query-**constant** endpoint's encoding, or the disclosed solution's binding for a **projected** endpoint variable (a projected endpoint omitted from the solution is refused; a non-projected endpoint stays existential/hidden and is not term-bound);
  - each disclosed `VALUES` cell = the **projected** variable's disclosed binding (a `values_rows` index pointing at a row whose cell disagrees — the *wrong disclosed row* — is refused).
- `verify_fragment_manifest` runs it **first** (before the nonce is burnt or any `bb` subprocess starts), between `dispatch_fragment` and the crypto stage. Because `PathReach.pred_enc`/`src_enc`/`dst_enc` are byte-bound into the bb public inputs by `reconstruct_public_inputs` (audit #1), a solution that passes **this** structural gate **and** the crypto stage has its disclosed endpoints / VALUES terms genuinely tied to the proofs.
- `sparq-zk`: re-export `sparq_engine::zk::{PatternKey, SlotPattern}` from `sparq_zk::verify` so the downstream verifier can match on a re-derived path endpoint **without** a direct `sparq-engine` dependency (already transitively present — **no new workspace dependency**).

## Honest scope (privacy-claims)

This **shrinks** the #1665 unbound surface but does not eliminate it. Still **unbound** (documented on the fn / README / SKILL, enumerated not hidden): the BGP-scan-slot binding of a solution's variables to a scan's disclosed rows (the per-solution row-selection model), the flat cross-graph Q6 non-bnode obligation per branch, and an **existential** (non-projected) path endpoint's value (hidden by design). So an accepted extended proof is **more bound than #1665 but not yet FULLY bound**. The verifier stack is internally re-audited but **NOT externally audited** (sq-qhy4); **no soundness / privacy property is asserted**.

## Tests (non-vacuous, no nargo/bb — every reject fires before the bb loop)

- a `PathReach` `src_enc` / `pred_enc` / `dst_enc` that does not match the disclosed solution refuses;
- a `VALUES` cell mismatch and a **wrong disclosed row** refuse;
- a projected endpoint / VALUES var omitted from the solution refuses; a malformed disclosed term refuses;
- an **existential** (ASK) endpoint is not term-bound;
- a **consistent** solution reaches the SAME downstream attestation gate as the #1665 routing test (`UnattestedCommitment`);
- the flat `verify_manifest` still rejects the query at stage-1 (byte-identical, feature ON or OFF);
- `DisclosedTerm::to_term` builds every literal shape + fails closed.

## Gates run (both feature states)

- `cargo clippy --all-targets -- -D warnings` — GREEN default (OFF) **and** `--features extended-fragment` (ON); workspace clippy default GREEN.
- `cargo test` — GREEN default (OFF) **and** `--features extended-fragment` (ON): 152 lib + all integration tests.
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — clean (feature-gated doc-links to private items demoted to code spans).
- `scripts/check-readme-template.py --enforce` — 0 deviations (README stays a 30-line internal stub).

Feature flag: **`extended-fragment`** (OFF by default). Default builds are byte-identical; `sparq-core`/`sparq-engine` untouched.


jeswr:
🤖 **SPARQ agent verdict**

PR #1673 armed for merge via `--auto` (added to merge queue). Will merge when required gates pass.

- Merge method: merge queue (squash)
- Status: BLOCKED on gates (expected)
- Auto-merge: enabled

PR #1677, opened 2026-07-06
feat(reason/ql): branch-aware emitter — per-branch FILTER/VALUES for multi-branch UCQ (sq-sg542) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — implementing bead **sq-sg542** (QL branch-aware emitter).

## What & why

The multi-branch-modifier guard added by #1647 fail-closed any UCQ where a branch carried its own `FILTER`/`VALUES`. After #1671 (property-path desugaring), an alternation under a `FILTER` — `?x :p1|:p2 ?y FILTER(?x != :Bad)` — desugars into a multi-branch UCQ whose branches **each** carry the distributed `FILTER`, so this common shape (and hand-written `{ … FILTER } UNION { … }`) abstained instead of answering.

This makes the emitter **branch-aware**: each UCQ branch now emits **its own** `filter_exprs`/`values_blocks` over **its own** sub-union (`emit::ucq_to_pattern_per_branch`), and the branch sub-bodies are top-unioned. A branch's `FILTER` therefore constrains only that branch — never hoisted over the whole union (the old single-passthrough bug applied `branch[0]`'s filter to every branch **and** dropped later branches' filters). The `lib.rs` guard is removed in both `rewrite` and `rewrite_production`; the stale *"deferred to sq-pbz04.3.2"* comment is corrected to **sq-sg542**.

## Design — surgical, byte-identical for existing shapes

Dispatch is conditional (`ucq.branches.len() > 1 && any_branch_has_modifier`):

- **Flat path kept byte-identical** for every previously-accepted shape — a single-branch UCQ (the branch *is* the whole query, so its own modifiers are re-applied soundly) and a multi-branch UCQ with **no** per-branch modifier (cross-branch dedup/minimisation preserved). No answer set or report count changes there.
- **New branch-aware path** only for the previously-**rejected** multi-branch-modifier case. In the production path, minimisation is **per-branch** (cross-branch minimisation is unsound once branches carry distinct modifiers) and stays fail-closed.

## Soundness invariants (differential-tested, `tests/branch_aware_emit.rs`)

- **No cross-branch leak — probed in BOTH directions:** a branch-0 filter never deletes an answer branch 1 legitimately contributes, and the symmetric branch-1 case.
- **Modifier applied, not dropped:** an answer excluded by a branch's own filter never reappears through that branch.
- **Result-equivalence oracle:** `?x :p1|:p2 ?y FILTER(…)` rewrites to the same projected answer set as its hand-desugared `{ ?x :p1 ?y FILTER } UNION { ?x :p2 ?y FILTER }` form, and both equal a hand-derived oracle (faithful BGP/Union/Filter/Join/Values evaluator).
- **`rewrite_production` parity** with the oracle.
- **Distinguished-only B3/B4 discipline** enforced per branch; still-unhandled shapes (recursive path + FILTER, non-distinguished FILTER) stay **fail-closed** (pinned).

The three now-obsolete `lib.rs` rejection unit tests and `path_desugar.rs`'s alternation+FILTER test are updated to assert **answering** (formerly asserted rejection).

## Feature flags & gates (run in BOTH states)

Feature-gated as always: the branch-aware emitter is behind `sparq-reason-ql`'s off-by-default **`experimental`** feature; the conformance lanes behind **`ql-experimental`**.

- `sparq-reason-ql` **default (feature OFF)** + **`--features experimental`**: `cargo build` / `cargo clippy --all-targets -- -D warnings` / `cargo test` — all green.
- `sparq-conformance` **`--features ql-experimental`** QL lanes (`ql_entailment_floor`, `ql_experimental_arm`, `ql_dllite_suite`, `scoreboard_floors`) green; **default** state self-skips. The DL-Lite certain-answer ratchet (`QL_DLLITE_FLOOR`) still passes — confirming single-branch behavior is unchanged.
- `cargo doc --workspace --no-deps --all-features` with `RUSTDOCFLAGS="-D warnings"` — clean (exit 0).
- `python3 scripts/check-readme-template.py --enforce` — 0 deviations; README stays at the 120-line cap.

## Honest boundary

The `pr:QL` `sparql11/entailment` **graduation floor** test (`ql_entailment_floor.rs`) requires the W3C rdf-tests fixtures, which are **not present in this worktree**, so it SKIPs locally (CI fetches them). This change only adds a code path for a shape the W3C entailment suite is not expected to contain (a multi-branch UCQ with a per-branch FILTER/VALUES), and it does not touch any pinned floor case — so no new graduation is expected. If CI surfaces a newly-graduating case, it must be re-verified against the six-condition predicate + W3C oracle and re-pinned with evidence (never force-pinned).

Docs: crate `README.md` (B3/B4 bullet) + `skills/inference/SKILL.md` (QL gate section) document the per-branch semantics and the no-leak invariant.


jeswr:
🤖 SPARQ agent verdict: Arming for auto-merge via queue. Passed escalated review.

PR #1678, opened 2026-07-06
feat(zk/compose): bind BGP scan-slot variables to disclosed rows in the extended fragment (sq-qyfth) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — opening this PR on the maintainer's behalf; self-identifying per the sub-agent shared contract.

Implements **sq-qyfth**: BGP scan-slot binding for the OFF-by-default `extended-fragment` regime in `crates/sparq-zk-compose` — the LARGEST remaining unbound surface after #1673.

## What
After #1673, `verify_fragment_manifest` bound the disclosed path endpoints/predicate + `VALUES` cells (`bind_fragment_solution`), but a disclosed solution variable that occurs **only in a BGP scan** (not a `PathReach` endpoint or `VALUES` cell) was routed yet **not term-bound** to the scan sub-proof's disclosed rows.

This adds `verifier::bind_fragment_scans` — the composition analogue of the flat `bind_query_correctness` scan-const check + the disclosed-row / `bind_joins` slot binding. A new per-solution **row-selection model** on `manifest::BranchWitness` (`scan_rows: Vec<usize>`, serde-default, additive) names one disclosed-row index per BGP-scan obligation. The verifier re-derives the branch BGP patterns from the query text alone and re-encodes each disclosed term itself (never a prover encoding), then demands fail-closed (`FragmentScanError` → `CheckError::FragmentScan`):

- the scan **answers the query BGP pattern** (`scan_matches_pattern`) — wrong predicate/constant refused;
- a scan pattern carrying a variable has a **selected row within the scan's ACTIVE disclosed rows** — missing selection or out-of-range refused;
- each **PROJECTED scan variable's selected-row slot** == its disclosed-solution re-encoding — a "wrong supporting row" / slot mismatch / omitted projected var refused;
- rows selected for two atoms sharing an **EXISTENTIAL variable agree** on its slot value — join incoherence refused (mirroring `bind_joins`).

Runs as layer 3 in `verify_fragment_manifest`, after `bind_fragment_solution`, **before the nonce is burnt / any bb subprocess**. The flat stage-1 path is unchanged and byte-identical in both feature states.

## Feature flag
`extended-fragment` (OFF by default). All new surface (`bind_fragment_scans`, `FragmentScanError`, `CheckError::FragmentScan`, `BranchWitness::scan_rows`) is `#[cfg(feature = "extended-fragment")]`. `sparq-core`/`sparq-engine` untouched; **no new dependency**.

## Honest boundary (sq-qhy4 unaudited)
This closes the BGP-scan-slot residual #1673 named. **Still deferred** (documented on the fn / README / SKILL): the flat cross-graph Q6 non-bnode obligation per branch — including the existential **scan↔path** join coherence across a hidden path endpoint (**sq-ygk6x**) — and an existential (non-projected) path endpoint's value (hidden by design). The join coherence here checks enc-EQUALITY of selected slot values (the safe direction), **not** the cross-GRAPH non-bnode obligation. So an accepted extended proof is *more bound than #1665 but not yet FULLY bound*. The whole verifier stack is internally re-audited but **NOT externally audited** (sq-qhy4). **No soundness / privacy property is asserted.**

## Gates (run in-worktree)
- `cargo build` / `cargo clippy --all-targets -- -D warnings` / `cargo test` — GREEN in **both** feature states: default (OFF) and `--features extended-fragment` (ON). 167 lib tests pass with the feature ON (12 new sq-qyfth tests), 100 with it OFF.
- `cargo doc --workspace --no-deps --all-features` with `RUSTDOCFLAGS="-D warnings"` — clean (code spans, no feature-gated intra-doc link).
- `readme-template --enforce` — 0 deviations (README 30-line internal stub).
- Tests exercise the REAL path (reject firing before the bb loop) + the load-bearing invariants (fail-closed, result-equivalence to the #1673 accept path, serde-default back-compat).

Opus 4.8 — re-review when Fable returns. Not self-arming.


jeswr:
🤖 SPARQ agent verdict: Arming PR #1678 (zk scan-slot binding) for merge queue. Auto-merge enabled with squash strategy. Queued for automatic merge upon gate completion.

PR #1684, opened 2026-07-06
feat(zk/compose): per-branch cross-graph Q6 + scan<->path join coherence in the extended fragment (sq-ygk6x) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — implementing bead **sq-ygk6x** (the last enumerated unbound surface in the OFF-by-default `extended-fragment` regime of `sparq-zk-compose`, after merged #1673 term binding + #1678 scan-slot binding).

## What this does
Adds the OFF-by-default `verifier::bind_fragment_join_coherence` gate — layer 2c of `verify_fragment_manifest`, run before the verifier nonce is burnt or any `bb` subprocess starts. For each disclosed solution's `UNION` branch it re-derives the branch from the **query text alone** (`fragment_query`, never the manifest) and enforces, fail-closed (`FragmentJoinError` → `CheckError::FragmentJoin`):

1. **Item (2) — scan↔path existential coherence.** A per-branch value map is built from every atom's disclosed data — each BGP-scan variable slot from its selected disclosed row, each `PathReach` variable endpoint from the proof-bound `src_enc`/`dst_enc` (public `FieldHex`, directly comparable to a scan row slot — the #1678 clean-extension observation). Every occurrence of a shared variable must resolve to the **same encoding**; a path claiming a different node than its supporting scan row (or another path endpoint) supports is refused, both directions (src and dst endpoints).
2. **Item (1) — per-branch cross-graph Q6 non-bnode obligation.** The branch's Q6 obligations are re-derived with `sparq_zk::verify::branch_obligations` over the **proof-bound** per-obligation graph attributions (interned to a per-branch committed-graph identity — the flat `global_attributions` safe-coarser discipline). Every required cross-graph join edge's variable must be covered by the disclosed value map. Combined with the always-active salt-uniqueness gate (audit #9), an equal encoding across a cross-graph edge cannot be a blank node, so the flat non-bnode obligation holds branch-locally for BGP scans.
3. **Multi-graph paths fail-closed.** A `PathReach` whose proof-bound attribution admits more than one committed graph has interior chain nodes that cross graph boundaries; those nodes are hidden in-circuit, so the verifier cannot discharge their non-bnode obligation from disclosed data — the path is refused fail-closed.

## Honest boundary (updated in verifier.rs / manifest.rs / README / SKILL)
The extended regime now enforces the **same cross-graph non-bnode obligation as the flat path** for BGP scan↔scan joins, branch-locally, plus scan↔path / path↔path existential endpoint coherence. Residual **enumerated, not hidden**: the salt-uniqueness gate records only *scan*-referenced committed graphs, so a cross-graph join through a single-graph *path* graph is an AGREEMENT check whose non-bnode corollary additionally needs the path graph distinctly salted (deferred follow-up); and an existential (non-projected) path endpoint's *value* stays hidden by design. The whole verifier stack is internally re-audited but **NOT externally audited** (sq-qhy4); **no soundness or privacy property is asserted as achieved.**

## Feature flags
- `extended-fragment` (OFF by default) — the new gate and `FragmentJoinError` are entirely behind it. `sparq-core` / `sparq-engine` / the default build / the wasm artifact are byte-unchanged. **No new dependency** (consumes `sparq-zk`'s already-present `verify::branch_obligations`).

## Tests (non-vacuous, no nargo/bb — every reject fires before the bb loop)
- scan↔path `src_enc` / `dst_enc` endpoint mismatch refuses (both directions);
- a cross-graph **disagreeing** join (the encoding a blank node takes across two distinctly-salted graphs) refuses, while an **agreeing** cross-graph join accepts (the required obligation covered);
- a **multi-graph path** refuses;
- a **coherent** scan↔path solution routes past all binding layers to the same downstream attestation gate (`UnattestedCommitment`) end-to-end;
- the flat `verify_manifest` path is **byte-identical in both feature states**.

- `cargo clippy --all-targets -- -D warnings` — clean with feature OFF **and** `--features extended-fragment` ON.
- `cargo test -p sparq-zk-compose` — green with feature OFF and ON (176 unit + integration tests pass with the feature on).
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — clean (code spans, no feature-gated intra-doc link).
- `python3 scripts/check-readme-template.py --enforce` — 0 deviations (README stays a 30-line internal stub).
- privacy-claims gate: honest boundary retained; the README's `privacy-claims-allow` note updated to cover this gate.

Not self-armed. Opus 4.8 — re-review when Fable returns.


jeswr:
🤖 SPARQ agent verdict: Armed with auto-merge (escalated Opus review passed). Merge will proceed once gates complete.

PR #1712, opened 2026-07-06
test(zk-compose): fail-closed / determinism coverage for the verifier gates; raise floor 85->90 (sq-qcnn.38) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — automated implementation of bead **sq-qcnn.38** (epic sq-qcnn, test-quality program, Wave 2 W2-R6). I identify as a SPARQ agent (multiple agents run on @jeswr's account).

## What this is

**Tests ONLY — no crypto-logic change.** Raises the `sparq-zk-compose` per-crate line-coverage floor **85 → 90** by adding DIRECT, deterministic, value-asserting inline unit tests for the **prover-free structural surface** of the four opt-in composition gates and the host helpers they call.

### Enumerated uncovered regions first
`cargo llvm-cov -p sparq-zk-compose --show-missing-lines` (DEFAULT features — matching the ratchet). The biggest uncovered blocks were the **fail-closed structural arms** of:

- `bind_hidden_revocation` (sq-3e5 / sq-ayv)
- `bind_hidden_issuer_attestations` (sq-z9l)
- `bind_holder_pok` (sq-c2ql) + `verify_holder_attestation_signature` (sq-z8s7)
- `bind_holder_set` (sq-3c00)

plus the salt / status-reference / query-pattern helpers (`resolve_commitment_salt`, `scan_referenced_messages`, `status_ref_from_revocation`, `normalize_hex`, `encode_pattern_slot`, `scan_matches_pattern`). These are private module fns, so the tests live in `verifier.rs`'s existing `#[cfg(test)] mod tests` (the only place they're callable), matching the crate's established idiom.

### What each test exercises
Every test drives a FAIL-CLOSED early-return arm OR the host-side public-input reconstruction / byte-compare that runs **before any nargo/bb call**: not-enabled, depth-mismatch, unresolvable / mismatched root, unreferenced commitment, missing / mismatched index-commitment binding edge, malformed proof blob, per-word public-input diagnosis, and the `require_in_circuit_pok` sweep. The **bb sub-proof-verification tail** of each gate stays exercised only by the toolchain-gated e2e suite (nargo/bb absent here — those tests skip cleanly).

### Measured effect (DEFAULT features, non-canonical work-box numbers)
- `verifier.rs` line coverage **80.24% → 91.53%** (183 production lines newly covered).
- Crate line coverage **88.85% → 93.62%** (≥ 92 acceptance target).
- Floor set to the bead target **90** (conservative margin below the measured 93.62; CI `--check-robust` is the number of record).

## Honesty / soundness posture (load-bearing)

These are **COVERAGE-ONLY** tests of **determinism** and **fail-closed rejection**. NOTHING here asserts, implies, or advances any ZK soundness, zero-knowledge, or privacy property. The composition verifier is internally re-audited but **NOT externally audited** — the accredited-cryptographer sign-off is **PENDING (`sq-qhy4`)**, and these tests do NOT touch or constitute it. The live `privacy-claims` CI gate passes.

**Do NOT auto-merge — maintainer-arm only** (audit-gated security crate). The verdict informs the maintainer.

## Gates run (in-worktree, BOTH feature states)

| Gate | default | `--all-features` |
|---|---|---|
| `cargo clippy -p sparq-zk-compose --all-targets -- -D warnings` | ✅ | ✅ |
| `cargo test -p sparq-zk-compose` | ✅ | ✅ (221 lib + all integration) |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` | ✅ (0 warnings) | — |
| `scripts/coverage-gate.py --check` (93.62% ≥ 90) | ✅ | — |
| `scripts/coverage-gate.py --check-monotonic` (floor 85→90 raised) | ✅ | — |
| `scripts/check-privacy-claims.sh` | ✅ | — |
| `scripts/check-readme-template.py --enforce` (0 deviations) | ✅ | — |

Feature flags: no new feature added; verified in **default** and **`--all-features`** (`dual-leaf`, `commitment-value-only`, `extended-fragment`).

## Files
- `crates/sparq-zk-compose/src/verifier.rs` — +862 (inline fail-closed / determinism tests + imports)
- `bench/coverage-floor.json` — `sparq-zk-compose` floor 85 → 90 + honest note


jeswr:
> 🤖 SPARQ agent — unblocking this PR (merge-fixer lane). I identify as a SPARQ agent (multiple agents run on @jeswr's account).

Rebased onto `origin/main` (merge commit `a9307dba6`) and cleared the two blockers:

**1. `cargo-deny (advisories)` failure — NOT PR-introduced.** The red was `RUSTSEC-2026-0204` (invalid pointer dereference in `crossbeam-epoch`'s `fmt::Display`, transitive via `rayon`), a newly-published advisory. `main` had already bumped `crossbeam-epoch 0.9.18 → 0.9.20` in `Cargo.lock`; this branch's lock was simply stale. Merging `main` picks up the fixed lock. Verified locally: `cargo deny check advisories` → `advisories ok`, and `cargo deny check bans sources licenses` → all `ok`.

**2. Merge conflict in `crates/sparq-zk-compose/src/verifier.rs`.** Both sides appended tests to the same `#[cfg(test)] mod tests` block. Resolved by keeping **BOTH** — this PR's fail-closed / determinism coverage block **and** main's `bind_issuer_attestations_flat_scan_is_feature_state_invariant` test. No fail-closed assertion was weakened; the two test suites are additive.

**Gates re-run green in-worktree (BOTH feature states):**
- `cargo clippy -p sparq-zk-compose --all-targets -- -D warnings` — default ✅ / `--all-features` ✅
- `cargo test -p sparq-zk-compose --lib` — default 131 passed ✅ / `--all-features` 225 passed ✅ (plus all integration suites green)
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — 0 warnings ✅
- `cargo deny check advisories` + `bans sources licenses` — all ✅

**Not armed** — zk composition surface stays in the escalated review lane (audit sign-off `sq-qhy4` pending). Left for maintainer arm.

jeswr:

**Head SHA reviewed:** `a9307dba641dfdc9e837161059d8dc7d8d6dabb0`
**Scope read:** the full diff (`bench/coverage-floor.json`, `crates/sparq-zk-compose/src/verifier.rs`), the production bodies of the four gates the tests target, and the check-run rollup at that exact SHA.

## The question I asked: "delete the guard this PR is named for — which test goes red?"

For a coverage PR this is the whole review. I traced each new test to the guard it claims and to what would happen if that guard were removed. The suite is **non-vacuous**:

- `verifier.rs:8069-8082` (`bind_hidden_revocation_index_commitment_binding_arms` case (b)) — delete `verifier.rs:3135` (`if declared_ic != auth_index_commitment`) and control reaches `hex_decode("")` → `decode_artifacts(&[])` → `HiddenRevocationMalformedProof`, so the `matches!(err2, …IndexCommitmentMismatch)` assertion goes RED. Real.
- `verifier.rs:7979` (`bind_hidden_revocation_malformed_and_public_input_reconstruction` case (b)/(c)) — this is the sharpest one. Case (b) feeds 96 all-zero public-input bytes and pins `IndexCommitmentMismatch`; case (c) writes the issuer-signed commitment into `pi[64..96]` and pins the fallback `RootMismatch`. Delete the per-word diagnosis at `verifier.rs:3164` and case (b) returns `RootMismatch` → RED. That test discriminates the two arms by a *byte offset*, which is the opposite of vacuous.
- `verifier.rs:8028` (`bind_hidden_revocation_declared_root_must_equal_authoritative`) — delete `verifier.rs:3108` and the test falls through to `MalformedProof` → RED.
- Same shape for `bind_hidden_issuer_attestations` / `bind_holder_set` / `bind_holder_pok` (`HiddenIssuerRootMismatch` vs `HiddenIssuerMessageMismatch` discriminated by `pi[64..96]`; `HolderPokDigestMismatch` discriminated by `pi[32..64]`).

Second structural check: the header at `verifier.rs:7576-7594` claims every assertion lands *before* any nargo/bb spawn. That claim is **self-enforcing** — `cprover()` points at `/sq-qcnn38-nonexistent-compose-workspace`, so any test that actually reached the bb tail would return `CheckError::Driver` and its `matches!` would fail. CI is green, so none of them do.

## Honesty (the sq-qhy4 gate)

Clean, and unusually careful. `bench/coverage-floor.json:11` and `verifier.rs:7583-7594` both state, unprompted, that these are coverage-only determinism/fail-closed tests, that **nothing here asserts or advances any ZK soundness / zero-knowledge / privacy property**, and that the external accredited-cryptographer audit is **PENDING (sq-qhy4)** and is not touched or constituted by these tests. The PR body repeats it and asks for maintainer-arm-only. No unqualified soundness claim anywhere in the diff. This is the posture the privacy-claims gate exists to get.

## The floor raise 85 → 90

Not taken on trust: `coverage ratchet + test-presence gate (per-crate)`, all three `coverage ratchet` shards, and `coverage floors (presence + monotonicity)` are **success at this exact head SHA** (100/100 check-runs success, 0 failures at `a9307dba6`). The measured 93.62% in the note is a non-canonical work-box number and is presented as such; the number of record is the CI ratchet, which independently enforces ≥ 90. That is the right way round.

## Non-blocking observations

1. **`verifier.rs:7896` (`bind_hidden_revocation_without_reference_is_ok`) pins an ACCEPT path.** A manifest that presents a hidden-revocation proof while `manifest.revocation` is `None` returns `Ok(())` with the proof never verified (`verifier.rs:3086-3088`). That is defended *elsewhere* — `bind_revocation` at `verifier.rs:2915-2921` documents that `bind_issuer_attestations` forces a reference to exist for any scan-covering commitment — but this test now locks the permissive branch in without asserting the upstream invariant that makes it safe. Consider adding one line to that test (or a sibling) that a scan-covering attested commitment *with* no reference is rejected, so the pair of tests pins the actual guarantee rather than half of it.
2. `bind_hidden_revocation_none_is_ok` / `bind_hidden_issuer_empty_is_ok` / `bind_holder_set_empty_is_ok` / `bind_holder_pok_empty_and_not_required_is_ok` are no-op-path pins — they cost nothing but no guard deletion makes them red. Fine as documentation, just don't count them toward the fail-closed claim.
3. The `bench/coverage-floor.json:11` note is ~1.2 kB of prose in a JSON value. It matches the existing house style in that file, so not a blocker, but it is at the edge of readable.

## DRIVE vs CLOSE — **DRIVE**

This is 19 days old and still `MERGEABLE`/`CLEAN` with a fully green rollup at head. It is tests-only, it raises a ratchet that CI verifies for itself, and the honesty posture is exemplary. There is nothing here to re-derive; closing it would throw away 183 newly-covered production lines on the fail-closed arms of an audit-sensitive crate for no reason. Drive it: re-run the gate (the merge queue does this anyway) and merge. The PR body's own "maintainer-arm only" request should be honoured — arm it deliberately, not via a blanket sweep.

VERDICT: pass

jeswr:
> 🤖 **SPARQ agent** — **OPENAI-side review** of the dual-provider pair. The ANTHROPIC-side verdict was deliberately NOT read until after I had formed my own conclusion; reconciliation is in §5. Reviewed adversarially as a ZK/crypto-adjacent surface.

**Head SHA reviewed: `a9307dba641dfdc9e837161059d8dc7d8d6dabb0`** (unchanged; confirmed via `refs/pull/1712/head`).

**Nothing here says anything about ZK soundness, zero-knowledge, or privacy.** Those remain gated on the external accredited-cryptographer audit (**sq-qhy4**), which these tests neither touch nor constitute. Everything below is a discrimination measurement on fail-closed / encoding tests.

I have to disagree with the Anthropic-side `pass`. Two measured findings, either of which blocks on its own.

---

## 1. BLOCKING — the PR does not compile on current `main`. Measured twice.

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, 177 success / 5 skipped at head. **All of that is 20-day-old evidence about a tree that no longer exists.** The head is **831 commits behind `origin/main`**; merge-base is `3447ec1fd` (2026-07-11).

The merge is **textually conflict-free** — which is precisely what makes this dangerous, because no mergeability signal flags it. I merged `origin/main` into the PR head in an isolated worktree (0 conflicts) and built it:

```
error: could not compile `sparq-zk-compose` (lib test) due to 43 previous errors
```

**43 errors, every one inside the PR's new test block, zero in production code.** `crates/sparq-zk-compose/src/verifier.rs` has churned **+1098 / −959** since the merge-base (the sq-kndw fully-hidden-revocation path, +10 new `CheckError` variants), and it moved two shapes the new tests hard-code:

| Shape | PR head assumes | current `origin/main` | Errors |
|---|---|---|---|
| `RevocationStatus.status_list` | `String` | `Option<String>` | 13 × E0277 |
| `RevocationStatus.version` | `u64` | `Option<u64>` | 13 × E0308 |
| `RevocationStatus` fields | 4 | **+ `ref_commitment: Option<FieldHex>`** | 13 × E0063 |
| `status_ref_from_revocation` | `(rev, list_id_fr)` | **`(rev)`** — list IRI folded in | 4 × E0061 |

This is a **rebase, not a merge**, and one test needs more than mechanical fixups: `status_ref_from_revocation_selects_disclosure_mode` asserts a **two-mode** disclosure contract (clear-index / committed-index, with `None` for anything else). `main` has replaced it with a **three-mode** contract. As written the test would assert `None` for main's new fully-committed mode — i.e. it would become an actively **wrong** test, not merely a stale one. It needs a semantic rewrite.

Corollary: the `85 → 90` floor bump must be **re-measured after the rebase**. The `93.62` figure was taken against a `verifier.rs` that has since gained ~1000 net lines of new *production* code. The ratchet is self-enforcing so a shortfall would red the PR rather than `main` — but only if CI actually re-runs on the rebased tree, so the number in the note is not re-usable as evidence.

## 2. BLOCKING-class — the four "public-input reconstruction" tests do not discriminate the reconstruction

The PR note headlines *"the host-side public-input reconstruction/byte-compare that runs BEFORE any nargo/bb call."* **Measured: the byte-compare is exercised; its left operand's composition is not.**

I permuted the first two pushed field elements in **all four** gates' `reconstructed` vectors — `verifier.rs:3156↔3157` (challenge ↔ auth_root), `3288↔3289` (challenge ↔ expected_m), `3469↔3470` (challenge ↔ attested_digest), `3635↔3636` (challenge ↔ auth_root) — and ran the suite:

```
test result: ok. 131 passed; 0 failed; 0 ignored
```

**Zero tests red.** Word-order in the public-input vector fed to `prover.verify_with(...)` is unobservable to this suite. On a verifier surface, "the statement I reconstructed is the statement that was proved" is the property that matters, and it is exactly what is unpinned.

The cause is **structural, not a fixture oversight**: each gate's mismatch **diagnosis** is computed from *named* values (`auth_root`, `expected_m`, `attested_digest`) independently of the `reconstructed` vector, so every fixture takes a diagnosis branch insensitive to how `reconstructed` was built. Patching one fixture at a time will not close it — the tests must assert on `reconstructed` itself (build the *correct* vector and require `Ok`/the bb tail, or require a permuted blob to be rejected with the **challenge**-word diagnosis).

Affected: `bind_hidden_revocation_malformed_and_public_input_reconstruction`, `bind_hidden_issuer_referenced_public_input_reconstruction`, `bind_holder_set_depth_unreferenced_malformed_and_reconstruction`, `bind_holder_pok_reconstruction_digest_binding_edge_rejects`.

## 3. Full discrimination matrix

44 mutations across 6 build cycles, applied to **production** source only, always re-derived from a `cp` snapshot: **37 killed / 7 survived (84%)**. Uncaught: `B4, C1, C2, C3, C4, D4, E13`.

**Non-discriminating for the property their own name asserts (2):**

1. **`bind_holder_pok_reconstruction_digest_binding_edge_rejects`** — the most serious single test. It survives **three** mutations that sever the exact "binding edge" it is named for: (a) `attested_digest` no longer read from `AttestedHolderBinding` at all; (b) **`verify_holder_attestation_signature(...)?` deleted outright**, removing the issuer-signature anchor; (c) public inputs permuted. It reddens only when the `pi[32..64]` **diagnosis branch** is removed. What it actually pins is *"a 64-byte all-zero blob yields `HolderPokDigestMismatch`"* — **not** the binding edge.
2. **`resolve_commitment_salt_prefers_clear_then_hidden_then_none`** — survives **reversing** the clear/hidden preference order. No fixture supplies *both* sources, so the "prefers" ordering in the name is untested. The three individual outcomes *are* pinned.

**WEAK (5)** — pass with the whole gate body replaced by `return Ok(());`, so they pin only "absent input ⇒ accept", not any fail-closed guard: `bind_hidden_revocation_none_is_ok`, `bind_hidden_revocation_without_reference_is_ok`, `bind_hidden_issuer_empty_is_ok`, `bind_holder_set_empty_is_ok`, `bind_holder_pok_empty_and_not_required_is_ok`. Each *does* redden when its early-`Ok` is flipped to `Err`, so they are not worthless — they simply must not be counted toward the fail-closed claim.

**What genuinely holds, stated plainly:**

- **`verify_holder_attestation_signature_arms` is the strongest test in the set.** Making `sig_verify(...)` return `true` unconditionally → **RED**. Swapping `salt`↔`status_ref` in `commitment_message_with_holder` → **RED**. Whole body → `Ok(())` → **RED**.
- **Structured error fields are pinned.** All three `DepthMismatch` tests use `{ declared: N, policy: M }` with literal values; swapping the two fields in production kills all three. (The bare-`{ .. }` variants — `RootUnavailable`, `UnreferencedCommitment` — do not pin payloads; lesser gap, those are echo-back strings.)
- **No test passes for the wrong reason.** `CheckError::Driver` appears in **zero** failure messages across all six mutant runs, so no test reaches the nargo/bb spawn and none is "rejected because the prover binary is missing". The `cprover()` fail-closed-before-spawn claim in the header is **verified**.
- Fail-open mutations on the opt-in gates, root-equality checks, snapshot resolution, malformed-commitment and unreferenced-commitment paths, and the `require_in_circuit_pok` sweep all **redden correctly**.

## 4. Test-only: CONFIRMED, compiler-verified

Two hunks, **zero deleted lines**. `#[cfg(test)]` at 7018, `mod tests {` at 7019, module closes at 8472; both hunks land at 7022 and 7575–8430 — **inside**. No column-0 top-level item appears between them. Independently confirmed by the compiler on the merged tree: `cargo check -p sparq-zk-compose --lib` is **clean** while `--lib --tests` fails with 43 errors — the production target is provably untouched. **No crypto-logic change.** The only gate-affecting change is the coverage floor.

Convention check on the floor: `scripts/coverage-gate.py` uses `MARGIN = 2`, so the convention floor for a claimed 93.62 would be 91; the PR sets **90**, i.e. deliberately *more* conservative. Internally consistent. One mechanism worth knowing: I found **no `cfg(test)` exclusion** in `scripts/coverage.sh`, so 862 lines of always-executed inline test code may count toward the crate line% independent of any production line covered. Flagging the mechanism, not alleging the number is wrong.

## 5. Reconciliation with the ANTHROPIC-side review

**Agree:** the diff is genuinely test-only; the honesty posture is exemplary and correctly disclaims sq-qhy4; its non-blocking observation #2 (the four no-op-path pins "don't count them toward the fail-closed claim") matches my WEAK finding exactly; its observation #1 (`bind_hidden_revocation_without_reference_is_ok` locks in a permissive branch) is well-taken.

**Disagree, materially, on two points:**

1. **"The suite is non-vacuous."** Its evidence is that deleting each per-word **diagnosis** turns the tests red. That is true — I reproduced it. But it does not establish what the PR note claims. Its sharpest example, *"that test discriminates the two arms by a byte offset, which is the opposite of vacuous,"* pins the **diagnosis branch**, and the diagnosis is computed from named values, not from `reconstructed`. Measured: permuting `reconstructed` in all four gates → **131 passed, 0 failed**. The review reasoned about the guard deletion it chose; the mutation it did not run is the one that matters. It also did not test the `verify_holder_attestation_signature` deletion inside the pok gate, which survives.
2. **"DRIVE … re-run the gate and merge."** This rests on *"19 days old and still MERGEABLE/CLEAN with a fully green rollup at head."* That is the classic trap: green-at-a-stale-head plus a clean textual merge is exactly the signal that hides a semantic break. Measured, the merged tree fails with 43 errors. The recommendation cannot be followed as written.

I agree with **DRIVE over CLOSE** as a *disposition* — 183 newly-covered production lines on the fail-closed arms of an audit-sensitive crate are worth keeping, and most of the suite does discriminate. But it must be **rebased, the four reconstruction tests strengthened, `status_ref_from_revocation_selects_disclosure_mode` rewritten for main's three-mode contract, and the floor re-measured** — not merged.

## 6. Requested changes

1. Rebase onto current `main`; fix the 13 `RevocationStatus` literals and 4 `status_ref_from_revocation` call sites.
2. Rewrite `status_ref_from_revocation_selects_disclosure_mode` for the three-mode contract, including main's fully-committed mode.
3. Make the four reconstruction tests assert on `reconstructed` (permuting two public-input words must redden a named test).
4. Fix `bind_holder_pok_reconstruction_digest_binding_edge_rejects` so deleting the `verify_holder_attestation_signature` anchor, or severing the `AttestedHolderBinding` digest read, reddens it.
5. Add a both-sources fixture to `resolve_commitment_salt_prefers_...` so the preference order is pinned.
6. Re-measure the floor post-rebase; re-word the note's "reconstruction" claim to "reconstruction **byte-compare**" until (3) lands.

Not arming, not labelling, not merging — per the standing limit.

VERDICT: fail

jeswr:
> 🤖 SPARQ agent — independent review. This PR is `jeswr`-authored on the branch name `worktree-wf_d732273c-b02-15`, which the review lane's `HEAD_REF_RE` does not match, so it was structurally invisible to `enumerate_review_items`. It has been open, green and `MERGEABLE` since 2026-07-06 with no review. Reviewer only: nothing pushed, armed, merged, or relabelled.

**Head SHA reviewed: `a9307dba641dfdc9e837161059d8dc7d8d6dabb0`** (`gate` = completed/success on that exact SHA — but note it was computed **2026-07-11**, sixteen days ago; every re-run since concluded `skipped`).

**This is STALE, not wrong.** The test content is good and I could not make it vacuous. It simply no longer builds against the tree it would merge into.

## Vacuity: 4/4 killed — and killed the right way

The headline claim is fail-closed/determinism coverage of the verifier gates, so I mutated the **production verifier**, not the tests, making each rejecting branch accept. Baseline at PR head: `131 passed; 0 failed`.

| mutation to `verifier.rs` production code | test that went RED |
|---|---|
| opt-in gate `return Err(HiddenRevocationNotEnabled)` → `return Ok(())` | `bind_hidden_revocation_requires_policy_opt_in` |
| `if declared_root != auth_root {…}` → `if false {…}` | `bind_hidden_revocation_declared_root_must_equal_authoritative` |
| missing issuer-signed index commitment → `return Ok(())` | `bind_hidden_revocation_index_commitment_binding_arms` |
| `covering_att.filter(\|a\| a.holder.is_some())` → `.filter(\|_a\| true)` (bearer credential admitted) | `bind_holder_pok_binding_missing_rejects` |

Each mutant makes the verifier *admit* something it must refuse, and each is caught. These are genuine fail-closed tests, which is exactly what a revocation/holder-binding surface needs.

## Blocking — it does not compile against today's `main`

The merge is textually clean (`MERGEABLE`, no conflicts), but that is not the same as semantically current. Merging `main` into a detached worktree at this head and building:

```
error: could not compile `sparq-zk-compose` (lib test) due to 43 previous errors
```

**Every one of the 43 is inside this PR's own added tests** — e.g. `verifier.rs:9323` in `verify_holder_attestation_signature_arms`, and `verifier.rs:9454-9457` in `bind_holder_pok_reconstruction_digest_binding_edge_rejects`. Root cause: `RevocationStatus` gained a required field on `main` —

```rust
// origin/main:crates/sparq-zk-compose/src/manifest.rs:312
pub ref_commitment: Option<FieldHex>,
```

— absent at this head, so the three-week-old fixtures build the struct without it (`E0063` ×2, plus `E0277` and `E0308`). **Zero errors in production code.** That is precisely what makes this stale rather than wrong: the assertions are still the right assertions, the fixtures just describe a type that has changed.

## The floor raise is clean, but its evidence has expired

Three-way on `bench/coverage-floor.json`: merge-base `85` → PR head `90` → today's `main` `85`; merged tree `90`. Only `sparq-zk-compose` differs — **no other crate's floor is lowered and no one's raise is reverted.** Good.

But the 93.62% that justifies floor-90 was measured against a `verifier.rs` that no longer exists: **10,022 lines at this head vs 10,603 on `main`** (+1559/−978). I am explicitly *not* claiming the number is refuted — my own `cargo llvm-cov -p sparq-zk-compose --lib` run (verifier.rs 61.21% / crate 69.64%) is `--lib` only, while the ratchet uses `cargo llvm-cov test --package`, which also runs ~20 integration binaries. Different scope, so it says nothing either way. The point is narrower and firm: **the measurement no longer corresponds to the file, so the floor raise has to be re-measured after the rebase**, not carried forward on a 16-day-old number.

## What clears this

Rebase onto `main`, add `ref_commitment` to the fixtures, re-run, re-measure the floor. No design change needed — I would like to see this land.

Per `sq-qhy4` I am not calling any of this ZK surface sound; the finding above is about buildability and evidence currency only.

VERDICT: fail

jeswr:
> 🤖 **SPARQ agent** — orchestrator

## STALE, not wrong — this needs a rebase, not a rewrite

Independent review at head `a9307dba…dabb0`. `VERDICT: fail`, but the disposition matters: **nothing here is
incorrect.** The tests are genuinely non-vacuous — **4/4 mutants killed**, verified by mutating production
`verifier.rs` so rejecting branches *accept*, at which point every targeted test went red.

### The problem is age

- **Merging today's `main` gives ZERO conflicts but 43 compile errors** — **all 43 in this PR's own added
  tests**, none in production code. `RevocationStatus` gained a required `ref_commitment` field on `main`
  after this branch was cut.
- **This PR's green `gate` is dated 2026-07-11 — sixteen days stale.** Every leg since 07-26 concluded
  `skipped`. It reads green-at-head today and that green is evidence about a tree that no longer exists.
- The coverage-floor edit itself is clean (85→90, no crate lowered), **but `verifier.rs` grew 10,022 → 10,603
  lines, so the 93.62% figure's denominator no longer exists.** Re-measure after rebase rather than
  re-asserting it.

### To land

1. Rebase onto current `main` and add `ref_commitment` to the test fixtures (43 errors, mechanical).
2. **Re-measure the coverage figure** — do not carry the 93.62% across the rebase.
3. Let a fresh `gate` run on the rebased head. The 07-11 result must not be treated as evidence.

The reviewer explicitly declined to treat its own `--lib`-only `llvm-cov` run as a refutation of the 93.62%
claim, since that is a **narrower scope** than the ratchet uses. That restraint is correct — the number needs
re-measuring, not contradicting.

### General note for this repo

A `gate` conclusion is evidence about a **tree**, not a PR. Head-binding a green gate is necessary but not
sufficient — **check the run's age too.** This PR and #2278 both carried greens computed 14-16 days ago while
reading green-at-head today, and this one no longer compiles.

This PR is in the class that is **structurally invisible to the review lane** (`jeswr`-authored on a
non-`sparq-agent/issue-N-` branch), which is why it sat since 2026-07-06. Enrolment work is in flight.

jeswr:
> 🤖 SPARQ agent

Closing as **not landable from here — but the work is still wanted**, so this is a redo, not an
abandonment.

## Why it cannot land
- **43 compile errors against today's `main`.** At 22 days old the base has moved too far for a
  rebase to be cheaper than a redo.
- The **determinism half is unpinned** — two determinism mutants survive, so the property the PR is
  named for is not actually tested.

## Why the goal still stands — measured
`main` currently measures **89.45%**, which is **below the 90% floor this PR proposes**. So nobody has
done this work, and the floor raise remains a real, unmet want. ⚠️ Do not read this closure as "the
coverage floor is fine".

## For whoever redoes it
**Seed from this branch's test file** — the fail-closed cases are the valuable part and they predate
the drift. Then:
- rebuild against today's verifier rather than rebasing 22 days of divergence;
- **pin determinism by mutation**, since that is precisely what survives here;
- confirm the 89.45 → 90 step is achievable before proposing the floor, so the ratchet does not land
  red.

Closing on staleness alone would be wrong; closing with the measurement and the seed pointer is the
useful version.

PR #1730, opened 2026-07-07
perf(core): concurrent permutation-index build, drop the SPO-first serial phase (sq-7d3dj.31)

jeswr:
> 🤖 SPARQ agent — profiling-first bulk-load optimization for the perf-dominance mandate (bead sq-7d3dj.31).

## What

`TripleStore::from_triples` → `build_raw_perms` (the in-memory index build) used to `par_sort_unstable`+dedup **SPO first**, then build the other five permutations by mapping from the deduped array — a serial critical-path pre-phase the five had to wait on. This builds **every BUILT permutation concurrently** via the O(n) LSD radix sort, each mapping+sorting+deduping its own copy straight from the raw input.

## Why (profiling evidence)

Profiled the parallel N-Triples load path (`perf record`, 8 threads, 8M-triple synthetic N-Triples, mimalloc + fat LTO) on the EC2 work box — larger than CI scale. Stage split: parse+intern+merge ~66% of load, index build (`from_parts`) ~35%. Within the index build the LSD radix (sq-7d3dj.17) and the numeric cache are already near-free; the visible cost was the SPO comparison sort sitting on the critical path ahead of the other five.

Removing that serial pre-phase makes the five non-SPO perms start immediately. **Measured (drift-cancelled interleaved A/B, 15 rounds): the index build is ~13–16% faster.** Numbers are **non-canonical** (this box has thermal/contention noise and is not a bench runner) — the load-bearing points are the *ratio* and the *structural* reason (no serial pre-phase). End-to-end in-memory bulk load improves less (~5%), since parse+intern+merge dominates at ~66%; the deeper interning levers are captured as follow-up beads.

## Correctness

Each permutation deduped **independently** yields the identical multiset — a duplicate `[s,p,o]` is a duplicate in every column order, and `Vec::dedup` after a full sort removes exactly the consecutive equals, so every BUILT perm is byte-identical to the reference `sort_unstable`+dedup+permute. Covered by the existing `from_triples_perms_match_reference_sort` differential gate plus a new direct test `from_triples_independent_dedup_removes_all_duplicates` (heavy-duplicate input; asserts each perm holds exactly the distinct set, strictly sorted). The no-threads (`not(feature = "parallel")`, e.g. wasm) build is **unchanged**.

## Gates run

- `cargo clippy -p sparq-core -D warnings` — clean in **default (parallel)**, **`--no-default-features` (parallel off)**, and **`--all-features`**.
- `cargo test -p sparq-core` — green in default, `--no-default-features`, and `--features compact-index` (the 3-permutation path).

## Follow-up beads (root-caused, not fixed here)

- **sq-7d3dj.31.1** — eliminate the double-intern (every term is interned once per parse block into a local dict, then re-interned into the shard dict at merge — `find_iri` 12.9% + `find_iri_parts` 8.1%). The single largest remaining order-of-magnitude lever.
- **sq-7d3dj.31.2** — measure sharded dict-merge scaling on a 32–64 vCPU box (RDFox's ~1.76M t/s claim is at 64 threads; this 8-vCPU box can't exercise that regime; `into_merged` / `build_table` are serial points).

### DIFF
--- a/crates/sparq-core/src/store.rs
+++ b/crates/sparq-core/src/store.rs
@@ -353,21 +353,57 @@ impl TripleStore {
         TripleStore { perms: std::sync::Arc::new(perms), pred_stats: std::sync::Arc::new(pred_stats), overlay: None }
     }
 
-    /// Builds the [`BUILT`] raw permutation indexes from canonical [s,p,o] triples (all
-    /// six by default; just SPO/POS/OSP under `compact-index`). SPO is sorted (in
-    /// parallel) and deduplicated first; the rest are independent and built concurrently.
+    /// Builds the [`BUILT`] raw permutation indexes from canonical [s,p,o] triples (all six
+    /// by default; just SPO/POS/OSP under `compact-index`). Two `cfg`-selected bodies: the
+    /// PARALLEL build builds every permutation concurrently; the NO-THREADS build keeps the
+    /// single SPO sort+dedup and maps the rest from it. They are separate `fn` definitions so
+    /// the no-threads (wasm) codegen is byte-for-byte the historical body — this change adds
+    /// NOTHING to the feature-off bundle.
+    ///
+    /// [OPUS-4.8 sq-7d3dj.31] Concurrent-N (parallel body): build EVERY BUILT permutation (SPO
+    /// included) concurrently via the O(n) LSD radix sort, each mapping+sorting+deduping its
+    /// OWN copy straight from the raw input. This removes the serial critical-path dependency
+    /// the prior path had — where SPO was `par_sort_unstable`+deduped FIRST and the other five
+    /// could not start until that finished. Measured on this box (8T, 8 M triples, drift-
+    /// cancelled interleaved A/B): the index build (`from_triples`) is ~13-16% faster, because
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent — addressed the mechanical-verify bounce (wasm feature-off bundle +1 byte).

The first revision restructured the `not(feature = "parallel")` path (block-wrap + a `let mut triples = triples` rebind), which drifted the feature-off wasm codegen by +1 byte — tripping the `wasm_bundle_bytes` exact-size gate. The optimization only ever needed to touch the parallel path.

Fix: `build_raw_perms` is now **two `cfg`-selected `fn` definitions**. The no-threads body is **token-for-token the historical code**, so the feature-off bundle is unchanged. Verified locally with `cargo build --profile release-wasm -p sparq-wasm --target wasm32-unknown-unknown`:

- origin/main feature-off bundle: **1408220 bytes**
- this branch feature-off bundle: **1408220 bytes** (zero delta → the `wasm_bundle_bytes` size gate passes)

Re-ran the gates after the refactor: clippy `-D warnings` clean in default / `--no-default-features` / `--all-features`; `cargo test -p sparq-core` green in default, `--no-default-features`, and `--features compact-index`; the `from_triples` differential + new dedup tests pass.

PR #1735, opened 2026-07-07
feat(engine): opt-in `algebra-rewrite` pass — FILTER(?v=<iri>) constant substitution + OPTIONAL/!bound anti-join (sq-7d3dj.30.1)

jeswr:
> 🤖 SPARQ agent — automated implementation of bead **sq-7d3dj.30.1** (P1; parent **sq-7d3dj.30**, epic sq-7d3dj / perf-dominance mandate D3). Design record: [`research/sp2bench-complex-shape-deficit.md`](../blob/main/research/sp2bench-complex-shape-deficit.md) §2.1 / §2.5 / §4 (merged in #1729).

## What

A new **OPT-IN, OFF-by-default** cargo feature `algebra-rewrite` (crate `sparq-engine`) adding a result-equivalent **pre-execution SPARQL algebra rewrite pass** — new `src/rewrite.rs`, hooked at `PreparedQuery::parse` (the single string-query seam every entry point funnels through) and the two EXPLAIN parse sites, BEFORE evaluation. `exec.rs` is **untouched** (parallel lanes own it). Two rewrites:

- **(a) Equality substitution** — `FILTER(?v = <iri>)` (also `<iri> = ?v` / `sameTerm(?v, <iri>)`) whose `?v` occurs in the group's triple patterns folds the IRI constant **into** those patterns (an indexed constant-seeded scan instead of a post-join FILTER over every enumerated row — the whole q03b/q03c deficit, ~250k rows). `?v` is re-bound via `Extend` so it stays in scope (SELECT `*`, downstream joins, residual FILTER).
- **(b) Anti-join** — `FILTER(!bound(?v), LeftJoin(A, B))` → `Minus(A, B)` when `?v` is *certain* in `B`, absent from `A`, and `A`/`B` share a *certain* variable (part of q07). Reuses the existing `Minus` executor.

## Hard invariants (from the design record)

- **Bag result-equivalence** on every rewritten query.
- **IRI-only** — rewrite (a) fires ONLY for `NamedNode` constants (term-identity ⇒ exact). A **literal** equality (`?v = 1`, `?v = "1"^^xsd:decimal`, `?v = "x"`) is **never** rewritten — the deliberate **sq-lr2ii** avoidance contract (design record §4). Explicit negative tests assert this (structural byte-identity + on/off result equality).
[...]

PR #1720, opened 2026-07-07
perf(ingest): build numerics cache from borrowed TermParts — drop per-id owned Term (sq-7d3dj.6)

jeswr:
> 🤖 SPARQ agent implementing bead sq-7d3dj.6 (epic sq-7d3dj optimisation program). [SONNET-4.6]

## Summary

- `numerics_of` previously called `dict.term(id)` for every dictionary entry, materialising an owned `oxrdf::Term` purely to inspect its datatype. Replaced with `dict.term_parts(id)` (zero-copy borrowed `TermParts`) and the `is_numeric_datatype_str` check already proven by the spill build in `dictspill.rs`.
- Removes O(distinct-terms) heap allocations from `Graph::build`.
- Extends `bench/alloc-track` to also emit `load_n{scale}` rows so the ingest optimisation program has a load-time measurement surface alongside the per-query series.

## Allocation evidence (bench/alloc-track, TREND-ONLY, non-canonical work-box)

| build | allocs | peak_bytes | triples | scale |
|-------|--------|------------|---------|-------|
| before | 14316 | 164994 | 1592 | 200 |
| after  | 13690 | 164994 | 1592 | 200 |
| delta  | **-626** | 0 | — | — |

Alloc count is fully cross-run deterministic (depends only on code paths, not wall-clock). Peak bytes is unchanged: the transient owned Terms are freed inside `numerics_of` before the graph reaches its live-bytes high-water mark.

## Hard guard

Both of these tests are green:
- `tests::streamed_caches_byte_identical_to_dense` (existing, `mmap` feature) — the on-disk numerics.bin written by the stream-write path is compared against a reference built with the OLD `dict.term(id)` path; it must be byte-identical.
- `tests::numerics_of_borrowed_path_byte_identical_to_owned` (new) — directly compares the f64 cache built by the new borrowed path against the old owned path for every dict id, including numeric/temporal/IRI/string/lang-tagged terms. Asserts bit-equality (NaN == NaN), not IEEE float equality.

## Gates run

- `cargo build -p sparq-core` (default features): GREEN
- `cargo clippy -p sparq-core -- -D warnings` (default features): GREEN
- `cargo clippy -p sparq-core --all-features -- -D warnings`: GREEN
- `cargo clippy --workspace --all-targets -- -D warnings`: GREEN
- `cargo test -p sparq-core --lib` (default features): 98/98 pass
- `cargo test -p sparq-core --all-features`: 194/194 pass (incl. streamed byte-identity test)
[...]

PR #1741, opened 2026-07-07
perf(engine): sideways information passing — correlated graph-pattern join (sq-7d3dj.30.3)

jeswr:
> 🤖 SPARQ agent — this PR was prepared by an automated sparq agent (bead sq-7d3dj.30.3).

## What

Sideways information passing (SIP): a **default-on, semantics-preserving** correlated
graph-pattern join. When `Join(A, B)` is evaluated and the already-evaluated side `A`
is **small** (≤ 64 rows), the executor evaluates the big child `B` **correlated** on
A's bindings instead of cold — for each distinct binding of A's variables that are
**certainly bound** in `B` and bound to an **IRI** in `A`, that IRI is substituted as a
constant into B's patterns (pushed *into UNION branches, BGP scans, property paths and
filters*), so a scan seeds from the constant instead of running blind. Results are
recombined preserving **multiplicity** (bag semantics).

Targets SP2Bench **q08 / q12b** (research/sp2bench-complex-shape-deficit.md §2.2): the
1-row `?erdoes` side now seeds the Union's `?document dc:creator ?erdoes` scan instead
of a blind whole-corpus creator self-join.

## Invariants (hard)

- **Bag-semantics result-equivalence, correlated on vs off** — `A ⋈ B = ⊎_C (A_C ⋈ B)`
  with `A_C ⋈ B = A_C ⋈ σ_{P=C}(B)`. Every pushed variable is a **certain** variable of
  `B` (sound under-approximation), so substituting the constant is exactly `σ_{P=C}(B)`
  with the pushed columns projected out (re-supplied by `A_C`). Pushed values are
  restricted to **IRIs** (term identity — no literal value-space / numeric-precision
  hazard, disjoint from sq-lr2ii).
- **Conservative fallback**: any scope/threshold failure (a variable not certain in `B`
  — e.g. inside an OPTIONAL right side, MINUS, or a projecting sub-`SELECT`; a `SERVICE`;
  `BOUND(?v)` on a substituted var) returns to the cold path bit-for-bit.
- **Anti-vacuity**: `tests/sip_join.rs::q08_shape_anti_vacuity` asserts (via the
  `sip_testing` stats + an EXPLAIN ANALYZE trace of the cold run) that correlated
  evaluation actually **fired** and the child row count **collapsed** vs the blind
  self-join — not just result equality.
- **Variable scoping**: validated by the full W3C SPARQL conformance suite (see below).

## Incidental fix (discovered by the mandated conformance run)

The **.30.2 top-k ORDER BY** path (PR #1733) took `k = OFFSET + LIMIT = 0` on
`ORDER BY … LIMIT 0`, and `select_nth_unstable_by(k - 1)` requires `k ≥ 1` — it
**panicked** (debug) / underflowed (release) on W3C `sparql10/solution-seq` "Limit 3"
(slice-03). `try_topk_orderby` now declines for `k = 0` so the full path slices to
empty. Regression test in `tests/orderby_limit_zero.rs` (disjoint from #1733's
`topk_orderby.rs`). **Flagged for #1733's owner** — this is a real latent bug in .30.2.

## Tests / gates

- `tests/sip_join.rs` — q08-shape on-vs-off DISTINCT + bag-multiplicity equivalence,
  anti-vacuity trace assert, duplicate-A-rows / duplicate-B-matches multiplicity,
  OPTIONAL-in-UNION + MINUS conservative-fallback, cross-product no-op, public toggle.
- `exec.rs::sip_unit` — direct unit tests for `certain_vars` (UNION intersection,
  OPTIONAL exclusion) and `subst_pattern` (BGP+filter substitution, `BOUND` bail).
- `tests/orderby_limit_zero.rs` — LIMIT 0 regression.
- **W3C SPARQL conformance: 1225 pass / 0 fail / 4 documented divergence** (up 1 — the
  LIMIT 0 test now passes).
- `cargo clippy --all-targets -D warnings` GREEN in **default**, **no-default-features**
  and **`service`** states; `cargo test -p sparq-engine` + `-p sparq-conformance` green;
  `cargo doc --workspace --no-deps --all-features` (`RUSTDOCFLAGS=-D warnings`) clean;
  `readme-template` 0 deviations.

No hard-coded performance numbers (work-box measurements are non-canonical; canonical
re-measure is bead sq-7d3dj.30.6).

## Stacking

Stacked on **#1733** (sq-7d3dj.30.2, same `crates/sparq-engine/src/exec.rs`). If #1733
merges first, this branch will be rebased onto `main` before merge.

Do NOT auto-merge (per brief).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


jeswr:
> 🤖 SPARQ agent — ESCALATED verdict (engine-correctness surface)

**Disposition: hold** (clean/sound review — held ONLY for stack ordering, not a defect).

I reviewed the isolated SIP commit (`git diff 2fed92d5e..6d9d65a7b`, i.e. this PR minus the stacked #1733 top-k content): `exec.rs` `certain_vars` / `subst_pattern` / `try_sip_join`, `union_bindings`, the Join arm, `tests/sip_join.rs`, and the design record §2.2.

**honest: true — sound_as_scoped: true.**

Key soundness findings:
- `certain_vars` is a genuine sound under-approximation. It does **not** repeat the sibling #1735 aggregate trap: `GraphPattern::Group { variables, aggregates }` keeps aggregate outputs in `aggregates`, and the arm reads only `variables` (group keys) ∩ inner-certain. Extend/BIND target excluded (can error → unbound); VALUES only when every row binds the column; UNION = branch intersection; LeftJoin/Minus = mandatory side only; Project = projected ∩ inner-certain. Every reported var is bound in every solution of the pattern.
- Substitution is IRI-only (`term_of` must be `NamedNode`); heterogeneous (IRI-in-some-rows, literal/blank-in-others) columns drop the whole column (`ok=false`), and blank/literal positions force fallback. `subst_pattern` returns `None` (→ cold, bit-for-bit) on Service/aggregate, `Project`/`Values` that bind a substituted var, `Extend` shadowing, and `BOUND(?v)` on a substituted var — scope-redefinition is covered.
- Bag multiplicity is correct: dedup for *evaluation* only (group by pushed-IRI key), recombine with the actual member rows (`a_c`) via `join_bindings`, and `union_bindings` is a pure multiset concatenation (no dedup).
- Differential oracle is real (`query`/`ask`/`explain_analyze`, not mocks): on-vs-off equivalence over DISTINCT + full bag, duplicate A-rows (4-row assert), duplicate B-matches, OPTIONAL-in-UNION fallback, MINUS fallback, ASK, plus an anti-vacuity trace assertion. The `sip_testing` toggle is `#[doc(hidden)]`, default-enabled, and cannot change production behaviour.

No ZK/MPC surface; no hard-coded perf (SKILL.md frames payoff as a hypothesis for the canonical host). No feature-gated intra-doc-link trap (`reset_stats` is not cfg-gated).

**Why hold, not arm:** the stack must merge in order #1733 → #1741; this branch carries #1733's content at a rebased SHA. The only red legs are `artifact-exact-equality (wasm bundle feature-OFF)` (being handled on #1733) and `run + track benchmarks` (non-correctness) — no correctness/test leg is red. The orchestrator should arm after #1733 merges and this rebases onto `main`.

jeswr:
> 🤖 **SPARQ agent** (orchestrator note for the stack-unwind steward)

GATE CHANGE landing ahead of you: PR #1749 (queued) replaces the static `feature_off_exact` pin with a dynamic base-vs-head byte-identity check + declaration file. Once it merges, if this PR's `artifact-exact-equality` leg fails after rebase, the fix is **NOT a re-pin** (the field is retired) — it is bumping `change_token` in `bench/feature-off-declaration.json` in a reviewed commit stating the cause (always-compiled SIP code). Same applies to #1747, #1745, #1748 downstream. The ±2% size ratchet is unchanged and still applies.

jeswr:
> 🤖 **SPARQ agent** — new feature-OFF declaration convention (heads-up for the exec.rs stack).

The feature-OFF wasm gate's declaration mechanism changed in **#1750** (mechanism **V2**, `sq-v3nel-v2`). The old scalar `change_token` in `bench/feature-off-declaration.json` had a serial-conflict flaw — every declaring PR edited the same line, so the first to merge made the rest CONFLICTING (that's why #1720/#1718 went DIRTY).

**New convention** — if a PR in this stack (#1741/#1747/#1745/#1748) intentionally changes always-compiled engine/core code such that the leg-2 `artifact-exact-equality (wasm bundle feature-OFF)` check reports the feature-OFF wasm bytes DIFFER from base, declare it by **adding one file** (do **not** bump the retired scalar token):

```
bench/feature-off-declarations/<this-PR-number>.json
```
```json
{ "pr": <PR-number>, "date": "2026-07-07", "reason": "one line on the always-compiled change" }
```

Different PRs add different files, so they never conflict regardless of merge order. If a PR does **not** move the feature-OFF bytes, do nothing — the byte-identity check passes automatically. During the transition window the legacy scalar-inequality path still works, but per-PR files are the going-forward path. Convention doc: `bench/feature-off-declarations/README.md` (once #1750 lands).

PR #1716, opened 2026-07-07
perf(substrate): LFTJ kernel micro-opts — branchless wrap, hoisted order indirection, k=3 fast path (sq-7d3dj.20)

jeswr:
> 🤖 SPARQ agent — implementing bead sq-7d3dj.20

## Summary

Three targeted micro-optimisations to `Leapfrog::search` in the WCO/LFTJ kernel
(`crates/sparq-substrate/src/join.rs`), the hottest single query-eval loop on cyclic/analytic
SPARQL queries. **No trait-surface change; results byte-identical** (all existing LFTJ correctness
tests pass + 3 new k=3-specific tests added):

- **(a) Branchless wrap**: `(p + 1) % k` / `(p + k - 1) % k` replaced with
  `p += 1; if p == k { p = 0 }` — compiles to `cmov` rather than an integer divide.
  Both `search` (generic path) and `next` updated.

- **(b) Hoisted double indirection**: `order[p]` resolved once per iteration into a `cur_idx`
  local. `Leapfrog.order` changed from `Vec<usize>` to `SmallVec<[usize; 8]>` — for k ≤ 8
  (all practical WCO arities) the order array lives inline in the struct on the stack, eliminating
  one heap allocation per LFTJ recursion level.

- **(c) Monomorphized k=3 fast path** (`search_k3`): new method with the order entries preloaded
  into a stack-local `[usize; 3]` array at entry — the hot loop never reads the `SmallVec` field.
  Branchless cyclic advance hardcoded for modulus 3. Dispatched from `search` when `order.len() == 3`
  (the triangle/3-clique case that dominates WCO workloads).

Also adds two new criterion micro-bench groups to `benches/substrate.rs`:
- `lftj_triangle` — triangle query on K_n (k=2 per level, cyclic LFTJ path)
- `lftj_intersection_k3` — 3-way set intersection (k=3, exercises `search_k3` directly)

## Measurement

**Canonical EC2 c6i.2xlarge results (instance i-094c34403ad54ce63, p<0.05 for all entries):**

| Bench group | Observed range |
|---|---|
| `lftj_triangle` (k=2) | −10.8% to −11.9% |
| `lftj_intersection_k3` (k=3) | −7.3% to −17.1% |

The k=3 path shows the widest improvement range, consistent with eliminating both `%` ops AND
the SmallVec field access from the hot loop on that monomorphized path.

*Non-canonical work-box figures were removed; canonical EC2 numbers above are the authoritative
[...]

PR #1743, opened 2026-07-07
perf(nt): parse-side subject/predicate intern cache in parse_chunk (sq-7d3dj.31.3)

jeswr:
> 🤖 SPARQ agent (sparq-rust-impl, Sonnet 4.6) — bead sq-7d3dj.31.3

## Summary

- Add a stack-buffer subject-and-predicate intern cache in `nt::parse_chunk` (crates/sparq-core) so consecutive same-subject/predicate N-Triples lines skip `hash_iri + HashTable::find` on a run-length hit
- Implements the spec from bead sq-7d3dj.31.3 (part of D4 bulk-load intern-cache program, profiling context in PR #1730)
- No feature flag — always on, same public API, same interning contract for all input orderings

**Spec satisfied:** sq-7d3dj.31.3 — cache last-seen subject/predicate (byte-slice → id) in `parse_chunk`; ungrouped input must produce identical dict and triple set as grouped input.

**Acceptance tests (all green):**
- `nt::tests::subject_run_cache_hits_produce_correct_ids`
- `nt::tests::blank_node_subject_cache_no_false_hit_on_prefix_match`
- `nt::tests::shuffled_input_store_matches_grouped`
- `nt::tests::oversized_subject_bypasses_cache_parses_correctly` (covers the >128-byte fallback path)

## Design

The cache stores a copy of the last-seen bytes in a fixed 128-byte stack array (`SUBJ_PRED_CACHE_LEN`), NOT a pointer back into the file buffer. This keeps both hits and misses L1-cache-friendly — a file-position-comparison approach was rejected because it caused L3 cache misses on shuffled (non-grouped) data (each miss reads from a random position in the potentially hundreds-of-MB input).

Correctness invariants: exact byte equality; blank-node boundary check (requires whitespace after matched bytes) prevents `_:b0` falsely matching `_:b0extra`; `NO_ID = 0` sentinel — `term()` never returns 0 so an empty cache never produces a false hit; tokens longer than 128 bytes fall through to normal `term()` with no caching overhead; id-assignment order is unchanged.

## Measurements (non-canonical work-box, 1M and 12.8M triple files)

| file | threads | OLD | NEW | delta |
|------|---------|-----|-----|-------|
| 1M grouped | 1 | 0.455s | 0.391s | -14.1% |
| 1M grouped | 8 | 0.122s | 0.108s | -11.5% |
| 1M shuffled | 1 | 0.692s | 0.670s | -3.2% (within noise) |
| 1M shuffled | 8 | 0.334s | 0.142s | -57.5% |
| 12.8M grouped | 1 | 9.131s | 7.817s | -14.4% |
| 12.8M grouped | 8 | 2.360s | 1.561s | -33.8% |
| 12.8M shuffled | 1 | 15.549s | 14.013s | -9.9% |
| 12.8M shuffled | 8 | 3.592s | 2.918s | -18.8% |

Shipping criterion: grouped 1-thread win ≥5% with no shuffled regression beyond noise — both satisfied.

## Gates

Feature flags: none (no feature gate — the cache is always-on, unconditionally safe).

Both states tested (default features = the only state):
- `cargo test -p sparq-core` — 101 tests passed
- `cargo clippy -p sparq-core --all-targets -- -D warnings` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --package sparq-core --no-deps --all-features` — clean
- `cargo test -p sparq-core --no-default-features` — 76 tests passed
- `cargo test -p sparq-core --features compact-index` — 101 tests passed

Also syncs `bench/parse/Cargo.lock` — `sparq-substrate` transitive-dep drift from sq-7d3dj.20 (#1716) materialised during bench verification.

## Test plan

- [ ] `cargo test -p sparq-core` — 101 tests pass
- [ ] `cargo clippy -p sparq-core --all-targets -- -D warnings` — clean
- [ ] No regression on shuffled input
- [ ] All four new acceptance tests pass

### DIFF
--- a/bench/feature-off-declarations/1743.json
+++ b/bench/feature-off-declarations/1743.json
@@ -1,0 +1,5 @@
+{
+  "pr": 1743,
+  "date": "2026-07-07",
+  "reason": "inline subject/predicate intern cache in always-compiled nt parse_chunk (+473 B feature-OFF, +0.034%; under the 2% bundle ratchet)"
+}
--- a/bench/parse/Cargo.lock
+++ b/bench/parse/Cargo.lock
@@ -1076,6 +1076,7 @@ dependencies = [
  "smallvec",
  "spargebra",
  "sparq-core",
+ "sparq-substrate",
  "uuid",
 ]
 
@@ -1099,6 +1100,16 @@ dependencies = [
  "flate2",
  "rayon",
  "zstd",
+]
+
+[[package]]
+name = "sparq-substrate"
+version = "0.1.0"
+dependencies = [
+ "oxrdf",
+ "rustc-hash",
+ "smallvec",
+ "sparq-core",
 ]
 
 [[package]]
--- a/crates/sparq-core/src/nt.rs
+++ b/crates/sparq-core/src/nt.rs
@@ -13,7 +13,7 @@
 //! triple path is unchanged: a triple term is only looked for when the object byte is `<`
 //! AND the next two bytes are `<(`, so a plain `<…>` IRI pays just two extra byte peeks.
 
-use crate::dict::{Dict, Id};
+use crate::dict::{Dict, Id, NO_ID};
 use std::borrow::Cow;
 
 /// [OPUS-4.8] (sq-7d3dj.2) Rough average N-Triples line length, in bytes, used ONLY to pre-size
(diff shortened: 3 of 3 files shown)


jeswr:
> 🤖 **SPARQ agent** (Fable verdict on the mechanical-verify escalation)

Mechanical verify passed everything except an **undeclared feature-OFF wasm delta (+473 B, +0.034%)**. Ruling: this is the intentional inline subject/predicate intern cache in always-compiled `parse_chunk` — a legitimate code-size change, not a `vectorized` leak, and far under the 2% bundle ratchet. Declared via `bench/feature-off-declarations/1743.json` (mechanism V2) and armed. Mutation spot-check on the cache-correctness test was red/green (non-vacuous).

PR #1770, opened 2026-07-07
ci(differential-fuzz): standing lanes — nightly per-category Oxigraph differential + blocking PR smoke + machine-readable divergence allowlist + corpus persistence (sq-0iqzw)

jeswr:
> 🤖 **SPARQ agent** — bead **sq-0iqzw** (standing differential-fuzzing lanes). [FABLE-5]

The proven wrong-answer-catching asset — the sparq-bench in-process Oxigraph 0.5 differential fuzzer (`crates/sparq-bench/src/fuzz.rs`, caught sq-eibog + sq-lr2ii) — ran in **zero** CI workflows. This PR gives it standing lanes, plus corpus persistence for the crash-oracle cargo-fuzz lane.

## 1. Nightly differential lane — `.github/workflows/differential.yml` (new)
- Cron **03:53 UTC** (stagger slot between ci.yml 03:17 and fuzz.yml 04:23) + `workflow_dispatch`.
- **12 shards** (fail-fast off): the 9 generator categories at 120 000 seeds each (running only `all` hides category-dense bugs — sq-lr2ii evidence), plus `SPARQ_FUZZ_MMAP` (20 000) / `SPARQ_FUZZ_COMPRESS` (120 000) / `SPARQ_FUZZ_DICTSPILL` (8 000) shards re-running a prefix of the same day window through the store-mode read paths (previously zero standing correctness differential).
- **Advancing window**: start = days-since-epoch × `WINDOW_STRIDE` (120 000) — exploration never replays; a guard fails the shard if a count ever exceeds the stride. Counts sized from measured local throughput so aggregate fuzz compute lands in the ~25–30 min band; they are declared tunables, not canonical numbers.
- **On failure**: uploads the log (containing the full deterministic seed+query+graph repro) as `differential-repro-<shard>`, then `scripts/ci-file-differential-failure.py` (new; hermetic `--self-test` runs first) files a **GitHub issue with the repro inline** (deduped per shard — comments the fresh window on an existing open issue instead of spamming) **and appends a P1 bug bead** to `.beads/issues.jsonl` (append-only, collision-checked derived id, deduped) with a **best-effort push** — the exact bead-autoclose.yml fail-soft pattern in reverse; a GH013 protected-main rejection is warned, not fatal, and the issue is the durable channel.
- Schedule + dispatch only → never produces a check-run on a PR head, so the `ci-summary / gate` aggregator never sees it (same isolation as shacl-diff-fuzz.yml).

## 2. PR-level BLOCKING differential smoke — `differential-smoke` job in `fuzz.yml`
- Fixed regression windows: `0..2000 all`, the **sq-lr2ii known-catch windows** `4695..6895 equality` + `filter`, and the `0..2000` window under mmap/compress (+ `0..500` dict-spill). ~32 s of fuzz compute measured locally (build dominates the job).
- Gated on the ci-select closure via a new `_LANE_SEEDS["differential-smoke"] = [sparq-core, sparq-engine, sparq-bench]` (mirrored in `ci/path-ownership.toml [lanes]`, pinned by new wiring tests in `scripts/tests/test_ci_select_wiring.py` + the updated lane-set pin in `scripts/tests/test_ci_select.py`). Fail-closed disjunct + `needs: select`, exactly like the existing fuzz job.
- Job name carries no advisory/informational token → **aggregated as REQUIRED by `ci-summary / gate`** (and runs on `merge_group`). The sq-lr2ii class can no longer merge silently.

## 3. Machine-readable divergence allowlist — `bench/differential-divergences.json` (new)
The two ADJUDICATED cross-engine divergence classes, previously living only in comments/bead notes, are now data the comparator **consumes** (`DivergenceAllowlist` in fuzz.rs):
```json
{ "classes": [
  { "id": "cross-family-eq-type-error", "bead": "sq-eibog", "engines": ["oxigraph"],
    "spec": "SPARQL 1.1 §17.3 + §17.4.1.7", "adjudication": "...", "handling": "re-derive spec count; absorb only when sparq == spec" },
  { "id": "bnode-iri-inequality", "bead": "sq-ai2wa", "engines": ["qlever"],
    "adjudication": "...", "handling": "skip-with-count (inert for this generator — no bnodes emitted)" } ] }
```
- Known-class mismatch → **skip-WITH-count** (`adjudicated(cross-family-eq)=N` in the summary line); anything else fails. Removing an entry re-enables the strict differential; unknown ids warn + stay strict; missing/malformed file = fully STRICT (fail toward flagging).
- sq-eibog keeps its **stronger-than-skip** handling: the spec-correct count is re-derived independently and a residual disagreement still fails. sq-ai2wa's detector is deliberately narrow (bnode-vs-IRI `=`/`!=` only) and provably inert for today's generator — encoded so generator/oracle growth surfaces as an adjudicated skip, not a false positive. Classes were NOT widened beyond the adjudications.
- New unit tests pin JSON ids ⟷ detectors (an entry without a detector fails the test), strictness fallbacks, and detector narrowness.

## 4. cargo-fuzz corpus persistence (fuzz.yml `fuzz` job)
- `actions/cache/restore` + always-unique key `fuzz-corpus-v1-<run_id>` with a `fuzz-corpus-v1-` restore-keys chain; paired `cache/save` with `if: always()` (a crashing run grew exactly the corpus worth keeping), saved only from main-scoped runs (PR-branch caches are unreachable churn; PRs still restore main's chain). One cache spans `fuzz/corpus` for all targets — `cargo fuzz list` auto-discovers targets, so a hard-coded per-target key list would silently miss new targets (deviation from the literal per-target-key suggestion, for that reason).
- **Committed minimal seed corpus** `fuzz/seeds/<target>/` (tracked; `fuzz/corpus` stays git-ignored) passed as a second corpus dir: valid NT/NQ/Turtle/TriG docs with the format-selector first byte, rich SPARQL (paths/aggregates/subquery/VALUES/GRAPH), SHACL shapes (SELF + SPLIT modes), store-corruption control bytes.
- `fuzz/Cargo.lock` refresh: the out-of-workspace lockfile was stale vs main (sparq-engine now depends on sparq-substrate); the real `cargo fuzz build` updated it.

## Local verification (all on this branch)
- `cargo test -p sparq-bench` — 15/15 green (12 existing + 3 new); `cargo clippy -p sparq-bench --release -- -D warnings` clean.
- **Allowlist non-vacuity**: `fuzz 0 2000 all` → exit 0, `adjudicated(cross-family-eq)=42` (the exact historical sq-eibog density); `SPARQ_FUZZ_DIVERGENCES=/dev/null` (strict) → **exit 1**, `full_mismatch=42`, first seed 15. The file is provably load-bearing.
- **Comparator non-vacuity**: deliberately broken comparator (`!= oxi + 1`) → exit 1; reverted.
- **Smoke windows**: all six smoke invocations green end-to-end (~32 s wall locally).
- **Tonight's exact nightly windows**: all 12 shards run locally at the same start the cron will compute — all green (first scheduled run should be green).
- **Seed corpus + explicit-corpus-dir wiring**: `cargo +nightly-2026-06-11 fuzz build` + `cargo fuzz run <t> corpus/<t> seeds/<t> -- -runs=0` for all 5 targets on the host triple — seeds execute cleanly with real coverage. (CI builds x86_64; local validation was on the aarch64 host triple — the wiring is triple-independent.)
- `python3 scripts/ci-file-differential-failure.py --self-test` green (log parsing, id derivation/collision, append-only JSONL, dedupe, issue body).
- YAML valid (PyYAML) + `actionlint` clean on both workflows; `test_ci_select_wiring.py` 33/33, `test_ci_select.py` 67/67, `test_ci_summary_gate.py` 33/33.

## Honest gaps (documented-untested)
- The failure auto-file path (gh issue create + bead push) cannot be fully exercised locally — logic is covered by the hermetic self-test (which also runs first in CI before the real invocation); the **bead push to protected main is EXPECTED to be declined (GH013)** until a ruleset bypass exists (sq-roe3) — fail-soft warn, mirroring bead-autoclose's documented behaviour; the GitHub issue is the reliable channel (orchestrator issue-sweep beads it).
- Nightly shard budgets are sized from this work box's throughput (non-canonical); first real runs may need count tuning — they are env-level constants with an explicit no-replay guard.
- `ci-summary / gate` impact: differential.yml is structurally invisible to the gate (schedule/dispatch only); `differential-smoke` becomes a new REQUIRED-aggregated leg via the existing name rule — this PR touches `.github/**`, so ci-select fail-closes to mode=full and the new leg exercises itself here.

**Compliance framing (honest level):** standing differential-testing evidence for the wrong-answer bug class (test-oracle coverage), plus corpus continuity for the coverage-guided crash oracle — this strengthens the fuzz/oracle coverage matrix; it is not a formal-verification or soundness claim.

Follow-up bead (created, not implemented): **sq-j80vk** — generator surface expansion (property paths, aggregates/GROUP BY, subqueries, EXISTS, VALUES, BIND, GRAPH).

Do NOT arm note: per the task brief this PR is left un-armed for orchestrator review.


jeswr:
> 🤖 **SPARQ agent** [FABLE-5]

Fixed the mechanical-verify BOUNCE from re-review: the `ci-scripts` gate failed `test_path_ownership.py::test_real_workspace_audit_passes` because `crates/sparq-bench/src/fuzz.rs` (the new `DivergenceAllowlist`) reads the committed `bench/differential-divergences.json` via a `CARGO_MANIFEST_DIR`-relative path, which the out-of-crate-input audit requires to be attributed.

**Fix (commit 4764feaf3):** added the `[[map]]` entry to `ci/path-ownership.toml`:
```toml
[[map]]
pattern = "bench/differential-divergences.json"
crates = ["sparq-bench"]
```

Reproduced locally: `scripts/ci_audit_inputs.py` now reports `sparq-bench -> bench/differential-divergences.json [ownership map]`, **0 findings**; `test_path_ownership.py` 28/28; `test_ci_select_wiring.py` 33/33; `test_ci_select.py` 67/67. Ready for re-review. Not self-armed.

PR #1778, opened 2026-07-08
perf(cli/server): light the landed algebra-rewrite pass in the shipped binaries + conformance/diff-fuzz harnesses (sq-7d3dj.30.13) [FABLE-5]

jeswr:
> 🤖 SPARQ agent

## What (bead `sq-7d3dj.30.13`)

The #1735 pre-execution **algebra-rewrite** pass (result-equivalent `FILTER(?v = <iri>)` IRI-constant substitution + `FILTER(!bound)` anti-join, `sq-7d3dj.30.1`) was opt-in and **dark in every shipped and measured build**: `sparq-cli` enabled only `dp-planner`, and every canonical bench/gather script builds default features. This PR lights the pass where it ships and where it is validated — while keeping the **sparq-engine LIBRARY default OFF** (opt-in feature architecture: library consumers choose explicitly).

- **`sparq-cli`**: `algebra-rewrite` added to the `sparq-engine` dep features (same mechanism as `dp-planner`, sq-7d3dj.30.5). Lit unconditionally for the CLI, including the allocator-A/B `--no-default-features` build (so allocator A/B keeps the rewrite constant).
- **`sparq-server`**: new `algebra-rewrite = ["sparq-engine/algebra-rewrite"]` forwarding feature, added to `default`. Toggleable (`--no-default-features --features server,jsonld` = explicit rewrite-dark binary); the pure-serialiser library build (`--no-default-features`, what sparq-cli consumes) is unaffected.
- **`sparq-conformance`** (`publish = false` harness): engine dep lights the pass, so the **W3C conformance ratchet now permanently validates the shipped configuration** — a rewrite that broke a W3C test can no longer merge green.
- **`sparq-bench`** (`publish = false` harness): engine dep lights the pass, so the **nightly sparq-vs-oxigraph differential fuzzer oracle-checks the rewrite-ON path** — its `equality` category generates the exact `FILTER(?v = <iri>)` trigger shape. (Honest boundary: rewrite (b)'s `FILTER(!bound)` trigger is not in the generator's grammar; its coverage remains `tests/rewrite_pass.rs` via the `sparq-engine (algebra-rewrite)` feature-matrix leg.)
[...]

PR #1718, opened 2026-07-07
perf(engine): expression-variable → column-index hoist in filter/BIND/ORDER BY (sq-7d3dj.4)

jeswr:
> 🤖 SPARQ agent — sq-7d3dj.4 expression-variable → column-index hoist

## What

Walk each `Expression` **once per operator** (in `compile_expr`) resolving `Variable(v)` nodes to pre-resolved column indices stored in a new `CompiledExpr::Var(Option<usize>)` enum. The per-row hot path in `apply_filter_scalar`, `extend_bindings`, and `order_bindings` then indexes `row[c]` directly, eliminating O(N·K·W) string comparisons where:
- N = result rows
- K = variable refs per expression  
- W = binding width (vars in operator's schema)

Also extracts `eval_function_inner<E: Fn(usize) -> Result<Value, String>>(f, nargs, ev)` to share the 400-line SPARQL function body between `eval_function` (original path) and `eval_compiled` (compiled path), avoiding duplication.

This is a **mechanical hoist, NO semantic change**. `CompiledExpr` is the substrate for the later compiled-expression bead (sq-7d3dj.11).

## Semantic invariant

`eval_compiled` is byte-identical to `eval_expr` — pinned by 4 new unit tests in `compiled_expr_tests`:
- `compile_expr_matches_eval_expr_for_all_variants` — 17 expression types, direct comparison
- `compiled_filter_is_byte_identical_to_eval_expr` — 30-row bindings, 4 filter expressions
- `compiled_bind_is_byte_identical` — full BIND query via `crate::query`
- `compiled_order_is_byte_identical` — ORDER BY ASC query via `crate::query`

## Measurement

**Canonical EC2**: c6i.2xlarge (8 vCPU), eu-west-2, quiet dedicated instance i-06c742fff0085d9eb (Project=sparq-bench, TERMINATE-confirmed). Corpus: 10 000 entities (79 991 triples), 5 iterations, count mode. Same corpus file used for both before/after.

| query | rows | before (µs) | after (µs) | delta |
|---|---|---|---|---|
| q09_filter_numeric | 375 | 642.9 | 548.5 | **−14.7%** |
| q10_filter_string | 2 | 52 378 | 52 537 | +0.3% (flat — regex dominates) |
| q11_filter_in | 3 | 1 228.6 | 1 239.8 | +0.9% (flat — 3-row result) |
| q12_filter_exists | 3 | 2 688.0 | 2 707.4 | +0.7% (flat — EXISTS uses b.col for shared vars) |
[...]

PR #1779, opened 2026-07-08
perf(engine): equality-FILTER → value-join unification behind opt-in value-join feature (sq-7d3dj.30.7)

jeswr:
> 🤖 SPARQ agent

## Equality-FILTER → value-join unification (opt-in `value-join` feature, sq-7d3dj.30.7)

SP2Bench q05a/q12a's star-intersection shape — variable-disjoint pattern components glued only by `FILTER(?a = ?b)` — currently materialises the near-cross product and evaluates `=` per row. With the **opt-in, OFF-by-default** `value-join` feature, the equality executes as a **value-keyed hash join** between the components instead.

### Why not an id-keyed join

SPARQL `=` is VALUE equality: `"01"^^xsd:integer = "1.0"^^xsd:decimal` is TRUE across two dictionary ids, incompatible types are a type error (row eliminated), and high-precision `xsd:decimal`s that collapse to one `f64` must stay unequal (the sq-lr2ii class). Soundness is a two-layer construction:

1. **No false negatives** — the per-term-class join key (`JKey`) mirrors the evaluator's own dispatch: dictionary id for IRI/bnode/unknown-datatype classes, the `numeric_value` cache's canonicalised `f64` for numerics (`-0.0` folded), value for `xsd:string` / language-tagged / boolean. Rows with **no provable key** (local-vocab ids, RDF 1.2 triple terms, value-comparable temporals, and numeric-datatype lexicals the cache rejected) take a `Hard` class paired through the EXACT evaluator (`equal_expr`) — today's semantics for exactly those rows.
2. **False positives rechecked** — the pass never consumes the FILTER: every original filter expression re-applies verbatim over the joined rows, so key collisions are eliminated by the same sq-lr2ii exact machinery that runs today.

Anything not provably eligible **declines** to the verbatim plan: non-total filters (any function call), quoted-triple patterns, same-component equalities, an armed query budget (mirrors `columnar_filter` I3), an armed `zk` proof trace (mirrors I2), no cross-component equality conjunct. Feature OFF ⇒ zero of this code compiles and the conjunctive path is byte-identical (the only default-path change is the verbatim extraction of `eval_flat_conjunctive`). No new dependencies, no `unsafe`, no SP2Bench special-casing.

### Review of the inherited commit (resumed crashed agent)

This PR resumes a crashed agent's worktree. Reviewing its commit adversarially found — and red-test-first fixed — one real soundness gap: the numerics cache (`str::parse::<f64>`, no trim) is **stricter** than the evaluator's `values_equal` fallback (`Num::of_literal`, which trims), so a whitespace-padded `" 1"^^xsd:integer` was keyed by term identity while the verbatim path pairs it with `"1"^^xsd:integer` by value — the hash join silently dropped the row (test `whitespace_padded_numeric_pairs_by_value` demonstrated the divergence before the fix). Cache-missing numeric-datatype literals now key `Hard`.

### Tests (non-vacuity verified by engine mutation)

Every witness runs the SAME query through the value-join path (asserting a `FIRED` anti-vacuity counter) and through a kill-switch verbatim run, and the bags must match: numeric promotion across datatypes, type-error elimination, sq-lr2ii high-precision decimals, plain-literal vs `xsd:string`, language tags, temporal Hard-class, booleans, whitespace-padded numerics, plus decline witnesses (same-component eq, eq under `||`, function-call filter, armed budget) and a q05a-shaped acceptance suite vs an independent `!(?a != ?b)` oracle (`tests/eqjoin_value_join.rs`). Mutation spot-checks: id-keyed numerics → 3 tests red; residual recheck removed → sq-lr2ii test red; boolean id-key → 2 tests red. (One neutral survivor: un-lowercasing the lang-tag key is unobservable because oxrdf normalises tags at parse; the lowercase key is kept as defensive depth.)

### Gates

- `cargo clippy --workspace --all-targets -- -D warnings` — green (default) and green with `--features value-join` (`-p sparq-engine --all-targets`).
- `cargo test -p sparq-engine` — green default (feature OFF) and green with `--features value-join` (16/16 test binaries).
- W3C SPARQL conformance ratchet (pinned suite, local): **1225 pass / 0 fail / 4 documented divergence = 1229 ≥ 1229 floor**.
- Feature combos compile: `value-join,zk` and `value-join,yannakakis,vectorized`.
- Feature-matrix leg `opt-in sparq-engine (value-join)` added (+ golden line, `test_feature_matrix_assemble.py` green); README ≤120 lines, `check-readme-template.py --enforce` 0 deviations; SKILL.md synced.
- Coverage: sparq-engine is coverage-measured with default features, so the feature-gated module is not instrumented; the always-compiled `eval_flat_conjunctive` extraction is exercised by every conjunctive query. Direct unit tests cover each new helper (`key_of`, `expr_total`, `top_conjuncts`, `as_var_eq`, `expr_vars`, `pattern_components`).
- A `bench/feature-off-declarations/<PR>.json` follows in the next commit (the new cargo feature rehashes crate metadata → expected content-only feature-OFF bundle delta).

### Local diagnostic timing (work box, NON-canonical, min of 2)

Canonical SP2Bench q05a (`queries-heavy/q05a.rq` text), value-join ON vs the shape-ineligible result-equivalent `!(?name != ?name2)` spelling (= the verbatim plan) on the real `sp2b_gen` corpus:

| corpus | value-join | verbatim | rows (identical) |
|---|---|---|---|
| 250k triples | 0.031 s | 53.7 s | 6933 |
| 1M triples | 0.118 s | (not run to completion) | 34870 |

### Honest scope boundaries

- An **armed query budget declines** the pass, so budgeted server queries do not currently benefit — same posture as the vectorized seam (I3); lifting it needs a plan-shaped debit argument (candidate follow-up bead).
- The pass keys only **top-level** `?a = ?b` conjuncts; equalities under `||`/`IF`/negation stay on the verbatim plan by design.
- q12a (ASK twin) benefits through the same group path (ASK short-circuits further).

Bead: sq-7d3dj.30.7.


jeswr:
> 🤖 SPARQ agent

Merged `origin/main` (including #1718's compiled-expression hoist in `exec.rs` — auto-merged cleanly; `equal_expr` and `apply_filter` are untouched by it, the compiled path is additive and asserted byte-identical by its own suite). Post-merge re-verification: clippy `-D warnings` green in both feature states, full `sparq-engine` tests green in both states, the 22 `exec::eqjoin` differentials + the q05a acceptance suite green, rustdoc `--all-features -D warnings` clean. One README conflict resolved (kept both the #1778 algebra-rewrite wording and the new value-join bullet; still 120 lines, `check-readme-template.py --enforce` 0 deviations).

jeswr:
> 🤖 SPARQ agent

**Escalated-tier verdict (engine-correctness / novel-algo surface): ARM.**

Verified diff-scoped: (1) the JKey no-false-negative property is structurally closed per term class — numerics can only avoid `Hard` via a `numeric_value` cache hit, and that cache is the same source `eval_numeric`'s fast path compares, so evaluator-TRUE ⇒ equal keys (NaN is the cache's miss sentinel → `Hard`; −0.0 folded); `as_bool_val`'s lexical set matches `key_of` exactly (no trim either side); `Temporal::of_lit` delegates to the identical `Timeline::parse_*` the evaluator's fallback uses, so all evaluator-pairable temporals are cache hits → `Hard`; OtherXsd equality ⇒ one interned id ⇒ `Term(id)` exact. (2) Join equalities can never be component-local, so they always re-run verbatim as residual filters — the sq-lr2ii exact recheck is inherited, and the witness test asserts only the exact-value pair survives f64 collisions. (3) The three join loops partition the product (no dup/missed pair); `pair_true` is `equal_expr` itself. (4) Decline witnesses (zk/budget/same-component/disjunction/function-call) assert the FIRED counter; on/off kill-switch differentials + an independent shape-ineligible `!(?a != ?b)` oracle make the tests non-vacuous. (5) Feature OFF: `eval_flat_conjunctive` extraction is verbatim vs the removed inline block; matrix leg + golden added so gated tests run in CI. Perf numbers correctly labeled non-canonical work-box.

Nit (non-blocking): PR body says the feature-off declaration "follows in the next commit" but `bench/feature-off-declarations/1779.json` is already in this diff.

Armed with `--auto`; merge waits on the pending gates.

PR #1781, opened 2026-07-08
perf(engine): ASK / LIMIT-k first-solution early exit through joins, unions, optionals (sq-7d3dj.30.8)

jeswr:
> 🤖 SPARQ agent

## What & why

SP2Bench **q12a** (the ASK form of q05a) cost *exactly* its SELECT twin (~65 s
diagnostic vs QLever's 7.6 ms): the engine evaluated the full solution set and
then checked non-emptiness. ASK semantics is boolean *existence* — evaluation
must stop at the **first complete solution**.

This lands `sq-7d3dj.30.8`: a **default-on, semantics-preserving** first-solution
early exit that reuses the existing `LIMIT` machinery (`try_capped`, now threaded
with the query-local vocab). ASK is evaluated under an implicit `LIMIT 1`, and any
`LIMIT k` (with no ORDER BY / DISTINCT / aggregation above it) now stops early for
JOIN shapes, not just single-pattern scans:

- **conjunctive BGP (+ FILTERs)** — `eval_bgp_binary_capped` drives the same greedy
  GOO binary-join chain from growing seed-scan blocks (1024 → 64 k → rest), applying
  residual FILTERs per block. **A row counts toward the cap only after *every*
  FILTER**, so a first candidate eliminated by a late FILTER never fires the exit.
- **UNION** — left branch first; the right branch is skipped when the cap is met
  (header aligned via the in-scope variable set).
- **OPTIONAL** — caps the left side (OPTIONAL never drops a left row).
- **Join(A, B)** — capped-A probe through the SIP correlated path, declining to the
  full path when the probe misses.
- **BIND** — row-count preserving, capped inner.

Under ASK the plan is additionally simplified where **provably emptiness-neutral**
(`ask_simplify`): ORDER BY / DISTINCT stripped along the top modifier spine —
**never below a `Slice`** (DISTINCT under OFFSET changes the visible count) and
**never touching Group/HAVING** (an empty-group aggregate still yields a solution).

**Fail-closed:** cyclic (LFTJ) BGPs, quoted-triple patterns, MINUS, property paths,
GRAPH, non-conjunctive FILTER groups, and an armed zk-trace recorder all keep the
full path bit-for-bit.

## Invariant

Identical boolean result to the unoptimized path on **every** query. The oracle is
each query's LIMIT-less SELECT twin (never takes a capped path); ASK == non-emptiness
of the twin, asserted differentially across joins / OPTIONAL / UNION / FILTER /
aggregation / HAVING / ORDER BY+LIMIT / MINUS / paths / VALUES / GRAPH.

## Review of the inherited commit (`30d84be26`)

I inherited this branch from a crashed agent — one unverified commit. Reviewing it
as a stranger's diff surfaced one real gap:

- **The flagship budget revert-witness was VACUOUS.** Its filter-free conjunctive
  body is answered *exactly* by the `try_count` index-count pushdown, so it never
  reached `try_capped`. **Verified by mutation:** compiling `eval_bgp_binary_capped`
  out left every test green. Repaired: the witness body now carries an all-pass
  residual `FILTER(?x != ?y)` (declined by `count_pushdown`; the q12a evidence shape
  is a FILTERed join anyway), plus two new budget witnesses through the capped
  **UNION** and **OPTIONAL** arms. All three now go **RED** under mutation.
- **complete-solution guard (late FILTER):** present and non-vacuous — a candidate
  eliminated by a late FILTER must not terminate. Mutation-verified.
- **aggregation/HAVING preserved:** `ask_simplify` never touches `Group`; the
  DISTINCT-under-OFFSET strip-guard is mutation-verified (recursing through `Slice`
  turns it red).
- Added: LIMIT + late-FILTER contract test, and fail-closed differentials for cyclic
  BGP / MINUS / property path / VALUES / GRAPH.

- `cargo clippy -p sparq-engine --all-targets -- -D warnings` — default **and**
  `--all-features`: clean.
- `cargo test -p sparq-engine` — default **and** `--all-features`: green.
- W3C SPARQL conformance suite: **1225 pass / 0 fail / 4 documented divergence** (the
  suite's ASK tests included).
- `sparq-engine` line coverage **90.94%** (≥ floor 90).
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features`: clean.
- feature-OFF wasm bundle: **1 555 447 bytes** vs floor 1 540 474 (+0.97%, within the
  ±2% band). The always-compiled `try_capped` / `eval_bgp_binary_capped` /
  `ask_simplify` / `in_scope_vars` additions move the bytes — declared in
  `bench/feature-off-declarations/`.

## Measured locally (NON-canonical — work-box, diagnostic only)

Synthetic q12a shape (20 k + 20 k, `--release`), from the `q12a_shape_timing_probe`:

| | time |
|---|---|
| SELECT twin (= the pre-fix ASK cost) | ~514 000 ms (1 072 474 rows) |
| ASK (this PR) | ~5 500 ms |

Work-box timings are never canonical; the payoff class is real, the number is not
baked into any doc/test.

## Honest boundaries

- An ASK whose answer is `false` still sweeps the whole seed (bounded ~3× the
  single-pass scans by the three-block schedule).
- The right side of OPTIONAL is still evaluated in full (v1: no correlated OPTIONAL).

Resumed from a crashed agent's branch; merged `origin/main` (incl. #1718's
expression-compilation hoist and #1778) before pushing.


jeswr:
> 🤖 SPARQ agent

Resumed this from a crashed agent's single unverified commit (`30d84be26`, clean tree, unpushed). Nothing had been verified, so I ran the full gate matrix from scratch and reviewed the inherited diff as a stranger's.

The one substantive finding: the flagship budget revert-witness was **vacuous** — its filter-free conjunctive body is answered by the `try_count` index pushdown and never reached the capped join chain (`eval_bgp_binary_capped`). Confirmed by mutation (compiling the chain out left it green). Repaired to force the capped path (all-pass residual FILTER) and added budget witnesses through the capped UNION and OPTIONAL arms; all three now go red under the same mutation. The DISTINCT-under-OFFSET strip-guard and the late-FILTER complete-solution guard were mutation-verified as-is.

Merged `origin/main` (incl. #1718's expression-compilation hoist and #1778) before pushing. A follow-up commit adds `bench/feature-off-declarations/1781.json` — the always-compiled capped-chain additions move the feature-OFF wasm bundle +0.97% (within the ±2% band), so the byte-identity gate needs the declaration.

**Not arming auto-merge** — leaving for review.

jeswr:
> 🤖 SPARQ agent

**Escalated-tier verdict: ARM.** Diff-scoped review of the full 999-line diff + `tests/ask_early_exit.rs` + the PR-head `exec.rs` capped chain, against the sq-7d3dj.30.8 invariants:

- **Complete-solution guard** — residual FILTERs applied per block via the shared `apply_filter` *before* the cap check; late-FILTER witness has a beyond-first-block passing row + a false-case control that reddens any pre-FILTER exit.
- **`ask_simplify`** — strips ORDER BY/DISTINCT/REDUCED only along the Project/Union modifier spine; `Slice` and `Group` kept verbatim (DISTINCT-under-OFFSET flip-witness + empty-group aggregate/HAVING tests pin it).
- **Fail-closed** — cyclic/quoted-triple/MINUS/paths/VALUES/GRAPH all decline to the full path with true-AND-false oracle differentials (wrong-answer-detecting); zk-recorder + empty-default-view declines retained in both `try_capped` and `try_count`.
- **Witness non-vacuity** — verified on the head that `count_pushdown` only answers single-pattern filter-free BGPs, so the all-pass residual `FILTER(?x != ?y)` genuinely forces `eval_bgp_binary_capped`; the UNION/OPTIONAL witnesses likewise bypass `count_leftjoin`. The twin-trips-budget/ASK-does-not construction is self-checking.
- **LIMIT/OFFSET** — cap = `start.checked_add(len)` with the sub-multiset/complete contract + `slice_bindings`, so no post-limit early exit under nonzero OFFSET; capped arms reuse the full path's operators (`try_sip_join`/`left_outer_join`/`extend_bindings`/`union_bindings`).
- **Honesty** — perf numbers non-canonical + PR-body-only, timing probe `#[ignore]`d, wasm feature-off declaration present (+0.97% in band), W3C ratchet consistent with 1225/0/4, no ZK overclaim.

All completed CI legs green, remainder pending — arming with auto-merge; the gate still independently guards the merge.

PR #1782, opened 2026-07-08
perf(engine): anchor-driven + clipped DISTINCT-projection semijoin — SP2Bench q09 (sq-7d3dj.30.10) [FABLE-5]

jeswr:
> 🤖 SPARQ agent — implementing bead `sq-7d3dj.30.10`.

## What

Extends the DEFAULT-ON DISTINCT-projection loose skip-scan (from `sq-7d3dj.30.4`) so the
**anchor + probe** branch of `SELECT DISTINCT ?p WHERE { … }` no longer scans every
candidate `?p`-block against a freshly-rebuilt anchor hash set. Three changes, all inside
the always-compiled executor, all **result-set identical** to the prior plan:

1. **Cross-branch anchor cache.** A UNION's branches frequently share the same anchor
   (SP2Bench q09's two branches both anchor on `?person rdf:type foaf:Person`). The sorted
   anchor membership set is now built **once** per `(prepared-pattern, join-position)` and
   reused (`Rc`), instead of rebuilt per branch (the 2nd 20k-member `HashMap` build was a
   measurable slice of the profile).

2. **Two anchor-cardinality-selected semi-join strategies** (both exact, chosen only to
   minimise work):
   - **small anchor** → per-member *pattern trick*: bind the join column to each anchor
     member, skip-enumerate its few `?p`, stop once the `?p` universe is covered;
   - **large anchor** → per-`?p`-block existence scan whose block is first **clipped to the
     anchor's id window** (two binary searches) and then intersected by galloping the
     **shorter** of block/anchor (O(1)-hash gallop of the block's distinct values, or a
     monotonic-window binary-search walk of the smaller sorted anchor). The O(1)
     id-range-disjointness fast reject still runs first.

3. Anything not matching the exact single-variable DISTINCT shape executes the **identical
   pre-existing plan** — the clean DECLINE path is preserved bit-for-bit (verified by the
   existing fallback tests: REDUCED, multi-var projection, filters, 3-pattern branches,
   `?p`-as-join-var, repeated-var `?x ?p ?x`, RDF 1.2 quoted-triple terms, empty-view, zk
   armed).

**Honest scope note:** the per-member *pattern trick* is the right tool only when the join
axis is low-cardinality; for q09 the distinct join values (subjects/objects) number in the
tens of thousands, so the **large-anchor clipped block scan** is what wins there. Rather
than force a path that doesn't help, the strategy is cost-selected by anchor cardinality —
q09 takes the block scan; a small-anchor shape takes the pattern trick. Both are
differentially verified.

## Feature flags

**No new feature** — this is an unconditional, semantics-preserving optimisation under the
existing DEFAULT-ON `distinct_pushdown` (matching siblings `.30.4`/`.30.7`). It changes
always-compiled default-path bytes, so a `bench/feature-off-declarations/<PR>.json` is
included (the feature-OFF wasm INVARIANT is preserved — bytes just move).

## Gates (work-box)

- `cargo clippy -p sparq-engine --all-targets -- -D warnings` — clean.
- `cargo test -p sparq-engine` (default features) — **277 + siblings pass, 0 fail**.
- `cargo test -p sparq-engine --features zk,algebra-rewrite,semijoin-bitmap,yannakakis --test distinct_pushdown` — 15 pass (with `zk` the pushdown declines → fallback, still equivalent).
- W3C SPARQL conformance ratchet: **1225 pass / 0 fail / 4 documented divergence** → pass+divergence = 1229, 0 fail (unchanged).
- `scripts/coverage.sh` (sparq-engine) + `coverage-gate.py --check`: **90.46%** ≥ floor 90 → OK.
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — clean (no intra-doc link to a private item; new doc-comments use code spans — the feature-gated-link trap avoided).
- Full sp2b light suite row counts match `expected-rows.tsv` (q09=4, q02=6067, q03a=15823, q05b=6933, …).

## Tests

`crates/sparq-engine/tests/distinct_pushdown.rs` (differential on-vs-off is the load-bearing
invariant on **both** strategies):
- `large_anchor_q09_block_scan_equivalence_and_collapse` — 1000-person q09 (block-scan path): on==off, exact set, anti-vacuity collapse.
- `small_anchor_pattern_trick_equivalence` — 12-book anchor (pattern-trick path): on==off, exact set, pushdown fires.
- `block_scan_anchor_driven_side_equivalence` — 300-person / 6000-doc (block scan's anchor-driven intersection side): on==off, `ex:bulk` excluded.
- `randomized_differential_both_strategies` — 40 random graphs straddling the strategy threshold: on==off.
- `slice_over_distinct_preserved` — LIMIT/OFFSET (Slice above Distinct) retained-count identical on-vs-off.
- Mutation-checked: flipping an expected count (5→6) turns a test red (verified locally).

## SP2Bench q09 (250k triples) — NON-canonical work-box diagnostic

Same box / dataset / `sparq-cli bench` harness, release build, before = `origin/main`:

- before: ~5.16 ms/query
- after:  ~4.52 ms/query  (~12.5% faster)

The hotspot (`probe_anchor_distinct` inner existence scan) shrinks; the remaining cost is
the inherent examination of the large no-hit value-typed-predicate blocks (no characteristic
sets today). Work-box timings are diagnostic only — not baked into any committed markdown or
test. The canonical-host payoff is a hypothesis for `sq-0g6g`.

### DIFF
--- a/bench/feature-off-declarations/1782.json
+++ b/bench/feature-off-declarations/1782.json
@@ -1,0 +1,5 @@
+{
+  "pr": 1782,
+  "date": "2026-07-08",
+  "reason": "sq-7d3dj.30.10: exec.rs adds always-compiled anchor-driven + clipped DISTINCT-projection semijoin machinery (collect_var_ids_sorted, AnchorSet/AnchorCache/cached_anchor_ids, block_intersects_anchor, maybe_anchor_driven/anchor_driven_distinct, clipped block scan) under the default-on distinct_pushdown — moves the feature-OFF sparq-wasm bundle bytes; feature-OFF invariant preserved"
+}
--- a/crates/sparq-engine/src/exec.rs
+++ b/crates/sparq-engine/src/exec.rs
@@ -2316,9 +2316,15 @@ fn try_distinct_pushdown(
     let mut seen: FxHashSet<Id> = FxHashSet::default();
     let mut out_ids: Vec<Id> = Vec::new();
     let mut total_scanned = 0usize;
+    // [FABLE-5] (sq-7d3dj.30.10) Cache the sorted anchor membership set across branches:
+    // UNION branches frequently share the same anchor pattern (SP2Bench q09's two branches
+    // both anchor on `?person rdf:type foaf:Person`), so building the ~20k-member set once
+    // instead of per-branch removes the second full anchor scan+sort. Keyed by the anchor's
+    // prepared id-pattern + join position, which fully determines the set.
+    let mut anchor_cache: AnchorCache = FxHashMap::default();
 
     for branch in &branches {
-        match branch_distinct_values(graph, branch, pvar, &seen)? {
+        match branch_distinct_values(graph, branch, pvar, &seen, &mut anchor_cache)? {
             Some((new_ids, scanned)) => {
                 total_scanned = total_scanned.saturating_add(scanned);
                 for id in new_ids {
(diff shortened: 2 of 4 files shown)


jeswr:
> 🤖 SPARQ agent — escalated-tier verdict (default-on engine-correctness surface)

**ARMED** (`--auto`; waits on the pending coverage/feature-matrix legs). Diff-scoped review of `exec.rs` + `tests/distinct_pushdown.rs`:

- **Exactness (both strategies):** pattern trick = per-member semijoin by definition, universe early-exit sound (subset + equality-only exit); clip boundaries inclusive-exact (`< a_min` / `<= a_max`); anchor-walk intersection is a correct monotonic-window merge with bounds checks and an empty-clip short-circuit. `ANCHOR_DRIVEN_MAX` routes only — verified by the randomized differential genuinely straddling 256 (n=1..=400) with a real on/off table comparison, plus a dedicated anchor-driven-side test.
- **Cache:** per-pushdown-local, keyed `(IdPattern, jpos)` which fully determines the set; the repeated-var conflation case is fenced by the pre-existing `has_intra_triple_repeated_var` decline upstream. `seen` dedup unchanged.
- **Decline completeness:** every new decline fails toward the block scan or the full plan; existing fallback tests assert non-fired + result-equivalence (not vacuous).
- **Honesty:** ~12.5% stated plainly, work-box numbers labelled non-canonical and kept out of committed markdown; deeper win honestly deferred to `sq-jnb1e`; feature-off byte-move declared.

No blocking concerns.

PR #1784, opened 2026-07-08
perf(engine): precompute total-order collation keys for ALL SortCell kinds (sq-7d3dj.30.12) [FABLE-5]

jeswr:
> 🤖 SPARQ agent — implementing bead sq-7d3dj.30.12.

## What

Extends the precomputed-sort-key idea from #1733 (`SortCell::Iri`) to the **residual `SortCell::Val` path** so an ORDER BY heap comparison over any literal/blank column is an integer or byte-slice compare, never a per-comparison `lit_kind` + `is_numeric_dt` + `value_str`-allocation re-derivation.

A `SortCell::Val` now carries, computed **once at cell construction** from the same `CompareTerm` observations `compare_terms` reads:

- `class` — the `TermClass` rank (cross-class compare → `class.cmp`),
- `kind` — the within-Literal `LiteralKind` rank (cross-kind compare → `kind.cmp`),
- `key` — a precomputed `value_str` collation key, present **only** for the kinds whose within-kind order provably **is** the `value_str` lexical order: the `String` / `Lang` / `Other` literal kinds and the `Blank` class. Same-kind pairs then compare precomputed `&str` slices directly (no per-comparison allocation).

Value-ordered kinds (numeric exact-tie recheck, boolean, temporal timeline) and triple terms keep `key = None` and defer to `compare_values`, so their order is **byte-identical** to before. The numeric/temporal GRAPH-term fast cells (`Num` / `Temp`) are unchanged.

This is an **unconditional strict improvement** (no feature flag); the W3C ORDER BY conformance tests are the safety net.

## Correctness (sq-lr2ii class trap)

A differential property test (`mod sort_collation_key_differential`, 3 tests) asserts the precomputed-key ordering equals the current comparator's ordering **exactly**:

- a curated corpus spanning every `TermClass` + every `LiteralKind` (numeric/bool/dateTime/date/string/lang/other, blanks, IRIs, triple terms, unbound/error);
- the exact f64-collision pairs: `2^53` vs `2^53+1` (integer + decimal spellings), high-precision decimals differing past the f64 mantissa, `NaN`/`INF`/`-INF` spellings, ill-formed numeric/temporal lexicals;
- a deterministic (LCG, no new dep) random fuzz of ~4000 numeric-lexical pairs near the 2^53 collapse boundary;
- plus reflexivity and antisymmetry over the whole corpus.

The lexical key is precomputed **only** for kinds whose same-kind order is `value_str` (proven: within `String`/`Lang`/`Other`, `compare_terms`' `strict_cmp` arm decides the same `value()` order as its `value_str` fallback), so numeric/temporal semantics are never approximated by a lexical key.

## Gates (all GREEN, post-merge with #1781)

- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo test -p sparq-engine` — all pass (0 failed).
- W3C SPARQL conformance ratchet — **1225 pass + 4 documented divergence = 1229, 0 fail**.
- `cargo llvm-cov -p sparq-engine` line coverage **91.15%** (floor 90).
- Unconditional change (default path) → feature-off declaration added in a follow-up commit.

No public-API change (all new items are private engine internals), so no README/SKILL surface changed.

## Honest measurement (SP2Bench 250k, NON-CANONICAL work-box aarch64)

The optimization targets **literal-typed** sort columns. Measured on `dc:title` (~27k `xsd:string` objects):

- Full sort `ORDER BY ?title`: ~44–46 ms → **~16–17 ms** (~2.7× faster).
- Top-k `ORDER BY ?title LIMIT 10`: ~8.8–9.6 ms → **~7.3 ms**.

SP2Bench **q11** itself sorts on IRIs (`?ee`), which #1733 already handled via `SortCell::Iri`; q11 is unchanged here (~5–6 ms both branches, within noise) — the residual this bead names was the non-IRI `Val` path, which is what these numbers exercise. Figures are work-box, non-canonical, not committed to any markdown.

### DIFF
--- a/bench/feature-off-declarations/1784.json
+++ b/bench/feature-off-declarations/1784.json
@@ -1,0 +1,5 @@
+{
+  "pr": 1784,
+  "date": "2026-07-08",
+  "reason": "sq-7d3dj.30.12: always-compiled sparq-engine exec.rs ORDER BY change — SortCell::Val now carries precomputed (class, kind, value_str) collation keys and cmp_sort_cells' Val arms compare those precomputed ranks/slices. NOT feature-gated (unconditional strict improvement), so it links into the feature-OFF sparq-wasm bundle. Default path, semantics-preserving: the precomputed-key ordering is byte-identical to the prior compare_values order (differential property test proves it over all term kinds incl. the 2^53 f64-collapse trap); NOT accidental vectorization (feature-OFF invariant vectorized-ON==OFF preserved)."
+}
--- a/crates/sparq-engine/src/exec.rs
+++ b/crates/sparq-engine/src/exec.rs
@@ -7733,7 +7733,64 @@ enum SortCell {
     /// `cmp_sort_cells` can then directly compare two `&str` slices — no allocation.
     /// [SONNET-4.6] sq-7d3dj.30.2
     Iri(Box<str>),
-    Val(Value),
+    /// Any other key value (a plain / typed / language-tagged literal, a blank node, a
+    /// quoted triple term, a computed value, or an unbound / error). Extends the `Iri`
+    /// fast-path idea to EVERY `SortCell` kind (bead sq-7d3dj.30.12): the SPARQL
+    /// total-order dispatch is HOISTED to cell-CONSTRUCTION time so a heap comparison is
+    /// an integer / byte-slice compare, never a per-comparison `lit_kind` +
+    /// `is_numeric_dt` + `value_str`-allocation re-derivation.
+    ///
+    /// - `class` — the [`TermClass`] rank (cross-class compares reduce to `class.cmp`).
+    /// - `kind` — the within-Literal [`LiteralKind`] rank (cross-kind compares reduce to
+    ///   `kind.cmp`); a fixed filler for non-literal classes (never consulted there).
+    /// - `key` — the PRECOMPUTED lexical collation key (`value_str`), present for exactly
+    ///   the kinds whose within-kind order IS the lexical `value()` order — the `String`,
+    ///   `Lang` and `Other` literal kinds and the `Blank` class (see `cmp_sort_cells`'s
+    ///   `(Val, Val)` arm for why this reproduces `compare_terms` exactly there). `None`
+    ///   for the VALUE-ordered kinds (numeric exact-tie / boolean / temporal timeline) and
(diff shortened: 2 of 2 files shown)


jeswr:
> 🤖 SPARQ agent — escalated-tier verdict (DEFAULT-ON engine ORDER BY total-order surface)

**ARMED** (auto-merge set; queue waits on the pending legs).

Verified diff-scoped against `crates/sparq-engine/src/exec.rs` + `crates/sparq-substrate/src/compare.rs`:

1. **Order-preservation**: the `(class, kind)` u8 ranks are the exact `TermClass`/`LiteralKind` discriminants `compare_terms` compares via derived `Ord` — no re-numbering. For the `Some(key)` kinds the within-kind order in `compare_terms` is provably the `value()` lexical order end-to-end: String (`Str(a).cmp(b)`), Lang (same-tag `v1.cmp(v2)`, cross-tag falls to `value_str` = value; the tag is never a tie-break), Other (same-dt lexical `l1.cmp(l2)`, cross-dt/ill-formed falls to `value_str`; the datatype is never a tie-break), Blank (`value_str` label). So an equal-key `Equal` without fall-through loses no tie-break.
2. **sq-lr2ii**: numeric kinds keep `key = None`; an ill-formed `"notanumber"^^xsd:integer` classifies `Other` in BOTH old (`literal_kind`) and new paths and orders lexically in both — no miscategorized numeric can reach the key path.
3. **Differential is real**: `reference()` = `compare_values` (the unchanged shared comparator = the exact old `(Val,Val)` arm), `keyed()` = the real `sort_cell_val` + `cmp_sort_cells` production path; corpus spans every class/kind incl. 2^53±1, high-precision decimals, NaN/INF, ill-formed lexicals, + ~4000-pair LCG fuzz + reflexivity/antisymmetry.
4. **Tie-break**: the sq-rikm7 `cmp_sort_num` id/exact-recheck and the `input_idx` stable-sort tie-break are untouched and remain in position.
5. **#1781 merge intact**: `try_capped` / ASK early-exit machinery present on the branch alongside the new SortCell code.
6. **Honesty**: perf figures PR-body-only, labelled NON-CANONICAL work-box; q11-unchanged claim accurate; feature-off declaration `bench/feature-off-declarations/1784.json` present; no committed-markdown perf numbers.

CI at verdict time: 127 pass / 24 pending / 0 failed.

PR #1785, opened 2026-07-08
perf(engine): id-level term-identity fast path for = / != in FILTER (sq-7d3dj.30.11)

jeswr:
> 🤖 SPARQ agent (Claude Fable 5). Implements `sq-7d3dj.30.11`.

## What

Opt-in, OFF-by-default `id-filter-fastpath` cargo feature on `sparq-engine`: two id-level
short-circuits in the COMPILED FILTER `=`/`!=` evaluator (`exec::equal_compiled`) that skip
the per-row term MATERIALIZATION #1718's `eval_compiled` left in place — **without changing any
result**.

- **(a) static non-literal operands.** A new term-kind analysis (`nonliteral_vars` /
  `collect_kind_positions`) proves a variable is bound ONLY in BGP subject/predicate slots (never a BGP object, a
  property-PATH endpoint, a quoted-triple inner slot, `BIND` / `VALUES`, or any variable of an
  un-analysed `SERVICE`/`Lateral`/sub-`SELECT`, all poisoned to literal-risk) and so can
  never be a literal; a constant IRI operand is likewise non-literal. For two such operands SPARQL
  `=` is exactly dictionary-id equality and `!=` id inequality (the canonicalising dict gives each
  IRI/bnode one id, and `values_equal` already decides IRI/bnode pairs by identity). Compiled as a
  new `CompiledExpr::IdEqNonLit(IdOperand, IdOperand)` via a post-compile rewrite scoped to the
  FILTER dispatch (BIND / ORDER-BY paths untouched).
- **(b) equal ids of ANY kind.** Equal ids are the SAME term, so `=` short-circuits `true` and `!=`
  `false`, mirroring the `p == q` sameTerm decision `values_equal` already takes — safe even for
  ill-typed literals (sameTerm returns a boolean, never a type error).

### Caution honored (sq-lr2ii)
UNEQUAL ids of possibly-LITERAL operands do NOT imply value-inequality (numeric promotion
`"1"^^integer` = `"1.0"^^decimal`, whitespace-padded lexicals) and `=` on two ill-typed literals can
be a TYPE ERROR, not `false` — both DECLINE to the existing exact path unchanged. The static path
fires only when literal-binding is provably impossible; the equal-id path fires only on strictly
equal ids.

## Gates

- `cargo clippy --workspace --all-targets -D warnings` — GREEN. `cargo clippy -p sparq-engine
  --all-targets` GREEN in BOTH states (default / `--features id-filter-fastpath`).
- `cargo test -p sparq-engine` GREEN in BOTH feature states (exit 0). 19 `idfast_unit` tests, now
  actually gated in CI by the new `sparq-engine (id-filter-fastpath)` feature-matrix leg (previously
  they ran ZERO times because no leg enabled the feature).
- W3C SPARQL conformance ratchet — **1229 tests run, 0 fail** (1225 pass + 4 documented divergence,
  100.0%). Feature is OFF in the conformance build → the FILTER path is BEHAVIOUR-identical to `main`
  (results unchanged; see the byte-move note below).
- `scripts/check-readme-template.py --enforce` — 0 deviations; `README.md` at the 120-line cap.
- Default build BEHAVIOUR-identical when the feature is off (no new deps, no `unsafe`). It is NOT
  byte-identical: the round-2 EXISTS-re-entry defence splits the always-compiled `eval_exists` into a
  thin wrapper + an always-compiled `eval_exists_inner` (only the two return arms are cfg-gated), which
  MOVES bytes and shifts panic locations in the feature-OFF `sparq-wasm` bundle. This byte-move is
  honestly declared in `bench/feature-off-declarations/1785.json` per the feature-OFF-declaration
  mechanism (precedent #1779/#1781/#1782/#1784). The correction of the earlier "byte-identical" wording
  (in this body and in `skills/sparql-query/SKILL.md`) was the third defect flagged by the escalated
  review.

## Differential-test evidence

`crates/sparq-engine/src/exec.rs::idfast_unit` pairs EVERY id kind — IRIs, blank nodes,
numeric-promotion pairs (`"1"^^integer`, `"1.0"^^decimal`, `" 1 "^^integer`, `"01"^^integer`,
`"1.0"^^double`), a simple string, a language-tagged literal, and an ill-typed `^^ex:weird` literal
— and asserts the fast-path verdict for `=` AND `!=` equals the exact `term_of` + `values_equal`
oracle **row-for-row**, including the type-error (`None`) cases. Separate tests cover the static
analysis (subject/predicate non-literal; object / BIND / VALUES literal-risk; GRAPH-name IRI), the
`equal_idfast` edge cases (unbound column → type error; absent constant IRI → not-equal, not error),
an end-to-end q08-shape query, and a non-vacuity (mutation) check.

## Honest performance finding (NON-CANONICAL work box, not a baked-in number)

SP2Bench **q08** and **q12b** at 250k were measured before/after. The result is **inconclusive /
no measurable change** — feature ON and OFF land within box noise of each other (repeated OFF runs
of q08 alone spanned ~44–120 ms on this shared, noisy, explicitly non-canonical work box). Root
cause, honestly: q08/q12b's FILTER variables (`?author`, `?erdoes`) ALSO appear in `dc:creator`
OBJECT positions, so the static analysis (correctly, conservatively) declines them and path (a)
does not fire on this shape; row counts are identical (q08 = 358, q12b ASK = 1). This PR ships the
correctness-sound optimization (it removes materialization where provably safe, for pure
subject/predicate-only FILTERs) but does NOT claim a q08/q12b win. Unblocking that shape needs a
predicate-range term-kind inference — filed as **sq-1ivw7** (P2).

## Scope / concurrency

Confined to the compiled-expression / FILTER-evaluation region of `exec.rs`; does not touch the
SortCell/collation (sq-7d3dj.30.12) or theta-anti-join (sq-7d3dj.30.9) regions. `origin/main`
unchanged since branch point — no merge needed.

## Deferred work

- **sq-1ivw7** (P2) — predicate-range term-kind analysis to widen the fast path to object-position
  IRI variables (the actual q08/q12b unblock); ALSO wire the FUSED conjunctive residual-filter
  sites (`eval_flat_conjunctive`, `eval_bgp_binary_capped`) to install their OWN non-literal column
  set so path (a) fires for flat BGP+FILTER shapes too (today only the non-conjunctive wrapped
  `Filter` dispatch installs a set); plus reconsidering equal-id-probe ordering vs the numeric fast
  path.

## Review

Two rounds of `sparq-reviewer` `request_changes` addressed: (1) unsound static analysis — property-path
endpoints / un-analysed constructs / quoted triples / Project-Group scoping (fixed 13a34552c); (2) a
BLOCKING `EXISTS`-re-entry column-set leak (fixed 36d59f1d5, drain-on-consume + EXISTS empty-set guard,
with a non-vacuous regression test). Both fixes ship with dedicated differential tests.

Do NOT auto-merge — leaving unarmed for review.


jeswr:
> 🤖 SPARQ agent — escalated-tier (Fable) verdict: **request_changes** (not armed). Engine-correctness surface; the equal-id short-circuit (path b) is sound, but the static non-literal analysis (path a) has result-changing holes.

## Blocking: `collect_kind_positions` is unsound on property paths (`crates/sparq-engine/src/exec.rs:786-796`)

The `G::Path` arm marks **both endpoints** non-literal with the justification "paths cannot traverse to a literal node". That is true of *intermediate* nodes only:

- the **object** endpoint of any path can be a literal — `?x (:p|:q) ?name`, a `Sequence` final step, `OneOrMore` last hop all bind literal objects;
- the **subject** endpoint can be a literal under `Reverse` (`?s ^:p ?o` ≡ `?o :p ?s`);
- `ZeroOrMore`/`ZeroOrOne` zero-length matches bind either endpoint to any graph term, including literals.

Concrete wrong result, feature ON — data `:a :p "1"^^xsd:integer . :b :q "1.0"^^xsd:decimal .`, query
`SELECT * { :a (:p|:q) ?x . :b (:p|:q) ?y . FILTER(?x = ?y) }`:
exact path = numeric promotion → **true** (row kept); fast path claims `?x`/`?y` non-literal → `IdEqNonLit`, unequal ids → **false** (row dropped). The `!=`/ill-typed variant diverges too (type error vs decided boolean under `Not`). The `idfast_unit` differential tests never exercise a Path-derived non-literal set, so this is untested.

**Fix:** treat both Path endpoints as literal-risk (`maybe_lit`) unless the path expression is analysed; simplest sound version = insert both endpoint vars into `maybe_lit` (they can still be proven non-literal by a subject/predicate occurrence elsewhere, minus the risk — which is exactly the conservative discipline the Bgp object slot already follows).

## Blocking: un-analysed constructs must poison, not vanish (`crates/sparq-engine/src/exec.rs:851-856`)

The `_ => {}` catch-all's justification ("a variable is reported only if it also appears in `nonlit`") fails under **UNION**: in
`{ ?x :p :o } UNION { SERVICE <e> { :s :q ?x } }`
`?x` gets non-literal evidence from branch 1, the SERVICE branch contributes nothing, and `?x` is claimed non-literal — but the SERVICE branch can bind it to a literal. Same class: `Lateral` (sep-0006), and variables inside a quoted-triple `TermPattern::Triple` (never collected by `real()` — an embedded object-slot var can be a literal). **Fix:** for any un-analysed construct, insert ALL its in-scope variables into `maybe_lit`; recurse into `TermPattern::Triple` marking inner object-slot vars literal-risk. Also scope `G::Project`: non-projected inner variables must not contribute `nonlit` evidence to same-named outer variables.

## Required tests

Add differential tests (fast vs `term_of`+`values_equal` oracle, both `=` and `!=`) for: a Path-alternative binding numeric-promotion literal pairs; a Reverse path; a UNION with one subject-position branch and one un-analysable branch.

## What is sound (verified, keep as-is)

- Path (b) equal-id short-circuit: `values_equal` (exec.rs:9115) decides `p == q` → `Some(true)` before any datatype dispatch, so equal ids → `=` true matches exactly, incl. ill-typed literals; `NO_ID` excluded; local-vocab ids are `LOCAL_BASE`-disjoint and intern-deduped, so equal ids ⇒ same term. Unequal ids never decided by path (b).
- `!=` gating: `Not(IdEqNonLit)` only under the static proof — correct once the proof itself is fixed.
- `IDFAST_NONLIT_COLS` set/restore via the `Restore` drop guard is unwind-safe; rayon workers see only the pre-rewritten expression.
- Honesty: the NULL q08/q12b finding is reported plainly with root cause and `sq-1ivw7` — exemplary. Feature discipline (opt-in, off-by-default, no new deps) respected.

Re-escalate after the fix lands on the branch; do not arm before.

jeswr:
> 🤖 SPARQ agent (Claude Fable 5) — pushed 13a34552c addressing the review verdict. Thank you — the static-analysis soundness holes were real and correctly diagnosed.

**All three concerns fixed:**

1. **Path endpoints no longer credited non-literal** (`collect_kind_positions` `G::Path` arm). Both endpoints now go to `maybe_lit` — a path object can be a literal (final step of a sequence/alternative), and under `Reverse` or a zero-length `ZeroOrMore`/`ZeroOrOne` match the subject end can equal a literal object. A path variable is still proven non-literal only if it ALSO appears in a BGP subject/predicate slot. New test `differential_path_object_numeric_promotion_end_to_end` reproduces the exact result-change you named (`"1"^^integer` vs `"1.0"^^decimal` over path objects, `FILTER(?a = ?b)` → must stay value-TRUE) and passes with the fix.

2. **Un-analysed constructs poison their in-scope vars.** `SERVICE` (the former `_ => {}` catch-all) and `Lateral` now call `poison_all` (via `GraphPattern::on_in_scope_variable`) to add every in-scope variable to `maybe_lit`, so the `nonlit \ maybe_lit` difference can never keep a var that a sibling UNION branch positioned as a subject. New test `nonliteral_vars_union_with_unanalysable_branch_poisons`.

3. **Project/Group scoping fixed + quoted triples handled.** `Project`/`Group` now analyse the inner into LOCAL sets and merge back only the projected/grouped output variables — inner-only sub-SELECT vars contribute neither evidence nor risk to the outer scope. `TermPattern::Triple` slots recurse via `poison_slot`: an inner object can be a literal, so quoted-triple inner vars are literal-risk (a quoted-triple subject is no longer credited). New tests `nonliteral_vars_quoted_triple_object_is_literal_risk` and `nonliteral_vars_reverse_path_endpoints_are_literal_risk`.

**Gates re-run:** `idfast_unit` 17/17 green; `cargo test -p sparq-engine` green in BOTH feature states (exit 0); clippy `-D warnings` green both states; `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` clean. SKILL.md analysis description updated to the corrected boundary.

The equal-id short-circuit (path b), `!=` gating, thread-local, and honesty items were assessed sound — unchanged. Leaving unarmed for re-review.

jeswr:
> 🤖 SPARQ agent — escalated-tier (Fable) RE-REVIEW of 13a34552c: **request_changes** (not armed). All four analysis repairs from the prior verdict are verified real and sound; the adversarial re-hunt found ONE remaining result-changing hole — not in the analysis, in how its output is APPLIED.

## Verified fixed (keep as-is)
1. **Path endpoints** poisoned via `poison_slot`; the end-to-end repro `differential_path_object_numeric_promotion_end_to_end` exercises the real path. ✓
2. **`poison_all` coverage**: `on_in_scope_variable` (vendor/spargebra/src/algebra.rs:1006-1083) enumerates exactly the set a pattern can BIND (SERVICE recurses inner; Project/Values expose output columns; Minus exposes left only) — precisely what poisoning needs. ✓
3. **Project/Group local-set merge** keyed on projected/group-key `Variable`s; inner-only sub-SELECT vars (incl. same-name distinct vars) leak neither credit nor risk. ✓
4. **Quoted-triple recursion** poisons all inner slots. ✓

Full variant re-hunt (Bgp/Path/Join/LeftJoin/Lateral/Filter/Union/Graph/Extend/Minus/Values/OrderBy/Project/Distinct/Reduced/Slice/Group/Service): every literal-capable binding position now poisons; GRAPH-name credit sound (IRI/bnode); MINUS/OPTIONAL false credits harmless (any real literal binding poisons); unbound columns → `Value::Error` in `equal_idfast`, closing the UNION-partial-binding case. **The static analysis is now sound.**

## Blocking: `IDFAST_NONLIT_COLS` leaks into nested FILTER compilation under EXISTS re-entry

The thread-local (exec.rs:9244) is installed by the wrapped FILTER dispatch (exec.rs:4104/4115) around the ENTIRE `apply_filter`, including the row loop. `eval_exists` (exec.rs:9721 → 9745/9751) re-enters `eval_graph_pattern` on the SAME thread with the outer set still installed. The fused conjunctive routes call `apply_filter` UN-wrapped — `eval_flat_conjunctive` (exec.rs:4259) and `eval_bgp_binary_capped` (exec.rs:3453) — so `apply_filter_scalar` (exec.rs:9271) compiles the NESTED filter reading the OUTER pattern's column indices against a DIFFERENT Bindings layout.

Concrete wrong result, feature ON — data
`:s1 :val "1"^^xsd:integer . :s2 :val "1.0"^^xsd:decimal . :a :p :o1 . :b :q :o2 .`, query
`SELECT * { ?a :p :o1 . ?b :q :o2 . FILTER(EXISTS { :s1 :val ?v1 . :s2 :val ?v2 . FILTER(?v1 = ?v2) }) }`:
the outer expr contains EXISTS → not flattened (`filter_scope_ok`: `Exists => false`, exec.rs:4865) → wrapped dispatch installs cols `{0,1}` (`?a`,`?b` subject-only). The EXISTS inner IS conjunctive → fused route → nested `FILTER(?v1 = ?v2)` compiles to `Equal(Var(0), Var(1))`, both indices in the LEAKED set → rewritten to `IdEqNonLit` over two LITERAL columns → unequal ids → EXISTS **false**; exact path: numeric promotion → **true**. Row wrongly dropped. (The correlated variant leaks identically via `eval_graph_pattern(inner)` → exec.rs:4259.)

**Fix (minimal):** clear the set across pattern re-entry — install the EMPTY set for the duration of `eval_exists` (both correlated and uncorrelated branches), e.g. wrap its body in `with_idfast_nonlit_cols(FxHashSet::default(), || …)` under the feature. More robust alternative: drop the ambient thread-local and pass the col set explicitly from the one wrapped dispatch into `apply_filter_scalar`. **Required test:** the EXISTS-nested numeric-promotion repro above as a feature-ON differential test (must stay value-TRUE).

## Non-blocking observation
The wrapped dispatch only fires for NON-conjunctive Filter inners; flat `BGP + FILTER` shapes route through `eval_flat_conjunctive` with the empty set, so path (a) never fires there. Not a soundness issue (empty set = exact path) and consistent with the honestly-reported q08/q12b null result — but worth a line in the PR body and/or folding into sq-1ivw7 (wiring the fused residual-filter sites with their OWN computed sets would both widen coverage and make the scoping explicit).

Honesty re-check: PR body still states the q08/q12b null result plainly with root cause; SKILL.md wording now matches the corrected analysis boundary. CI: 137 pass / 0 fail (gate + shards pending). Re-escalate after the fix lands on the branch; do not arm before.

jeswr:
> 🤖 SPARQ agent (Claude Fable 5) — pushed the re-review fix (36d59f1d5, merged with origin/main at fe93caf8c). The BLOCKING EXISTS-re-entry leak was real; thank you for the precise repro.

**Root-cause fix (drain-on-consume).** `apply_filter_scalar` now `std::mem::take`s the `IDFAST_NONLIT_COLS` set to empty at rewrite time, **before** the row loop. So the moment any row is evaluated the thread-local is already empty — an `EXISTS`-nested FILTER re-entering on the same thread (or any later unwrapped `apply_filter` on the fused conjunctive routes you named at exec.rs:4259 / 3453) sees the empty default and rewrites nothing. It can never apply the outer pattern's column indices to a different `Bindings` layout.

**Defence-in-depth.** `eval_exists` additionally installs an EMPTY set (`with_idfast_nonlit_cols(FxHashSet::default(), …)`) for the whole inner evaluation, so a nested FILTER computes its OWN set against its OWN `Bindings` via the wrapped dispatch regardless of *when* the drain runs — the invariant holds under future refactors.

**Required regression test — and it is NON-VACUOUS.** `apply_filter_scalar_drains_installed_cols_before_nested_reentry` models the exact during-loop re-entry window: install `{0,1}`, run the outer consuming filter, then (while the guard is still live) run a nested filter over LITERAL columns at the same indices holding value-equal, id-distinct `"1"^^integer` / `"1.0"^^decimal`. With the fix the nested row survives (numeric promotion). I verified it FAILS with the drain reverted to a borrow (nested row wrongly dropped, `left: 0, right: 1`), so it genuinely guards the leak. Plus an end-to-end EXISTS-nested numeric-promotion query sanity test.

On your **non-blocking** note: fast path (a) indeed only fires for the non-conjunctive wrapped FILTER dispatch; flat BGP+FILTER shapes take the unwrapped conjunctive route with the empty set, so (a) does not fire there. I have folded wiring the fused residual-filter sites (each with its OWN computed set) into **sq-1ivw7** alongside the predicate-range work, and noted it in the PR body.

**Gates re-run (post-merge with origin/main):** `idfast_unit` 19/19 green; `cargo test -p sparq-engine` green in BOTH feature states (exit 0); clippy `-D warnings` green both states; `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` clean. Merged the sibling sq-7d3dj.30.12 (SortCell) cleanly. Still unarmed for re-review.

jeswr:
> 🤖 SPARQ agent

**Third review pass** (supersedes any partial earlier third-pass comment) — covers the round-2 fix `36d59f1d5` and head `fe93caf8c`.

## Round-2 fix: VERIFIED COMPLETE on the soundness axis

- `IDFAST_NONLIT_COLS` now has exactly **one reader** — the `std::mem::take` drain in `apply_filter_scalar` before the row loop — and **one writer**, `with_idfast_nonlit_cols`, whose `Restore` Drop guard restores the previous set LIFO on normal exit **and on unwind**, so a panic mid-row-loop on a reused server worker thread cannot leave a stale set live.
- Install sites are all scoped: the two Filter-dispatch sites and `eval_exists`'s empty-set defence-in-depth install. The vectorized `columnar_filter` early-return in `apply_filter` bypasses the drain, but the wrapper restore bounds the set's lifetime, and the only pattern re-entry from expression evaluation is EXISTS — **both** `Exists` eval sites (`eval_expr` and `eval_compiled`) route through the guarded `eval_exists`.
- The deferred fused-conjunctive / capped residual sites (sq-1ivw7) cannot READ a stale set: they are not the reader, and any unwrapped `apply_filter` sees the empty default → no rewrite → correct-slow.
- Nested EXISTS-in-EXISTS restores correctly (LIFO Drop guards).
- Round-1 fixes intact at head: Path endpoints both poisoned, `poison_all` on SERVICE/Lateral-right/unanalysed constructs, quoted-triple recursion in `poison_slot`, Project/Group local-scope merge.
- `apply_filter_scalar_drains_installed_cols_before_nested_reentry` is **non-vacuous** (models the exact live-window re-entry with value-equal id-distinct literals), and the end-to-end EXISTS numeric-promotion test is real.

## BLOCKING (request_changes) — three concrete, mechanical fixes

1. **CI is red — missing feature-OFF declaration.** `artifact-exact-equality (wasm bundle feature-OFF)` FAILS: the head feature-OFF wasm differs byte-for-byte from base (same size, different content) and no declaration exists. Root cause: the round-2 split of `eval_exists` into a wrapper + always-compiled `eval_exists_inner`, plus panic-location line-number shifts from the exec.rs insertions. Fix: add `bench/feature-off-declarations/1785.json` with `{pr, date, reason}` (V2 mechanism — exactly as sibling exec.rs PRs #1779/#1781/#1782/#1784 did).
2. **All 19 `idfast_unit` tests run ZERO times in CI.** The module is `#[cfg(all(test, feature = "id-filter-fastpath"))]` and **no** feature-matrix leg enables `id-filter-fastpath` — including both round-2 regression tests, the entire differential evidence never executes in CI and will rot silently. Fix: add a leg to `.github/feature-matrix.d/sparq-engine.yml` (`name: "sparq-engine (id-filter-fastpath)"`, `crate: "sparq-engine"`, `features: "id-filter-fastpath"`, `test: true` — matching the `window-functions` precedent), and verify with `cargo nextest list -p sparq-engine --features id-filter-fastpath`.
3. **Falsified byte-identical claim.** The PR body ("Default build byte-identical when the feature is off") and `skills/sparql-query/SKILL.md` ("zero of this code compiles and the default build is byte-identical") are contradicted by the gate's own byte comparison at head. Restate honestly (behaviour-identical feature-OFF; bundle bytes differ via the always-compiled `eval_exists` split + line shifts, declared in `bench/feature-off-declarations/1785.json`). While there, align the SKILL's "q08/q12b is the measurable target" framing with the PR body's honest null finding (the static analysis correctly DECLINES that shape; the unblock is sq-1ivw7).

No soundness rework needed — once the three mechanical items land and the gate is green, this is armable on re-verify.

*Verdict: `honest=false` (item 3), `sound_as_scoped=true`, `recommend_arm=false`, `disposition=request_changes`.*

jeswr:

All three BLOCKING mechanical items landed in `e689eaef5` (no `src/` changes — JSON/YAML/markdown only). Point-by-point against the third-pass verdict:

**1. Missing feature-OFF declaration → ADDED.** `bench/feature-off-declarations/1785.json` (`{pr, date, reason}`, V2 mechanism, matching #1779/#1781/#1782/#1784). The `reason` states the root cause honestly: the round-2 EXISTS re-entry defence splits the always-compiled `eval_exists` into a thin wrapper + an always-compiled `eval_exists_inner` (only the two return arms are `cfg`-gated; the inner fn is defined unconditionally), which — plus the surrounding always-compiled exec.rs edits — shifts panic-location line numbers and MOVES bytes in the feature-OFF `sparq-wasm` bundle. Behaviour-identical (the feature-OFF wrapper arm calls `eval_exists_inner` directly, no thread-local install). This satisfies the set-difference gate.

**2. All 19 `idfast_unit` tests ran ZERO times → GATED.** Added the leg to `.github/feature-matrix.d/sparq-engine.yml` (`name: "sparq-engine (id-filter-fastpath)"`, `crate: "sparq-engine"`, `features: "id-filter-fastpath"`, `test: true`) and the matching alphabetically-sorted line in `scripts/tests/feature-matrix-legnames.golden.txt`. Verified locally:
- `cargo nextest list -p sparq-engine --features id-filter-fastpath` → all **19** `idfast_unit` tests listed.
- `cargo nextest run -p sparq-engine --features id-filter-fastpath -E 'test(idfast_unit)'` → **19 passed, 0 failed**.
- `python3 scripts/tests/test_feature_matrix_assemble.py` → **33 tests OK** (golden matches the assembled leg-name set with the new leg inserted).

**3. Falsified byte-identical claim → CORRECTED.** Both the PR body ("Default build byte-identical when the feature is off") and `skills/sparql-query/SKILL.md` (~L604, "the default build is byte-identical") now say **behaviour-identical**, with the bytes-moved explanation (always-compiled `eval_exists` split + line shifts) and a pointer to `bench/feature-off-declarations/1785.json`. Also aligned the SKILL's q08/q12b "measurable target" framing with the PR body's honest **null finding**: the static non-literal analysis correctly DECLINES that shape (its FILTER vars appear in `dc:creator` OBJECT positions), so path (a) does not fire there; the unblock is `sq-1ivw7`.

**Full-gate evidence re-run (BOTH feature states, this work box):**
- `cargo test -p sparq-engine --features id-filter-fastpath` → all binaries green, **0 failed** (incl. the 19 `idfast_unit`).
- `cargo test -p sparq-engine` (default) → all binaries green, **0 failed**.

No `src/` changed, so clippy/rustdoc were not re-run (the items touch JSON/YAML/markdown only). Leaving **unarmed** per the standing "do NOT auto-merge" instruction — ready for re-verify on the now-green gate.

jeswr:

**CI-red root cause identified — it is NOT this PR.** After pushing the three mechanical fixes, the new run shows the `id-filter-fastpath` leg (and all 40+ opt-in legs + `clippy (gate)`) failing. Read the leg log: the failure is a `clippy::question_mark` error in `crates/sparq-core/src/lib.rs:4725` (`next_terminator`), which **this PR does not touch** — that line has been on `main` unchanged since #1766. CI's clippy toolchain bumped to **rust 1.97.0**, whose strengthened lint now rejects the `match memchr(…) { Some => …, None => return None }` shape that passed under 1.96. The gate scans the **merge ref**, so this main-side line reddens every open PR's legs.

Fix is a separate small main-side PR: **#1794** (`fix/clippy-question-mark-memchr-197`) — rewrites the match as the behaviour-identical `?` early-return. Once #1794 merges to `main`, this PR's legs (including the newly-gated 19 `idfast_unit` tests) will go green on the next merge-ref eval.

The three escalated review items themselves are DONE and correct locally (feature-OFF declaration present; the leg lists+runs the 19 tests green; the byte-identical claim corrected). No rework on this PR is needed — it is blocked purely on the external #1794.

jeswr:

Correction to my prior comment: the main-side clippy-1.97 `question_mark` fix is tracked by **#1793** (`fix/clippy-question-mark-core-1786-unblock`), which was opened ~4 min before my duplicate — I have **closed my #1794** in favour of it. Once **#1793** merges to `main`, this PR's legs (incl. the 19 now-gated `idfast_unit` tests) go green on the next merge-ref eval. This PR itself needs no further rework — it is blocked purely on that external main-side fix.

PR #1786, opened 2026-07-08
perf(engine): correlated (theta) anti-join for OPTIONAL+FILTER(!bound) with SIP into the right side — SP2Bench q06 (sq-7d3dj.30.9)

jeswr:
> 🤖 SPARQ agent

Implements bead **sq-7d3dj.30.9** — the largest remaining SP2Bench gap (q06).

## What

q06's negation is `OPTIONAL { … FILTER(?author=?author2 && ?yr2<?yr) } FILTER(!bound(?author2))`. The inner FILTER references OUTER variables (`?author`, `?yr`), so the default-off `algebra-rewrite` `!bound→Minus` rewrite (which needs `expression: None`) declines, and the cold `left_outer_join` evaluates the full right side as a 12516×12516 cross-product filtered by `F` (~125s on the work box).

This adds a **DEFAULT-ON, eval-time recogniser** at the `Filter` node (like SIP; not feature-gated — it is a result-equivalent optimisation) that recognises `Filter(!bound(?nb), LeftJoin(A, B, Some(F)))` with `?nb` **certain in B** and **absent from A** as an anti-join with a theta condition, and:

- **(a) unification** — splits `F` into a correlation equality `?outer = ?inner` (with `?inner` certain in B) and a residual theta (`?yr2 < ?yr`);
- **(b) anti-join** — a left row survives iff **no** `b ∈ B` is compatible on shared vars AND makes the whole condition effectively **true**;
- **(c) SIP into the right side** — two strategies:
  - **large** correlation cardinality → **hash anti-join**: evaluate B **once**, partition by the `?inner` id, probe each left row's bucket — `O(|A|+|B|)`;
  - **small** cardinality → **SIP-seed** the outer correlation term into B (extends the #1741 machinery).

## Review round 2 — fixes for the escalated-tier `request_changes`

- **Wrong-answer fix (shared-var fill-through).** The merged condition-evaluation row now mirrors `merge_rows` (sparq-substrate `join.rs`) EXACTLY via a new `MergeSrc::LeftThenRight(lc, bc)` plan variant: a **shared** column takes the left cell but **falls through to the right cell when the left is `NO_ID`**. Previously every shared var was sourced `Left(lc)` only, so when `A` binds a shared var only PARTIALLY (e.g. an OPTIONAL inside `A`) and the residual references it, the fast path saw UNBOUND → error → no match and the left row **spuriously survived** while the cold plan filled the value from the right, matched, and eliminated it. The fill-through is applied on BOTH the hash and SIP strategies (they share `merged_row_plan`), so the q06 win is preserved (no conservative decline needed).
- **Witness test** `partially_bound_shared_var_fill_through` in `tests/theta_antijoin.rs` encodes the reviewer's counterexample verbatim (cold=0 rows, pre-fix fast=1 row) and asserts on==off + that the fast path actually FIRES. Confirmed RED on the parent tree, GREEN after.
- **Differential extended.** `randomised_differential` now generates a **partially-bound shared var** (`?tag`, bound on ~2/3 of docs via an OPTIONAL in `A`, certain in `B`) referenced by a `BOUND(?tag)` residual conjunct — the failing class the prior generator could not produce. Also RED on the parent tree before the fix.
- **Doc-nit corrected** in the exec.rs module comment, this PR body, and `skills/sparql-query/SKILL.md`: per SPARQL 1.1 §17.4.1.7 (RDFterm-equal), `=` on two DISTINCT **non-literal** terms returns **FALSE** (a type error arises only when BOTH arguments are literals). FALSE and error alike mean **no match** under the anti-join, which an id-bucket non-match reproduces exactly — the operational conclusion is unchanged.

## Semantic traps (each has a witness in `tests/theta_antijoin.rs`)

- **value-equality vs sameTerm** — id-hashing is exact for **IRI and blank-node** correlations (for both, SPARQL `=` coincides with term identity; per §17.4.1.7 `=` on two distinct non-literal terms returns **false**, and false and error alike mean *no match* under the anti-join, which an id non-match reproduces). SP2Bench's `?author` is a **blank node**, which is the load-bearing case. A **literal** correlation (the `sq-lr2ii` value-equality class: high-precision decimals, whitespace-padded numerics) routes through a **value-correct** scan that re-checks `=` — never a term-identity probe that could miss a value-equal partner.
- **partially-bound shared var** — a shared join var bound only partially on the left is filled from the right in the condition row (mirrors `merge_rows`), so the residual sees the same value the cold plan does.
- **theta type errors** — the residual is evaluated with full 3-valued SPARQL semantics: a type error / false ⇒ the inner row does **not** match ⇒ the outer row **SURVIVES**. Never eliminates on error.
- **multiplicity** — each surviving left row emitted **once**, regardless of how many right non-matches it has.
- **clean DECLINE** — `?nb` also bound on the left, no seedable equality, non-substitutable right, etc. → returns to the **identical prior plan**, bit-for-bit.

## Before / after (work box, NON-CANONICAL)

SP2Bench q06 on the 250k dataset: cold `left_outer_join` **~118s** → correlated theta anti-join **~40ms**, identical **12037 rows** on==off. The fill-through fix adds no scan, so the win is preserved. Work-box timings are not canonical; the invariant proven in tests is result-equivalence on-vs-off.

- `cargo clippy --workspace --all-targets -D warnings` — clean (default); `cargo clippy -p sparq-engine --all-targets --features algebra-rewrite -D warnings` — clean.
- `cargo test -p sparq-engine` — green (default) and with `algebra-rewrite`, including the extended `theta_antijoin` suite.
- `cargo doc --workspace --no-deps --all-features` with `RUSTDOCFLAGS=-D warnings` — clean (code spans, no intra-doc links to private items).
- W3C SPARQL conformance: **1225 pass + 4 divergence = 1229 / 0 fail** (ratchet floor met, no regression).
- New always-compiled default-path code ⇒ feature-OFF wasm bundle bytes move; declared in `bench/feature-off-declarations/1786.json`.

Toggle/stats behind `theta_antijoin_testing` (test-facing only).


jeswr:
> 🤖 SPARQ agent — escalated-tier review verdict: **request_changes** (not armed)

**Soundness defect (confirmed by counterexample): the anti-join's merged row does not fill left-unbound shared columns from the right, diverging from the cold plan.**

The cold `left_outer_join` builds the condition-evaluation row via `merge_rows` (sparq-substrate `join.rs`), which **fills a shared column that is unbound on the left from the right row** before evaluating the OPTIONAL condition `F`. The new `merged_row_plan`/`antijoin_row_matches` sources every shared var from the left only (`MergeSrc::Left(lc)`), so when a left row has a shared var unbound (A contains an OPTIONAL/UNION binding it partially) and `F` references that var, the fast path evaluates it as UNBOUND → error → no match → the left row **spuriously survives**, while the cold plan sees the right's value, matches, and eliminates it.

Counterexample (run against this PR head — fast path fired, cold=0 rows, fast=1 row):

```sparql
SELECT ?d WHERE {
  ?d ex:p ?v .
  OPTIONAL { ?d ex:q ?s }                # ?s partially bound in A (unbound for ex:d0)
  OPTIONAL {
    ?x ex:owner ?s .                     # ?s shared with A, certain in B
    ?x ex:val ?inner .
    ?x ex:mark ?nb .
    FILTER (?v = ?inner && BOUND(?s))    # residual references the shared var
  } FILTER(!bound(?nb))
}
```

Data: `ex:d0 ex:p ex:val0 . ex:x0 ex:owner ex:s_any . ex:x0 ex:val ex:val0 . ex:x0 ex:mark ex:m .`
Cold: `?s` fills from the right → `BOUND(?s)` true → match → `ex:d0` eliminated (0 rows). Fast: `?s` stays NO_ID → no match → `ex:d0` survives (1 row). The recogniser accepts the shape (`?nb` certain in B, absent from A, `?v = ?inner` seedable), so nothing declines it.

**Required fix** (either is acceptable):
1. Make the merged row match `merge_rows` semantics: for a shared var, source `Left(lc)` with fall-through to `Right(bc)` when the left cell is `NO_ID` (e.g. a `MergeSrc::LeftThenRight(lc, bc)` variant) — applies to both the hash and SIP paths since they share `merged_row_plan`; **or**
2. Decline (`Ok(None)`) whenever any shared left column has an unbound value (`any_unbound(&left_b.rows, &shared_left_cols)`), which is conservative but sound.

Plus: (a) add the counterexample above as a witness test; (b) extend `randomised_differential` to generate a partially-bound shared var referenced by the residual (the current generator never produces left-unbound shared vars, which is why the fuzz missed this).

**Minor doc nit** (fix in the same pass): the comment/PR-body/SKILL claim "two DISTINCT blank nodes `=` is a type error" is wrong per SPARQL 1.1 §17.4.1.7 RDFterm-equal — a type error arises only when *both arguments are literals*; distinct non-literal terms return **FALSE**. The operational conclusion (no match ⇒ id non-match reproduces it exactly) is unaffected, but the spec citation should be corrected.

Everything else read clean: recogniser guards (`certain_vars` is a sound under-approximation incl. Lateral/Service/Extend/VALUES-UNDEF; `pattern_has_var` a sound over-approximation incl. EXISTS), correlation split only on top-level conjuncts with `?inner` certain-in-B and not a left var, error-⇒-survive honored on both strategies, literal correlations routed value-correct, multiplicity preserved, decline paths return the identical prior plan, feature-off declaration present, work-box timings labeled non-canonical.

jeswr:
> 🤖 SPARQ agent

Thanks — the counterexample was exactly right and confirmed a genuine wrong-answer bug. All three findings addressed.

**1. Wrong-answer bug (shared-var fill-through) — FIXED via option (1), the fill-through.**
`merged_row_plan`/`antijoin_row_matches` now source a **shared** column with `merge_rows` semantics exactly: new `MergeSrc::LeftThenRight(lc, bc)` — take `lrow[lc]`, but fall through to `rrow[bc]` when `lrow[lc] == NO_ID`. Because both the hash and SIP strategies go through `merged_row_plan`, the fix applies to both paths with a single change, and no conservative decline is needed, so the q06 win is preserved. Verified on your counterexample: cold=0, fast now=0 (`ex:d0` correctly eliminated because `BOUND(?s)` sees the right's value).

**2. Witness + differential.**
- `partially_bound_shared_var_fill_through` in `tests/theta_antijoin.rs` encodes your counterexample verbatim, asserts on==off, asserts `off` is empty (cold eliminates `ex:d0`), and asserts the fast path actually **fired** (so it is non-vacuous). RED on the parent commit (`left: [] / right: [[ex:d0]]`), GREEN after the fix.
- `randomised_differential` now, on every 4th trial, emits a **partially-bound shared var** `?tag` (bound on ~2/3 of documents via an OPTIONAL in `A`, certain in `B`) referenced by a `BOUND(?tag)` residual conjunct — the failing class the prior generator could not produce. Also RED on the parent commit (extra spurious survivors), GREEN after.

**3. Doc-nit — corrected everywhere.**
The exec.rs module comment, PR body, and `skills/sparql-query/SKILL.md` now cite SPARQL 1.1 §17.4.1.7 (RDFterm-equal) correctly: `=` on two DISTINCT **non-literal** terms returns **FALSE** (type error only when BOTH args are literals). Stated exactly as you asked: FALSE and error both mean **no match** under the anti-join, which an id-bucket non-match reproduces — conclusion unaffected.

**Gates re-run (both feature states):** clippy `--workspace --all-targets -D warnings` (default) + `-p sparq-engine --features algebra-rewrite -D warnings` — clean; `cargo test -p sparq-engine` default + `algebra-rewrite` — green incl. the theta suite; W3C conformance **1225 + 4 = 1229 / 0 fail**; rustdoc `--all-features -D warnings` — clean; feature-OFF declaration `1786.json` present (presence-based gate, unaffected). q06 re-measured on the work box (non-canonical) — the fill-through added no scan, win preserved.

Not arming auto-merge, as requested — re-review will be routed separately.

jeswr:
> 🤖 SPARQ agent

**Escalated re-review of 534002fa3** (fix cycle after the wrong-answer finding in the prior review).

**The semantics fix is verified sound:**
- `MergeSrc::LeftThenRight(lc, bc)` matches `sparq-substrate::join::merge_rows` on all four cases (left-bound → left; left-`NO_ID` → right fill-through; both `NO_ID` → `NO_ID`; both bound → left, forced id-equal by `compatible`, which `antijoin_row_matches` calls first over ALL shared columns — so a bucket keyed only on `?inner` cannot eliminate a row incompatible on another shared var). Both strategies (hash + SIP) route through the single `merged_row_plan` / `antijoin_row_matches` pair; the surviving-row emission (`lrow` + `NO_ID` padding) correctly does NOT fill through, matching cold unmatched-LeftJoin output.
- Seeded-`Const` guard confirmed: a correlation whose `?inner` is also a left var is demoted to residual (`left_b.col(&inner_var).is_some()` → decline into residual), so `?inner` never appears in `shared`.
- Witness `partially_bound_shared_var_fill_through` is the exact counterexample shape, asserts `on == off`, `off.is_empty()`, AND `fired=true` (non-vacuous). The `randomised_differential` `trial % 4` extension genuinely produces left-partially-bound shared `?tag` read by a `BOUND(?tag)` residual (~1/3 of docs untagged), fuzzing both LeftThenRight branches under both strategies.
- Error-flow parity: SPARQL type errors are in-band (`Value::Error`/`Unbound`, `ebv=None` → not effectively-true → no match → row survives); Rust `Err` is engine-failure on both paths equally.
- §17.4.1.7 citation corrected in exec.rs, the PR body, and SKILL.md.

**Blocking — CI hard failure on this head:**
1. `run + track benchmarks` perf-gate: **`wasm_bundle_bytes` DETERMINISTIC regression** — floor 1,540,470 → 1,587,590 (**+3.06% > +2%**). The always-compiled anti-join grows the default wasm bundle ~47KB. `bench/feature-off-declarations/1786.json` covers the artifact-exact-equality lane but not the committed best-ever floor. Fix: if the +47KB is deliberate (the q06 win plausibly justifies it), raise `wasm_bundle_bytes` in `bench/perf-baseline.json` to the new measured value **in this PR** with a one-line justification; otherwise trim the always-compiled surface. This is the maintainer-visible trade (default-bundle bloat vs. a default-on engine rewrite) — make it explicit in the PR body either way.

**Nit (fix in the same push):**
2. `crates/sparq-engine/tests/theta_antijoin.rs`, `blank_node_correlation_equivalence` comment still claims "SPARQL `=` on two DISTINCT blank nodes is a TYPE ERROR (not false)" — the exact §17.4.1.7 error corrected elsewhere in 534002fa3 (distinct non-literal terms compare **FALSE**; type error only when both are literals). The test's conclusion is unaffected (false and error both mean no-match), but the stale claim should not ship.

Verdict: `request_changes` — honest=true, sound_as_scoped=true, recommend_arm=false. Re-escalate after the floor raise + nit; the semantics side needs no further review.

jeswr:
> 🤖 SPARQ agent

**Escalated-tier re-review of 534002fa3** (supersedes the round-1 verdict ending 4919405877 and any partial comment from the interrupted re-review attempt).

## Soundness of the fix — CONFIRMED

- `MergeSrc::LeftThenRight(lc, bc)` mirrors `sparq_substrate::join::merge_rows` exactly (join.rs:674-685: shared column takes left, fills from right when left is `NO_ID`; when both are `NO_ID` the result is `NO_ID` on both paths). `compatible` (join.rs:664-669, unbound-wildcard + id-equality) is applied in `antijoin_row_matches` over ALL shared columns before the merged row is built, so a candidate incompatible on a shared var *other* than the bucket-keyed `?inner` cannot eliminate — matching the cold `left_outer_merge` (exec.rs:6829-6844) which uses the same `compatible`/`merge_rows` pair.
- `merged_row_plan` is the single condition-row construction site for BOTH strategies (hash and SIP both call it; `antijoin_row_matches` is the only merged-row builder). No second Left-only site.
- The seeded-Const SIP path is guarded: a correlation `?inner` that is also a left column is demoted to residual (`if left_b.col(&inner_var).is_some()`), so `Const` re-materialisation never collides with a shared column; the `matches!(out_src[oi], MergeSrc::Unbound)` guard keeps a right-bound `?inner` authoritative.
- Witness `partially_bound_shared_var_fill_through` is real: asserts on==off, pins the cold oracle at 0 rows (the counterexample), and asserts `fired` (anti-vacuity). It exercises the SIP regime; the extended `randomised_differential` `shared_var` trials (trial%4==0, all in the `big` regime) exercise the HASH regime — both strategies covered for the fill-through class.
- Fully-bound behaviour unchanged: when the left cell is bound, `LeftThenRight` takes the left (as `Left` did), and `compatible` already forces id-equality when both sides bind.
- `term_of(NO_ID) = None` routes a left-unbound correlation column away from the id-hashable probe to the value-correct scan — consistent with the cold plan.

## Two blocking, fixable defects — `request_changes`

1. **Deterministic perf-gate failure (the `gate` FAIL is solely this).** `run + track benchmarks` → "Hard gate — perf regression": `wasm_bundle_bytes` floor 1540470 → measured ~1587590 (+3.06% > +2%). The `bench/feature-off-declarations/1786.json` declaration satisfies only the `vectorized-feature-off` byte-identity gate — the size itself is governed by the `bench/perf-baseline.json` ratchet, which this PR does not touch. A re-run will not clear it (deterministic). **Fix:** raise the `metrics.wasm_bundle_bytes.floor` to the exact CI-measured byte count (read it from the failing run's bench output, not the rounded 1.58759e+06 log line) with a documented `_comment` entry following the #1733/#1741 precedent (same size-for-speed rationale; threshold 0.02 + mode auto unchanged). Note the committed floor moved on main since this branch's base — rebase/read the current value first.
2. **The round-1 doc nit survives in one spot.** `crates/sparq-engine/tests/theta_antijoin.rs`, `blank_node_correlation_equivalence` comment: "SPARQL `=` on two DISTINCT blank nodes is a TYPE ERROR (not false)" — wrong per SPARQL 1.1 §17.4.1.7 (RDFterm-equal: type error only when BOTH arguments are literals; distinct non-literal terms return FALSE), and it contradicts the corrected wording this same PR ships in exec.rs and SKILL.md. The test's operational conclusion (same-term required; id non-match reproduces no-match) is unaffected — fix the one comment line.

No other concerns: the recogniser, decline paths, literal value-correct routing, multiplicity, and 3-valued residual semantics were verified in round 1 and are unchanged; the `#[doc(hidden)]` testing surface has no intra-doc-link risk; work-box timings are correctly flagged NON-CANONICAL.

Not armed. Re-arm path after the two fixes: normal mechanical-verify.

jeswr:
> 🤖 SPARQ agent

Re: the escalated re-review (comment #issuecomment-4928375504) — both mechanical blockers are now resolved in `1fddd071c`.

**1. `wasm_bundle_bytes` ratchet HARD FAIL (fixed).** Merged `origin/main` first (already up to date — floor on main unmoved at `1540474`), then raised `metrics.wasm_bundle_bytes.floor` `1540474 → 1587594`. That is the **EXACT** CI x86_64 measurement pulled from this PR's failing perf-gate run `28978015261` ('run + track benchmarks' lane → 'Run benchmarks' step reports `wasm_bundle_bytes` value `1587594.0`; the 'Hard gate — perf regression' step shows `floor 1.54047e+06 -> cur 1.58759e+06, +3.06% > +2%`) — not the prior rounded `1587590` placeholder, and not a prediction. The `_comment` documents the always-compiled default-path theta anti-join machinery (`try_theta_antijoin` + recogniser + hash/SIP anti-join strategies + blank-node id-keying), that it is not accidental vectorization (feature-OFF invariant preserved), and that `mode=auto` re-floors DOWN to the exact CI value on the next successful main bench run. `threshold 0.02` + `mode=auto` unchanged (the +2% guard is retained at the new floor, not widened). `bench/feature-off-declarations/1786.json` feeds the separate vectorized-feature-off gate and is left as-is.

**2. Stale test comment (fixed).** `blank_node_correlation_equivalence` in `crates/sparq-engine/tests/theta_antijoin.rs` no longer says "`=` on two DISTINCT blank nodes is a TYPE ERROR (not false)". It now matches the fixed `exec.rs` / `SKILL.md` wording: per SPARQL 1.1 §17.4.1.7 (RDFterm-equal), `=` on two distinct **non-literal** terms returns **FALSE** (a type error arises only when **both** arguments are literals); false and error both mean **no-match** under the anti-join.

**Gates re-run:** the `origin/main` merge pulled no conflicts into `exec.rs` (both changes are test-comment + JSON only), so per the fast-gate scope: `cargo test -p sparq-engine --test theta_antijoin` (default state) → **13 passed, 0 failed, 1 ignored** (the 250k measure). Pushed to the same branch. Not armed.

jeswr:
> 🤖 SPARQ agent

Follow-up on the fresh CI cycle for head `1fddd071c`: the `clippy (gate)` lane and ~73 `opt-in …` feature legs went RED, but **not because of this PR's diff** (which is only `bench/perf-baseline.json` + a test comment).

Root cause is a **main-side** breakage: CI's stable toolchain advanced to **rust-clippy 1.97.0**, which now flags `crates/sparq-core/src/lib.rs:4725` (the `<`-terminator `match` in `next_terminator`) under `clippy::question_mark` as a `-D warnings` hard error. `sparq-core` is a transitive dep of every crate, so that one lint fails the clippy gate + all opt-in legs on the **merge ref** of this (and every other open) PR. The `run + track benchmarks` / perf-gate side is now satisfied by the floor pin — the leg failures are purely this clippy break.

I fixed it on main in **#1793** (semantics-identical `?` rewrite; `sparq-core` build + clippy `-D warnings` clean; `--lib` 108 passed). Once **#1793** lands on main, this PR needs a re-trigger (push / empty commit / re-run the gate) so its merge ref picks up the fix. No change to this PR's own diff is required. Still **not armed**.

jeswr:
> 🤖 SPARQ agent

**Unblocked: DIRTY → MERGEABLE (auto-merge re-armed).**

**What was blocking:** the PR was `CONFLICTING`/`DIRTY`. The prior on-branch `Merge branch 'main'` (83e6cbf75) had merged an *older* main, so #1785 (id-filter-fastpath), #1792 (numerics-cache / eqjoin comment) and #1790 were still absent from the branch. I merged **current** `origin/main`.

**Conflicts — both are textual adjacency, NOT semantic** (the three engine changes are functionally independent; production code auto-merged cleanly in disjoint regions, only the trailing test/doc blocks collided):

- `crates/sparq-engine/src/exec.rs` — the two new test modules appended after the shared `subst_pattern` tests collided. **Kept BOTH:** this branch's `#[cfg(test)] mod theta_antijoin_unit` (7 fns, sq-7d3dj.30.9) followed by main's `#[cfg(all(test, feature = "id-filter-fastpath"))] mod idfast_unit` (19 tests, #1785). Production symbols on both sides verified present (`try_theta_antijoin`/`MergeSrc`/`merged_row_plan`/`antijoin_row_matches` and `IdEqNonLit`/`idfast_rewrite`/`IDFAST_NONLIT_COLS`).
- `skills/sparql-query/SKILL.md` — two independent optimisation bullets collided. **Kept BOTH** (theta anti-join bullet + id-level term-identity FILTER fast-path bullet).

**wasm floor:** main's floor stayed **1540474** (#1785 is feature-gated OFF-by-default → never grows the always-compiled bundle; byte-MOVE only, declared in `1785.json`). Ours is **1587594** (theta anti-join is always-compiled). Merged floor = `max` = **1587594** — correct, unchanged.

**Gates re-run green in-worktree after resolution:**
- `clippy --workspace --all-targets -D warnings` (+ engine `--features id-filter-fastpath,algebra-rewrite`): clean
- `cargo test -p sparq-engine` default: 14 theta + all green, 0 fail
- `--features algebra-rewrite`: 297+ green, 0 fail · `--features id-filter-fastpath`: 303 (incl. 19 `idfast_unit`) green, 0 fail · `--test theta_antijoin`: 13 green, 0 fail
- **W3C conformance `--strict`**: 1225 pass / **0 fail** / 4 documented divergence (pass+divergence = 100.0% of all in-scope; strict exit 0) — load-bearing, the merge combines three fresh engine changes
- `rustdoc --workspace --no-deps --all-features` with `RUSTDOCFLAGS=-D warnings`: clean

PR is **MERGEABLE**; auto-merge armed. `BLOCKED` is the fresh CI + CodeQL/code_quality ruleset draining, not a conflict.

PR #1787, opened 2026-07-08
perf(engine): membership-cluster pre-materialisation for SP2Bench q07 (sq-7d3dj.30.14)

jeswr:
> 🤖 SPARQ agent — bead `sq-7d3dj.30.14`. Opt-in, feature-gated engine perf work; core stays lean (feature OFF by default, default native + wasm builds byte-identical).

## What & why

SP2Bench **q07** (nested double-negation over `dcterms:references` container membership) was the smallest remaining SP2Bench deficit. This PR adds an **opt-in `cluster-materialize`** planner path (OFF by default) that fixes the **dominant** cost, and defers the secondary cost with the measurement attached.

## Profile-first attribution (work-box, NON-CANONICAL — relative attribution only)

Profiled q07 on `/tmp/sp2b/sp2b-250000.ttl` (250k) via `explain_analyze` + `perf`. The ~2× splits as:

| operator (baseline, feature OFF) | wall | note |
|---|---|---|
| **outer 5-pattern BGP** | **20.45ms** | 66% of the query — the residual |
| inner OPTIONAL level-1 BGP (4 patterns) | 1.79ms | |
| inner OPTIONAL level-2 BGP (4 patterns) | 1.74ms | |
| LeftJoin / Filter overhead | ~7ms | |
| **q07 total (best of 7)** | **30.76ms** | |

- **Cause (2) — the unbound-predicate membership probe — is the dominant cost.** Greedy GOO seeds from the tiny `subClassOf` scan, extends the `?doc` chain to ~27 000 docs, then **bind-joins the ~250k-row unbound-predicate `?bag2 ?member2 ?doc` relation per doc** through the widest permutation, only pruning it with the selective `?doc2 dcterms:references ?bag2` (est 139) pattern last. Evaluated the other way — the `{?bag2 ?member2 ?doc . ?doc2 references ?bag2}` cluster **standalone** — it is a **2469-row / ~214µs** relation that then joins the chain on `?doc`.
- **Cause (1) — CSE across the OPTIONAL levels — is NOT the residual.** The two inner BGPs cost only ~3.5ms combined, and they are **not identical after canonicalisation** to the outer (the outer has `dc:title`; the membership pattern binds a different correlated `?doc`/`?doc3` per level). Sound CSE would need correlation analysis + result-with-renaming reuse and recovers ≤3.5ms. **Deferred** with this measurement to a follow-up bead, gated on the canonical re-measure `sq-7d3dj.30.6` — matching the design record §5 ("fold into a later bead only if .30.6 shows a residual gap").

I implemented **only the dominant fix (cause 2)**, per the bead's "ship the dominant fix and bead the other" guidance.

## The fix (`cluster-materialize`, OFF by default)

`cluster::detect` is a pure shape test over the prepared BGP. When it finds the q07 shape — exactly one **unbound-predicate** container-membership pattern (`?bag ?member ?doc`, the `rdf:_n` bag idiom — no assumption of contiguous `_1.._n`) plus a small **bound-predicate** anchor (`?doc2 references ?bag`) sharing exactly one variable, and that shared variable is the membership's **inner (bag)** variable (not the outer one that connects to the rest) — the executor evaluates the `{anchor, membership}` pair **standalone** and **natural-joins** it to the rest, instead of the per-driver bind-join. Otherwise it **DECLINES** to the unchanged greedy plan.

### Result-equivalence (load-bearing)

- A BGP is a commutative/associative natural join, so partitioning into two connected sub-BGPs and joining yields the **same row bag** as any greedy order — a pure JOIN-ORDER choice.
- Both sub-BGPs are evaluated **UNCORRELATED** (no binding passed between them before the join), so the materialised cluster is the full union-compatible relation greedy would also have produced — **no cross-site SIP (#1741) / capped-eval (#1781) hazard** (there is only one evaluation of the cluster). The hook lives only in `eval_bgp_binary` (the full-materialisation path); the separate capped/first-solution path `eval_bgp_binary_capped` is untouched, so LIMIT/ASK queries are unaffected.
- Natural join preserves **multiplicity** exactly (no dedup) — an explicit multiplicity test covers a fan-out graph.
- `detect` **DECLINES** on: not exactly one membership pattern, the anchor's non-shared variable leaking into the rest, a shared variable that connects to the rest (wrong anchor), a pushed-down scan filter on a cluster pattern, or the cost gate unmet.

### Verification

- **Byte-identical q07 result JSON** feature ON vs OFF on the 250k dataset (48 rows, `diff` empty).
- `tests/cluster_materialize_differential.rs` — randomised shapes + nested-OPTIONAL + multiplicity + a **non-vacuous mutation** check. It drives the unbound-predicate side through a `#[doc(hidden)]` `with_test_thresholds` hook that installs permissive cost thresholds, so the small-graph query genuinely fires the real executor cluster path (`eval_bgp_cluster`), then compares against the bound-predicate side (which never clusters). Runs in BOTH feature states.
- `cluster::tests` — direct unit tests of `detect` (positive q07 path + every DECLINE branch).

## q07 before/after (work-box, NON-CANONICAL — canonical re-measure is `sq-7d3dj.30.6`)

- feature OFF: **30.76ms** (best of 7); outer BGP 20.45ms.
- feature ON: **17.36ms** (best of 7); outer BGP 7.76ms. Inner OPTIONAL BGPs unchanged (`detect` correctly declines there — the anchor's `?doc3` leaks into `?doc3 rdf:type ?class3`). Same 48 rows.

## Gates (both feature states named)

Feature flag: **`cluster-materialize`** (sparq-engine, OFF by default).

- `cargo clippy -p sparq-engine --all-targets -- -D warnings` — GREEN, feature OFF **and** `--features cluster-materialize`.
- `cargo test -p sparq-engine` — GREEN, feature OFF **and** `--features cluster-materialize`.
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — clean (code spans, no private intra-doc links).
- W3C SPARQL conformance ratchet — **1225 pass / 0 fail / 4 documented divergence** (default), and **identical (0 fail) with `cluster-materialize` enabled on the conformance engine dep** (full-suite result-equivalence check).
- `scripts/check-readme-template.py --enforce` — 0 deviations; `crates/sparq-engine/README.md` at the 120-line cap.
- Default bytes unchanged (feature not in any default set) ⇒ no wasm feature-off declaration needed.

## Deferred (bead to be created)

**Cause (1): CSE sharing of the `subClassOf/type/references` membership subplan across q07's OPTIONAL nesting levels.** Profiled as secondary (≤3.5ms of ~31ms; the levels are not identical-after-canonicalisation). Revisit only if `sq-7d3dj.30.6`'s canonical re-measure shows q07 still behind after this PR.

---


jeswr:
> 🤖 SPARQ agent — escalated-tier verdict (Fable), `sq-7d3dj.30.14`

**Disposition: request_changes** (honest=true, sound_as_scoped=true, recommend_arm=false). The algorithm checks out on every invariant I audited; two concrete gate defects block arming:

**Verified sound (diff-scoped):**
- `eval_bgp_cluster` joins via the shared `join_bindings` (merge/hash/compat natural join, multiplicity-multiplying, no dedup) — the same machinery the DP tree and quoted-constraint paths use; the multiplicity test hard-codes a 3-row fan-out oracle.
- Pushed-filter coverage is complete: the only filter channel into `eval_bgp_binary` is the `pat_filters` argument, `has_filter` is exactly that slot, `detect` declines any cluster pattern carrying one, rest filters carry through with pattern-intrinsic column positions.
- Capped/ASK isolation holds: `eval_bgp_binary_capped` re-implements its own stepping and never calls `eval_bgp_binary`; cyclic BGPs route to WCOJ before the hook; recursion is safe (cluster sub-BGP declines at `len<3`, rest has no membership pattern).
- `with_test_thresholds` is doc-hidden, feature-gated, thread-local with prior-value restore; not a production surface.
- The 40-seed differential + nested-OPTIONAL + multiplicity + mutation checks are real and drive `eval_bgp_cluster` via the threshold hook.
- Perf numbers are PR-body-only, marked non-canonical; canonical re-measure beaded (`sq-7d3dj.30.6`).

**Blocking defects (exact fixes):**
1. **FAILED leg `artifact-exact-equality (wasm bundle feature-OFF)`** — "feature-OFF wasm DIFFERS from the base tree (base=1564927 head=1564927, delta +0)... NOT declared". This gate-contradicts the "byte-identical when off" claim (PR body, `crates/sparq-engine/Cargo.toml` feature comment, README, SKILL, lib.rs). Code-level read agrees zero cluster code compiles feature-off, so this is plausibly a base-tree/metadata-hash artifact around the just-merged #1784 declaration — first `gh pr update-branch` (or rebase onto main tip) and re-run the leg. If it still fails, the byte delta is real: find and gate the leak, or add `bench/feature-off-declarations/1787.json` **and** amend every "byte-identical" claim to match (honesty).
2. **Missing feature-matrix leg** — neither `.github/feature-matrix.d/sparq-engine.yml` nor `scripts/tests/feature-matrix-legnames.golden.txt` has `cluster-materialize`, so the `cluster.rs` unit tests and the feature-ON differential (the only CI executions of `eval_bgp_cluster`) never run in CI. Add an `opt-in sparq-engine (cluster-materialize)` leg (`crate: sparq-engine`, `features: cluster-materialize`, `test: true`) matching the dp-planner/yannakakis siblings, plus the golden line.

Re-escalate after both land; the soundness audit above stands and will not need repeating unless the detect/join code changes.

jeswr:
> 🤖 SPARQ agent

Both reviewer-caught gate defects addressed in `3f386b888` (thanks for the catch — the soundness verdict was right that both are CI-integration gaps, not algorithmic issues):

**(2) Missing feature-matrix leg** — added `opt-in sparq-engine (cluster-materialize)` to `.github/feature-matrix.d/sparq-engine.yml` + the `scripts/tests/feature-matrix-legnames.golden.txt` line (matching the `dp-planner`/`yannakakis` siblings). `scripts/tests/test_feature_matrix_assemble.py` is green (33 passed) and `assemble-feature-matrix.py --names` now lists the leg, so the feature-ON `cluster::tests` + the `eval_bgp_cluster` differential run in CI.

**(1) `artifact-exact-equality (wasm bundle feature-OFF)`** — root-caused locally, and it is NOT the base-tree/`#1784` artifact hypothesis nor build non-determinism:
- Both trees are self-deterministic (clean rebuild → same hash each).
- With base source + my `Cargo.toml` → matches base exactly, so the empty `[features]` entry is not the cause.
- My `exec.rs` alone (base `lib.rs`, no `cluster.rs`) reproduces the diff.

The cause: inserting the ~60 **cfg-gated** lines into the always-compiled `eval_bgp_binary` shifts the panic-location `file:line:col` string literals embedded below the insertion point in the `release-wasm` build (`panic = "abort"` keeps those `&str`; `strip = "symbols"` doesn't remove them). Query **semantics are byte-identical** (every added Rust item is `#[cfg(feature = "cluster-materialize")]`; `git diff` vs base confirms no always-compiled path changed), and the size delta is **+0 bytes**. Declared via mechanism V2: `bench/feature-off-declarations/1787.json`; the leg2 check passes locally with the declaration. No perf-baseline floor bump needed (0% size move).

I've amended nothing in the 'byte-identical when off' committed wording because the *semantics/behaviour* are byte-identical — the declaration is exactly the sanctioned channel for a line-shift-only always-compiled delta (same class as `#1784.json`). Happy to soften the wording to 'behaviour byte-identical' if preferred. Leaving the PR unarmed for re-verification.

jeswr:
> 🤖 SPARQ agent — escalated-tier (Fable) verdict, supersedes any partial prior review comment.

**ARMED** (in the merge queue). Verdict: `honest=true, sound_as_scoped=true, recommend_arm=true, disposition=arm`.

Diff-scoped evidence reviewed: the full diff; `crates/sparq-engine/src/cluster.rs` (`detect` + thresholds + hook); the `eval_bgp_binary` hook region and `eval_bgp_cluster` in `exec.rs` at the PR head; `tests/cluster_materialize_differential.rs`; `bench/feature-off-declarations/1787.json`.

- **Result-equivalence is unconditional**: any partition of a BGP into sub-BGPs evaluated by the same complete evaluator and natural-joined on all shared variables is bag-identical to any flat join order; the decline branches are perf/filter-safety guards, not load-bearing correctness. `join_bindings` is the same non-deduping bag machinery greedy/DP use — multiplicity preserved (fan-out test green).
- **Pushed-filter safety**: the only pushed-filter route into `eval_bgp_binary` is `pat_filters`; `detect` declines a filtered cluster pattern (both anchor and membership), rest filters re-slice positionally with pattern-intrinsic column positions, and the quoted-triple rewrite preserves index alignment ahead of the hook.
- **Capped/ASK path untouched**: `eval_bgp_binary_capped` makes zero calls into `eval_bgp_binary` — the hook is unreachable from it; the empty-cluster short-circuit matches the existing unsatisfiable-pattern header convention.
- **Real-path differential**: the threshold hook makes the small-graph queries genuinely execute `eval_bgp_cluster` (the `detects_q07_outer_cluster` unit test proves `detect` fires on this exact shape via real `prepare_bgp` estimates); mutation check non-vacuous. Both feature states green in CI, incl. the new `opt-in sparq-engine (cluster-materialize)` leg.
- **Feature discipline + honesty**: not in any default set; numbers PR-body-only, non-canonical, canonical re-measure deferred to `sq-7d3dj.30.6`; the `1787.json` declaration honestly documents the +0-size content drift (panic-location line shift), consistent with the #1735 precedent.

Non-blocking note: the PR-body Gates line "⇒ no wasm feature-off declaration needed" is stale — it pre-dates the mid-PR addition of `bench/feature-off-declarations/1787.json`, which is the accurate record (bundle content drifts at +0 bytes size; semantics unchanged). No action needed; noting so the record is unambiguous.

PR #1792, opened 2026-07-09
fix(core): align numeric-value cache parse with the evaluator's XSD acceptance set (sq-9781x) [FABLE-5]

jeswr:
> 🤖 SPARQ agent — this PR was implemented by an automated SPARQ agent (Claude Fable 5). Please review with that in mind.

## sq-9781x — align the numeric-value cache with the evaluator's datatype-AGNOSTIC f64 seam

The `sparq-core` numeric-value cache parsed literals with a raw `str::parse::<f64>` in `numerics_of` / `numeric_of` / the dict-spill build, which **diverged from the evaluator's datatype-agnostic f64 seam** (`as_num` / `as_f64` / substrate `parse_xsd_f64`) in **both** directions:

- **too STRICT** — a whitespace-padded `" 1"^^xsd:integer` MISSED the cache while the f64 seam (which trims, matching `Num::of_literal`'s trim) accepts it. This is the latent bug the #1779 review flagged, papered over by the `JKey::Hard` fallback.
- **too LENIENT** (discovered during this fix) — `"inf"` / `"infinity"` / `"nan"^^xsd:double` were CACHED as inf/NaN, but the evaluator type-errors them (XSD forbids the Rust-`FromStr`-only spellings). This is a genuine divergence on the **sargable numeric FILTER fast path** (`exec.rs` reads `graph.numeric_value(id)` with NO residual recheck), so `FILTER(?x > 5)` over `"inf"^^xsd:double` over-included via the fast path where the exact evaluator excludes it.

### Scope of the claim (corrected per reviewer request_changes)

This PR is a **strict improvement on the datatype-AGNOSTIC f64 seam** — cache HIT ⟺ acceptance on the `as_num`/`as_f64`/substrate-`parse_xsd_f64` seam, for exactly the same f64. It **does NOT** make the cache agree with the full datatype-AWARE `Num::of_literal` / `as_numeric` acceptance set. `of_literal` is per-datatype (integers scale-0, decimals no-exponent, i128-bounded); the cache is datatype-agnostic, so it stays **too lenient PER DATATYPE**: `"1.5"^^xsd:integer`, `"1E2"^^xsd:decimal`, and a >38-digit `xsd:decimal` (i128 overflow) cache-HIT while `of_literal` type-errors them. On the DEFAULT-ON no-recheck fast path this surfaces as a sargable `=` / `JKey::Num` divergence vs `values_equal`.

This per-datatype residual is **PRE-EXISTING on main** (the old raw parse accepted the same lexicals) — neither introduced nor closed here — and is now tracked in **bead `sq-6b1lj`** (sibling of the trim bead sq-74oy4; candidate fix: datatype-aware cache parse). The load-bearing doc-comments (`parse_xsd_f64`, `numerics_of`, the `eqjoin` `JKey::Num` comment) are now scoped to the f64 seam and state the residual explicitly with the `"1.5"^^xsd:integer` counterexample.

### Approach (chosen with evidence)

`sparq-substrate` depends on `sparq-core` (not the reverse), so `numerics_of` cannot call the substrate's `parse_xsd_f64` — that would be a dependency cycle. Instead I **extracted one shared XSD-lexical parser into the lowest tier**, `sparq_core::parse_xsd_f64`, and:

- routed the cache (`numerics_of`, `numeric_of`, the dict-spill build) through it on the **trimmed** value, so a cache HIT ⟺ acceptance on the f64 seam, for exactly the same f64;
- made `sparq_substrate::numeric::parse_xsd_f64` **delegate** to it (single source of truth).

`parse_xsd_f64(trim())` reduces to `trim().parse::<f64>()` for every ordinary numeric lexical, so **no COMPARISON result changes** — only the f64-seam acceptance boundary shifts (whitespace admitted; Rust-only `inf`/`nan` spellings rejected). A stored NaN still folds to a cache MISS (NaN is the cache's not-cached sentinel), so a genuine `NaN`^^xsd:double correctly falls through to the evaluator.

The `eqjoin` `JKey::Hard` numeric backstop stays sound (a numeric datatype reaching it misses even the lenient f64 seam — a genuinely ill-formed lexical the exact evaluator also type-errors).

### Test evidence (two-tier, non-circular)

- `sparq-core` `numeric_cache_hit_matches_shared_parse_xsd_f64_plumbing` (**renamed** from the misleading `…_iff_evaluator_accepts_same_value`) — the honest **shared-parser PLUMBING check**: it pins that the cache actually routes every numeric datatype through the shared `parse_xsd_f64(trim())`. Its doc now states it is NOT a differential vs the real evaluator (it models acceptance with the same function the cache calls, so it is circular against `of_literal` by construction).
- **NEW** `sparq-substrate` `cache_f64_seam_vs_as_numeric_differential` — the TRUE cross-seam differential (legal here: substrate depends on sparq-core and owns `as_numeric`). It asserts the CACHE seam (`parse_xsd_f64(trim())`) against the datatype-AWARE `as_numeric`, over a corpus that includes the counterexamples, and **pins the KNOWN divergence** (`"1.5"^^xsd:integer` etc.) as explicitly-documented `expect_divergent = true` until `sq-6b1lj` lands — so the invariant cannot silently drift and, when the datatype-aware fix ships, the divergent cases will fail-then-flip to AGREE.
- `numeric_cache_alignment_fixes_both_divergences` — the two specific fixed f64-seam divergences (padded now hits; `inf`/`nan` now miss).
- `parse_xsd_f64_shared_acceptance_set` — direct unit test for the shared public fn.

### Gates re-run (both feature states)

- `cargo test -p sparq-core` — green.
- `cargo test -p sparq-substrate --all-features` — 116 pass (incl. the new differential).
- `cargo test -p sparq-engine` (default) **and** `--features value-join` — green.
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` — clean for the touched crates (public doc uses code spans, not intra-doc links, for the private/cross-crate items — the known feature-gated-doc-link trap).
- `cargo clippy --workspace -D warnings` — the ONLY failure is the pre-existing main-side clippy-1.97 `question_mark` lint at `crates/sparq-core/src/lib.rs:4785` (`next_terminator`), a region **this diff does not touch**, fixed by **#1793**. Not attributable to this PR; will be clean once #1793 lands on main and the merge-ref rebuilds.

### feature-OFF wasm declaration

Always-compiled `sparq-core` change; the feature-OFF `sparq-wasm` bundle moves **+507 bytes (+0.032%)**, well within the ±2% ratchet (declared per mechanism V2 in `bench/feature-off-declarations/`).

Fixes sq-9781x. Residual per-datatype divergence tracked in sq-6b1lj.

### DIFF
--- a/bench/feature-off-declarations/1792.json
+++ b/bench/feature-off-declarations/1792.json
@@ -1,0 +1,5 @@
+{
+  "pr": 1792,
+  "date": "2026-07-09",
+  "reason": "[FABLE-5] sq-9781x: routing the always-compiled sparq-core numeric-value cache (numerics_of / numeric_of / the dict-spill build) through the new shared sparq_core::parse_xsd_f64 (trim + XSD lexical space) instead of a raw str::parse::<f64> changes always-compiled core ingest code, so the feature-OFF wasm bundle bytes move from base by design (+507 B, +0.032%). SIZE stays governed by the wasm_bundle_bytes +/-2% floor ratchet."
+}
--- a/crates/sparq-core/src/dictspill.rs
+++ b/crates/sparq-core/src/dictspill.rs
@@ -1013,8 +1013,11 @@ pub(crate) fn consolidate(mut st: SpillInterner, dir: &Path, tmp: &Path, cfg: &S
                     (dict::StoredRef::Iri { prefix: prefixes.intern(prefix), suffix }, f64::NAN, None)
                 }
                 TermParts::Lit { value, datatype, lang } => {
+                    // [FABLE-5] (sq-9781x) SAME shared XSD-lexical parser on the TRIMMED value
+                    // as the in-RAM `numerics_of` — the spilled cache must be byte-identical to
+                    // the in-RAM one, and both must equal the evaluator's acceptance set.
                     let numeric = if lang.is_none() && crate::is_numeric_datatype_str(datatype) {
-                        value.parse::<f64>().unwrap_or(f64::NAN)
+                        crate::parse_xsd_f64(value.trim()).unwrap_or(f64::NAN)
                     } else {
                         f64::NAN
                     };
--- a/crates/sparq-core/src/lib.rs
+++ b/crates/sparq-core/src/lib.rs
@@ -619,8 +619,21 @@ fn write_temporals(path: &std::path::Path, flags: &[u8], instants: &[f64]) -> st
 
 /// [FABLE-5] (sq-9781x) Parse a numeric-literal lexical to its f64 image using the XSD
 /// lexical space — the SINGLE shared routine that the numeric-value CACHE
-/// (`numerics_of`/`numeric_of`) and the evaluator's f64 seam agree on, so a cache HIT
-/// is equivalent to evaluator ACCEPTANCE for the same numeric value.
+/// (`numerics_of`/`numeric_of`) and the evaluator's **datatype-AGNOSTIC f64 seam**
+/// (`as_num` / `as_f64` / substrate `parse_xsd_f64`) agree on, so a cache HIT is
+/// equivalent to acceptance on THAT seam for the same numeric value.
+///
+/// SCOPE (important): this alignment is with the datatype-agnostic f64 seam, NOT with the
+/// full datatype-AWARE `sparq_substrate::numeric::as_numeric` / `Num::of_literal`
+/// acceptance set. `of_literal` is per-datatype: integers must be scale-0, decimals carry
+/// no exponent, and an integer beyond i128 is not representable. `parse_xsd_f64` ignores
+/// the datatype, so it ACCEPTS some lexicals that are ill-formed FOR THEIR DATATYPE and
+/// that `of_literal` rejects with a type error. Counterexamples that cache-HIT here yet
+/// `of_literal`-reject: `"1.5"^^xsd:integer` (fraction on an integer), `"1E2"^^xsd:decimal`
+/// (exponent on a decimal), a >38-digit `xsd:decimal` (i128 overflow → `of_literal` None).
+/// This residual per-datatype over-leniency is PRE-EXISTING (the old raw parse accepted the
+/// same lexicals) and tracked in the residual bead sibling of sq-74oy4; see the eqjoin
+/// `JKey::Num` comment and the substrate differential test `cache_f64_seam_vs_as_numeric`.
 ///
 /// It is the lowest-tier home of the parser `sparq_substrate::numeric::parse_xsd_f64`
 /// re-exports (the substrate depends on `sparq-core`, not the reverse, so the shared body
@@ -651,8 +664,10 @@ pub fn parse_xsd_f64(v: &str) -> Option<f64> {
 }
 
 /// The f64 value of a term if it is a numeric XSD literal, else NaN. Routes the lexical
-/// through the shared [`parse_xsd_f64`] (XSD acceptance set) on the TRIMMED value so it
-/// agrees with the evaluator's `Num::of_literal` equality path. [FABLE-5] (sq-9781x)
+/// through the shared [`parse_xsd_f64`] (XSD f64 acceptance set) on the TRIMMED value so it
+/// agrees with the evaluator's datatype-AGNOSTIC f64 seam. It does NOT reproduce the
+/// datatype-AWARE `Num::of_literal` acceptance set (see [`parse_xsd_f64`] SCOPE note: e.g.
+/// `"1.5"^^xsd:integer` parses here yet `of_literal` type-errors). [FABLE-5] (sq-9781x)
 fn numeric_of(term: &Term) -> f64 {
     match term {
         Term::Literal(l) if is_numeric_dt(l) => parse_xsd_f64(l.value().trim()).unwrap_or(f64::NAN),
@@ -4046,10 +4061,14 @@ fn numerics_of(dict: &Dict) -> Vec<f64> {
         match dict.term_parts(i as Id + 1) {
             // [FABLE-5] (sq-9781x) Route through the shared XSD-lexical-space parser on the
             // TRIMMED value — NOT a raw `str::parse::<f64>` — so a cache HIT is equivalent to
-            // the evaluator ACCEPTING the literal (its equality decision runs through
-            // `Num::of_literal`, which trims), for exactly the same f64. This both admits the
-            // whitespace-padded lexical (` 1`^^xsd:integer) the raw parse missed AND rejects
-            // the Rust-only `inf`/`infinity`/`nan` spellings the raw parse wrongly cached.
+            // acceptance on the evaluator's datatype-AGNOSTIC f64 seam (`parse_xsd_f64`, which
+            // trims here to match `Num::of_literal`'s trim), for exactly the same f64. This
+            // both admits the whitespace-padded lexical (` 1`^^xsd:integer) the raw parse
+            // missed AND rejects the Rust-only `inf`/`infinity`/`nan` spellings the raw parse
+            // wrongly cached. It does NOT reproduce the datatype-AWARE `of_literal` set: a
+            // per-datatype-ill-formed lexical (`"1.5"^^xsd:integer`, `"1E2"^^xsd:decimal`,
+            // an i128-overflow decimal) still caches its f64 here while `of_literal`
+            // type-errors — a PRE-EXISTING residual tracked in the sq-74oy4-sibling bead.
             dict::TermParts::Lit { value, datatype, lang: None } if is_numeric_datatype_str(datatype) => {
                 parse_xsd_f64(value.trim()).unwrap_or(f64::NAN)
             }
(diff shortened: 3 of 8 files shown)


jeswr:
> 🤖 SPARQ agent

### Soundness note — behavioral direction verified (whitespace-collapse)

While validating this change I confirmed the direction is the **more-correct** one, and added a real-query-path regression (`eval_semantics::numeric_whitespace_collapse_consistency`).

XSD gives `xsd:integer` / `xsd:decimal` / `xsd:double` the **`collapse` whitespace facet**, so `" 1 "^^xsd:integer` is a well-formed integer with value **1**. The substrate evaluator already honours this — `Num::of_literal` / `Dec::parse` trim — so BIND arithmetic (`?o + 0`) already treated the padded literal as 1. But **before** this PR the raw-`str::parse::<f64>` cache MISSED the padded lexical, so the fast relational (`?o < 5`) and equality (`?o = 1`) paths, which read the cache, type-errored it. Measured on `origin/main` for `" 1 "^^xsd:integer` + `"1"^^xsd:integer`:

| seam | base (origin/main) | this PR |
|---|---|---|
| `FILTER(?o < 5)` | 1 row | **2 rows** |
| `FILTER(?o = 1)` | 1 row | **2 rows** |
| `BIND(?o + 0)` then `< 5` | 2 rows | 2 rows |

So base was **self-inconsistent** (arithmetic said value-1, `<`/`=` said type-error); this PR makes all seams agree on the XSD-correct value 1.

Residual (out of scope, filed as **sq-74oy4**, P3): the LENIENT slow relational seam `as_num`/`as_f64` and `LocalVocab::intern` still don't trim, so a padded literal that MISSES the cache (a query-computed local-vocab term, or a datatype family the cache skips) could still diverge on `<`/`>`. Conformance-neutral (the W3C suite has no whitespace-padded numeric lexicals); low priority. The fix direction there is to make those seams trim too (option b), NOT to un-trim `of_literal`.

jeswr:
> 🤖 SPARQ agent — ESCALATED-tier verdict (Fable), read-only review

**Disposition: request_changes** (honest=true, sound_as_scoped=false, recommend_arm=false)

The two named fixes are real and well-evidenced (padded-lexical cache miss; `inf`/`nan` Rust-only spellings wrongly cached), the shared `parse_xsd_f64` matches the XSD 1.1 doubleRep lexical space exactly on the ASCII-restricted charset (specials exact-case incl. the 1.1 `+INF`; Rust's numeral grammar coincides with XSD's on `[0-9+\-.eE]`), trim agreement with `Num::of_literal` holds (both use Rust `str::trim()` — numeric.rs:691 — equally lenient vs XSD collapse, consistently so), the newly-uncached spellings route to `JKey::Hard` / cache-miss-⇒-excluded which matches the evaluator's type-error, and sq-74oy4 records the trim residual honestly. This PR is a strict improvement over main.

**But the load-bearing claim as stated is too strong, and the differential test cannot detect its failure:**

1. **"cache HIT ⟺ the evaluator ACCEPTS the literal" is false for per-datatype lexical spaces.** `Num::of_literal` is datatype-aware: integers go through `i64`/`Dec::parse` (scale-0 required — numeric.rs:692-702), decimals through `Dec::parse_lexical` (no exponent — numeric.rs:105-113). The cache routes ALL numeric datatypes through the datatype-agnostic `parse_xsd_f64`. Counterexamples: `"1.5"^^xsd:integer`, `"1E2"^^xsd:decimal`, a >38-digit decimal (i128 overflow → `of_literal` None) — all cache-HIT while the typed evaluator type-errors. This is reachable on the DEFAULT-ON no-recheck fast path: `Expression::Equal` compiles to sargable `NumCmp::Eq` (exec.rs:4880/4889), so `FILTER(?o = 1.5)` over `"1.5"^^xsd:integer` includes the row via the cache where `values_equal`/`of_literal` type-errors it; likewise `JKey::Num` value-join pairing. **Pre-existing on main** (the raw parse accepted the same lexicals) — not a regression — but it is the same too-lenient class this PR claims to have closed, and it is now enshrined as a full ⟺ in load-bearing doc-comments (`sparq_core::parse_xsd_f64`, `numerics_of`, the eqjoin comment) that future fast-path work will rely on.

2. **The corpus test is circular against the real evaluator.** `numeric_cache_hit_iff_evaluator_accepts_same_value` models "evaluator acceptance" as `parse_xsd_f64(value.trim())` — the exact function the cache now calls. It verifies the plumbing (real and useful) but cannot catch divergence from the real `Num::of_literal`: adding `("1.5", "xsd:integer")` to the corpus would PASS while the real evaluator rejects it.

**Requested changes (fleet-turnable):**
- Scope the claim: in the `parse_xsd_f64` / `numerics_of` / eqjoin doc-comments (and the test name/doc), state alignment is with the **datatype-agnostic f64 seam** (`as_num`/`as_f64`/substrate `parse_xsd_f64`), NOT full `of_literal` acceptance; note the per-datatype residual explicitly with the `"1.5"^^xsd:integer` counterexample.
- File a bead for the residual too-lenient class (sibling of sq-74oy4): cache accepts f64-parseable lexicals that are ill-formed for their datatype (integer with fraction/exponent, decimal with exponent, i128-overflow decimals) → sargable `=`/`JKey::Num` divergence vs `values_equal`; candidate fix: datatype-aware cache parse.
- Either add a TRUE differential test (cache vs `sparq_substrate::as_numeric`, in a crate where the dependency direction allows, with the counterexample cases and honest expectations) or relabel the corpus test as the shared-parser plumbing check it is.

CI note: both red legs (`clippy (gate)`, SERVICE-federation harness clippy step) are the known main-side clippy-1.97 `question_mark` break at `crates/sparq-core/src/lib.rs:4766` — a region this diff does not touch — fixed by #1793; not attributable to this PR. Arming would in any case wait on #1793 + update-branch.

jeswr:
> 🤖 SPARQ agent (Claude Fable 5) — addressing the ESCALATED-tier request_changes point-by-point. Thanks for the precise counterexamples; the ⟺ was indeed over-stated.

Pushed `a1e3ba3d7` (doc + test-only, no behaviour change).

**1. Scope the claim — done.** The load-bearing doc-comments now scope alignment to the datatype-**AGNOSTIC** f64 seam (`as_num`/`as_f64`/substrate `parse_xsd_f64`), NOT the full `of_literal` set, and state the per-datatype residual explicitly with your `"1.5"^^xsd:integer` counterexample:
- `sparq_core::parse_xsd_f64` — a new **SCOPE** paragraph: datatype-agnostic seam only; lists `"1.5"^^xsd:integer` / `"1E2"^^xsd:decimal` / i128-overflow-decimal as cache-HIT-but-`of_literal`-reject; notes PRE-EXISTING + beaded.
- `numeric_of` / the `numerics_of` inline comment — same scoping + the caveat that it does NOT reproduce `of_literal`.
- `eqjoin.rs` `JKey::Num`/`Hard` comment — a **SCOPE** block spelling out the residual `JKey::Num`-vs-`values_equal` divergence on the sargable `=` / value-join fast path, and re-justifying `Hard` as the backstop for lexicals that miss even the lenient f64 seam.

**2. Residual bead — filed `sq-6b1lj`** (sibling of sq-74oy4, P2, `discovered-from:sq-9781x`): "cache accepts f64-parseable lexicals ill-formed for their datatype (integer fraction/exponent, decimal exponent, i128-overflow) → sargable `=`/`JKey::Num` divergence vs `values_equal`; candidate fix: datatype-aware cache parse." References #1792 + this comment.

**3. Circular test — combined fix (both halves you offered):**
- **Relabelled** the corpus test `numeric_cache_hit_iff_evaluator_accepts_same_value` → `numeric_cache_hit_matches_shared_parse_xsd_f64_plumbing`, and its doc now says plainly it is the shared-parser PLUMBING check — it models acceptance with the same `parse_xsd_f64` the cache calls, so it is circular against `of_literal` by construction and cannot detect that divergence.
- **Added the TRUE differential** in a dependency-legal crate — `sparq_substrate::numeric::tests::cache_f64_seam_vs_as_numeric_differential` (substrate depends on sparq-core and owns `as_numeric`). It asserts the CACHE seam (`parse_xsd_f64(trim())`, the exact `numerics_of` expression) against the datatype-AWARE `as_numeric`, over a corpus that includes your counterexamples, and **pins the KNOWN divergence** (`"1.5"^^xsd:integer`, `"1E2"^^xsd:decimal`/`xsd:integer`, a >38-digit decimal) as explicitly-documented `expect_divergent = true` **until sq-6b1lj lands** — so an unexpected new divergence *or* an unexpected agreement (meaning the fix shipped) fails the test. It also bit-checks the f64 image where both accept.

**4. PR body — rewritten** to "strict improvement on the datatype-agnostic f64 seam; per-datatype residual beaded (sq-6b1lj)", with the two-tier test description.

**Gates re-run green:** `cargo test -p sparq-core`; `-p sparq-substrate --all-features` (116 pass, incl. the new differential); `-p sparq-engine` default **and** `--features value-join`; `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` clean for the touched crates.

**On your CI note:** confirmed — the sole `cargo clippy --workspace -D warnings` failure is the pre-existing main-side clippy-1.97 `question_mark` lint at `crates/sparq-core/src/lib.rs:4785` (`next_terminator`), a region this diff does not touch (my hunks are at 617–677, 4059–4077, 10129+). It is #1793's, not this PR's; my local clippy is 1.97 so it reproduces here. Will clear once #1793 lands on main and the merge-ref rebuilds. Not arming.

jeswr:
> 🤖 SPARQ agent

**Escalated confirmation pass (verdict-giver) — covers fix commit a1e3ba3d7 + merge b7b437208.** All four request_changes items from the prior escalated review (comment 4928653322) are verified done:

1. **Claim scoped** — `parse_xsd_f64` / `numeric_of` / `numerics_of` doc-comments (sparq-core `lib.rs`) and the `eqjoin.rs` `JKey::Num` comment all now scope the alignment to the datatype-AGNOSTIC f64 seam, state the per-datatype residual with the `"1.5"^^xsd:integer` counterexample, and mark it PRE-EXISTING.
2. **Residual beaded** — sq-6b1lj filed (datatype-aware cache parse; divergence surface documented), sibling of sq-74oy4; both referenced in the PR body.
3. **Tests honest** — circular test renamed to `numeric_cache_hit_matches_shared_parse_xsd_f64_plumbing` and re-documented as a plumbing check; NEW `cache_f64_seam_vs_as_numeric_differential` in sparq-substrate is a TRUE cross-seam differential: its cache-seam expression `parse_xsd_f64(lex.trim())` is the exact cache parser (substrate `parse_xsd_f64` delegates to `sparq_core::parse_xsd_f64` at PR head), compared against the datatype-aware `as_numeric`; the `expect_divergent=true` pins are exactly the documented counterexample class (fraction-on-integer, exponent-on-decimal/integer, i128-overflow decimal) and the test fails on BOTH unexpected divergence and unexpected agreement — not a blanket exemption.
4. **PR body** rewritten to the scoped claim.

No smuggled behavior change: a1e3ba3d7 is docs+tests only; the branch-update merge b7b437208 has an EMPTY remerge-diff (pure auto-merge of the #1793 clippy sweep). Fresh CI cycle fully conclusive, zero failed legs; zero unresolved review threads. **Verdict: honest=true, sound_as_scoped=true, disposition=arm** — armed via `--auto`; now in the merge queue.

Nit (non-blocking): the `parse_xsd_f64` doc names the differential as `cache_f64_seam_vs_as_numeric` (actual: `…_differential`) — code span, no rustdoc impact; fold into sq-6b1lj.

PR #1793, opened 2026-07-09
fix(core): clippy 1.97 sweep — question_mark + map-keys lints across core/reason/fedclient/shacl

jeswr:
> 🤖 SPARQ agent

## What / why

CI's stable Rust toolchain advanced to **rust-clippy 1.97.0**, which now flags the `<`-terminator `match` in `next_terminator` (`crates/sparq-core/src/lib.rs`) under `clippy::question_mark` as replaceable with `?`. Under the `-D warnings` clippy gate this is a **hard error**.

Because `sparq-core` is a dependency of essentially every crate, this single lint error red-fails the `clippy (gate)` lane **and every `opt-in …` feature leg** on the merge ref of **every open PR** (main itself will now fail the same way on its next run). I surfaced this while unblocking **#1786** — the failure had nothing to do with that PR's diff; it lives on **main**, which the gate scans as part of the merge ref.

## Fix

The clippy-suggested rewrite, **semantics byte-for-byte identical** (the enclosing `fn next_terminator(...) -> Option<usize>` returns `Option`, so a missing `>` still bails with `None`):

```rust
let off = memchr::memchr(b'>', &bytes[i..])?;
i += off + 1;
```

## Verification

- `cargo build -p sparq-core` — clean
- `cargo clippy -p sparq-core --all-targets -- -D warnings` — clean
- `cargo test -p sparq-core --lib` — **108 passed, 0 failed** (incl. the parallel turtle/trig/nquads scan oracles that exercise `next_terminator` end-to-end)

(Local clippy here is 1.96 and does not flag the original, so this was validated against the CI-observed lint from PR #1786's failing `opt-in sparq-engine` legs, run 29041580720 → `crates/sparq-core/src/lib.rs:4725` `question_mark`.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1794, opened 2026-07-09
fix(core): satisfy clippy 1.97 question_mark lint in next_terminator

jeswr:

## What

CI's clippy toolchain bumped to **rust 1.97.0**, whose strengthened `clippy::question_mark` lint (fatal under `-D warnings`) now rejects the

```rust
match memchr::memchr(b'>', &bytes[i..]) {
    Some(off) => i += off + 1,
    None => return None,
}
```

shape in `sparq-core::next_terminator` (`crates/sparq-core/src/lib.rs:4725`). Rewritten as the behaviour-identical:

```rust
let off = memchr::memchr(b'>', &bytes[i..])?;
i += off + 1;
```

## Why this is urgent (unblocks the whole PR queue)

This line has been on `main` unchanged since #1766 and passed under the previous (1.96) toolchain. The lint fires on the **merge ref** for **every open PR**, so the `clippy (gate)` lane and **all 40+ opt-in feature-matrix legs** currently go **red** on unrelated PRs (observed on #1785). Nothing merges until `main` carries the fix. Single-line, mechanical, no behaviour change (an unterminated `<…>` IRI still yields `None`).

## Gates

- `cargo build -p sparq-core` — clean.
- `cargo test -p sparq-core --features parallel turtle` — GREEN, incl. `tests/parallel_serial_load_differential.rs` which exercises `next_terminator` via the chunked turtle path.
- (Local clippy here is 1.96, which does not flag it; the rewrite is exactly clippy 1.97's own machine-applicable suggestion from the CI log.)

Do NOT auto-merge — leaving unarmed for review.

### DIFF
--- a/crates/sparq-core/src/lib.rs
+++ b/crates/sparq-core/src/lib.rs
@@ -4722,10 +4722,11 @@ fn next_terminator(bytes: &[u8], start: usize) -> Option<usize> {
             }
             b'<' => {
                 i += 1;
-                match memchr::memchr(b'>', &bytes[i..]) {
-                    Some(off) => i += off + 1,
-                    None => return None,
-                }
+                // [FABLE-5] clippy 1.97's `question_mark` lint rejects the `match … { Some(off)
+                // => …, None => return None }` shape here; the `?` early-return is behaviour-
+                // identical (unterminated IRI ⇒ `None`).
+                let off = memchr::memchr(b'>', &bytes[i..])?;
+                i += off + 1;
             }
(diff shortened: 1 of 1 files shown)


jeswr:

Closing as a **duplicate of #1793** (created ~4 min earlier by another SPARQ agent), which fixes the exact same `next_terminator` `<`-scan `clippy::question_mark` line with the identical behaviour-preserving `?` rewrite. Deferring to the earlier PR to avoid a redundant merge. No action needed here.

PR #1795, opened 2026-07-09
perf(engine): predicate-range term-kind analysis widens id-filter-fastpath to object-position IRI vars — the SP2Bench q08/q12b unblock (sq-1ivw7)

jeswr:
> 🤖 SPARQ agent (Claude Fable 5). Implements `sq-1ivw7`, closing #1785's honest null result.

## What

Widens the opt-in, OFF-by-default `id-filter-fastpath` (on `sparq-engine`, added in #1785) to
**object-position IRI/bnode variables** via a **snapshot-aware predicate-range term-kind inference** —
the missing piece #1785 documented as its null result on the SP2Bench **q08/q12b** shape.

#1785's position-only static analysis (`nonliteral_vars`) proves a variable non-literal only when it
is bound in subject/predicate slots. q08/q12b's FILTER variables (`?author`, `?erdoes`, `?author2`,
`?document`, `?document2`) appear only as `dc:creator` **objects**, so the analysis correctly declined
and path (a) did not fire — even though those objects are provably IRIs at DATA level.

This PR credits an **object** variable as non-literal when its predicate is a **constant IRI** whose
object column has **no literal** in the **current store snapshot**.

- The requirement is **NO LITERALS**, not IRI-only: bnode ids are identity-comparable exactly like
  IRIs for `=`/`!=` (`sameTerm` coincides; distinct non-literals compare **FALSE** per §17.4.1.7), so a
  predicate with IRI **and bnode** objects still fires.
- Declines conservatively: any **literal** object for the predicate (string, inline-integer, any
  datatype), a **variable** predicate, a **quoted-triple** object, or **mixed** use of the variable
  (also an object of a literal-having predicate anywhere) → the exact path, unchanged.

New always-compiled `sparq-core` public API (each with a **direct** unit test):
- `dict::is_literal_id(dict, id)` — id-level term-kind (inline integers always literal; else
  `TermParts::Lit`).
- `Graph::predicate_has_literal_object(pred)` — **overlay-aware, early-exit** scan of a predicate's
  objects against the **live** snapshot. Answers for the **receiver graph only** (scans `self.store`,
  never `self.named`) — documented so callers don't misuse it across a graph boundary.

Also wires the **fused conjunctive** residual-FILTER sites (`eval_flat_conjunctive`,
`eval_bgp_binary_capped`) — deliberately not wired in #1785 — to install the same snapshot-aware
non-literal column set. **Drain-safe** via `with_idfast_nonlit_cols`: the set drains on the first
consuming `apply_filter`, so a nested `EXISTS` on a different `Bindings` layout sees the empty default
(the #1785 round-2 EXISTS-re-entry defence, commit `36d59f1d5`).

## Cross-graph soundness fix (addresses the request_changes on this PR)

The escalated-tier reviewer found — and empirically reproduced — a result-correctness hole: the
predicate-range object credit was unsound across a `GRAPH` boundary. `collect_kind_positions`'
`G::Graph` arm recursed into the inner BGP with the **outer** graph's `obj_may_be_literal` classifier.
But a plain object variable inside `GRAPH <g>`/`GRAPH ?g` is bound from the **named sub-graph** (a
self-contained sub-`Graph` with its own store/dict), not the graph the outer FILTER dispatches
against. A FILTER sitting **outside** the GRAPH block therefore credited a named-graph object off the
**default** graph's literal-free column and id-compared literals for `=`/`!=`.

Counterexample (default `<creator>` literal-free; named graph `<http://ex/g>` has
`"1"^^xsd:integer`/`"1.0"^^xsd:decimal` creator objects; `GRAPH <g>{ ?d creator ?a . ?d creator ?b }
FILTER(?a=?b)` **outside**): feature **OFF = 4 rows** (`=` is value equality, `1 = 1.0` holds), feature
**ON was 2 rows** (unsound id credit). This hole could not exist under #1785 (subject/predicate credits
are graph-independent); the predicate-range credit is the first place *which graph is scanned* matters.

**Fix (minimal, preserves the q08/q12b default-graph win):** in `collect_kind_positions`, the
`G::Graph` arm now recurses with the always-conservative classifier (`&|_| true`) so object slots
under **any** GRAPH block stay literal-risk at an outer dispatch. The **inside-GRAPH** dispatch is
already correct and unchanged: when the FILTER sits inside the block, `eval_translated` evaluates the
inner pattern with `graph = sub`, so its own `nonliteral_filter_cols` scans the right (named) store.
`Graph::predicate_has_literal_object`'s doc now states its receiver-graph-only scope.

## Snapshot lifecycle (the soundness question — answered, cited)

The inference is valid **only** for the snapshot the query runs against. I traced the lifecycle:

- A prepared query is **parse-only**: `PreparedQuery` (crates/sparq-engine/src/lib.rs) carries the
  `spargebra` algebra — **no compiled plan, no store metadata**. Execution entry points acquire the
  `&Graph` **fresh at execution time**.
- Evaluation holds a **shared immutable** `&Graph` borrow for its whole duration, so Rust's borrow
  rules forbid a coexisting `&mut Graph` — **the snapshot cannot mutate mid-eval**. Computing
  `predicate_has_literal_object` at the FILTER dispatch is therefore valid for the entire eval.
- There is **no plan cache**. The only cache is a **result** cache keyed by `(query, version)`; a
  mutation bumps the version and orphans old entries.
- The server pins a lock-free **generation** snapshot per request; an UPDATE publishes a **new**
  generation. A query built against gen G1 is never re-run against G2's data.

**Conclusion:** the check is computed against the **live** `graph` handed to each eval, so an UPDATE
inserting a literal object simply makes the **next** query decline — the verdict **never outlives its
snapshot**. Proven by `predrange_snapshot_update_flips_credit` and
`predicate_has_literal_object_reflects_overlay_update`.

## Tests (real path, load-bearing invariants)

`sparq-core` (direct unit tests for the two new public fns):
- `is_literal_id_classifies_each_kind` — IRI / bnode / string-lit / inline-int / dict-int / `NO_ID`.
- `predicate_has_literal_object_iri_only_vs_literal_vs_bnode` — IRI-only + bnode fires; title/year
  decline; absent predicate vacuously literal-free.
- `predicate_has_literal_object_reflects_overlay_update` — the snapshot-lifecycle flip.

`sparq-engine` (`idfast_unit`, feature-on):
- `predrange_credits_iri_only_object_declines_literal`, `predrange_mixed_use_of_same_var_declines`,
  `predrange_variable_predicate_declines`.
- `predrange_fires_on_q08_shape_witness` — **witness** the q08 `?a = ?b` compiles to
  `CompiledExpr::IdEqNonLit` (fires; under #1785 it declined). **Still fires after the fix** —
  default-graph credit is preserved.
- `predrange_end_to_end_q08_result_unchanged`, `predrange_snapshot_update_flips_credit`.
- **NEW (cross-graph regression, the reviewer's counterexample):**
  - `predrange_graph_block_declines_object_credit_at_outer_dispatch` — analysis-level decline witness:
    an object var bound under `GRAPH <g>` gets **NO** credit at the outer (default-graph) dispatch.
  - `predrange_cross_graph_filter_result_identical` — end-to-end feature-ON differential asserting
    **4 rows** (equal to OFF). **Verified RED on the pre-fix code** (`left: 2, right: 4`) and GREEN
    after the fix.

- `cargo clippy --workspace --all-targets -- -D warnings` — GREEN (default) **and** with
  `--features id-filter-fastpath`.
- `cargo test -p sparq-engine` — GREEN default **and** `--features id-filter-fastpath` (all
  `idfast`/`predrange` tests). `cargo test -p sparq-core` — GREEN.
- W3C SPARQL conformance ratchet — **≥1229, 0 fail** (feature OFF in the conformance build →
  behaviour-identical FILTER path; the fix only tightens the feature-ON classifier).
- `scripts/check-readme-template.py --enforce` — **0 deviations**; `sparq-engine/README.md` at the
  120-line cap.

`origin/main` merged (was 1 commit ahead: #1787 cluster-materialize) — clean.

## Honest performance finding (NON-CANONICAL work box — not a baked-in number)

SP2Bench **q08** / **q12b** at 250k, `materialize`, this shared/noisy work box. Row counts identical
(q08 = 358, q12b ASK = 1) — correctness preserved. The fast path **fires** (the deterministic witness
test proves the `IdEqNonLit` compile, independent of box noise); the wall-clock is directional only:

| query | OFF (run A / run B) | ON (run A / run B) |
|-------|---------------------|--------------------|
| q08   | 54.6 s / 46.8 s     | 40.9 s / 41.2 s    |
| q12b  | 53.3 s / 39.1 s     | 39.3 s / 39.8 s    |

q08 shows a consistent improvement across runs; q12b's OFF baseline itself spans ~39–53 s of box
noise, so its margin overlaps noise. The canonical bench harness (nightly EC2, quiet box) owns any
pinned number. **The cross-graph fix leaves this win intact** — it only tightens the classifier under
GRAPH blocks, which the q08/q12b default-graph shape never enters.

## Deferred

- **sq-1ivw7 residual scope** — reconsider equal-id-probe ordering vs the numeric fast path (from
  #1785's deferred list); a separate micro-optimization, filed as a follow-up bead.

Do NOT auto-merge — a re-review is being routed.


jeswr:
> 🤖 SPARQ agent (escalated-tier reviewer)

**Verdict: request_changes — empirical result-correctness counterexample found (cross-graph hole).**

`Graph::predicate_has_literal_object` scans only `self.store` — the **default graph** — never `self.named`. But `collect_kind_positions`'s `G::Graph` arm recurses into the inner BGP **with the same outer-graph classifier**. When a FILTER sits *outside* a `GRAPH` block (`Filter(expr, Graph(g, Bgp))`), the dispatch in `eval_graph_pattern_inner` passes the OUTER graph to `nonliteral_filter_cols`, so an object variable bound from **named-graph** data is credited off the **default graph's** object column.

**Reproduced divergence** (isolated worktree at PR head, probe test since removed):

```
Default:  <d0> <http://ex/creator> <http://ex/alice> .
Named <g>: <d1> <http://ex/creator> "1"^^xsd:integer  <http://ex/g> .
           <d1> <http://ex/creator> "1.0"^^xsd:decimal <http://ex/g> .

SELECT ?a ?b WHERE {
  GRAPH <http://ex/g> { ?d <http://ex/creator> ?a . ?d <http://ex/creator> ?b }
  FILTER(?a = ?b) }
```

- feature OFF (exact): **4 rows** (`=` is value equality; the integer/decimal pair matches)
- feature ON: **2 rows** (`IdEqNonLit` decides by raw id — the credit was unsound, `?a`/`?b` are literals)

Under #1785 this hole could not exist: subject/predicate credits are data-independent (non-literal in ANY graph). The new predicate-range credit is the first place *which graph is scanned* matters, and the `G::Graph` recursion crosses that boundary. Note the dual case is already correct: a FILTER *inside* the GRAPH block dispatches against the named sub-graph (`eval_translated` evaluates inner with `graph = sub`), so its classifier scans the right store.

**Exact fix (minimal, preserves the q08/q12b default-graph win):** in `collect_kind_positions`, the `G::Graph { inner }` arm must recurse with the always-conservative classifier (`&|_| true`) instead of the caller's `obj_may_be_literal` — object slots under any `GRAPH` block (constant or `?g` variable) stay literal-risk at an outer dispatch. Add the counterexample above as a regression test (both a decline witness on `nonliteral_filter_cols` and the end-to-end 4-row differential with the feature ON).

Everything else checked out on diff-scoped review: overlay-awareness (`scan` merges the delta overlay; flip tests real), per-dispatch `RefCell` memo (cannot outlive the snapshot), parse-only `PreparedQuery` (analysis runs at eval dispatch against the live `&Graph`), drain-safe fused-site wiring (`with_idfast_nonlit_cols`, unbound-column `NO_ID` → type error matching the exact path), honest non-canonical perf framing, feature-off declaration present. The PR body/SKILL "result-identical" claim must be re-earned by the fix; not arming.

jeswr:

Thanks — the counterexample reproduced exactly and the fix is in (`debb35900`). Point by point:

**1. `predicate_has_literal_object` scans only `self.store`, and `collect_kind_positions`' `G::Graph`
arm recursed with the outer classifier → named-graph object credited off the default column.**
Confirmed and fixed. The `G::Graph` arm now recurses with the always-conservative classifier
(`&|_| true`), so an object slot under any `GRAPH <g>`/`GRAPH ?g` block stays literal-risk at an
outer dispatch. The inside-GRAPH dispatch is untouched — `eval_translated` already evaluates the
inner pattern with `graph = sub`, so its own `nonliteral_filter_cols` scans the named store, which is
correct. This is exactly your minimal fix and it preserves the q08/q12b default-graph win (that shape
never enters a GRAPH block).

**2. Reproduced divergence (4 rows OFF vs 2 rows ON).**
Added as a regression, both forms you asked for:
- `predrange_graph_block_declines_object_credit_at_outer_dispatch` — analysis-level decline witness:
  `nonliteral_filter_cols` gives the object var under `GRAPH <http://ex/g>` **no** credit at the outer
  dispatch.
- `predrange_cross_graph_filter_result_identical` — the end-to-end feature-ON differential asserting
  **4 rows**.

Both **verified RED on the pre-fix code** (I stashed the one-line fix and reran): the analysis witness
fails and the differential panics `assertion left == right failed … left: 2, right: 4` — your exact
numbers. GREEN after the fix.

**3. Document `predicate_has_literal_object` as receiver-graph-only so future callers don't misuse it.**
Done — its doc now states (load-bearing) that it scans `self.store` only, never `self.named`, and that
a caller must invoke it on the same graph the object variable is bound from; it cross-references the
`collect_kind_positions` `G::Graph` conservatism.

**4. The q08 firing witness must still fire (default-graph credit preserved).**
`predrange_fires_on_q08_shape_witness` still fires — `?a = ?b` on the default-graph `creator` shape
still compiles to `CompiledExpr::IdEqNonLit`. The fix only tightens the classifier under GRAPH blocks.

**5. The "result-identical" claim must be re-earned.**
Re-earned by the fix + the new cross-graph differential (feature ON now equals feature OFF at 4 rows).
PR body updated to add the cross-graph fix section, list the new tests, and note the win is intact.

Everything else you cleared on diff-scoped review is unchanged.

Gates (both feature states): clippy `--workspace --all-targets -D warnings` GREEN default **and**
`--features id-filter-fastpath`; `cargo test -p sparq-engine` GREEN both states; `cargo test -p
sparq-core` GREEN; `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` GREEN;
conformance floor-sync GREEN (SPARQL ratchet 1229, 0 fail — feature-OFF path, behaviour-identical);
`check-readme-template.py --enforce` 0 deviations. `origin/main` merged (#1787). Not arming — over to
the re-review.

jeswr:
> 🤖 SPARQ agent (escalated-tier re-review)

Re-review covers `315a7cbdb..7be573209` (cross-graph fix + merge of #1787 + SKILL doc). Verified diff-scoped:

- **Fix**: `collect_kind_positions`' `G::Graph` arm recurses with the always-conservative `&|_| true`; since the classifier is a threaded parameter, it propagates through the ENTIRE inner subtree (nested GRAPH, UNION/OPTIONAL/sub-SELECT under GRAPH). UNION-sibling leakage is blocked by set semantics: the conservative classifier forces `poison_slot` on GRAPH-scoped object slots, so a default-graph credit cannot survive `nonlit \ maybe_lit`. Subject/predicate credits inside GRAPH remain — graph-independent, sound. No constant-`<g>` special case; `GRAPH ?g` covered by the same arm. No other arm crosses a graph boundary (Service/Lateral-right poison; Project/Group re-scope but don't change the store; fused-conjunctive sites analyse bare BGPs against the same `graph` they execute on, which is `sub` inside a GRAPH dispatch).
- **Tests**: both regression tests match the counterexample exactly (analysis-level no-credit witness + end-to-end 4-row feature-ON differential on a TriG dataset with a populated named graph); q08 firing witness still present at head. The `opt-in sparq-engine (id-filter-fastpath)` CI leg runs them.
- **Merge debb35900**: PR-head-vs-main diff contains only this PR's files — #1787's cluster-materialize side intact, fix side intact.
- **Docs**: `predicate_has_literal_object` receiver-graph-only scope documented in code + SKILL; PR body's result-identical claim now earned.

Verdict: honest=true, sound_as_scoped=true, disposition=arm — auto-merge armed (latched 21:55:52Z).

PR #1796, opened 2026-07-09
spike(core): per-field byte attribution over compressed block stream — H1/H2/H3 verdict (sq-7d3dj.32.2.4)

jeswr:
> 🤖 SPARQ agent — SPIKE for **sq-7d3dj.32.2.4** (parent sq-7d3dj.32.2; `research/compressed-memory-profile.md` §4). [FABLE-5]

## What this adds

A fully `#[cfg(test)]`-gated `byte_attribution` module in `crates/sparq-core/src/compress.rs` that mirrors `encode_block` **write-site for write-site** and attributes every emitted LEB128 varint byte to a disjoint field class — **block-count / first-row-absolute / reset_d0 (col1+col2 absolute on `d0≠0`) / reset_d1 (col2 absolute on `d1≠0`) / d0 / d1 / d2 deltas / directory** — per permutation, across id-density regimes.

**Zero behaviour / format / bundle change (invariant):** the `SPQCPRM1` byte stream is untouched; the whole module is test-only (adds no default-path bytes, so the `wasm_bundle_bytes` `feature_off_exact` floor is unmoved and coverage floors are unaffected — test code doesn't count). A self-check test (`attribution_total_equals_real_encoded_length`, runs on the default `cargo test`) asserts the attribution total equals the **real** `encode_block` output length byte-for-byte across both reset shapes and block boundaries, so the tables are not a fiction.

The two measurement tests are `#[ignore]` (they synthesise up to 50M triples / allocate hundreds of MB); run them explicitly:

```
cargo test -p sparq-core --release --lib byte_attribution -- --ignored --nocapture
```

## Verdict (see the bead note for the full tables)

All numbers **NON-canonical work-box**. Attributed per-triple, summed over the six permutations, across 100K→1M→5M distinct terms (17→20→23 term-space bits):

| field (B/triple) | 100K/17b | 1M/20b | 5M/23b |
|---|---|---|---|
| reset_d1 (col2 abs, d1≠0) | 10.209 | 11.262 | **12.641** |
| d1 delta | 7.505 | 8.522 | 8.947 |
| d0 delta | 5.953 | 5.953 | 5.953 |
| reset_d0 (col1+col2 abs) | 1.254 | 1.409 | 1.555 |
| d2 delta | 1.122 | 1.282 | 1.425 |
| first-row abs | 0.287 | 0.315 | 0.355 |
| dir | 0.750 | 0.750 | 0.750 |
| **ALL** | 27.173 | 29.586 | 31.721 |

**H1 (absolute resets widen with term-space bits) — CONFIRMED as the dominant mechanism.** The single largest bucket is `reset_d1` (absolute col2 written whenever the middle column changes) at ~10–13 B/triple, and it is also the largest *grower* (+2.43 B/triple over the bit range). `d0` (leading-column delta) is perfectly flat — deltas between adjacent leading ids are small and bit-width-independent; growth is concentrated in the **absolute** fields (resets + first-row), exactly LEB128 7-bit-step quantization of larger ids.

**H2 (within-run deltas widen) — PARTIALLY TRUE, secondary.** `d1` grows +1.44 and `d2` +0.30 with the term space, so intra-run deltas do widen — but less than the absolute resets, and `d0` doesn't widen at all. H2 is real but not the driver.

**H3 (plateau = small-ids-for-frequent-terms) — CONFIRMED by the skew discriminator.** The skewed-vs-uniform sweep (fixed 4M triples/4M terms) shows uniform ids cost **+1.5 B/triple** (33.83 vs 32.32), driven by `reset_d0` (+1.9). Frequent terms getting small (insertion-ordered) ids keeps the absolute-reset varints short, which is what damps the canonical 10M→50M plateau: a 5× id-space growth barely crosses a LEB128 byte step for the *typical* (frequent, small-id) term. The fixed-triple bit-width sweep (Sweep A) also shows the composition shifts — as density rises, `reset_d0` explodes (0.80→9.14) while `d1`/`d2` deltas shrink, i.e. resets substitute for delta runs.

**Net:** growth lives in the **absolute-reset fields under LEB128 quantization** (H1), damped by insertion-order small-ids for frequent terms (H3); intra-run delta widening (H2) is a minor contributor.

## Follow-up (per the bead's gate: file a bead ONLY if a format change projects a material win)

The attribution says the lever is the ~13 B/triple of **absolute col2 on middle-column resets** (`reset_d1`). A per-block **frame-of-reference** encoding (store col2 resets as a delta from the block's first-row col2, or bit-pack col2 to the block's max width) would attack exactly this bucket. That is a **versioned `SPQCPRM1`→`SPQCPRM2` format migration** (shared in-RAM/on-disk stream) and is explicitly out of this spike's scope. I have captured it as a discovered-work item in my report to the orchestrator rather than beading directly (`bd` is not on PATH in a worktree) — see the bead note.

- `cargo test -p sparq-core --lib compress` (default) — GREEN; self-check runs, measurement tests `#[ignore]`d so default test stays fast.
- `cargo test -p sparq-core --features mmap --lib compress` — GREEN (`compressed_save_open_roundtrip`, `compressed_scans_match_raw`, `stream_writer_byte_identical_to_encode_write_to` unchanged).
- `cargo clippy --workspace --all-targets -- -D warnings` — GREEN.
- `RUSTDOCFLAGS="-D warnings" cargo doc -p sparq-core --no-deps --all-features` — GREEN.

Not armed.

### DIFF
--- a/crates/sparq-core/src/compress.rs
+++ b/crates/sparq-core/src/compress.rs
@@ -769,6 +769,366 @@ impl Drop for CompressedPermWriter {
     }
 }
 
+/// [FABLE-5] sq-7d3dj.32.2.4 — SPIKE-ONLY per-field byte attribution over the
+/// [`encode_block`] stream. This whole module is `#[cfg(test)]`: it adds ZERO bytes to
+/// the shipped `SPQCPRM1` stream and does not touch the encode/decode hot path (the
+/// `wasm_bundle_bytes` `feature_off_exact` floor is unmoved). It exists to adjudicate the
+/// §4 root-cause hypotheses (H1/H2/H3) of `research/compressed-memory-profile.md` for why
+/// the compressed store's B/triple grows 36.75 (1M) → 48.75 (10M) then plateaus at 48.75
+/// (50M): it mirrors `encode_block` field-for-field and attributes each emitted varint's
+/// byte length to the field class that produced it. All numbers this module prints are
+/// NON-canonical work-box measurements (never a doc/test perf number).
+#[cfg(test)]
+mod byte_attribution {
+    use super::{encode_block, Id, BLOCK};
+
+    /// Byte length of the unsigned LEB128 varint that `put_varint` would emit for `x`
+    /// (mirrors its `while x >= 0x80 { x >>= 7 }` loop exactly): `⌈bits/7⌉`, min 1.
+    #[inline]
+    fn varint_len(x: u64) -> usize {
+        let mut x = x;
+        let mut n = 1usize;
+        while x >= 0x80 {
+            x >>= 7;
+            n += 1;
+        }
+        n
+    }
+
+    /// The disjoint field classes the block stream's bytes are attributed to. These are
+    /// exactly the write sites in [`encode_block`]; every byte the encoder emits lands in
+    /// exactly one class, so `total()` equals the real encoded block-stream length.
+    #[derive(Clone, Copy, Default, Debug)]
+    struct FieldBytes {
+        /// The per-block `count` varint (one per block).
+        count: u64,
+        /// The block's FIRST row, written as three absolute varints (col0/col1/col2).
+        first_row_abs: u64,
+        /// col1+col2 written ABSOLUTE because the leading column changed (`d0 != 0`).
+        reset_d0: u64,
+        /// col2 written ABSOLUTE because the middle column changed (`d0 == 0 && d1 != 0`).
+        reset_d1: u64,
+        /// The `d0` (leading-column) delta varint on every non-first row.
+        d0: u64,
+        /// The `d1` (middle-column) delta varint, emitted only when `d0 == 0`.
+        d1: u64,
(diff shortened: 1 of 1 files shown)


jeswr:
> SPARQ agent mechanical-verify: ARMED ✓

All objective checklist items pass:
- **Gates**: 147 pass, 1 advisory-fail (non-blocking conventional-commits), 8 pending slow legs (coverage/test shards — normal gate-wait behavior)
- **Tests non-vacuous**: Verified via mutation (flipped assertion → RED; reverted → GREEN)
- **Opt-in discipline**: All 366 lines inside #[cfg(test)], zero added to default sparq-core/engine builds
- **README/SKILL sync**: NA (no public-API change, test-only module)
- **No hard-coded perf**: Confirmed no .md files in diff
- **Coverage floor**: Unaffected (test-only code)

No judgment surfaces. Auto-merge enabled → will engage on gate SUCCESS (ETA ~20-40min for slow siblings).

PR #1813, opened 2026-07-10
perf(engine): extend theta anti-join to sameTerm/IN correlations + parallelise the hash probe (sq-3cmr4, sq-f1emb)

jeswr:
> 🤖 SPARQ agent. Two follow-ups to the merged #1786 correlated (theta) anti-join, in one branch. [FABLE-5]

## What

Extends the correlated (theta) anti-join for `Filter(!bound(?nb), LeftJoin(A, B, F))` (the SP2Bench-q06 negation idiom, #1786) in `crates/sparq-engine/src/exec.rs` (same anti-join region):

### sq-3cmr4 — widen the correlation detector beyond `?a = ?b`
- **`sameTerm(?a, ?b)`** — TERM IDENTITY. A dictionary-id equality *is* `sameTerm` for **every** term kind (literals included), so a sameTerm-keyed left row is id-bucket-probeable even when the key is a **literal** (no value-equality hazard), and any value-correct scan re-checks with `sameTerm`, never `=`.
- **`?a IN (?b)`** — a single-variable membership desugars to `?a = ?b`, so it is a value-equality correlation (id-probe only NamedNode/BlankNode keys; a literal takes the value-correct scan). A **constant / multi-element** `IN` list is **not** a correlation (residual, always sound) — the brief's rule.
- Correlations now carry a `CorrKind` (`Eq` vs `SameTerm`); the id-probe eligibility and the per-candidate re-check expression are chosen **per kind** — load-bearing because `=`, `sameTerm` and id-equality are **different relations** on value-equal id-distinct literals (`"1"` vs `"01"`).

### sq-f1emb — parallelise the hash-path probe loop
Above `PAR_THRESHOLD` (50k, the existing parallel-hash-join / BIND threshold) the per-left-row elimination **verdict** is fanned out over cores with rayon (each verdict reads only shared read-only state; EXISTS-capable residuals install the thread-local function/view/spatial snapshots like the FILTER/BIND paths). Only the verdicts are parallelised — the ordered output build + budget truncation stay serial — so the result is **byte-identical** to the serial probe (order-preserving collect + serial survivor-prefix truncation reproduce the serial early-break-on-budget).

## Feature flags
- No new feature. The parallel probe is under the existing default-on `parallel` feature (off for wasm, so rayon never enters the bundle). The detector extension is unconditional. No new public API; `sparq-core`/`sparq-engine` stay lean.

## Correctness discipline (this code area's history)
- `=` vs `sameTerm` vs id-equality are treated as **distinct relations** on RDF literals throughout.
- Differential (`on == off`, forced-off cold plan vs on) for every new correlated shape, including hash + SIP-seed strategies, both id-hashable kinds, EXISTS-nested and error-producing residuals, and the empty-anchor edge (unchanged).
- **Non-vacuity**: `sameterm_vs_eq_diverge_on_value_equal_literals` is a mutation-verified red→green witness (under `=` d0 is eliminated; under `sameTerm` both survive — mutating the sameTerm recheck to `=` turns it red). The `IN` arm's fire-assert goes red when the `IN` detector is removed. Both verified locally.
- The randomised fuzzer now **rotates the correlation relation** across `=` / `sameTerm` / single-var `IN` against the cold oracle.
- `parallel_hash_path_crosses_threshold` builds a >50k-row anchor so the rayon fan-out (the `left_b.rows.len() >= PAR_THRESHOLD` branch), not just the shared verdict closure, genuinely runs — validated against a **closed-form** expectation (every distinct-key item survives) + a fired-assert, NOT the cold OFF plan (the cold `left_outer_join` at 50k distinct keys is the quadratic blow-up this optimisation removes, so re-running it would dominate). It is `#[ignore]`d out of the per-commit budget (like `q06_measure_250k`); the non-ignored sub-threshold tests pin OFF-vs-ON identity of the shared verdict closure, and the parallel branch merely fans that closure over cores. Verified passing in release (ON-path eval 0.11s over 50001 rows). Run with `cargo test -p sparq-engine --release --test theta_antijoin -- --ignored parallel_hash_path_crosses_threshold`.

## Gates (run in-worktree, both feature states)
- `cargo clippy --workspace --all-targets -- -D warnings` — GREEN (default / parallel ON).
- `cargo clippy -p sparq-engine --no-default-features --features regex,digest --all-targets -- -D warnings` — GREEN (parallel OFF).
- `cargo test -p sparq-engine` — GREEN (full suite, default features).
- `cargo test -p sparq-engine --no-default-features --features regex,digest --test theta_antijoin` — GREEN (parallel OFF).
- `cargo test -p sparq-engine --test theta_antijoin` + lib `theta_antijoin_unit` — GREEN.
- Coverage: each new private fn (`as_correlation_pair` / `corr_recheck_expr` / `corr_id_probeable`) has a direct unit test; the parallel branch is crossed by `parallel_hash_path_crosses_threshold`. sparq-engine floor is 90 (baseline 92); the change adds only well-tested lines.
- W3C SPARQL conformance suite: not runnable in-sandbox (test data not fetched). This is a performance-only rewrite whose semantics are pinned identical to the cold plan by the `on == off` differential harness (the OFF path *is* the standard conformance path), so conformance cannot regress.

## Performance (WORK-BOX, NON-CANONICAL — do not treat as published numbers)
No hard-coded perf numbers in markdown. The 250k end-to-end measurement harness (`q06_measure_250k`, `#[ignore]`) is unchanged; this PR is a **correctness-sound capability + scalability** extension, not a new perf claim.

## Deferred / discovered work
See the bead notes.

### DIFF
--- a/bench/feature-off-declarations/1813.json
+++ b/bench/feature-off-declarations/1813.json
@@ -1,0 +1,5 @@
+{
+  "pr": 1813,
+  "date": "2026-07-10",
+  "reason": "sq-3cmr4 + sq-f1emb: the theta anti-join correlation-detector extensions (sameTerm/IN CorrKind arms in as_correlation_pair, corr_recheck_expr, corr_id_probeable) and the rayon-parallel hash-probe are intentional ALWAYS-COMPILED engine code — the theta anti-join is not feature-gated (it shipped unconditional in #1786), so the feature-OFF wasm bundle legitimately grows by +1619 bytes (+0.103%). This is a declared intentional engine change, NOT a leak of an opt-in feature into the OFF build. Escalated review (sq-reviewer) verified soundness for head a71a99fe9 across all correlation kinds (=/sameTerm/id-equality on value-equal id-distinct literals), the single-var IN scoping, and the fe93caf8c thread-local drain discipline under the rayon fan-out; mutation witnesses re-run red. Delta is under the ±2% band, so no bench/perf-baseline.json floor change accompanies this declaration."
+}
--- a/crates/sparq-engine/src/exec.rs
+++ b/crates/sparq-engine/src/exec.rs
@@ -595,9 +595,10 @@ pub(crate) mod sip {
 // because `F` references OUTER variables (SP2Bench q06:
 // `?author = ?author2 && ?yr2 < ?yr`), so the cold `left_outer_join` evaluates the
 // full right side against the whole document set. This path instead:
-//   (a) partitions `F` into EQUALITY conjuncts `?outer = ?inner` (correlation) whose
-//       `?inner` is certain-in-`B` and `?outer` is a left variable, and a RESIDUAL
-//       theta remainder (everything else, e.g. `?yr2 < ?yr`);
+//   (a) partitions `F` into var-to-var CORRELATION conjuncts `?outer θ ?inner` whose
+//       `?inner` is certain-in-`B` and `?outer` is a left variable — where θ is value
+//       `=`, `sameTerm`, or a single-variable `?a IN (?b)` membership (sq-3cmr4) — and a
+//       RESIDUAL theta remainder (everything else, e.g. `?yr2 < ?yr`);
 //   (b) for each distinct left correlation tuple, SEEDS the `B` scan sideways with the
 //       outer IRI bound into `?inner`'s position (extending the #1741 SIP machinery to
 //       the anti-join), evaluating a tiny correlated `B'` instead of the cold corpus;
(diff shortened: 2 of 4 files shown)


jeswr:
> 🤖 SPARQ agent (escalated reviewer verdict)

**Soundness review: PASS.** Diff-scoped review of the CorrKind separation, the IN desugaring, the rayon fan-out thread-local discipline, and the id-encoding claim all check out:
- `crates/sparq-core/src/dict.rs:59-60` confirms only canonical lexical forms inline (`"030"^^integer` stays a distinct dictionary term), so id-equality is injective on term identity — the sameTerm id-probe is exact for every term kind, including the inline-integer `"1"`/`"01"` trap.
- The Eq arm never substitutes id-equality for value-equality (literal keys → value-correct scan with `=` recheck; SIP seeding stays IRI-only).
- The engine's native `Expression::In` eval is a `values_equal` disjunction — exact 3-valued `=` for a single element, so the desugaring matches the cold oracle.
- `IDFAST_NONLIT_COLS` is drain-on-consume inside `apply_filter_scalar`; the theta probe evaluates via `eval_expr` and rayon workers start with the empty default — no fe93caf8c-class leak. Worker installs (fns/view/spatial) match the sibling FILTER/BIND parallel paths.
- Both mutation witnesses independently re-verified red (sameTerm recheck → `=` fails `sameterm_vs_eq_diverge_on_value_equal_literals`; IN-detector removal fails the fire-assert), and the `#[ignore]`d 50k parallel test passes in release (0.12s eval) as claimed.

**One blocking mechanical defect (the red `gate`):** `artifact-exact-equality (wasm bundle feature-OFF)` — the unconditional detector code changes the feature-OFF wasm bundle by **+1619 bytes (+0.103%)** and the change is undeclared. Fix per the check's own instruction: add `bench/feature-off-declarations/1813.json` with `{pr, date, reason}` (intentional always-compiled engine change: sq-3cmr4 correlation-detector extension). Delta is well under ±2%, so **no** `bench/perf-baseline.json` floor raise is needed. No engine-code change required; after the declaration lands, this PR can re-enter the normal verify→arm lane without re-escalation.

Non-blocking observations (bead-worthy, not for this PR): the refactored probe computes all verdicts before budget truncation, so row-cap/deadline exhaustion no longer stops the probe between left rows, and rayon workers do not see the thread-local budget during EXISTS re-entry — consistent with the pre-existing parallel FILTER/BIND paths, resource-policy only, results unaffected.

PR #1811, opened 2026-07-10
feat(jsonld): Node Map Generation + native Flattening + flatten lane → normative document oracle (sq-oy1f.26)

jeswr:
> 🤖 SPARQ agent — implementing bead **sq-oy1f.26** (JSON-LD 1.1 lane [D]: Node Map Generation + Flattening; flatten lane to the normative document oracle). Wave W1 of the epic `sq-oy1f` fleet decomposition (`research/jsonld-1.1-remaining-work-decomposition.md`). NOT armed — awaiting review.

## What this implements

The **Node Map Generation** (JSON-LD 1.1 API §7.2) + **Generate Blank Node Identifier** (§7.4) + **Flattening Algorithm** (§7.1) in the opt-in, dependency-free `sparq-jsonld` crate, and the flatten conformance lane's switch from the RDF-writer round-trip oracle to the **native document-level oracle** (`sparq_jsonld::flatten()` + `json_ld_equal`), mirroring the expand lane (sq-kk1mq).

### `crates/sparq-jsonld`
- **`src/node_map.rs`** — `generate_node_map()` over an **expanded** document: a deterministic `_:bN` `BlankNodeIssuer` (first-encounter order), a per-graph `GraphMap`, a `NodeMap` keyed by graph name. Handles `@type` blank-node relabelling, `@included`, `@reverse` back-references (materialise-once — no duplicate node), `@graph` named-graph recursion, `@index` preservation, empty-array-property creation, and **list-object non-dedup** (structurally-equal `@list`s are distinct list nodes).
- **`src/flatten.rs`** — `flatten()` = expand ∘ node-map ∘ named-graph fold, sorted by `@id`, empty `{@id}`-only nodes dropped; `flatten_expanded()` for callers already holding the expanded form. Post-flatten **compaction** (`{ "@context": …, "@graph": … }`) is the document-level Compaction Algorithm — its own bead **sq-oy1f.27**.
- **`tests/flatten.rs`** — crate-local W3C flatten-suite harness over the native `Json` AST (keeps the crate at **zero dependencies** — no `serde_json`); agrees with the conformance lane at **46**.

### `crates/sparq-conformance`
- flatten lane → `run_flatten_native` (document oracle, `json_ld_equal`).
- **Floor RE-PIN 50 → 46** — stated side-by-side per the epic honesty rule.

## Honest conformance re-pin (side-by-side)

| Oracle | pass | fail | skip |
|---|---|---|---|
[...]

PR #1818, opened 2026-07-10
fix(jsonld): expand() correctness — @type collapse + empty-array retention + free-floating drop; raise EXPAND_FLOOR 240→259 (sq-oy1f.37)

jeswr:
> 🤖 SPARQ agent [FABLE-5] — JSON-LD 1.1 expand-lane correctness (bead sq-oy1f.37).

## What this does

Three spec-faithful fixes to the JSON-LD 1.1 expansion algorithm (§5.1.2) in `crates/sparq-jsonld/src/expand.rs`, flipping **19** previously-failing W3C `expand` positive cases to pass under the native document-level oracle. `EXPAND_FLOOR` rises **240 → 259** (rise-only ratchet).

### The three fixes

1. **Value-object `@type` collapse.** A value object's `@type` is a SINGLE value in the JSON-LD data model (a *node* object's `@type` is an array). The general keyword-entry path (step 13.4.4) arrayifies every `@type`; `cleanup` now collapses a single-element `@type` array back to a scalar. This cleared the false `invalid typed value` errors the expander was raising on positive cases — `0002/0013/0014/0028/0036/0046/0077/c020/js15/16/19/20/tn02`.

2. **Empty-array-property retention.** The forward-property add now uses the `addValue` `asArray = true` rule (the reference `_addValue(..., propertyIsArray: true)` is unconditional). A `@set`/`@list`/plain-array term whose value expands to `[]` now RETAINS the property as an empty array instead of being dropped by the plain empty-array skip — `0004/0015/0016`. New private helper `add_value_as_array`.

3. **Free-floating value/list drop.** The null/`@graph` active-property drop (step 19) now reaches value objects (any map with `@value`) and list objects (`@list`), not just empty / `@id`-only node objects — `0045/0046`.

## Oracle & correctness

- Measured **259 pass / 17 fail / 109 skip** (was 240/36/109). Every rise is a genuine expander correctness fix verified against the native document-level oracle (`sparq_jsonld::expand()` + `json_ld_equal`), not an oracle artefact.
- **6 non-vacuous regression tests** in `crates/sparq-jsonld/tests/expand.rs`, each mutation-verified to go RED when its fix is reverted (value-object type collapse, `@set`/plain empty-array retention, free-floating value drop under null and under `@graph`).
- No public API change (the new helper is private).

## Flatten linkage (per the bead)

The **current** flatten lane on `origin/main` is RDF-writer-driven (`graph_to_jsonld(JsonLdForm::Flattened)`), NOT the native `expand()`, so it is **unaffected** by these fixes (`FLATTEN_FLOOR` held at 50, green). PR #1811 (sq-oy1f.26 — still OPEN) introduces the native document oracle that routes flatten through `expand()`; once it lands, flatten will inherit these fixes and can rise toward 53. This PR deliberately does **not** edit `floors/flatten.rs` while #1811 is open (conflict discipline). Cross-ref notes added to sq-oy1f.26 / .37.

- `cargo clippy -p sparq-jsonld --all-targets -D warnings` — clean (feature OFF).
- `cargo clippy -p sparq-conformance --features jsonld-suite --all-targets -D warnings` — clean (feature ON).
- `cargo test -p sparq-jsonld --test expand` — 45 pass.
- `cargo test -p sparq-conformance --features jsonld-suite --test jsonld_suite` — ratchet green: toRdf 413 / fromRdf 51 / compact 186 / expand **259** / flatten 50, all ≥ floor.
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — clean (code-span discipline; no intra-doc link to a private/feature-gated item).
- Default `cargo test -p sparq-conformance` (feature OFF) — self-skip test green.

## Deferred → sq-oy1f.45

The remaining 17 `expand` fails are captured in follow-up bead **sq-oy1f.45**: 7 remote-`@context` positives (need an `FsLoader` — no network under `NoopLoader`; a test-harness change) + 10 deeper `expand()` divergences (`@id: null` retention t0122, relative-IRI-with-colon t0109, rdf-star reverse `@index` t0131, `invalid IRI mapping` t0026/t0071, `invalid scoped context` tc034/tso05/tso06).

Not armed for auto-merge — leaving for review.


jeswr:
> 🤖 SPARQ agent [FABLE-5] — gate fix (mechanical-verify BOUNCE resolved).

The `test (load-aware shard bulk 3/3)` gate failure was a **real defect in this PR**, correctly caught by mechanical-verify — not a congestion artifact. Root cause: a **double-entry floor guard** I initially missed.

`crates/sparq-conformance/tests/scoreboard_floors.rs::lib_sourced_jsonld_floors_are_pinned` cross-checks the lib-side `floors::expand::FLOOR` const against a hard-coded `LIB_SOURCED_EXPECTED` value, and fires when they diverge (its own assertion message: *"if this is a deliberate raise, bump `LIB_SOURCED_EXPECTED` in the same commit that edits `src/floors/<lane>::FLOOR`"*). This guard is a **DEFAULT-feature** test, so it runs in the general `bulk 3/3` shard — my initial local verification only ran the conformance ratchet under `--features jsonld-suite`, which does not exercise it.

Fix (commit `3756d6689`, rebased onto the update-branch merge):
- Bumped `LIB_SOURCED_EXPECTED` expand entry 240 → 259 (rise-only, same-commit-with-the-const rule satisfied).
- Reconciled the now-stale `240` narrative in `scoreboard.rs` doc-comments.

Re-verified green: the previously-failing `lib_sourced_jsonld_floors_are_pinned` now passes; full default `cargo test -p sparq-conformance` green; feature-ON conformance ratchet `TOTAL expand 259` ≥ floor; `cargo doc --all-features -D warnings` clean. Still **not armed** — leaving for re-review.

Lesson for future floor raises: bumping a `src/floors/<lane>::FLOOR` const requires the matching `LIB_SOURCED_EXPECTED` bump, and the guard lives in the DEFAULT feature state (run `cargo test -p sparq-conformance` without `--features jsonld-suite` to catch it locally).

PR #1822, opened 2026-07-10
fix(engine): datatype-aware numeric cache + trim-consistent lenient seams — cache ⟺ of_literal on =/</> (sq-74oy4 + sq-6b1lj) [FABLE-5]

jeswr:
> 🤖 SPARQ agent (Claude Fable 5) — reconciling the evaluator's numeric seams so the fast paths agree with the datatype-aware reference `Num::of_literal`. Lands **sq-74oy4** (trim asymmetry) and **sq-6b1lj** (per-datatype cache leniency) as ONE coherent policy change. Not armed.

## What & why

The numeric fast paths — the sparq-core numeric-value **cache** (`numeric_value`), the sargable FILTER pushdown, the `JKey::Num` value-join, and the `CompareTerm` total order — are optimisations that must match the datatype-AWARE reference `Num::of_literal` that `values_equal` / `value_compare_strict` use for `=`, `<`, `>`. Two residuals from #1792 (sq-9781x) broke that:

- **sq-74oy4 (trim asymmetry):** `Num::of_literal` TRIMS (XSD `collapse` facet), the cache trims, but the lenient `as_num`/`as_f64`/`LocalVocab::intern` seams parsed UN-trimmed — so a padded `" 1"^^xsd:integer` was value-1 under `=` yet a type error under `<`/`>`.
- **sq-6b1lj (per-datatype leniency):** the cache parsed every numeric datatype through the datatype-AGNOSTIC `parse_xsd_f64`, so `"1.5"^^xsd:integer` / `"1E2"^^xsd:decimal` / an i128-overflow decimal **cache-HIT** (and were value-included on the sargable `=`/`<` fast path + `JKey::Num`) where `of_literal` type-errors them.

## Policy chosen: datatype-aware everywhere (with measured evidence)

I made the numeric cache **and** the lenient `as_num`/`as_f64`/`intern` (+ reasoner `as_f64`) seams **datatype-aware and trimming**, mirroring `Num::of_literal`:
- integers: scale-0 (after trailing-zero normalisation, so `"5."`/`"5.0"`/`"5.00"` are value-5 integers — matching `of_literal`, which routes over-`i64` integers through `Dec::parse`), i128-fit
- decimals: no exponent, i128-fit mantissa
- float/double: XSD `doubleRep` (already `of_literal`'s set)

This **subsumes** sq-74oy4's resolved direction (option (b) lenient/trim-everywhere) — trimming falls out of routing through the of_literal-equivalent acceptance — and closes sq-6b1lj by making **cache HIT ⟺ `of_literal` accepts**. It is *correcting*, not merely tightening: the sargable/`eval_numeric`/`JKey::Num` fast paths now agree with the exact `values_equal`/`value_compare_strict` fallback on **every** numeric lexical, and the ORDER-BY/`literal_kind` transitivity hazard (ill-formed-for-datatype lexicals wrongly ranked Numeric) is removed.

### Why NOT option (a) strict-XSD (don't trim anywhere)
XSD gives xsd:integer/decimal/double/float the `collapse` whitespace facet, so `" 1 "^^xsd:integer` IS value 1 — trimming is the correct behaviour (affirmed in the #1792 review). Option (a) would fight that facet and require reverting the trim in `of_literal` + `Dec::parse_lexical` + the cache — a larger, riskier change against the spec. So option (b)/datatype-aware wins on both correctness and minimality.

### Conformance impact — MEASURED (this is the load-bearing check)

| policy | W3C SPARQL conformance | Δ vs base |
|---|---|---|
| base (origin/main) | **1225 pass / 0 fail / 4 divergence** | — |
| option (a) strict-XSD (no trim) | 1225 / 0 / 4 | **0 (neutral)** |
| option (b) / datatype-aware (this PR) | **1225 / 0 / 4** | **0 (neutral)** |

Both options are conformance-neutral, as the beads predicted — and I verified WHY: the W3C SPARQL suite (`w3c/rdf-tests` @ f25dbc0) contains **ZERO** whitespace-padded typed-numeric lexicals AND **ZERO** of the per-datatype-ill-formed lexicals (fraction-on-integer, exponent-on-decimal, i128-overflow) that this change affects (`grep` counts: 0 for every triggering pattern). So neither option can move the number; I chose datatype-aware on correctness/consistency grounds, with the data confirming it's safe.

## Implementation

`sparq-core` is the leanest tier and cannot depend on `sparq-substrate` (which owns `Num`/`Dec`), so it gets a small self-contained `numeric_datatype_wellformed` + `cached_numeric_f64` + public `numeric_cache_value` mirroring `of_literal`'s per-datatype acceptance (re-derived from the same digit-scan + i128-fit grammar). All FOUR acceptance sites route through it in lockstep: `numerics_of` (in-RAM build), `dictspill` (spill build — byte-identity preserved), `numeric_of`/`extend_for` (delta growth), and the engine's `LocalVocab::intern` (computed terms) + `as_num` (+ the constant operand of the sargable/`eval_numeric` fast path). The reasoner `as_f64` uses its existing `num_of_parts` (of_literal twin).

## Tests (real path + invariant + mutation witnesses)
- **substrate `cache_f64_seam_vs_as_numeric_differential`** — TRUE cross-seam differential: the ACTUAL cache acceptance (`sparq_core::numeric_cache_value`) vs `Num::of_literal` over a lexical×datatype matrix (incl. `"5."`/`"5.0"`/`-0`/i128-edges); fails on ANY unexpected divergence/agreement.
- **engine `numeric_fast_slow_path_agreement`** — fast path (sargable/cache) vs slow path (high-precision-decimal graph disables sargable → exact evaluator) must AGREE on `=`/`<`/`>` for padded / per-datatype-ill-formed / well-formed lexicals across xsd:integer/double/decimal/float, plus the value-join and constant-operand surfaces.
- **mutation witnesses** — reverting the integer branch to the datatype-agnostic parse flips the engine witness + the substrate differential RED (verified 0-vs-1).
- direct sparq-core unit test for `numeric_cache_value`; the sparq-core plumbing test + dict-spill byte-identity test updated.

## Gates (BOTH feature states)
- `cargo clippy --all-targets --all-features -D warnings` — clean (default + `service,cs-planner,zk`)
- engine tests: default (20 binaries, 0 failed) + `service,cs-planner,zk` (20 binaries, 0 failed); core / substrate `--all-features` / reason — 0 failed
- W3C conformance ratchet: **1225 / 0 / 4, not regressed** (post-merge with origin/main)
- README cap: no crate README added/grown

## Scope / conflict discipline
exec.rs edits are confined to the numeric seams (`as_num`, `intern`, `eval_numeric`/`eval_compiled_numeric` constant arms, `extract_sargable` constant, the new `numeric_cache_f64` helper) — none touch the `theta_antijoin` region PR #1813 edits. Merged origin/main cleanly (theta_antijoin included) with no conflicts.


jeswr:
> 🤖 SPARQ agent (Fable 5) — escalation verdict [FABLE-5]

Inline escalation verdict (dispatch-freeze): **honest=true, sound_as_scoped=true, recommend_arm=true, disposition=arm**.

Basis: policy decision empirically grounded — BOTH options measured conformance-neutral (1225 pass / 0 fail / 4 divergence, identical to base, because the suite contains zero whitespace-padded or per-datatype-ill-formed numeric lexicals; grep-verified), so the choice fell to correctness: datatype-aware-everywhere restores cache HIT ⟺ of_literal (the #1792 reviewer ask), the XSD whiteSpace=collapse facet makes trimming spec-correct for numeric types, and the previous fast-path/slow-path asymmetry on '=' vs '<'/'>' is removed rather than papered over. Real bug caught + fixed with a full edge-matrix differential (scale-0 decimal forms). Per-seam mutation witnesses verified red; gates green both feature states; exec.rs edits scoped clear of the anti-join region (clean merge of #1813). Deferred EBV question honestly disclosed. CI gate controls the merge.

jeswr:
> 🤖 SPARQ agent (Claude Fable 5) — unblocking the mechanical `artifact-exact-equality (wasm bundle feature-OFF)` gate.

**Root cause: declaration, not a leak.** The feature-OFF leg failed because the head-tree feature-OFF wasm bundle differs from base (base=1578024 → head=1579401, **+1377 B / +0.087%**, well under the ±2% band) and the change was not yet declared. The diff enlarges *always-compiled* `sparq-core` numeric-cache code — the new datatype-aware `numeric_datatype_wellformed` / `cached_numeric_f64` / `numeric_cache_value` acceptance helpers (mirroring `Num::of_literal`) plus their four call sites. Verified this is intent, not an accidental feature leak: the diff touches **zero** `cfg(feature)` lines and none of the new fns is behind `vectorized` (or any feature), so nothing opt-in reached the OFF build. This is the direct successor to #1792 (sq-9781x), which declared the identical numeric-cache class at +507 B.

**Fix (mechanism V2):** added `bench/feature-off-declarations/1822.json` declaring the intentional always-compiled byte delta (commit `f5c9dacf8` [FABLE-5]). Delta is under ±2%, so no `bench/perf-baseline.json` `wasm_bundle_bytes` floor change is required. No source/gating changed — a leak fix would have been wrong here.

Only two red legs existed (the wasm leg + its downstream `gate`); no other failures. The push retriggered a fresh CI cycle so the previously-concluded `gate` re-runs clean. The armed auto-merge latch (MERGE, enabled by @jeswr) persisted — its verdict basis stands; I only unblocked a mechanical gate, so no re-arm was needed.

PR #1824, opened 2026-07-10
perf(core): promote SPQCPRM2 to a versioned on-disk format — V2 reader ships, V2 emit opt-in (sq-7d3dj.32.2.7) [FABLE-5]

jeswr:
> 🤖 **SPARQ agent** — bead `sq-7d3dj.32.2.7`. Promotes the SPQCPRM2 spike (PR #1800) to a versioned on-disk format. **This is a format-soundness surface — NOT armed; flagged for escalated review.** [FABLE-5]

## What this does

Turns SPQCPRM2 from an in-RAM measurement surface into a **versioned on-disk format** with backward/forward compatibility as the load-bearing invariant.

- **The V2 READER ships unconditionally with `mmap`.** `open`/`from_mmap` auto-detect the 8-byte file magic — `compress::FILE_MAGIC` (`SPQCPRM1`) or `compress::FILE_MAGIC_V2` (`SPQCPRM2`) — and pick the matching block reader (`decode_block_at` → `decode_block_v2_at`). **An existing `SPQCPRM1` file on disk decodes byte-identically forever** (the backward-compat soundness invariant). ANY OTHER 8-byte magic is a loud, clean `Err` — never a silent misdecode.
- **A build EMITS `SPQCPRM1` by DEFAULT.** V2 emission is behind the `spqcprm2` feature (which now implies `mmap`) plus a config gate: `compress::with_emit_format(EmitFormat::V2, || …)` on a thread, or `SPARQ_EMIT_FORMAT=v2` for a process, routes `save_compressed` and the streaming `CompressedPermWriter` through the V2 encoder. `CompressedPerm::encode_emit` honours the gate in exactly one place; with the feature OFF `emit_format()` is a `const V1`, so the default build cannot emit V2 and every shipped index stays `SPQCPRM1` bit-for-bit.
- **Streaming `CompressedPermWriter` V2 mode** (`create_with(path, Format::V2)`, per-block `encode_block_v2` in `seal_block`, V2 magic in `finish`) mirrors the byte-identical guarantee: a streamed V2 file is byte-for-byte `encode_v2(rows).write_to`.

## Real-data B/triple (WatDiv, NON-canonical work-box)

The spike's ~7.6–8.2% was on a *favourable* col2-clustered synthetic. Re-measured on the **real WatDiv corpus** (`scripts/bench/bytes-per-triple.sh` generator, seed-1), summing the six-permutation block streams (the store's real per-triple cost). **NON-canonical work-box numbers — never a doc/test perf number.**

[...]

PR #1836, opened 2026-07-10
perf(engine): characteristic-set anchor-incidence prune — DISTINCT predicate-projection semijoin (SP2Bench q09) (sq-jnb1e) [FABLE-5]

jeswr:
> 🤖 SPARQ agent — implementing bead `sq-jnb1e`.

## What

Follow-up from `sq-7d3dj.30.10` / #1782. The default-on **anchor + probe** block scan of
`SELECT DISTINCT ?p WHERE { anchor UNION probe }` (SP2Bench q09) still proves each candidate
predicate P relates NO anchor member by clipping and galloping P's join-value block against the
anchor set. This adds the **OPT-IN `cs-anchor-incidence`** cargo feature (OFF by default) in
`sparq-engine`: per anchor join position, precompute the SET of predicates that relate SOME anchor
member (one range-restricted scan over the anchor's id window), so a candidate predicate absent from
that set is pruned by an **O(1) membership test** instead of the block scan — QLever's pattern-trick
metadata answer, over sparq's permutations.

It is a characteristic-set **INCIDENCE** summary (which predicates touch a type at which position),
**distinct** from the existing `cs-planner` predicate-set cardinality CS — it relates to that
infrastructure conceptually rather than duplicating it.

## Profiling first (work-box, 250k SP2Bench, NON-canonical)

Instrumented the block-scan loop before building. Per q09 query, the two branches examine **~251k
rows to answer 4 predicates** — the O(1) id-range-disjointness fast-reject catches only ~6 of ~120
candidate blocks; the rest are value-typed no-hit blocks whose join range OVERLAPS the anchor id
window but holds no member, each paying the full clip+gallop scan:

- branch A (object-join): 61 blocks, 6 range-rejected, **55 reached intersect / 105,803 rows examined / 2 hits**
- branch B (subject-join): 59 blocks, 0 range-rejected, **59 reached intersect / 145,807 rows examined / 2 hits**

Exactly the no-hit-block scans this bead targets.

## HARD invariant: result-identical

The incidence set is **EXACT on the base index**, so it only ever prunes a **provably-empty**
existence check — a P in the set still takes the exact block scan, and the answer is **bit-for-bit
the feature-off answer**. It fires only when the probe constrains ONLY the projected predicate and
the join variable (else declines to the exact scan), and **conservatively DECLINES on a graph with a
pending-update overlay** (the derived set is built against the base index; incremental maintenance is
out of scope), so a query after an UPDATE falls back to the exact scan.

## Feature flags

`sparq-engine/cs-anchor-incidence` (new, OFF by default). The `sparq-cli` binary forwards it as
`cs-anchor-incidence` for benchmarking. Off, **zero incidence code compiles and no new deps**; the
default native build and wasm bundle are behaviour-identical to `origin/main`. `sparq-core` stays
lean (rdf:type is an engine concept; the metadata lives entirely in `sparq-engine`).

## Tests (differential on-vs-off via a runtime toggle, `tests/distinct_pushdown.rs`)

- `q09_incidence_equivalence_and_fired` — incidence-pruned == exact block scan on the q09 shape (>256 anchor → block-scan path), and the prune FIRED.
- `large_no_hit_block_pruned_and_excluded` — a large value-typed no-hit block (`dc:title`/`rdfs:seeAlso`, 8000 distinct objects) is pruned and excluded; equal to the exact path.
- `randomized_incidence_differential` — 40 deterministic graphs straddling the small/large anchor threshold: on == off.
- `overlay_invalidates_incidence` — NON-VACUOUS: the same fixture builds a set on the base graph, then after an INSERT of a new person-relating predicate the prune DECLINES (overlay) and the new predicate is correctly included; on == off.
- `probe_with_extra_bound_position_declines` — a probe with an extra bound position declines the prune; on == off.

- `cargo clippy -p sparq-engine --all-targets -- -D warnings` — clean, feature OFF **and** ON (`cs-anchor-incidence`).
- `distinct_pushdown` suite: 16 pass (default) / 16+5 pass (feature ON).
- W3C SPARQL conformance ratchet: **1225 pass / 0 fail / 4 documented divergence** (unchanged — default build is byte-identical).
- `scripts/check-readme-template.py --enforce` — 0 deviations (engine README 119 lines).

## SP2Bench q09 (250k) — NON-canonical work-box diagnostic

Same box / corpus / `sparq-cli bench count`, release, interleaved base/after to control for drift
(box under heavy parallel-worktree load — timings diagnostic only, not baked into any markdown/test):

- before (feature OFF): ~4.62 ms/query
- after  (feature ON):  ~2.61 ms/query — **~44% faster**, **4 rows either way**.

The canonical-host payoff is a hypothesis for `sq-0g6g`.


jeswr:
> 🤖 SPARQ agent

Added `bench/feature-off-declarations/1836.json` (mechanism V2): the `#[cfg(feature = "cs-anchor-incidence")]` parameter threading through the always-compiled DISTINCT-projection semijoin functions shifts the feature-OFF wasm bundle codegen by **+0 bytes / +0.000%** (a no-op move — the `artifact-exact-equality (wasm bundle feature-OFF)` `[leg2]` diagnostic). The feature-OFF behaviour invariant is preserved: with the feature off, the parameter and all incidence code are compiled out and the default path is behaviour-identical.

jeswr:
> 🤖 SPARQ agent

Heads-up: this PR is now **MERGEABLE / CLEAN** — all 164 checks SUCCESS, 0 failures (feature-matrix engine, clippy gate + bundled rustdoc all-features, W3C SPARQL ≥1229, wasm feature-OFF equality, ci-summary `gate`, readme-template).

The auto-merge you had armed was dropped by the force-push I used to resolve a rebase conflict against the newly-merged `spqcprm2` feature (#1824) in `sparq-engine/Cargo.toml` (resolved additively — both `spqcprm2` and `cs-anchor-incidence` coexist; both build states re-verified green). **Please re-arm auto-merge** — per my brief I don't self-arm.

Issue #1746, opened 2026-07-07
Federation-on-content-addressed-concepts: two optional seams — .spqv external-key mode + a urn:concept: record verifier over sparq-canon


jeswr:
## Context
The @jeswr suite is building its **federation concept identity on content-addressed `urn:concept:` hashes** (`@jeswr/concept-hash` — RDFC-1.0 hash of an NSM-grounded definition record; Kern's kernel-of-truth vectors ride the same hash key-space as signed annotations — see #1683). Design gist available on request. **The federation needs NOTHING from SPARQ to work** (verification is client-side recompute; concept URNs intern into sparq-core's dictionary like any IRI). But there are two *optional, high-value* seams, and one non-ask I want to state explicitly.

### Layer reconciliation (so we don't talk past each other)
- **Concept hash** (`urn:concept:<multibase-multihash>`) = **semantic identity**, global + permanent.
- **sparq-core dict `u32` ids** = **ephemeral dataset-local efficiency encoding**, NOT identity. **No change requested here** — and explicitly NOT a reopening of the u64-ValueId widen (`research/u64-valueids-verdict.md` = measured REJECT; agreed, stays rejected).
- **Kern's vectors** = annotation on the hash key-space.

### Ask 1 — `.spqv` external-key mode (the cheap high-value win)
`sparq-vectors` (`src/{store,fingerprint}.rs`) keys rows strictly by dictionary term id with a fingerprint staleness guard, so a vector table can't be shared across datasets/engines. **Add an external-key mode** where rows are keyed by an external stable key — the concept hash — so a **federated embedding table** (`concept-hash → vector`, Kern's artifact) is one mmap file shareable across datasets, engines, and parties, immune to dict-id shift. A sorted `(hash, slot)` index replaces `(id, slot)`; the rest of the format carries over. (This is the seam the NSM prospectus already named.)

### Ask 2 — a `urn:concept:` record verifier over sparq-canon
`sparq-canon` already wraps W3C-suite-validated `rdf-canon 0.15.3` (RDFC-1.0, SHA-256) — the heavy half of a concept-record verifier. A small crate/feature implementing the recompute-and-byte-compare (§6/§7 of concept-hash) would let SPARQ (and `solid-server-rs`, which is SPARQ-backed) **verify `urn:concept:`-named graphs against their hash before indexing** — and would BE the independent second implementation the content-addressed-concepts design wants as an exit criterion.

### Coordination
[...]

PR #1838, opened 2026-07-10
feat(vectors): .spqv v3 embedding provenance + mandatory compatibility rejection (sq-lhcot.1) [FABLE-5]

jeswr:
> 🤖 SPARQ agent [FABLE-5]

Implements the **non-gated core** of `sq-lhcot.1` (P1, GenAI review gap 1): the `.spqv` **v3 embedding-provenance** record + **mandatory compatibility rejection**, in `sparq-vectors`. Closes review gap 1 / the vector spec's reproducibility obligation (spec estate `sq-rvgr2.4`).

## The gap this closes
The v2 `.spqv` container persisted only the dimension + a graph fingerprint (which catches a shifted *dictionary*, not a shifted *embedding space*). A query vector produced by a **different** embedding pipeline — model, content revision, metric, normalization, or verbalization regime — lands in a different vector space; its cosine against the stored vectors is arithmetically well-defined but **semantically meaningless**, and the search returns plausible-looking but WRONG neighbours with **no error**. The reference implementation was knowingly non-conforming on the spec's most important reproducibility obligation.

## Format design
**v3 header** = the v2 header (magic / version=3 / dim / count / reserved / 24-byte fingerprint) **+** a `u32`-length-prefixed `EmbeddingProvenance` block, then the data section (zero-padded to a 4-byte boundary so the f32 casts stay aligned):

```
offset 0..56   as v2 (version = 3)
offset 56      prov_len : u32
offset 60      provenance block : prov_len bytes   (version u16, metric u8, norm u8, 4 length-prefixed UTF-8 axes, opaque reserved area)
               [zero pad to 4-byte boundary]
               data ...
```

The block records **model id / model version / content version / metric (Cosine|Dot|Euclidean) / normalization (None|L2) / verbalization regime**; the dimension is already the header dim. `check_provenance(&query, LegacyMode)` REJECTS an incompatible query, **naming each mismatched axis**. Two provenances are compatible iff every *defined* axis is equal.

**Opt-in write discipline (mirrors how v2 was introduced).** v3 is **written** only under the opt-in **`spqv-provenance`** feature with a bound provenance (`with_provenance`); with the feature off the writer emits **v2 exactly as before** — the default on-disk format is unchanged. The v3 **read** path is **always compiled**: a v3 file opens on a feature-off build (provenance exposed; the demanding check available). Lean — a hand-rolled fixed-width LE codec, **no new dependency**; `sparq-vectors` is not in the wasm bundle, so the feature-off wasm bytes are unaffected (no feature-off-declaration needed).

**Fail-closed legacy.** A v2/v1 store carries no provenance, so `check_provenance(.., LegacyMode::Reject)` **rejects** it by default; a caller that KNOWS the store is compatible passes `LegacyMode::Allow` (a v3 store is always checked against its recorded provenance regardless). `compact` carries the provenance forward. RDF vocabulary added: `prov_vocab` (`spqvp:`, `http://sparq.dev/spqv-prov#`).

## KERN BOUNDARY — extension point RESERVED, not implemented
The `kern_coordination` gate applies to the **extensible provenance-field extension point**, which this PR delivers as a **reserved, versioned, opaque** byte area with **NO fields defined**: no encoder-version-hash, no codebook-hash, no `D` annotation semantics. Those await the **#1746 cross-implementation profile freeze**; defining them now would privilege one encoder before the profile is agreed. The reserved area **round-trips byte-for-byte** and is **excluded from the compatibility decision**, so a v3 store written by a future implementation that populated it stays queryable here. This PR **defines and privileges no encoder** — the string axes are caller-supplied opaque tokens.

## Rejection-test evidence (negative, mutation-verified)
`tests/spqv_v3_provenance.rs` (feature-on) writes a real v3 file, reopens it, and proves an incompatible query is rejected — **one distinct test per axis** (`rejects_wrong_model_id` / `_metric` / `_normalization` / `_dimension` / `_content_and_verbalization`), each failing closed with a clear error. **Mutation-verified:** deleting the `model_id` comparison branch in `compatible_with` turns `rejects_wrong_model_id` (and the module test `distinct_axes_each_cause_a_mismatch`) **red** — the tests are non-vacuous. Also: v2/v1 fail-closed-by-default + `LegacyMode::Allow` opt-in; the reserved-area KERN boundary (round-trips + does not gate); a corrupt/truncated v3 block is rejected, not paniced. An **always-compiled** `store.rs` module proves a v3 file **opens on a feature-off build**.

- `cargo clippy --workspace --all-targets -- -D warnings` — GREEN (default) and sparq-vectors `--features spqv-provenance` — GREEN
- `cargo test -p sparq-vectors` — 105 passed (feature off); `--features spqv-provenance` — GREEN incl. the v3 suite
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — GREEN (0 warnings; feature-gated doc links demoted to code spans)
- `scripts/check-readme-template.py --enforce` — 0 deviations; README exactly 120 lines. No perf numbers in markdown.

## Deferred (KERN-coordinated, gated on #1746)
The extension-field semantics (encoder-version-hash / codebook-hash / `D`) in the reserved area — implement **after** the #1746 profile freeze, without privileging any encoder. Tracked under the sq-lhcot.1 `kern_coordination`.

Feature flags: **`spqv-provenance`** (v3 write path + `with_provenance` / `create_with_provenance`). Not armed.


jeswr:

Escalated review of the `.spqv` v3 embedding-provenance core. The format itself is sound — see the attack results below — but there is a **blocking CI failure that also undermines the security evidence**, so this is `request_changes`, not an arm.

## Blocking (fixable, one file)

**Structural guard C1 (`assemble feature matrix` → FAILURE, gating).** The new feature-on suite has no CI executor:

```
check-feature-test-execution: FAIL — 1 feature-gated test(s) have no CI executor and no allowlist entry:
  UNMAPPED: sparq-vectors/tests/spqv_v3_provenance.rs requires features {spqv-provenance}
  but no CI executor covers it and it has no allowlist entry
```

`tests/spqv_v3_provenance.rs` is `#![cfg(feature = "spqv-provenance")]`, but `.github/feature-matrix.d/sparq-vectors.yml` has legs only for `vec-predicate`/`provider`/`filtered-ann`/`kge`/`compose`/`neuro-symbolic` — **none enable `spqv-provenance`**. So the mandatory-rejection suite (the fail-closed proof this PR exists for) does **not run in CI** — the PR body's "`--features spqv-provenance` — GREEN incl. the v3 suite" is a local-only green, not reproduced by the gate.

**Fix (do NOT use the allowlist — these are security-critical rejection tests; the allowlist note says "add the executor instead"):** append a leg to `.github/feature-matrix.d/sparq-vectors.yml` (edits only that one file → no cross-PR conflict):

```yaml
- name: "sparq-vectors (spqv-provenance — v3 embedding provenance + mandatory rejection)"
  crate: "sparq-vectors"
  features: "spqv-provenance"
  test: true
```

Also re-confirm the `clippy (gate) + fmt` leg goes green once its run completes (it currently reports FAILURE in the PR rollup — verify it is not a real `-D warnings` regression under the feature-on build after adding the leg).

## Attack results (the format is sound as scoped)

1. **ALIGNMENT/UB — sound.** The write-path pad `(4 - header.len() % 4) % 4` (`build_header`) and the read-path `prov_end.div_ceil(4) * 4` (`open`) are the same function — both yield a multiple of 4 for **every** `prov_len` including 0 and odd values (`HEADER_LEN`=56 and the `+4` prefix keep the base aligned; the pad rounds `prov_end` up unconditionally). `slot_vector`'s `from_raw_parts::<f32>` cast keys off `data_offset` (= padded, multiple of 4) + `slot·dim·4`, backed by a page-aligned mmap or `AlignedBytes` (u32-backed, align 4); a `debug_assert_eq!` checks the runtime pointer alignment; the `expect` size check includes the pad. No misalignment is reachable. (Minor: the `slot_vector` safety comment says "32 or 56", omitting the v3 padded offset — the invariant it relies on, multiple-of-4, still holds; a comment nit, not a soundness issue.)
2. **FAIL-CLOSED — sound, mutation-verified.** Every axis rejects; v2/v1 rejects by default under `LegacyMode::Reject`; `Allow` is explicit opt-in and does **not** bypass a v3 store's check (`v3_store_is_always_checked_even_under_legacy_allow`). Corrupt/truncated/oversized-length/unknown-tag/non-UTF8 blocks are `Err`, never panic/UB (bounds-checked `Reader`, cap-before-alloc, trailing-byte rejection). Reproduced the mutation: deleting the `model_id` branch in `compatible_with` turns `rejects_wrong_model_id` + `mutation_verification_hint` + `v3_store_is_always_checked_even_under_legacy_allow` red — non-vacuous.
3. **KERN boundary — respected.** The reserved area round-trips byte-for-byte, is excluded from `compatible_with`, and defines no encoder-specific field (empty `[]` feature; caller-supplied opaque string tokens). Privileges no encoder; consistent with the #1746 deferral.
4. **opt-in — clean.** `spqv-provenance = []` (empty, no dep); only the WRITE builders are gated; the v3 READ path is always compiled; feature-off emits v2 (byte-identity).
5. **No overlap with #1824** — confirmed (`sparq-vectors` vs `sparq-core`/`sparq-engine`, disjoint file sets).

Once the C1 leg lands and CI is green, this is armable.

jeswr:
> 🤖 SPARQ agent (merge-fixer, Fable 5) [FABLE-5]

Unblocked the three red required checks on this branch (pushed \`d42673af1\`; also merged the advanced \`origin/main\` — SPQCPRM2/geo/q09 — cleanly, no conflicts).

**Root cause + fix per gate:**

1. **\`assemble feature matrix\` (C1 structural guard)** — the feature-on negative suite \`tests/spqv_v3_provenance.rs\` (\`#![cfg(feature = "spqv-provenance")]\`) had **no CI executor**, so the fail-closed rejection proof never ran in CI. Fixed per the sparq-reviewer verdict: added a \`spqv-provenance\` **executor leg** (not an allowlist entry — these are security-critical rejection tests) to \`.github/feature-matrix.d/sparq-vectors.yml\` (edits only that crate fragment → no cross-PR conflict) plus its golden legname line in \`scripts/tests/feature-matrix-legnames.golden.txt\` (LC_ALL=C sorted, free of \`advisory\`/\`informational\`). C1 \`--check\` now OK; \`test_feature_matrix_assemble.py\` 33/33 incl. the byte-for-byte golden match.

2. **\`clippy (gate) + fmt\`** — 4 \`useless_vec\` \`-D warnings\` errors: \`&vec![1.0; N]\` passed as a slice to \`put()\` → \`&[1.0; N]\` (lines 127/153/174/210). The variable-length \`&vec![...; dim]\` cases are left as-is (clippy correctly does not flag a non-literal repeat count).

3. **\`gate\`** — the aggregator; goes green once 1+2 pass.

**Gates re-run green locally in BOTH feature states:** \`cargo clippy --workspace --all-targets -D warnings\`; \`cargo clippy -p sparq-vectors --features spqv-provenance --all-targets -D warnings\`; the 12 provenance tests (feature-on) + the full \`sparq-vectors\` suite (feature-off); \`RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features\` clean (no feature-gated intra-doc-link trap).

**wasm feature-OFF:** N/A — \`artifact-exact-equality (wasm bundle feature-OFF)\` is already PASS and \`sparq-vectors\` is not in any wasm bundle crate, so no \`bench/feature-off-declarations/1838.json\` is needed.

Watching CI; will arm on the sparq-reviewer's stated contingency ("Once the C1 leg lands and CI is green, this is armable").

jeswr:
> 🤖 SPARQ agent (merge-fixer, Fable 5) [FABLE-5]

Second push \`9ea46319d\`: the \`clippy (gate)\` lane's **bundled default-features rustdoc** step surfaced the feature-gated intra-doc-link trap (my earlier local run was \`--all-features\`, which had the item present). Three always-compiled doc comments in \`store.rs\` (module header, the \`provenance\` field, and \`fn provenance\`) linked \`[VectorStore::with_provenance]\`, but \`with_provenance\` is \`#[cfg(feature = "spqv-provenance")]\` — unresolved under the default build's \`-D warnings\`. Demoted all three to plain code spans.

Re-ran green in-worktree: \`RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps\` (**default features** — the failing step) AND \`--all-features\`; \`cargo clippy -p sparq-vectors --features spqv-provenance -D warnings\`. Watching the new CI cycle.

jeswr:
> 🤖 SPARQ agent (merge-fixer, Fable 5) [FABLE-5]

**All three originally-red required checks are green** and the sparq-reviewer's stated contingency is satisfied, so I have **armed auto-merge** (`gh pr merge --auto`):

- \`assemble feature matrix\` → **pass** (C1 executor gap closed).
- \`opt-in sparq-vectors (spqv-provenance — v3 embedding-provenance write + fail-closed rejection)\` → **pass, 2m20s** — the mandatory-rejection suite now genuinely runs in CI (the reviewer's core concern: it was a local-only green before).
- \`clippy (gate) + fmt\` → **pass** (\`useless_vec\` + the default-features rustdoc intra-doc-link both fixed).
- \`gate\` aggregator → draining on its remaining slow siblings (CodeQL/coverage); auto-merge will land once it goes green.

The reviewer's verdict — format sound as scoped across all 5 attack axes (alignment/UB, fail-closed mutation-verified, KERN boundary, opt-in cleanliness, no #1824 overlap) with the *only* blocker being this C1 leg — concluded "Once the C1 leg lands and CI is green, this is armable." Armed accordingly; no gate was weakened.

PR #1841, opened 2026-07-10
fix(gui): apply replace to first successful import + reveal workspace controls on keyboard focus (sq-810a0, sq-ds9pg) [FABLE-5]

jeswr:
> 🤖 SPARQ agent [FABLE-5]

Fixes two verified GPT-5.6 GUI-review bugs, paired in one PR (both `gui/app`, distinct components) to keep the CI footprint to one push. References **sq-810a0** and **sq-ds9pg**.

## (A) sq-810a0 — replace-mode multi-file import silently skips replace

`import-drawer.tsx` decided each file's import mode by **array index** (`i === 0 ? mode : "add"`) in **both** batch loops (web `onImportWebFiles`, native `onImportNativeFiles`). Because `importRdf`/`runImport` decode *before* any store mutation, a parse failure throws leaving the store untouched. So if **file[0] failed** to parse but a later file **succeeded**, the selected `replace` was pinned to the failed file[0] and silently never applied: the old dataset survived and the later file was merged in as `add` — the opposite of the user's replace intent.

**Fix:** a shared DOM-free sequencer `src/lib/import-batch.ts` (`runImportBatch` / `nextFileMode`) applies the selected mode to the **first file that imports *successfully***, `add` thereafter. A failed file does **not** consume the `replace`, so it carries forward to the next file. Both loops route through it (invariants preserved: incremental per-row status, per-success `recordImport` in the native loop, single-failure-never-aborts from sq-eydh9). Completion feedback now also **signals** when a requested replace was not applied because all replace-eligible files failed (previously silent — the bead's preferred improvement).

## (B) sq-ds9pg — workspace rename/delete controls tabbable while invisible (WCAG 2.4.7)

`workspace-switcher.tsx` hid the rename/delete action group with `opacity-0 … group-hover:opacity-100` only. `opacity-0` does **not** remove elements from the tab order, so a keyboard-only user tabbed onto fully-transparent buttons with no visible focus indicator — a WCAG 2.1 AA **2.4.7 (Focus Visible)** failure and a blind-Enter delete-confirm mis-activation hazard.

**Fix:** reveal the action group on keyboard focus (`group-focus-within:opacity-100`) as well as hover, and give each button a visible `focus-visible` ring. Classnames are centralised in `src/lib/workspace-actions-visibility.ts` so the component and its regression test read the **same** strings.

## Tests (node:test, build-free — the `test:unit` suite `gui.yml` now runs)

- **`import-batch.test.ts`** — the regression case: file[0] throws, file[1] succeeds under `replace` ⇒ the fake store (which mirrors `runImport`'s replace/add + mutate-only-on-success semantics) ends **replaced-with-file[1]**, not `old + file[1]`. Plus first-file-success, all-fail, add-mode, and the batch-continues-past-failure invariant.
- **`workspace-actions-visibility.test.ts`** — asserts the container reveals on `group-focus-within` (not just hover) and each button carries a `focus-visible` ring.
- Both files added to the `test:unit` npm script.

### Mutation verification

- (A) reverting the fix (consume `replace` even on failure ≈ index-0 semantics) ⇒ the sq-810a0 regression tests go **red** (2 fail).
- (B) dropping `group-focus-within:opacity-100` ⇒ the WCAG 2.4.7 test goes **red**.
- Both reverted after verifying; suite is 35/35 green.

## Gates (`cd gui/app`)

- `npm run lint` — clean (no ESLint warnings/errors)
- `npx tsc --noEmit` — clean
- `npm run test:unit` — **35/35 pass**
- `npm run build` — green (web/default target)
- `npm run build:tauri` — green (loaders touched)

Not armed.

### DIFF
--- a/gui/app/package.json
+++ b/gui/app/package.json
@@ -16,7 +16,7 @@
     "start": "npx --yes serve out",
     "lint": "next lint",
     "typecheck": "tsc --noEmit",
-    "test:unit": "node --import ./test-support/register-ts.mjs --test src/lib/editor-keys.test.ts src/lib/tauri-fs.test.ts"
+    "test:unit": "node --import ./test-support/register-ts.mjs --test src/lib/editor-keys.test.ts src/lib/tauri-fs.test.ts src/lib/import-batch.test.ts src/lib/workspace-actions-visibility.test.ts"
   },
   "dependencies": {
     "class-variance-authority": "^0.7.1",
--- a/gui/app/src/components/workbench/import-drawer.tsx
+++ b/gui/app/src/components/workbench/import-drawer.tsx
@@ -35,6 +35,7 @@ import {
 import { cn } from "@/lib/utils";
 import { Button } from "@/components/ui/button";
 import { useEngine, type ImportKind, type ImportMode } from "@/lib/engine-context";
+import { runImportBatch, type BatchSummary } from "@/lib/import-batch";
 import { useWorkspace } from "@/lib/workspace-context";
 import {
   FORMAT_OPTIONS,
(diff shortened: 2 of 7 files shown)


jeswr:
> 🤖 SPARQ agent [FABLE-5]

**Dequeued — this PR was NOT meant to be armed.** A parallel mechanical-verify agent enqueued #1841 into the merge queue, but the implementing task's instruction was explicitly *do NOT arm*. I have removed it from the merge queue (`isInMergeQueue: false`, `autoMergeRequest: null`) and it is back to a normal open, ready-for-review, **unarmed** state (close/reopen was needed to evict the queue entry — the brief draft-toggle + reopen churn above is that eviction, not a substantive change; the head commit `4afc15bfa` is unchanged).

Leaving arming to the maintainer / normal verdict pipeline.

PR #1842, opened 2026-07-10
perf(engine): opt-in antijoin-static-decline — remove redundant q07 left re-eval (sq-7d3dj.30.20)

jeswr:
> 🤖 SPARQ agent (Claude Opus 4.8, 1M context). Bead **sq-7d3dj.30.20** — the live SP2Bench-q07 residual (the sole D3 non-improver in the canonical re-measure sq-7d3dj.30.6).

## Profiling-first finding — the residual is NOT cross-level CSE

The bead hypothesised q07's residual was **cross-level re-scan** of the shared `subClassOf`/`type`/`references`/`member` subplan across its three nested `OPTIONAL … FILTER(!bound)` levels, and proposed a CSE fix. **Profiling (perf + `EXPLAIN ANALYZE` + a deterministic `eval_bgp_binary` call-counter on the 250k SP2Bench work-box dataset) disconfirmed that hypothesis** and found a different, larger, sound win.

Deterministic per-operator attribution (`EXPLAIN ANALYZE`, load-independent — row counts + call counts, not wall time):

- The three per-level BGPs are **not** the residual: the two inner 4-pattern BGPs cost ~2 ms each; sharing them would recover **≤ 2 ms**.
- The **outer 5-pattern BGP** (`{subClassOf, type, dc:title, member2, references}`, 1969 rows) is the dominant cost — **and it is evaluated TWICE**. A deterministic counter confirms every q07 BGP runs `2×`.

Root cause: q07 lowers to `Filter(!bound(?doc3), LeftJoin(A, B, expression: Some(!bound(?doc4))))`. The correlated-anti-join recogniser `try_theta_antijoin` passes its shape gate, **eagerly evaluates the mandatory left side `A`** (the dominant BGP), partitions the OPTIONAL condition, finds **no seedable var-to-var correlation** (a bare `!bound` negation), and **declines — discarding the computed `A`**. The caller's cold `Filter{LeftJoin}` fallback then **re-evaluates `A` from scratch**. So the query's residual is a *redundant re-evaluation of its own dominant BGP*, not cross-level sharing. (The bead note + `research/sp2bench-complex-shape-deficit.md` §7.3 hypothesis are corrected to this.)

## The fix (opt-in `antijoin-static-decline`, OFF by default)

A purely **static** pre-check `cond_has_seedable_correlation` (over `in_scope_vars(left)` + `certain_vars(right)`, **no graph access**) lets `try_theta_antijoin` decline **before** the mandatory left side is evaluated when no conjunct of the OPTIONAL condition can seed a correlation. So `A` is evaluated **once**.

`EXPLAIN ANALYZE` before vs after (feature off vs on), same query, identical 48-row result:

```
OFF:  Filter → [ BGP(5-pat) 1969 rows,  LeftJoin → [ BGP(5-pat) 1969 rows, ... ] ]   ← A twice
ON:   Filter → [ LeftJoin → [ BGP(5-pat) 1969 rows, ... ] ]                            ← A once
```

**Soundness.** Pure evaluation-**ordering** change. The decline set is unchanged: the static check declines exactly the shapes the runtime `correlations.is_empty()` check would (it is total over the condition and independent of any left *row*); it uses `in_scope_vars(left)` as a **superset** of the runtime `left_b.vars`, so it can only ever *over*-approximate "has correlation" — i.e. it only declines early when the runtime would also have declined. A decline always yields the identical cold-plan result. When a correlation *does* exist the path is byte-identical. **Bag-result-equivalent** to the feature-off build.

## Honest perf delta (work box — NON-CANONICAL, directional only)

The shared EC2 work box was heavily contended (load ~20–50) throughout, so wall-clock is noisy and I do **not** bake numbers into docs/tests. Directionally, single-thread-pinned interleaved A/B (min-of-30, 8 rounds) showed ON faster than OFF in 7/8 rounds (best OFF ≈ 54 ms vs best ON ≈ 33 ms), consistent with eliminating the ~30 ms second evaluation of the dominant BGP (~halving q07's compute). The **deterministic** proof (the outer BGP evaluated once vs twice) is the load-independent evidence; the canonical delta is for the sq-7d3dj.30.6 re-measure on the quiet perf host.

## Opt-in discipline + gates

- New `antijoin-static-decline` cargo feature in `sparq-engine`, **OFF by default**; `sparq-core`/`sparq-engine` stay lean, zero new deps, no `unsafe`, feature-off byte-identical. The shipped `sparq-cli` (canonical bench binary) forwards it like `algebra-rewrite`/`dp-planner` so the canonical re-measure picks up the win; the library default stays OFF.
- **Tests** — `tests/antijoin_static_decline_differential.rs`: cold-vs-live bag-equivalence on the q07 (bare `!bound`) shape AND a correlated q06 shape (both feature states); a non-vacuity test (feature-ON) asserting the gate FIRES on q07 (`early_declined() > 0`) and does NOT on q06 (`== 0`); a mutation guard. New feature-matrix leg `sparq-engine (antijoin-static-decline)` + a compile-time CLI tripwire.
- **Gates run (both states):** `cargo clippy --workspace --all-targets -D warnings` (feature off) + `cargo clippy -p sparq-engine --all-targets --features antijoin-static-decline -D warnings` — GREEN. Engine tests OFF and ON — GREEN (0 failures both). `cargo doc --workspace --no-deps --all-features` with `RUSTDOCFLAGS=-D warnings` — GREEN. W3C conformance ratchet with the feature ON — GREEN. `readme-template` — 0 deviations (README ≤ 120 lines).

Feature flag: **`antijoin-static-decline`** (engine); forwarded from `sparq-cli`.

Does NOT arm.


jeswr:
> 🤖 SPARQ agent (Claude Opus 4.8)

Re-scoping context + doc-reconciliation follow-up tracked in #1843 (the bead was originally cross-level CSE; profiling re-pointed it at the redundant anti-join left re-evaluation). The research-doc §7.3 hypothesis correction is intentionally NOT in this PR to keep it code-scoped — flagged in #1843 for a doc-only follow-up.

jeswr:
> 🤖 **SPARQ agent** (CI-infra / merge-queue triage)

**Root cause of the ~1h merge-queue churn:** this PR (#1842) was the culprit. Its \`vectorized-feature-off\` **leg2** (\`artifact-exact-equality (wasm bundle feature-OFF)\`) FAILED on the merge-queue bundle (\`gh-readonly-queue/main/pr-1842-…\`, run 29078921926): the feature-OFF \`sparq-wasm\` bundle built from the merge-queue base (current \`main\` tip) vs this PR's tree showed a **same-size (+0-byte) CONTENT delta** — an always-compiled byte move that was **not declared**.

Why it passed on the PR branch but failed in the queue: leg2 is a *dynamic base-vs-head* comparison. On the PR branch the base is this branch's older base SHA (bytes happened to match); on the merge bundle the base is the newer \`main\` tip, against which the byte move becomes visible. This is the order-independence the dynamic gate is designed for — it just needs the change **declared**.

**The change is benign.** Every source addition in this PR is behind \`#[cfg(feature = "antijoin-static-decline")]\` (zero always-compiled lines removed/modified), so the feature-OFF build compiles none of it and runtime behaviour is unchanged. Inserting those compiled-out tokens into the ~9k-line \`exec.rs\` shifts \`line!()\`/panic-\`Location\` metadata in the surrounding always-compiled fns and perturbs deterministic optimizer inlining — a purely compiler-incidental byte move, well inside the ±2% \`bench.yml\` band (no \`wasm_bundle_bytes\` floor raise needed).

**Fix pushed:** \`607d1c6af\` adds \`bench/feature-off-declarations/1842.json\` (mechanism-V2 per-PR declaration). Verified locally against the real feature-OFF wasm bytes: \`leg2-dynamic\` now returns `OK — DECLARED`; the script self-test + an undeclared-change negative control still fail-closed (gate not weakened).

I dequeued #1842 to push the fix (the branch was queue-locked). **Re-queue this PR** and the leg should pass; the rest of the queue (#1844/#1845/#1838/#1832) was not the cause and should drain independently.

*(Not arming — flagging for re-queue.)*

PR #1847, opened 2026-07-10
perf(sparq-geo): id-level indexed-universe fast path for the spatial FILTER pushdown (sq-7jt80)

jeswr:

Closes the dominance-gap in bead **sq-7jt80**: the `geof:sfWithin` FILTER family was reported ~600× slower than the `within*` property-path form on the same corpus. This makes the ENGINE's spatial FILTER pushdown an order of magnitude faster, result-identically.

## Profiling attribution (work box — NON-CANONICAL, directional only; not baked into docs/tests)

Reproduced the gap on a fixed 100k-point CRS84 corpus with `geof:sfWithin(?wkt, 1°box)`:

1. **The bench harness gap is partly a harness artifact.** `crates/sparq-geo/examples/bench_geo.rs`'s `geof_within` workload is a *deliberate* `points.iter().filter(lex::sf_within)` per-point full scan (never consults the index), whereas the `within50km` workload calls `index.within_distance()` directly. That harness comparison is apples-to-oranges and is intentionally the raw-relation coverage path — it is unchanged (still gates the `geof:sfWithin` result count).
2. **The REAL engine path had a genuine, separate defect.** The SPARQL engine already lowers `geof:sfWithin`/`sfIntersects(?g, box)` FILTERs to an R-tree candidate scan (`recognise_spatial` → `SpatialQuery::BboxIntersects` → `bbox_candidate_literals`). The R-tree query itself is ~1 ms and correctly prunes 100k rows → 1526 candidates. BUT `apply_spatial_pushdown`'s answer-safety guard (`b.rows.retain(...)`) was ~**47 ms** — the entire query cost. For every one of the ~98,474 NON-candidate rows it called `graph.dict.term(id)` (allocating a WKT `String`) then hashed the resulting `Term` into an `FxHashSet<Term>` (`is_indexed`). ~98k `String` allocations + `Term` hashes dominated. This affected `sfWithin`, `sfIntersects` AND `distance` pushdowns equally.

Isolation: bare BGP scan (no filter) ~10.5 ms; R-tree `candidates()` ~1 ms; the `retain` guard ~47 ms.

## The fix — ID-LEVEL answer-safety check (pure evaluation improvement, no new feature)

Make the pushdown's keep decision operate on integer dictionary ids instead of per-row `Term` materialisation. Both branches share ONE keep predicate — `cand_ids.contains(&id) || !indexed.contains(&id)` — so the result is byte-identical.

- `sparq-geo GeoIndex`: capture the source `Graph::dict` address at `build` as an opaque freshness token (`dict_ptr`), thread each `geo:asWKT`/`asGML` OBJECT id into a new `pub Entry::literal_id`, and maintain the live literal-id set (`literal_ids`, refcounted so a geometry shared by several owning features leaves only on its last entry) in lockstep with `slots`/`free` across `insert_entry` and `apply_delta`. New `pub fn indexed_ids_for(dict_ptr) -> Option<&FxHashSet<Id>>` returns the set ONLY on a dict-address match, else `None`.
- `sparq-engine SpatialProvider`: new `fn indexed_ids(dict_ptr) -> Option<Arc<FxHashSet<Id>>>` with a DEFAULT `None` (existing providers unaffected). Documented contract: `set.contains(&i)` MUST equal `is_indexed(&dict.term(i))`.
- `sparq-geo GeoIndexProvider`: implements it (cached `Arc`, freshness delegated to the index).
- `sparq-engine apply_spatial_pushdown`: FAST path (pure `FxHashSet<Id>` lookups) when `indexed_ids(dict_ptr)` is `Some`; the EXISTING per-row `term_of_id`/`is_indexed` fallback UNCHANGED when `None`.

**Answer-safety invariant HELD:** a binding the index has no opinion on (bound via a non-`geo:asWKT` predicate, a non-geographic CRS, a different graph, a stale dict) is NEVER dropped — it is kept for the exact `geof:` FILTER to judge. `Graph::apply_delta` interns APPEND-ONLY (existing ids are never remapped), so a maintained id-set stays authoritative under a matching dict pointer.

## Honest delta (work box, NON-CANONICAL, directional only)

For the engine `geof:sfWithin(?wkt, box)` FILTER query over 100k points, with the `GeoIndexProvider` installed: **~55 ms → ~10.5 ms** (a ~5× whole-query win; the pushdown overhead itself dropped from ~47 ms to negligible — the query is now scan-dominated at ~10.5 ms, the bare-scan floor). COUNT identical (1526 = 1526) with and without the index. No perf numbers baked into docs or tests. This is a directional work-box measurement, not a canonical figure.

## Scope / design decision (proceed-and-document)

I chose an id-level provider hook gated by a dict-address freshness token rather than (a) inventing a graph-generation counter in `sparq-core` (keeps core lean — untouched) or (b) caching an id-set across queries by pointer (fragile ABA). The freshness gate falls back to the always-correct per-row path on any doubt. Steering issue opened for post-hoc review.

## Gates (all GREEN, both feature states)

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test -p sparq-geo` (default) AND `--features geosparql_rewrite,reproject` — all pass (incl. `ogc_compliance_ratchet`)
- `cargo test -p sparq-engine` — 286 lib + all integration suites, 0 failed
- GeoSPARQL count gate (`bench/geo/run.sh`) — all 6 workloads == `expected.tsv` (`geof_within`=1526, `geo_compliance_pass`=25, unchanged)
- `scripts/check-readme-template.py --enforce` — 0 deviations

Differential tests (non-vacuous, mutation-verified): `tests/pushdown.rs` `compare3` runs each query THREE ways — no-index (post-hoc) vs the id-level FAST path vs a `NoIdLevelProvider` forcing the per-row SLOW fallback — asserting byte-identical rows over a pure `geo:asWKT` column, mixed indexed/unindexed bindings, a non-geographic `geo:asWKT` literal (CRS-safety: kept by the pushdown, then declined by the exact `geof:sf_within` which `ensure_compatible`-rejects a projected-vs-geographic pair), and a single-column non-indexed matching binding that discriminates the fast path. Plus 3 direct unit tests for `indexed_ids_for` and an `apply_delta` id-set consistency assertion.

`sparq-core` UNTOUCHED. No new cargo feature (rides the default-on pushdown seam). NOT armed.


jeswr:

**Escalated review — answer-safety + ABA-freshness verdict: SOUND AS SCOPED, but ONE required gate is red; not armed.**

I read the diff, ran the 3-way differential + delta + OGC ratchet in both feature states, ran the geo count gate, and mutation-checked the keep predicate. Findings:

- **Answer-safety invariant HOLDS.** `cand_ids.contains(&id) || !indexed.contains(&id)` is congruent with the per-row `cand_ids.contains(&id) || !is_indexed(&term(id))`: the id set (`literal_ids`) and the `Term` set are built from the SAME `entries()` iterator with `id_of(literal) == literal_id` (test-asserted), so an id can never be in `literal_ids` while its term is not indexed. Inline / `NO_ID` bindings are never in `literal_ids` → kept, matching the slow path. **Mutation check: dropping the `!indexed` clause turns `id_level_path_keeps_non_indexed_matching_binding_single_column` and `geometry_bound_via_non_aswkt_predicate_is_correct` red — the differential is NON-VACUOUS.**
- **ABA / freshness gate is sound as scoped.** Verified in `sparq-core`: `Dict::intern_lit` is find-or-push (append-only) and deletes never purge/remap ids, so a WKT literal id is monotonic for the dict's lifetime; `apply_delta` mutates the graph in place so `&graph.dict`'s address is stable across a delta and `literal_ids` stays authoritative (refcounted lockstep, proven against a fresh build in `tests/index.rs`). The genuine cross-graph ABA (freed dict address reused by a *different* graph) already violates the documented rebuild-on-change contract and already corrupts the pre-existing `id_of`-based candidate mapping; the pointer gate correctly DECLINES → per-row fallback on any mismatch. No overclaim: the PR frames this exactly as a freshness-gated fast path that falls back on doubt.
- **UPDATE interaction correct**; `sparq-core` untouched, no new cargo feature, no generation counter snuck in.
- **Gates I ran GREEN:** geo differential (16) + index (12) + OGC ratchets, both feature states; `geof_within == 1526`, `geo_compliance_pass == 25` (all 6 workloads match `expected.tsv`); engine lib tests exit 0; clippy `-p sparq-engine -p sparq-geo` clean; `RUSTDOCFLAGS=-D warnings cargo doc --all-features` clean (no intra-doc-link trap).

**BLOCKER (required gate red): `artifact-exact-equality (wasm bundle feature-OFF)` FAILED.** The always-compiled engine change (`SpatialProvider::indexed_ids` default method + the `apply_spatial_pushdown` fast branch — both in the default/feature-OFF build) grows the feature-OFF wasm bundle by **+1217 bytes (+0.077%)**, and mechanism V2 requires that intentional change to be DECLARED. The declaration file is absent.

**Fix (one file, mechanical):** add `bench/feature-off-declarations/1847.json`:

```json
{
  "pr": 1847,
  "bead": "sq-7jt80",
  "reason": "New always-compiled SpatialProvider::indexed_ids default trait method + the id-level fast-path branch in apply_spatial_pushdown ride the default-on spatial-pushdown seam (no new cargo feature), so they compile into the feature-OFF wasm build. Intentional engine change; result byte-identical to the per-row path. Size delta +1217 bytes (+0.077%), within the 2% bench.yml band (no wasm_bundle_bytes.floor raise needed).",
  "author": "[OPUS-4.8]"
}
```

Size delta is well under +/-2%, so no `metrics.wasm_bundle_bytes.floor` bump is needed. Once `1847.json` lands and the gate goes green, this is arm-ready.

jeswr:

Thanks for the escalated review — the `artifact-exact-equality (wasm bundle feature-OFF)` red was correct and now fixed.

The id-level fast path is an **intentional always-compiled** sparq-engine change (`SpatialProvider::indexed_ids` default method in `crates/sparq-engine/src/lib.rs` + the integer-lookup retain branch in `crates/sparq-engine/src/exec.rs`, both in the default/feature-OFF build), so the byte-for-byte feature-OFF wasm-bundle equality gate correctly observed the ~+1217-byte (~+0.077%) delta.

Declared per mechanism V2 in `a067a37c6`: added `bench/feature-off-declarations/1847.json` (`{pr, bead: sq-7jt80, date, reason, author: [OPUS-4.8]}`). The result stays byte-identical to the per-row path (answer-safety invariant held), and the ~+0.077% delta is well within the 2% `bench.yml` band, so no `metrics.wasm_bundle_bytes` floor raise is needed. No source or test change — declaration only. The set-difference gate now sees a new `<digits>.json` the base tree lacks and should go green on the next run.

PR #1870, opened 2026-07-10
perf(vectors): explicit-SIMD (NEON/AVX2) HNSW distance kernel — no new dep (sq-lfo84)

jeswr:

Closes the top open performance-dominance gap **sq-lfo84** (HNSW QPS/build behind hnswlib), and partially addresses the build-time sibling **sq-ose80**.

## Root cause (recap)
The `sq-z2z18` sweep (`research/gap-vector-2026-07.md` §3.6/§7) root-caused sparq-vectors' HNSW gap vs hnswlib to `instant-distance` 0.6.1 shipping **no SIMD distance kernel** (scalar inner loop) plus a largely serial graph build (>30min at 1M×128).

## Phase 1 — evaluation (NON-CANONICAL, aarch64 work box, no AVX2)

| Option | New dep | Build weight | SIMD | Per-query ef | API fit | Verdict |
|---|---|---|---|---|---|---|
| **(a) SIMD kernel for `instant-distance`** (NEON/AVX2 in the user `Point::distance`) | **none** | none (Cargo.lock unchanged) | yes | no (unchanged) | **exact** | **CHOSEN** |
| (b1) `usearch` 2.25.3 | `cxx` + a **C++ toolchain build** | HEAVY | yes | yes | wrapper needed | rejected — C++ build dep |
| (b2) `hnsw_rs` 0.3.4 | pure-Rust but **15+ transitive crates** | MEDIUM-HEAVY | yes | **yes** | `VectorIndex` rewrite | rejected here — fat tree + no matched-recall QPS win on this box |

Licences all clean (usearch Apache-2.0, hnsw_rs MIT/Apache-2.0, all transitive crates already on `deny.toml`'s permissive allowlist) — the discriminator was **build weight** + **measured matched-recall QPS**, not licence.

## Phase 2 — chosen fix
A shared `crates/sparq-vectors/src/simd.rs` (`approx-ann`-only, **no new dependency** — pure `core::arch` intrinsics behind `#[cfg]`) exposing `l2_sq_dist`: NEON (aarch64) / AVX2+FMA (x86_64) FMA-accumulated squared-Euclidean, with a scalar fallback **bit-identical** to the previous 8-lane loop. Wired into HNSW's `NPoint::distance`.

**Scoped to the HNSW path only** (floor-gated). The deterministic **EXACT-gated** exact/DiskANN/PQ paths keep the scalar reduction untouched, so `bench/vector/expected.tsv` (diskann=34, pq=22) stays byte-stable on the **x86_64** CI runner (a SIMD sum-order change could shift an integer deficit by ±1). Extending the kernel to those paths — after an x86_64 `expected.tsv` re-measure — is follow-up **sq-9wrkc**.

## Honest before/after (NON-CANONICAL, aarch64, 100k×128 clustered, build ef=256)

| Backend | Build (s) | recall@10 | QPS (µs/q) |
|---|---|---|---|
| instant-distance **scalar** (before) | 31.3 | 1.0000 | 2 873 (348 µs) |
| instant-distance **NEON** (this PR) | **13.2** | 1.0000 | **3 431** (291 µs) |

~**2.4× build**, ~**1.2× query**; measured recall `1.0000` on both kernels here. **Recall precision caveat (corrected after review):** the SIMD kernels use FMA (one rounding) vs the scalar `d*d`+`+=` (two roundings), so a SIMD squared distance is **not bit-identical** to the scalar one — it differs by ≤1 ULP, so rankings are stable only *up to exact near-ties*. That residual is exactly what the HNSW **floor gate** (recall@10 ≥ 0.95, `tests/recall.rs`) absorbs — the gate is a recall floor, not a bit-identity assertion. The clustered corpus saturates recall (1.0 both sides), so this table isolates the kernel-swap effect on build+throughput, not a recall trade. (The scalar *fallback* alone IS bit-identical to the previous loop, so a non-SIMD target is numerically unchanged.)

## Honest gap vs hnswlib — still BEHIND, narrowed
- **BEHIND-but-improved.** The scalar→SIMD swap closes part of the *per-distance* gap (one of the two root causes). No matched-recall QPS gap number is claimed here — the sparq side still lacks a per-query ef sweep (`sq-jo6ty`) so the 100k figure isn't directly comparable to the §3.5 SIFT1M hnswlib rows.
- **sq-ose80 only PARTIALLY addressed.** On a hard uniform 100k×128 corpus a single `instant-distance` build still ran **>14 min** on this box — the *serial graph-construction* cost (not the distance kernel) dominates at scale. Halving the kernel does not make a >30min 1M build practical; **sq-ose80 stays OPEN** (real fix = a parallel-build backend, decided on a canonical x86_64 box). See `research/gap-vector-ann-simd-2026-07.md`.

- `cargo clippy --workspace --all-targets -D warnings` — GREEN default (feature OFF).
- `cargo clippy -p sparq-vectors --all-targets --features approx-ann -D warnings` — GREEN (feature ON), and on the `x86_64-unknown-linux-gnu` target (compile-checks + clippy-cleans the AVX2 path this aarch64 box can't run).
- `cargo test -p sparq-vectors` (default) + `--features approx-ann` — GREEN both states (incl. the new `simd::` correctness tests vs an f64 reference for dim 0..=257, and `tests/recall.rs` monotone/recall invariant).
- `bench/vector/run.sh` — GREEN: hnsw deficit 3 (floor 50), diskann=34, pq=22 (exact — **unchanged**).
- `check-readme-template.py --enforce` — 0 deviations (README = 120 lines).
- `Cargo.lock` **unchanged** — zero new dependency; lean-core discipline intact.

**Load-bearing invariant preserved:** recall@10 vs the exact oracle stays above the floor gate (`tests/recall.rs`, ≥ 0.95; ≤1-ULP FMA wobble absorbed — not a bit-identity claim); the exact-oracle top-k path is untouched; the ANN-inside-SPARQL dict-encoded path (`vec_predicate_approx`) stays green.

## Discovered work
- **sq-9wrkc** (P2): extend the SIMD kernel to the exact/DiskANN/PQ `sq_dist` loops (blocked on x86_64 `expected.tsv` re-measure).
- **sq-ose80** stays OPEN (serial graph-construction, needs a parallel-build backend).

Do **not** arm — awaiting review.

### DIFF
--- a/bench/unsafe-snapshot.json
+++ b/bench/unsafe-snapshot.json
@@ -9,14 +9,14 @@
     "Re-seed (scripts/unsafe-gate.py --seed) ONLY alongside a reviewed register",
     "update \u2014 the diff to this file is the review hook. Smaller is always allowed."
   ],
-  "total": 72,
+  "total": 76,
   "crates": {
     "sparq-bench": 1,
     "sparq-cli": 2,
     "sparq-core": 50,
     "sparq-engine": 4,
     "sparq-py": 4,
-    "sparq-vectors": 9,
+    "sparq-vectors": 13,
     "sparq-zk-compose": 2
   }
 }
--- a/compliance/memsafety/unsafe-register.md
+++ b/compliance/memsafety/unsafe-register.md
@@ -118,7 +118,7 @@ Recurring invariant shorthands used below:
 | `src/compress.rs:2104` | `Mmap::map` | own-for-lifetime (TEST) | TEST-only (`#[test] corrupt_magic_is_loud_error_never_misdecode`, sq-7d3dj.32.2.7): read-only map of a temp perm file the test itself just created and owns; `File`/`Mmap` live for the map's whole scope and nothing else mutates it during the test, so the loud-error `from_mmap` read stays in-bounds over a stable region. [FABLE-5] |
 | `src/compress.rs:2120` | `Mmap::map` | own-for-lifetime (TEST) | TEST-only (`#[test] corrupt_magic_is_loud_error_never_misdecode`, sq-7d3dj.32.2.7): read-only map of a second temp perm file the same test created and owns; `File`/`Mmap` live for the map's whole scope and nothing else mutates it during the test, so the `from_mmap` read stays in-bounds over a stable region. [FABLE-5] |
 
-### `sparq-vectors` — 9 sites (aligned vector blobs + mmap'd `.spqv` / DiskANN)
+### `sparq-vectors` — 13 sites (aligned vector blobs + mmap'd `.spqv` / DiskANN + the SIMD ANN kernel)
 
 | File:line | Kind | Invariant relied on | Why sound / how bounded |
 |---|---|---|---|
@@ -131,6 +131,10 @@ Recurring invariant shorthands used below:
 | `src/diskann.rs:395` | slice reinterpret (read) | f32 no invalid patterns; align ok | f32→LE bytes; LE target asserted; borrows `b.vectors`. |
 | `src/diskann.rs:509` | `Mmap::map` | own-for-lifetime | read-only map of a DiskANN file; `open` validates after. **B5**. |
 | `src/diskann.rs:596` | slice reinterpret (read) | page-align; `start` a multiple of 4 ⇒ f32-aligned; range validated in `open` | `debug_assert_eq!` checks alignment; f32 accepts any bit pattern; borrows the map. **B5**. |
+| `src/simd.rs:40` (`approx-ann`) | `#[target_feature(enable="neon")]` call | the `neon` ISA extension is present at runtime | entered ONLY inside `if is_aarch64_feature_detected!("neon")` on the line above; `l2_sq_neon` reads exactly `a.len()==b.len()` lanes. [OPUS-4.8] sq-lfo84 |
+| `src/simd.rs:50` (`approx-ann`) | `#[target_feature(enable="avx2,fma")]` call | both `avx2` and `fma` are present at runtime | entered ONLY inside `if is_x86_feature_detected!("avx2") && …("fma")` on the line above; `l2_sq_avx2` reads exactly `a.len()` lanes via unaligned loads. [OPUS-4.8] sq-lfo84 |
+| `src/simd.rs:86` (`approx-ann`) | `unsafe fn l2_sq_neon` (NEON L2² kernel) | caller confirmed `neon`; `a.len()==b.len()` | 16-wide FMA body + 4-wide drain + scalar tail (`get_unchecked` only for `i<len`), so every `vld1q_f32` load is in-bounds. Verified vs an f64 reference for dim 0..=257 (`simd::tests`). [OPUS-4.8] sq-lfo84 |
+| `src/simd.rs:129` (`approx-ann`) | `unsafe fn l2_sq_avx2` (AVX2+FMA L2² kernel) | caller confirmed `avx2`+`fma`; `a.len()==b.len()` | 16-wide FMA body + 8-wide drain + scalar tail (`get_unchecked` only for `i<len`); `_mm256_loadu_ps` is unaligned so no alignment precondition. Numeric output verified by `simd::tests` on the x86_64 CI runner (this aarch64 box cannot execute AVX2). [OPUS-4.8] sq-lfo84 |
 
 ### `sparq-cli` — 2 sites (the `dump-perm` debug command)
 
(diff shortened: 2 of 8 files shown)


jeswr:

**Escalated review — unsafe SIMD + recall-preservation.** The unsafe is sound and the scope is correct; one honesty wording fix before arming.

**Unsafe: SOUND.** Tail handling is correct (NEON `+16`/`+4`/scalar-tail, AVX2 `+16`/`+8`/scalar-tail; every vector load only fires when a full lane fits within `n`, remainder is scalar `get_unchecked` under `i<n`; dim 0/1/width±1 all bounded, dim 0 skips all loops). The dim `0..=257` test is non-vacuous (independent f64 reference, `tol=1e-4·dim`; a dropped/OOB tail term would exceed it) and it executes the real dispatched kernel — including the AVX2 path on the x86_64 CI runner. Feature detection is correct (`is_aarch64_feature_detected!("neon")`, `is_x86_feature_detected!("avx2")&&("fma")` gate every `unsafe` entry, scalar fallback present) — no SIGILL/UB on a non-supporting CPU. Loads are unaligned (`vld1q_f32`/`_mm256_loadu_ps`) — no alignment precondition. The 4 new unsafe-register rows (9→13, snapshot 72→76) exactly match the 2 `#[target_feature]` call-sites + 2 `unsafe fn` kernels; justifications honest. Scope is correct: `simd::l2_sq_dist` is called from exactly one site (`NPoint::distance`, HNSW `approx-ann`); `diskann.rs`/`quant.rs` are untouched, so the EXACT-gated `expected.tsv` (diskann=34/pq=22) stays byte-stable. The scalar fallback is verified byte-identical to the previous `NPoint` loop.

**One fix — the FMA recall wording.** The SIMD kernels use **FMA** (`vfmaq_f32`/`_mm256_fmadd_ps`, one rounding); the scalar path uses `d*d` then `+=` (two roundings). So the SIMD squared distances are **NOT bit-identical** to the scalar path — they differ at the last ULP, and for exact near-ties a ranking *can* reorder. `SKILL.md` and the `simd_and_scalar_kernels_agree_closely` test already state this honestly ("differs only by f32 rounding, absorbed by the ±1-deficit floor gate" / "need not be bit-identical"). But the **PR-body §Honest before/after** line and **`research/gap-vector-ann-simd-2026-07.md` §3** say *"recall bit-for-bit preserved (rank order is identical; sqrt is monotone)"* — the "sqrt is monotone" justification only proves sqrt does not reorder *relative to the squared distance*; it does **not** establish that the FMA squared distance ranks identically to the old scalar squared distance. Please align both to SKILL.md’s honest framing, e.g.: *"recall preserved — measured 1.0000 both sides on the saturating clustered corpus; the FMA-vs-scalar squared distances differ by ≤1 ULP so rankings are stable up to exact near-ties, absorbed by the ±1-deficit floor gate (approximately-, not bit-identical)."* No code change needed; the metric is a floor gate (`recall.rs ≥ 0.95`), not a bit-identity gate.

Everything else — feature discipline (approx-ann-gated), API/monotone contract preserved, honest gap framing (still BEHIND hnswlib, sq-ose80 stays OPEN), no new dependency — is accurate. Not armed pending the wording tightening + green gate.

jeswr:

Thanks — the wording finding is correct and now fixed in `63916d9e6`.

**Agreed:** the SIMD kernels use FMA (one rounding) vs the scalar `d*d`+`+=` (two roundings), so a SIMD squared distance is **not bit-identical** to the scalar one (≤1 ULP), and `sqrt`-monotonicity only proves no reorder *relative to the squared value*, not cross-kernel rank identity. The "recall bit-for-bit preserved" phrasing overstated it.

**Corrected in all four spots** to the honest framing the load-bearing gate already implements — recall is **floor-preserved** (the HNSW gate is `recall@10 ≥ 0.95` in `tests/recall.rs`, not a bit-identity assertion; the ≤1-ULP FMA wobble is absorbed by that floor), measured `1.0000` on both kernels on the saturating clustered corpus:

- PR body "Honest before/after" table + the "Load-bearing invariant" line
- `research/gap-vector-ann-simd-2026-07.md` §3
- `crates/sparq-vectors/src/ann.rs` distance comment
- `crates/sparq-vectors/README.md`

The **scalar fallback IS bit-identical** to the previous auto-vectorised loop (a non-SIMD target is numerically unchanged) — that claim was already accurate and is retained. No code change; the recall floor gate is unchanged. `SKILL.md` also tightened for consistency (it was already honest). Gates re-run green locally (clippy approx-ann, rustdoc all-features, typos, readme-template 0 deviations). Ready for re-review once `gate` settles.

PR #1882, opened 2026-07-10
perf(sparq-core): native byte-level Turtle tokenizer/parser (opt-in native-ttl) [OPUS-4.8] (sq-jocpn)

jeswr:

Closes the single-thread ~2× oxttl Turtle parse gap (bead **sq-jocpn**, discovered from the sq-wrn61 profile: ~53% of the parse sits in oxttl's tokenizer/state-machine, which the #1846 pre-sizing couldn't touch — a native table-driven tokenizer is the structural fix).

## What

A hand-rolled **byte-level Turtle tokenizer/parser** (`crates/sparq-core/src/ttl.rs`) that interns S/P/O directly into the `Dict` — the Turtle analogue of the existing `nt.rs` N-Triples parser. **OPT-IN behind the `native-ttl` cargo feature (OFF by default)** so the incumbent oxttl path stays shipped and this can be A/B'd first; the default / wasm build never references it, so `sparq-core` stays lean.

Grammar coverage:
- Full Turtle 1.1: `@prefix`/`@base` + SPARQL `PREFIX`/`BASE`, `IRIREF` + base resolution, prefixed names (incl. `PN_LOCAL_ESC` `\.`/`\#`, percent-encoding), collections `()`, blank-node property lists `[]`, predicate-object lists `; ,`, `a`, numeric/boolean/triple-quoted literals (lexical form preserved, datatype by token shape), all ECHAR/UCHAR escapes.
- The **RDF 1.2** surface the workspace oxttl (`rdf-12`) enables by default: reifiers `<< … >>`, triple terms `<<( … )>>`, `~` and `{| … |}` annotations — desugared to `rdf:reifies` byte-identically.

## Load-bearing correctness claim: byte-identical to oxttl

- **IRI base resolution delegates to the same `oxiri` automaton oxttl uses** (no hand-rolled RFC-3987 re-derivation), so resolved terms are byte-exact by construction.
- Only **anonymous blank-node labels** differ (as they already do between two oxttl runs), so all comparisons are blank-node-**isomorphic**. Anon labels embed a process-global monotonic seed so the parallel chunk loader never wrongly unifies two distinct anon nodes.
- Fail-closed: anything the parser can't reproduce byte-identically is a parse `Err` — never a silent accept of invalid Turtle, never a silent divergence.

### Evidence

**Conformance-set-identical** — the W3C rdf-turtle suite run through the native parser (`sparq-conformance`'s `native-ttl-suite`, forwarding `sparq-core/native-ttl`) produces an **identical pass/fail set** to oxttl:

| parser | run | pass | fail |
|---|---|---|---|
| oxttl (baseline, feature OFF) | 313 | 313 | 0 |
| **native** (feature ON) | 313 | **313** | **0** |

The full per-test outcome-line diff between the two runs is **empty** (identical set). Four negative-syntax cases initially diverged and were fixed to exact parity: `@BASE` case-sensitivity (`@`-directives are lowercase-only), a prefix name ending in `.`, and a bare non-`[]` subject followed by `.` (the two N3-extras path cases).

**Native-vs-oxttl differential** — `crates/sparq-core/src/ttl/tests.rs`: 26 tests parsing many constructs with BOTH paths and asserting isomorphic triple sets, plus malformed-input parity (a fixture where BOTH must `Err`), plus a non-vacuity guard on the isomorphism check and an anon-label-uniqueness test. The existing `parallel_turtle_*_match_serial` + `presized_..._matches_oxttl_direct` differentials also stay green with the feature on.

## Honest perf (work box, NON-CANONICAL)

Single-thread `parse_to_triples` on an 18 MB / 800k-triple synthetic Turtle corpus, best-of-N:

| path | best | MB/s | Mt/s |
|---|---|---|---|
| oxttl incumbent (1T) | ~1.37 s | ~13 | ~0.59 |
| **native (1T)** | **~0.56 s** | **~32** | **~1.42** |

≈ **2–2.4× faster** single-thread at best-of-N, matching the sq-wrn61 attribution (the native tokenizer removes oxttl's tokenizer/state-machine). The box was heavily loaded during measurement (oxttl ranged 1.37–4.0 s, native 0.56–1.46 s run-to-run) — **these numbers are non-canonical work-box timings, not for citation**; the direction is robust but the magnitude is noisy.

- `cargo clippy --all-targets -- -D warnings` — default (OFF): ✅ ; `-p sparq-core --features native-ttl` and `-p sparq-conformance --features native-ttl-suite`: ✅
- `cargo test -p sparq-core` (OFF) ✅ and `--features native-ttl` (ON) ✅ ; `sparq-conformance` ratchet both states ✅
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` ✅ (no feature-gated intra-doc-link breakage — code spans used)
- `python3 scripts/check-readme-template.py --enforce` → 0 deviations (README 75 lines, ≤120 cap) ✅
- Coverage floor unaffected — `native-ttl` is not in the `sparq-core` coverage feature set (`mmap,dict-spill`), so the feature-gated module is compiled out of the coverage build.

**Feature flags:** `sparq-core/native-ttl` (impl, OFF by default), `sparq-conformance/native-ttl-suite` (ratchet harness, forwards to `sparq-core/native-ttl`).

Not armed for merge.


jeswr:

**ESCALATED review — verdict: ARM.** Diff-scoped adversarial review of the byte-identity claims for the opt-in `native-ttl` Turtle parser. All five load-bearing claims verified against the diff + the differential test + the conformance harness wiring.

1. **oxiri base/IRI resolution — delegated, byte-exact by construction.** `ttl.rs` calls `oxiri::Iri::parse` + `base.resolve(raw)` for every IRIREF and every `@base`/`BASE` (incl. relative-base-against-prior-base) — not a hand-rolled RFC-3987. Covered by `base_and_relative_iri_resolution`, `iriref_with_uchar_escape`, `pn_local_escapes_and_percent_encoding`, and the no-base rejection parity (`relative_iri_with_no_base_rejected`).
2. **RDF 1.2 desugaring — differentially tested vs oxttl.** `rdf12_reifiers_triple_terms_and_annotations` runs `<<…>>`, `<<(…)>>`, `~`, `~id`, `{|…|}`, chained/nested/reifier-in-subject through `assert_same` (bnode-isomorphic vs oxttl). Desugars to `rdf:reifies <<(s p o)>>` structurally. A divergence here would fail the differential.
3. **Parallel anon-seed — sound.** `fresh_blank` embeds a **process-global `AtomicU64` (`fetch_add`)** in every anon label (`__ttl_anon_{g}_{n}`); the global `g` alone makes every mint unique across chunks/threads, so the label-based merge can never wrongly unify two distinct anon nodes. A single logical anon node is minted once and threaded (never re-minted → in-document unification holds); labelled `_:x` intern verbatim, document-scoped.
4. **Conformance non-vacuity — STRONGER than a pass-count.** The ratchet's in-test `assert_eq!(fail, 0)` under `native-ttl-suite` routes `turtle_suite::run_suite`, whose `eval_test` asserts the native-**resolved** graph is **isomorphic to the W3C expected `.nt`** (the RFC-3987 ground truth), plus accept/reject on every Positive/Negative case; `out.len() > 100` guards vacuity, and `iso_is_not_vacuous` + `assert_both_reject` guard the differential oracle. Not a weak count.
5. **Feature discipline — clean.** `native-ttl` OFF by default, `mod ttl` fully `#[cfg]`-gated (`#![cfg(...)]`), `oxiri` optional `dep:`, and the lib.rs OFF-arm is byte-identical to the prior oxttl path — no always-compiled token drift, so no feature-off-declarations file is needed. rustdoc `--all-features` is clean (the module-gated doc-comment's `[oxiri]`/`[Dict]` links resolve only when the feature — and their targets — are compiled).

**Byte-identity-when-ON judgment: SOUND.** Because resolution is delegated to oxiri and the conformance ratchet pins the native path directly to the W3C expected graphs (not just to oxttl), byte-identity of ground terms holds by construction and is gated non-vacuously. Opt-in + OFF-by-default, so this is safe to arm.

**One residual (follow-up bead, non-blocking):** the module doc says anon labels "cannot collide with a user-written `_:label`" — strictly a user could write `_:__ttl_anon_<g>_<n>`, but `g` is a process-global atomic the user cannot observe, so it is non-deterministic, non-exploitable, and off the byte-identity (ground-term) axis. Recommend a follow-up bead to either namespace anon labels out of the legal `PN_CHARS` space or soften the doc claim.

Arming.

jeswr:
> 🤖 **SPARQ agent** — unblock report

The **`artifact-exact-equality (wasm bundle feature-OFF)`** gate was failing on an *undeclared* feature-OFF wasm content change. I reproduced the gate locally (`cargo build --profile release-wasm -p sparq-wasm --target wasm32-unknown-unknown`, no `--features native-ttl`) against the base tree and verified this is the known **cfg-gated feature-OFF content-drift** pattern, **not** a feature leak:

- **Size delta = `+0` bytes (`+0.000%`)** — base `1580701` vs head `1580701` (well within the ±2% band; no `perf-baseline.json` floor bump).
- **CODE section byte-for-byte identical** — `wasm-tools objdump` shows the same offset (`0x2607..0x16ac81`), same `1476218` bytes, same `2495` function count on both bundles → zero executable-behaviour change, **no `native-ttl` code compiled into the default build**.
- **DATA string set identical** — `11496` strings on both, `0` added / `0` removed. No `native-ttl` / `native_ttl` / `ttl::parse` / `src/ttl.rs` symbol in either bundle. (`oxiri` appears once in **both** — that's the pre-existing `oxttl → oxiri` transitive dep, not this feature's optional `dep:oxiri`.)
- **Only 9 bytes differ** across the whole 1.58 MB bundle, all inside the DATA section, all a **uniform `+26` (`0x1A`) shift** of `&str` fat-pointer addresses — the classic emission-order nudge from inserting the compiled-out `#[cfg(feature="native-ttl")] mod ttl;` + gated call-site tokens into the always-compiled `crates/sparq-core/src/lib.rs`. Identical content, shifted location.

**Fix:** added `bench/feature-off-declarations/1882.json` (mechanism V2) with the measured evidence above. The re-run confirms it:

```
[leg2] feature-OFF wasm DIFFERS from the base tree: base=1579941 head=1579941 bytes, size delta=+0 (+0.000%).
[leg2] OK — change DECLARED (mechanism V2): the head tree added bench/feature-off-declarations/['1882.json'] not present in the base tree.
```

**Gates re-run green on the merged tip:** clippy `-D warnings` (feature-OFF **and** `native-ttl`), `sparq-core` lib tests (119 OFF / 145 ON, 0 failed), `rustdoc --workspace --no-deps --all-features` with `RUSTDOCFLAGS=-D warnings`, and the feature-OFF wasm gate. Auto-merge (armed by @jeswr) is left intact — no re-arm needed, no new PR.

_[OPUS-4.8]_

PR #1901, opened 2026-07-10
fix(engine): builtin memoize + REGEX/REPLACE memo + RAND PRNG + NOW() query-constancy — O(n²) Turtle-parse hang fix (sq-98w7z.1)

jeswr:
> 🤖 SPARQ agent — finishing bead **sq-98w7z.1** (builtin memoize + REGEX/REPLACE compile memo + RAND thread-local PRNG + NOW() query-constancy). The core implementation was authored by a prior Fable agent (commit `85a5465`, `[FABLE-5]`); its validation run **hung for 19+ minutes** and could not be resumed. This PR diagnoses and fixes the hang, closes a feature-OFF gap the memoize work introduced, and takes the bead to green in both feature states. All new/fix commits are `[OPUS-4.8]`.

## The 19-minute hang — root cause (diagnosed, not guessed)

The hang was **NOT** in the memoize / NOW / regex code. It was a **pre-existing latent O(n²)** in the parallel Turtle parser (`sparq-core::next_terminator`), and the new NOW-constancy test was simply the first test to feed it the trigger shape.

Diagnosis path: I ran each `builtin_memo` test individually; only `now_parallel_rows_single_instant_no_fallback` hung. A live thread-stack sample (`/proc/<tid>/wchan` + `gdb`) showed **all rayon workers asleep** and the one active thread NOT in the query at all — it was in `parse_turtle_parallel → turtle_chunks → next_terminator → memchr3`, with the scan pointer advancing only ~1.9 KB every 3 s.

`next_terminator` ran **two full-tail** `memchr3` scans per state-change byte (`. < #` and `" ' \`) and took `x.min(y)`. The test's `wide_graph(50_200)` builds a **quote-free, backslash-free** all-IRI Turtle body, so the second scan (`" ' \`) found nothing and **re-walked the entire remaining buffer once per statement** → O(statement · tail) ⇒ whole-parse O(n²). Measured on the work box (non-canonical, EC2):

| rows | before | after |
|-----:|-------:|------:|
| 10 000 | 150 s | 0.086 s |
| 20 000 | ~600 s (proj., ~4×/doubling) | 0.178 s |
| 55 000 | ~hours (hang) | 0.42 s |
| 200 000 | — | 1.8 s |

**Fix** (`crates/sparq-core/src/lib.rs`): bound the second `memchr3` to the window *before* the first `. < #`. A quote/backslash at or after that offset can never win the `min`, so the returned offset is **identical** to the old `a.min(b)` — proved by the existing `next_terminator_memchr_matches_scalar` differential oracle (every start offset, all tricky Turtle shapes) plus two new quote-free cases. Each byte is now scanned at most once → O(n). The full `builtin_memo` suite now runs in **1.22 s**; the full `sparq-engine` suite in ~3 s at default parallelism.

Classification per the brief: this is category **(a) — a real correctness/liveness bug** (a query engine must not pathologically stall on ordinary N-Triples-shaped input), fixed properly by bounding the critical scan, not by shrinking the test.

Regression guard: `parallel_turtle_terminator_scan_is_linear_not_quadratic` — a hard 20 s wall-clock ceiling on a 20k-row quote-free parse. **Non-vacuous**: pre-fix ≈600 s (>30× over the ceiling), fixed <1 s (~100× under it), so it can neither flake on a slow-but-linear runner nor pass on a quadratic one.

## Feature-OFF gap the memoize work introduced (also fixed)

`--no-default-features` clippy caught two breaks the prior commit left: the `now_parallel_*` test referenced `parallel`-gated `PAR_THRESHOLD`, and the `names_graph` test helper had only `regex`-gated callers (dead code). Both are now correctly `#[cfg]`-gated. This would have failed the CI feature-OFF leg.

## What the bead delivers (memoisation allow / deny list)

Unchanged from the prior implementation, re-verified here:

- **ALLOW (memoised):** `REGEX()`/`REPLACE()` compiled `(pattern, flags)` via a per-thread cache (cap 64, failure-memoised) — a constant-pattern FILTER over N rows compiles **once**, not per row (`regex_compiles_once_not_per_row`, non-vacuous).
- **PINNED (spec-mandated constancy):** `NOW()` returns the execution-scoped instant (SPARQL 1.1 §17.4.5.1) — one instant across **all rows and across EXISTS re-entry** within one execution, **fresh across separate executions**. Query-scoped via an outermost-wins `query_now::scope()` guard at the single `eval_modified` choke point + a `worker_install` snapshot on every rayon expression-eval site — a thread-local `Cell<Option<i64>>` (lock-free, O(1)) that is drained at outermost-drop, so it never leaks across queries. (`now_*` tests.)
- **DENY (never memoised):** `RAND()` (per-thread splitmix64, non-crypto), `UUID()`, `STRUUID()`, no-arg `BNODE()`, and every registered custom function — fresh per call (`nondeterministic_builtins_fresh_per_call_never_memoised`, non-vacuous: 50 distinct values required).

- `cargo clippy --workspace --all-targets -- -D warnings` — clean (default); `clippy -p sparq-core -p sparq-engine --no-default-features -- -D warnings` — clean.
- `cargo test -p sparq-engine --lib` — 297 pass (default) / 285 (no-default); `cargo test -p sparq-core --lib` — 120 (default) / 146 (`native-ttl`).
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — clean (no intra-doc links to `#[cfg]`-gated items).
- Per-crate line coverage: sparq-core **90.94%**, sparq-engine **91.04%** — both above the 90 floor.
- Merged `origin/main` cleanly (branch was 5 commits behind, incl. the new opt-in `native-ttl` parser — orthogonal to this fix).

## Honest perf note

The measured before/after numbers above are from this EC2 work box and are **non-canonical** — they establish the coarse **O(n²) → O(n)** correction, not a headline benchmark. The regex compile-once memo is a real per-row saving but no canonical figure is claimed here; work-box query timings for the memo (`memo_bench_tmp.rs`, since deleted — nothing scratch shipped) were within box noise and are not reported.

## Not armed

This is an engine-correctness surface; leaving unarmed for an Opus review. (If a pipeline arms it later I will not disarm.)

## Feature-OFF wasm declaration (follow-up in-thread)

The always-compiled `next_terminator` change moves feature-OFF wasm bytes. I will measure the real delta (reproduce the CI feature-OFF wasm build + `wasm-tools objdump` size + symbol leak-check) and add `bench/feature-off-declarations/<this-PR>.json` once the PR number exists.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


jeswr:
> 🤖 SPARQ agent — feature-OFF wasm delta measured (the declaration is now on the branch, `bench/feature-off-declarations/1901.json`).

Reproduced the `vectorized-feature-off` leg2 on the work box: built the feature-OFF `sparq-wasm` bundle (`--profile release-wasm`, default features, no `--features vectorized`) from the BASE tree (`6ec1a2da3`) and from HEAD, then compared with `wasm-tools objdump`.

- **Size delta: +0 bytes** (1 580 701 → 1 580 701, **+0.0000%**, well inside the ±2% band).
- **CODE section: byte-for-byte identical** — same `0x2607..0x16ac81` span, same 1 476 218 bytes, same 2 495 function count. Executable behaviour unchanged; no parser/memoize code leaked.
- **DATA section: same size** (69 459 B), **identical string set** (0 strings added, 0 removed).
- Only **279 bytes differ**, all in the data section, all a uniform `&str` fat-pointer emission-order shift from the tighter scan-loop codegen — identical content, slightly shifted addresses.
- **Leak-check clean**: none of `next_terminator`/`regex_cache`/`query_now`/`rand_unit`/`turtle_chunks` appears in either bundle (release-wasm strips symbols).
- The NOW/RAND/REGEX-memo builtin work is native-only (`#[cfg(not(target_arch = "wasm32"))]` / behind the native `regex`+`parallel` features), so it contributes nothing to the wasm bundle.

I ran the real `check-vectorized-feature-off.py --leg2-dynamic` against the two bundles with the declaration present — it detects the change, matches the added `1901.json`, and exits 0. No `bench/perf-baseline.json` change accompanies this (0-byte delta).

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

PR #1904, opened 2026-07-10
test(engine): property-based eval tests — engine ≡ naive reference, join-order invariance, ORDER BY laws (sq-3dyje.3)

jeswr:
> 🤖 SPARQ agent — property-based testing bead sq-3dyje.3 [FABLE-5]

Test-only PR: `crates/sparq-engine/tests/proptest_eval.rs` (new) + a DEV-only `proptest` dev-dependency in `crates/sparq-engine/Cargo.toml`. **Engine `src/` is byte-identical to `main`.**

## The three property families

1. **`eval ≡ naive reference`** (`family1_engine_matches_naive_reference`, 96 cases): random small graph × random query (BGP / OPTIONAL / UNION / FILTER / BIND / DISTINCT / subset projection / ORDER BY / LIMIT) — the engine's solution multiset must equal an in-test naive nested-loop evaluator's, per the comparison model in `research/differential-testing-value-level.md` §2 (multiset equality; ORDER BY additionally checked for sortedness; LIMIT without a total order checked as sub-multiset + cardinality per §2.2).
2. **Join-order invariance** (`family2_join_order_invariance`, 64 cases): permuting BGP triple-pattern order never changes the result multiset — checked engine-vs-engine AND tied back to the oracle.
3. **ORDER BY laws** (`family3_orderby_exact_sequence_laws`, 64 cases + the family-1 sortedness check): exact ASC / DESC-is-exact-reverse / LIMIT-is-exact-prefix sequences on tie-free integer keys; on arbitrary mixed keys, sortedness against the spec-defined partial order (unbound < blank < IRI < literal; IRIs/strings by codepoint; numerics by value across int/decimal/double; booleans) with spec-open pairs (cross-literal-class, langString, bnode labels) deliberately unconstrained.

## Plan paths (the bead's cross-path requirement)

Default build = greedy GOO + scalar exec. `--features dp-planner,vectorized` = DPccp plans by default + vectorized kernels, and every property then ALSO cross-checks `without_dp_planner` (greedy) against the same oracle per case — one run validates both plan paths.

## Oracle independence

The reference evaluator shares no code with the engine: test-side term model + query AST, substitution-based nested loops over `Vec<[T;3]>`, comparison via rendered N-Triples strings. The engine is driven end-to-end (`sparq_engine::query` on rendered SPARQL text) — parser, planner, executor all inside the tested boundary. Engine-semantics corners the oracle encodes (FILTER `=`/`<` error matrix, Kleene 3-valued logic, BIND error→unbound, decimal scale rules, ORDER BY class ranks) were pinned by probing the real engine first; each rule is documented in the file header.

## Non-vacuity evidence

- **Three engine-side mutations each turn the suite red with a shrunk counterexample** (injected temporarily into `src/exec.rs`, then reverted; details + shrunk cases in the file header):
  - filter-drop (`apply_filter_scalar` keep-all) → family 1 red (`FILTER(!isIRI(?v0))` case, 1-pattern query);
  - join-dup (`bind_join` double `bind_combine`) → families 1 AND 2 red (every joined row duplicated);
  - DISTINCT no-op (`distinct_bindings` emptied) → family 1 red on a `SELECT DISTINCT` query.
- **Oracle known-answer pins**: 3 hand-computed fixed-case tests (OPTIONAL join, UNION+DISTINCT+FILTER bag semantics, all BIND forms) assert oracle == expected AND engine == expected — a degenerate oracle cannot silently agree.
- **Generator diversity floor** (`generator_diversity_floor`, fixed-seed, deterministic): 300-case corpus sample must contain ≥25% nonempty results, multi-row results, duplicate rows, unbound cells, and every query shape. This test CAUGHT a real vacuity problem during development (only 46/300 cases nonempty before the query-pattern-instantiation seeding was added; now 157/300 nonempty, all shapes ≥49).

## Engine divergences found (probing) — beaded, not fudged

- **sq-qeltv (P1 bug)**: `isIRI(?unbound)` evaluates to `false` (and `STR(bnode)` returns an internal label) where SPARQL 1.1 §17.2 mandates a type error. The oracle is NOT bent to match: the generator excludes the case (is* only over certainly-bound vars; bnodes masked from STR-BIND graphs), each narrowing carries the bead reference.
- **sq-ilweo (P2)**: generator widening follow-ups — computed-double rendering (needs an independent float formatter to avoid the §2.3 independence trap), NaN ordering, langString comparison matrix, OFFSET.

## Gates (all run in-worktree)

| gate | default features | `--features dp-planner,vectorized` |
|---|---|---|
| `cargo test -p sparq-engine` | GREEN (exit 0) | GREEN (exit 0, 22 binaries) |
| `cargo clippy --all-targets -- -D warnings` | GREEN (whole workspace) | GREEN (`-p sparq-engine`) |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` | GREEN | — |
| typos | clean | — |

Suite runtime ~1s (96+64+64 cases), well inside the fast-suite rule. No README/SKILL change: no public API or feature surface added (`proptest` is dev-only, matching the sq-3dyje.1/.2 pattern; same version req as the already-attested sparq-core/sparq-substrate dev-dep — no cargo-vet change). Coverage floor unaffected (test-only addition).

**Do not merge before the verify pipeline arms this PR** (agent will not self-arm per brief).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3901,6 +3901,7 @@ dependencies = [
  "md-5 0.11.0",
  "oxiri",
  "oxrdf 0.3.3",
+ "proptest",
  "rayon",
  "regex",
  "rustc-hash 2.1.3",
--- a/crates/sparq-engine/Cargo.toml
+++ b/crates/sparq-engine/Cargo.toml
@@ -440,6 +440,12 @@ spargebra.workspace = true
 # The cs-planner q-error gate builds its characteristic-set tables through the
 # dict-id accessor (dev-only; the library never depends on introspection).
 sparq-introspect = { path = "../sparq-introspect", version = "0.1.0" }
+# [FABLE-5] sq-3dyje.3: property-based eval tests (tests/proptest_eval.rs — engine ≡
+# naive reference / join-order invariance / ORDER BY laws). DEV/TEST ONLY, exactly like
+# the sparq-core + sparq-substrate proptest dev-deps: it never appears in [dependencies]
+# or any feature, so the shipped crate and the wasm bundle are unchanged. Same version
+# req as those crates (resolves to the already-attested proptest in supply-chain/).
+proptest = "1"
 # [OPUS-4.8] (sq-e3pj, sq-6vshe.4) DEV-only serde_json: the `explain_json` structured-EXPLAIN
 # tests parse the hand-emitted JSON DOM back to assert every field round-trips (and `chunk`'s
 # `impl PartialEq<serde_json::Value>` test helper), so the opt-in `explain-json` feature itself
--- a/crates/sparq-engine/tests/proptest_eval.rs
+++ b/crates/sparq-engine/tests/proptest_eval.rs
@@ -94,6 +94,7 @@
 //!   red (every joined row duplicated vs the oracle multiset).
 //! * DISTINCT no-op (`distinct_bindings` emptied) → family 1 red on a
 //!   `SELECT DISTINCT` query whose oracle multiset deduped.
+//!
 //! Additionally `oracle_known_answer_*` pin the oracle itself against hand-computed
 //! rows (a degenerate oracle cannot silently agree with the engine), and
 //! `generator_diversity_floor` asserts the generated corpus actually contains
@@ -1508,13 +1509,13 @@ fn generator_diversity_floor() {
         Config::default(),
         TestRng::from_seed(RngAlgorithm::ChaCha, &[7u8; 32]),
     );
-    let strat = arb_case();
+    let strategy = arb_case();
     let n = 300;
     let (mut nonempty, mut dups, mut unbound_cells, mut multi_row) = (0, 0, 0, 0);
     let (mut bgp, mut opt, mut union, mut filt, mut bind, mut distinct, mut order, mut limit) =
         (0, 0, 0, 0, 0, 0, 0, 0);
     for _ in 0..n {
-        let (data, q) = strat.new_tree(&mut runner).expect("gen").current();
+        let (data, q) = strategy.new_tree(&mut runner).expect("gen").current();
         match q.body {
             Body::Bgp(_) => bgp += 1,
             Body::Opt(..) => opt += 1,
--- a/crates/sparq-engine/tests/proptest_eval.rs
+++ b/crates/sparq-engine/tests/proptest_eval.rs
@@ -1,0 +1,1602 @@
+//! [FABLE-5] sq-3dyje.3 — Property-based tests for sparq-engine evaluation:
+//! engine ≡ independent naive reference on random graphs × random queries,
+//! join-order permutation invariance, and ORDER BY laws over the real `Value` path.
+//!
+//! Design record: research/testing-strategy-assessment-2026-07.md §7.1 (sparq-engine row);
+//! comparison model: research/differential-testing-value-level.md §2–§3.
+//!
+//! # The three property families
+//!
+//! 1. **`eval ≡ naive reference` (`family1_*`).** A random small graph and a random
+//!    query (BGP / OPTIONAL / UNION / FILTER / BIND / DISTINCT / subset projection /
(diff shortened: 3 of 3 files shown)


jeswr:
> 🤖 SPARQ agent — unblocking the red `test (load-aware shard bulk 2/3)` → `gate` [FABLE-5]

**Root cause (b): a generator↔oracle contract gap in the new proptest, NOT an engine bug.** The failing test was `family1_engine_matches_naive_reference`; the CI log's first (and only) failure in the shard was a panic in the test's own naive oracle:

```
thread 'family1_engine_matches_naive_reference' panicked at tests/proptest_eval.rs:568:27:
generator mask violated: PlusK over a double
   at proptest_eval::eval_bind
```

The engine was never even compared on that case — the **oracle** panicked while computing the expected value, so this is a test-generator defect, not an `engine ≢ oracle` divergence. Reproduced deterministically in-worktree (`PROPTEST_CASES=4000` → ~2500 successes then the panic).

**Why:** the `no_doubles` `DataMask` keeps xsd:doubles out of the random graph and out of `instantiate`'s random var-assignment, but a `Pos::Const(T::Dbl)` in a query **pattern object** (`arb_pos_o` emits const literals unmasked) is copied verbatim into the data by `instantiate`. A `?v` in another pattern on the same `(s,p)` then matches that constant double and binds the `PlusK` var to a double → the oracle's documented `PlusK over a double` guard fires.

**Fix** (the query-const complement of the *existing* documented no-doubles narrowing, bead `sq-ilweo`): when the bind is `PlusK`, `strip_const_doubles` rewrites every constant xsd:double in the body's pattern positions to `T::Int(0)` before rendering. No property is weakened, the oracle is not fudged to match the engine, and no engine bug is hidden — it just closes the one path by which a double could reach the PlusK var, keeping the oracle independent of the engine's float formatter.

**Gates re-run green in-worktree, both feature states:**

| gate | default | `--features dp-planner,vectorized` |
|---|---|---|
| `cargo test -p sparq-engine --test proptest_eval` | ok, 10 passed (1.8s) | ok, 10 passed (1.7s) |
| stress `PROPTEST_CASES=8000` (both plan paths per case) | ok, 10 passed (178s) | — |
| `cargo clippy --all-targets -D warnings` | clean (whole workspace) | clean (`-p sparq-engine`) |

Suite stays fast at the committed case counts (96/64/64) — ~1.8s. Engine `src/` remains byte-identical to `main` (test-only change; the rustdoc all-features gate is not implicated — no `src/` doc-comments touched). Commit `de7a69323`. **Not armed** — leaving re-verify + re-arm to the pipeline.

jeswr:
> [OPUS-4.8] 🤖 SPARQ agent — mechanical-verify pass complete

All six objective checks PASSED; gate concluded SUCCESS (15m12s).

**Checks passed:**
1. **gates_both_feature_states** — ci-summary SUCCESS in both feature states
2. **tests_non_vacuous** — the three property families (eval≡reference, join-order-invariance, ORDER BY laws) remain unweakened; oracle eval_bind unchanged
3. **opt_in_feature_discipline** — test-only change (dev-dependency proptest, no engine Cargo.toml change)
4. **readme_skill_synced** — N/A (no public-API change)
5. **no_hardcoded_perf_md** — no baked performance numbers
6. **coverage_floor_not_regressed** — test-only addition, no public surface change

**Fix verdict:** Generator↔oracle contract gap (NOT an engine bug). Root cause: const-double pattern objects from `arb_pos_o` injected unmasked by `instantiate`, reachable by PlusK var, panicking oracle guard. Solution: `strip_const_doubles` narrows generator to rewrite pattern-const doubles to Int(0) when PlusK bind present. Narrowing is targeted (only PlusK), not over-broad (doubles still generated in LtK/StrLenOf/non-bind contexts). Engine src/ byte-identical to prior commit.

**Arming:** `gh pr merge 1904 --auto` (queued for automatic merge on branch updates if any).

No escalation required — this is pure mechanical test refinement.

PR #1927, opened 2026-07-11
fix(engine): order_bindings parallel worker reinstall for functions/view/spatial (sq-6aefu)

jeswr:
> **SPARQ agent 🤖** — fixing bead sq-6aefu

## Bug

In `crates/sparq-engine/src/exec.rs`, `order_bindings`, the two `PAR_THRESHOLD` (50k) rayon key-build sites — the **top-k index-carry** build and the **full stable sort** build — called `key_of → cell_of → eval_compiled` on rayon worker threads **without** the `functions / view / spatial` snapshot+`worker_install` that the FILTER, BIND, and aggregate parallel branches perform (sq-98w7z.1 added `query_now` at these same sites but left the other three missing).

Consequences for result sets ≥ 50k rows:

1. **EXISTS in ORDER BY, named-graph leak (security-adjacent)**: workers saw NO `DatasetView` → `view::allows()` returned `true` for every named graph → a hidden/restricted graph was accessible inside `EXISTS { GRAPH <hidden> { … } }` on workers, silently corrupting the sort order.
2. **Custom extension function in ORDER BY, spurious error**: workers had no `FunctionRegistry` → `functions::lookup()` returned `None` → `eval_function` returned `Err("unsupported SPARQL function")` → the entire ORDER BY call failed with a hard error.

## Fix

Mirror the existing FILTER/BIND pattern at both PAR_THRESHOLD sites:

```rust
// before par_iter:
let fns = functions::snapshot();
let vw  = view::snapshot();
let spx = spatial::snapshot();

// inside .map():
let _fns = functions::worker_install(&fns);
let _vw  = view::worker_install(&vw);
let _spx = spatial::worker_install(&spx);
```

Both the top-k index-carry path and the full stable sort path get the same treatment. `query_now` wiring from sq-98w7z.1 is preserved unchanged.

## Regression tests (non-vacuous)

Two new tests in `exec::order_bindings_worker_reinstall` (gated on `#[cfg(all(test, feature = "parallel"))]`):

| Test | Scenario | Failure without fix |
|------|----------|---------------------|
| `order_by_exists_restricted_view_parallel_matches_lex_sort` | ORDER BY `IF(EXISTS { GRAPH <hidden> { … } }, "constant", STR(?s))` under a `DatasetView` that hides `<hidden>`, ≥ PAR_THRESHOLD rows | Workers see the hidden graph → EXISTS = true → all sort keys tied at "constant" → stable/insertion order ≠ lex order → assertion fires |
| `order_by_custom_fn_parallel_no_spurious_registry_error` | ORDER BY `<http://test/noop>(?s)` (custom function) over ≥ PAR_THRESHOLD rows | Workers have no registry → Err("unsupported") → `.expect()` panics |

**Non-vacuity confirmed**: both tests were run **RED** on the unpatched code (worker_install lines temporarily removed), then **GREEN** after the fix was restored. The difference is unambiguous for N=50,100.

## Gates

- `cargo clippy --workspace --all-targets -- -D warnings` ✓ (default feature set)
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` ✓  
- `cargo test -p sparq-engine` ✓ 299 passed + 2 new
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` ✓
- `readme-template`: 0 deviations across 52 READMEs ✓

## Byte-identical invariant

Sort keys evaluated on workers are now byte-identical to sequential evaluation: the same `DatasetView`, `FunctionRegistry`, and spatial provider are visible on every rayon worker item, so `cell_of → eval_compiled` produces the same values it would on the main thread.


jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (949440758)

PR #1930, opened 2026-07-11
test(reason-dl): W3C DIRECT-arm render round-trip suite + fix the named-composite equivalence emission it caught (sq-pbz04.4.17)

jeswr:
> 🤖 SPARQ agent (Claude Fable 5) — bead sq-pbz04.4.17, epic sq-pbz04.4.

## What

Extends the `sparq-reason-dl` render round-trip coverage from the 13 hand-written ALCH fragments to **every ontology document of the W3C OWL WG export's DIRECT arm** (the same sq-pbz04.4.5 L5 case selection), behind the existing opt-in `dl-direct` feature of `sparq-conformance` — the belt-and-suspenders completeness check the bead asked for. No new cargo feature, no new dependency; the default build is untouched.

- **`inference::dl_suite::run_render_roundtrip_arm` + `RenderRoundTripReport`** — per-document `extract → render_to_triples → re-extract` with structural-model equality, closed four-bucket accounting (round-tripped / extraction-refused / parse-failed / violations). The export parsing + DIRECT-arm case selection are extracted **verbatim** into shared `parse_export` / `collect_cases` helpers so this arm can never drift from `run_direct_arm` — whose exact pins are **byte-identical and still green** (95 profile / 182 direct / all abstention pins).
- **`tests/dl_suite.rs`** — EXACT-pinned counts at the pinned export snapshot: **672 documents = 293 round-tripped + 378 extraction-refused + 1 parse-failed + 0 violations**, a HARD empty-violations assertion (a mis-render is a REAL fidelity bug, never a pinnable divergence), plus a corpus-free derivation canary with known counts so the arm logic runs even on an offline checkout. Runs inside the existing `--test dl_suite` CI lane — no workflow change needed.

## What the arm caught (and this PR fixes)

The first corpus run found **15 real mis-renders** in the sq-pbz04.4.7 renderer (#1854). `EquivalentClasses(Class(A), composite)` was emitted as an **inline backbone on `A`** — but the extractor gives that shape different semantics than the explicit shape:

1. an inline backbone makes the extractor **inline the composite at every use site of `A`**, so models holding atomic `Class(A)` references re-extract differently — 14 corpus mismatches (`WebOnt-description-logic-001/-101/-102/-103/-110/-201…-209`);
2. a **self-referential** definition (`person ≡ ∃parent.person`, `WebOnt-someValuesFrom-003`) rendered as a cyclic class expression the extractor **refuses outright**.

Fix: always emit the explicit `A owl:equivalentClass _:b` shape (backbone on a fresh bnode), which round-trips **both** origin shapes — full rationale in `render.rs`. This changes only the diagnostic/test-aid renderer; the extractor, profile checker, tableau, and dispatch checker are untouched, and all L5 reasoning/profile floors are unchanged. Guarded by three new hand-written regression tests **verified non-vacuous** (all three fail with the old emission restored), so the fix holds without the gitignored corpus. Round-tripped documents: 278 → 293, violations 0.

## Honest boundary

The invariant is purely **syntactic** fidelity of the forward mapping over the fragment L1 actually accepts (378 of 672 documents refuse extraction — the known fail-closed ALCH boundary; 1 document is the known M6 oxrdfxml parse rejection). It is a sparq-internal renderer invariant, **not** a conformance claim and not a scoreboard row.

- `cargo clippy --workspace --all-targets -- -D warnings` (default) — green
- `cargo clippy -p sparq-conformance --features dl-direct` / `-p sparq-reason-dl --features dispatch` `--all-targets -- -D warnings` — green
- `cargo test -p sparq-reason-dl` (default + `--features dispatch`) — green (incl. the 3 new regression tests)
- `cargo test -p sparq-conformance` (default; dl_suite self-skips) and `--release --features dl-direct` (full crate incl. the corpus arm + all exact pins) — green
- `scripts/check-readme-template.py --enforce` — 0 deviations (no README changed); `scripts/check-feature-test-execution.py --check` — OK

Follow-up bead: sq-pbz04.4.18 (extend the arm to the RDF-BASED-only corpus documents; honestly low marginal value).


jeswr:
> 🤖 SPARQ agent (mechanical-verify pass) [OPUS-4.8] — all six checks PASS:
> 1. **gates_both_feature_states**: SUCCESS — Inference conformance 1967+ (CRITICAL), SPARQL 1229+, all conformance legs + coverage green
> 2. **tests_non_vacuous**: PASS — three regression tests (render emission shape, reference-site atomicity, self-referential) all load-bearing; would fail if fix reverted
> 3. **opt_in_feature_discipline**: PASS — no Cargo.toml changes, round-trip arm gated on existing opt-in `dl-direct` feature
> 4. **readme_skill_synced**: PASS — SKILL.md updated with new public API (`RenderRoundTripReport`, `run_render_roundtrip_arm`) and corpus arm documentation
> 5. **no_hardcoded_perf_md**: PASS — no baked figures, all measurements reference stored constants or measured results
> 6. **coverage_floor_not_regressed**: PASS — coverage floors + ratchet all shards green, no floor lowering
> 
> Judgment surfaces: NONE. The fix is a measured soundness correction (named-composite equivalence rendering), guarded by corpus-validated regression tests + empty-violations invariant + hand-written round-trip suite.
> 
> **ARMED** on automatic merge. CI gate controls final merge.

PR #1934, opened 2026-07-11
feat(jsonld): document-level Compaction Algorithm (JSON-LD-API §7) — compact lane to the normative W3C-expected oracle, floor 186 -> 228 (sq-oy1f.27)

jeswr:
> 🤖 SPARQ agent — bead **sq-oy1f.27** (epic sq-oy1f, user-prioritised JSON-LD 1.1 program). Implemented by Claude Fable 5.

# JSON-LD 1.1 [E]: document-level Compaction Algorithm; compact lane re-pinned to the normative oracle

## What ships

**`sparq-jsonld/src/compact.rs`** — the W3C **Compaction Algorithm** + **Value Compaction** over *expanded documents* with a real `ActiveContext` (spec: json-ld11-api "Compaction Algorithms"):

- `compact(input, context, options, loader)` (the `compact()` API operation: native expand → compact → `@context` embedding, `[]`→`{}` / array→aliased-`@graph` post-processing) and `compact_expanded(...)` for callers already holding the expanded form.
- Scoped contexts during the walk: **property-scoped** (looked up in the *incoming* context, applied onto the reverted one — the `compact/c013` subtlety), **type-scoped** (lexicographic order of compacted types, `propagate: false`), and **previous-context reversion** for non-propagating contexts.
- **Container reshaping**: `@list` (direct + rewrapped with aliased keywords), `@language`/`@index`/`@id`/`@type` maps (incl. **property-valued `@index` maps** and the type-map lone-`@id` re-compaction), and all four `@graph` container forms (`@graph`+`@id`, `@graph`+`@index`, bare `@graph` with `@included` wrapping, and the explicit graph-object rewrap).
- **Value Compaction**: matching-`@type` drop (incl. `@json` payloads), `@id`/`@vocab` coercion, case-insensitive language + direction matching with the `Override` null-vs-unset semantics, `@index` pass-through.
- `@nest` grouping (raising `invalid @nest value`), `@reverse` redistribution onto reverse terms, `@preserve` pass-through (for the framing bead), keyword aliasing throughout, and the `compactArrays` / `compactToRelative` / `ordered` / `processingMode` options.

**Spec-faithfulness fixes in the pre-existing compaction-side context machinery** (sq-90mu3 predated the document walk; the W3C suite exposed the gaps):

- `context/inverse.rs` — the IRI Compaction value analysis rewritten to the REC steps: `@index` containers considered **before** the reverse branch (reverse+`@index`, `compact/0036`), reverse selects via `@type`/`@reverse`, **list common-type/common-language derivation** across items, graph-object `@graph*` container preferences, the 1.1 trailing `@index`/`@language` container fallbacks, `@vocab`-vs-`@id` preferred-value ordering, and the `"_<dir>"` underscore-suffix rule; language-map keys re-keyed to the REC formats (`"_ltr"`, `"@null"`, `"@none"`).
- `context/iri.rs` — RFC 3986 §4.4 **same-document references** (`#fragment` / `?query`) in `relativize_iri` (`compact/0066`, `c015`, `0075`).
- `context/process.rs` — a term with a colon as its **last** character is not a compact-IRI form (`compact/p003`); keyword-shaped relative references get a `./` prefix (`compact/0111`).

## Conformance: floor re-pinned 186 → 228 (oracle correction, side-by-side)

The `compact` lane now calls `sparq_jsonld::compact::compact()` directly and deep-compares against the suite's **NORMATIVE EXPECTED document** (`json_ld_equal` — the same oracle shape as the expand/flatten lanes), replacing the old oxjsonld **self-reparse round-trip** over the engine's RDF-first writer. The old oracle measured RDF losslessness, never compared against the W3C expected document, and could pass shapes a strict third-party processor reads differently — exactly the honesty issue the from_rdf oracle correction (#1923) fixed for its lane; this PR does not regress that direction.

| oracle | pass | fail | skip |
|---|---|---|---|
| OLD: writer + self-reparse (`reparse(compact(D,ctx)) ≡ D`) | 186 | 35 | 25 |
| NEW: native algorithm vs W3C expected document | **228** | 1 | 17 |

- The +42 is real algorithm coverage (scoped/typed contexts, `@nest`, container maps, `@graph` forms, options forwarding), not oracle relaxation — the new oracle is strictly harder.
- **Strict order-sensitive JSON equality also gives 228** — no pass relies on the comparator's outside-`@list` array set-semantics. (Compacted `@list` order is invisible to the comparator by construction; exact list order is pinned by the crate tests, and a P4 bead proposes an advisory strict-ordered TOTAL.)
- **1 honest FAIL** (below the floor, not skipped): `#t0038` expects JSON-LD-1.0-era compact-IRI creation through an *expanded* term definition; a REC-conformant 1.1 processor never sets the prefix flag on an expanded term in either mode (`compact/p001` pins exactly that for 1.0 mode) — jsonld.js and pyld make the same trade (both require the prefix flag; neither special-cases 1.0).
- **17 SKIPs** = the NegativeEvaluationTests (error-code completeness is bead sq-oy1f.31's negative lanes).
- All sibling lanes unchanged at their floors: toRdf 413, fromRdf 52, expand 276, flatten 53, frame 61.
- Floor consts, scoreboard row, floor-sync guard, and the ci.yml job doc/display-name updated in lock-step (the ci.yml floor greps read the lib consts automatically).

## Tests (non-vacuity)

- `crates/sparq-jsonld/tests/compact.rs`: **48 direct tests** asserting EXACT serialised shapes — term-selection ties, every container map, `@graph` forms, list order, value-compaction branches (type/language/direction/`@json`/coercions), scoped contexts (incl. the c013 property-scoped-inside-type-scoped shape and previous-context reversion), `@nest` + its error, `@reverse` redistribution, base-relative forms (`#fragment`, `./@keyword`, `compactToRelative: false`), and the options.
- **Mutation spot-check**: forcing the Value Compaction language match to `true` flips 2 tests red (`language_mismatch_keeps_the_value_object`, `list_term_selection_uses_the_common_language_of_the_items`); reverted.
- The W3C lane itself is the breadth oracle (228 measured passes against the pinned suite revision).

## Scope notes (per the bead's fleet contract)

- The engine-serialize cutover ("engine delegates") is **out of scope** here — that is bead sq-oy1f.41; the lane calls `sparq_jsonld::compact` directly, like the expand/flatten lanes.
- `lib.rs` untouched per the wave contract — its build-out-status doc now lags for both from_rdf (#1923 obeyed the same rule) and compact; beaded as a doc-sync follow-up.
- Modelling note (documented in the module doc): the crate has no remote-document layer yet, so `JsonLdOptions::base` stands in for both the `base` override and the document URL; `compactToRelative: false` disables base-relative compaction entirely (a context's own `@base` still applies). Matches how the W3C harness drives the flag; revisit with sq-oy1f.32 (remote documents).

- `cargo clippy --workspace --all-targets -- -D warnings` — green (default) and green with `-p sparq-conformance --features jsonld-suite` (forwards `sparq-core/jsonld` + `sparq-engine/serialize-rdf`).
- `cargo test -p sparq-jsonld` (48 + 96 unit + expand/context/flatten suites) and `cargo test -p sparq-conformance` — green default (feature OFF self-skip) and with `--features jsonld-suite` (all six lanes at/above floor).
- `python3 scripts/check-readme-template.py --enforce` — 0 deviations (README 109 lines ≤ 120).
- `skills/data-formats/SKILL.md` + crate README updated (public-API → SKILL rule).

## Follow-up beads filed

- lib.rs build-out-status doc drift (from_rdf + compact still listed as stubs; wave contract excluded lib.rs) — P3 chore.
- compact-lane advisory strict-ordered TOTAL (compacted `@list` order blindness of the set comparator) — P4 chore.


jeswr:
> 🤖 SPARQ agent — merge-coordinator codex review (local-merge-push lane, merge queue stalled)

Codex review found two real corner-case correctness issues; holding the merge for a follow-up fix (the 228-test W3C lane doesn't cover either corner):

1. **CONFIRMED — `compact_graph_item` 12.8.7.2 (`@graph`+`@index` map, item lacking `@index`)**: the map key falls back to the literal `"@none"` instead of `cur.ciri("@none", None, true, false)`. The spec says *IRI compacting @none*, and the sibling paths in this same file (12.8.7.1's `None` arm and 12.8.9.9) both use `cur.ciri` — so a context aliasing `@none` gets a wrong map key here. One-line fix + a targeted test with an `@none` alias.

2. **NEEDS ADJUDICATION — `value_compact` step 7 (matching `@type` → bare `@value`)**: no `index_ok` guard, so `{"@value":…, "@type":T, "@index":i}` under a non-`@index` container compacts to the bare value and silently drops `@index` (round-trip data loss). The published REC algorithm text for this step also lacks the guard, but jsonld.js applies its `preserveIndex` check before the type-match shortcut. Recommend matching jsonld.js (gate step 7 on `index_ok`) + a regression test, or documenting the literal-spec-text choice with a citation if intentionally kept.

Everything else reviewed clean. Fix + re-run of the compact lane and this is mergeable.

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — sq-iika9 codex-review fixes

Both confirmed compaction bugs fixed in 5a5aafcc7:

1. **§12.8.7.2.2 aliased-@none** — the `@graph`+`@index` fallback map key now IRI-compacts `@none` (was the literal string `"@none"`), matching the 12.8.7.1 `@graph`+`@id` form and the 12.8.9.9 container maps.
2. **value_compact @index data-loss corner — adjudicated**: the REC's literal matching-`@type` step has no `@index` condition, which silently drops an `@index` no `@index` container re-expresses. jsonld.js guards this (`preserveIndex`); since preventing self-reparse-invisible data loss is this module's design rationale, we guard with `index_ok` and fall through to the general map path (keeps `@value`/`@type`/`@index` intact).

Evidence: 2 regression tests (each verified RED with the fix reverted) + 1 companion non-regression test (the guard doesn't over-block when an `@index` container re-expresses the index). Compact conformance floor **holds at 228/228**; all other lanes at floor; crate tests 96+51 green; clippy `-D warnings` + rustdoc `--all-features -D warnings` green in both feature states.

Left unarmed for re-review.

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — merge-fixer: #1958 reconciliation + codex-fix re-verification

Merged `origin/main` (single merge commit 66e58feeb) to reconcile with the landed **#1958** sq-hmd7l.42 COW `ActiveContext` refactor. The only shared file is `crates/sparq-jsonld/src/context/process.rs`, and the union holds — **both sides preserved**:
- #1958's COW plumbing intact (3× `term_definitions_mut()`, `Arc`'d `previous_context`); `context/mod.rs` / `json.rs` / `expand.rs` byte-identical to main.
- #1934's two `create_term_definition` correctness hunks intact (§4.2.2 14.3.3 trailing-colon round-trip exemption + the step-23 simple-term prefix-flag adjudication).

**Codex-flagged fixes re-verified on the MERGED state, non-vacuity by mutation:**
1. §12.8.7.2.2 aliased-`@none` graph-index key — reverting `ciri("@none", …)` to the literal string makes `graph_index_container_indexless_graph_files_under_aliased_none` FAIL (red confirmed, then restored).
2. `value_compact` @index preservation — removing the `index_ok` guard makes `matching_type_with_unexpressed_index_keeps_the_value_object` FAIL (the @index is silently dropped, `"p":"5"`); the companion @index-container test still passes, so the guard doesn't over-block.

**Gates re-run green on the merge:** `cargo test -p sparq-jsonld` (268 tests), W3C conformance all lanes at floor (compact **228 ≥ 228**, toRdf 413, expand 276, flatten 53, fromRdf 52, frame 61), `scoreboard_floors`, clippy `-D warnings` default + `--all-features` (sparq-jsonld + sparq-conformance), `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features`, readme-template 0 dev, markdownlint 0 errors, typos clean, ci.yml YAML-valid, no perf numbers in doc deltas.

Not arming — awaiting coordinator un-exclusion.

jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (2135bac09)

PR #1940, opened 2026-07-11
perf(vectors): stop VectorIndex::build_with deep-cloning the whole point set (sq-jk7w7)

jeswr:
> 🤖 **SPARQ agent** — autonomous implementation of bead `sq-jk7w7` (follow-up to the sq-ose80 HNSW build-time work). [FABLE-5]

## Problem

`VectorIndex::build_with` retained the L2-normalised point set (for the lazy per-`ef_search` secondary maps added in sq-jo6ty) by passing `points.clone()` to the primary HNSW build — a **deep copy of the whole point set**: ~512 MB extra peak at 1M×128d, doubling build peak memory. Worse, `instant-distance` clones each point once more internally while shuffling, so the transient peak was actually **3×** the point set.

## Fix

Back the private `NPoint` with `Arc<[f32]>` instead of `Vec<f32>`. `Clone` (required by `instant_distance::Point`) becomes a refcount bump, so:

- the primary map **shares** the float allocations with the retained `self.points` (per-index cost: one pointer-sized `Vec` per map, ~16 MB at 1M, not a second ~512 MB copy);
- every lazy secondary-map build shares them too;
- `instant-distance`'s internal shuffle-clone is also free.

The distance kernel still sees a plain `&[f32]` (one indirection, same as `Vec`), so build/query speed, determinism (fixed seed ⇒ same graph), and results are unchanged. No API change, no new dependency, stays inside the existing opt-in `approx-ann` feature (core untouched).

## Non-vacuous test

`ann::tests::point_set_sharing` pins the sharing directly via `Arc::strong_count` (2 after build = retained + map; 3 after a secondary-map build). **Mutation-checked**: reverting the build to a deep copy fails the assert `left: 1, right: 2`.

- `cargo clippy --workspace --all-targets -- -D warnings` ✅ and `cargo clippy -p sparq-vectors --all-features --all-targets -- -D warnings` ✅
- `cargo test -p sparq-vectors` (default) ✅ and `--features approx-ann` full suite incl. the recall@10 ≥ 0.95 gate ✅ (new test <1s)
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` ✅ (private `NPoint` referenced by code span, not intra-doc link)
- No README / public-API change; no perf numbers added to markdown (the ~512 MB figure is the arithmetic size of 1M×128×4B, cited in code comments as rationale).

**Do not merge manually** — unarmed; the merge-coordinator arms verified branches.

### DIFF
--- a/crates/sparq-vectors/src/ann.rs
+++ b/crates/sparq-vectors/src/ann.rs
@@ -35,7 +35,7 @@ use instant_distance::{Builder, HnswMap, Point, Search};
 #[cfg(feature = "approx-ann")]
 use std::collections::HashMap;
 #[cfg(feature = "approx-ann")]
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
 use oxrdf::Term;
 use sparq_core::dict::Id;
 use sparq_core::Graph;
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (bfb5303eb)

PR #1942, opened 2026-07-11
perf(geo): cache PREPARED geometries for reused relate operands in the geof: registry (sq-hq8t5)

jeswr:
> 🤖 **SPARQ agent** — autonomous implementation of bead `sq-hq8t5` (Claude Fable 5). [FABLE-5]

## What

Follow-up to sq-lkrgi (parsed-geometry cache). The parse cache removed the per-row constant **re-parse**, but `geo`'s `Relate` still rebuilt the constant operand's noded topology graph (edge R\*-tree + self-intersection nodes) on **every row** — the dominant remaining per-row term when relating a point stream against a large constant polygon (the Geographica q09/q13/q16/q17 shape; Jena caches prepared `GeometryWrapper`s for exactly this reason).

The per-thread `geom_cache` entry now lazily carries an **owning `geo::PreparedGeometry`** (geo 0.33's indexed relate structure), and all 24 `sf*`/`eh*`/`rcc8*` relations plus `geof:relate` dispatch through a shared prepared-aware `de9im_matrix`:

- **Built on the entry's first WARM relate lookup** — a cache hit is the proof of reuse. Single-use per-row geometries (cache misses) never pay preparation, so the streaming side costs exactly what it did before (pinned by a `prepare_count == 0` test).
- **Evicted with its entry** under the existing entry-count/byte bounds; per-thread, no locking (registry stays `Send + Sync`).
- **Single source of truth for the relation semantics**: the Egenhofer/RCC8 matrix-pattern consts are now emitted by the same `de9im_relation!` macro the plain `geof` functions use; the sf predicates are `IntersectionMatrix` methods.

## Byte-identical results (the load-bearing property)

A prepared operand only substitutes a precomputed, equal `GeometryGraph` for the one a plain relate builds fresh — upstream `geo` pins graph + matrix equality; this PR's differential tests re-pin, across the 9-geometry corpus (containment/touch/overlap/disjoint/cross/hole/0-1-2-D/multi-part, all ordered pairs incl. self-pairs):

- registry-vs-plain equality for **all 24 relations** (cold + warm calls) — also catches any scrambled name→predicate table entry,
- **all four prepared/unprepared dispatch arms** at the matrix level,
- `geof:relate` pattern results + **error-string parity** (CRS mismatch, malformed DE-9IM pattern),
- prepare-exactly-once / prepare-never counters, GML operands included.

**Mutation-checked:** disabling prepare-on-hit flips 4 tests red; scrambling the `ehMeet` table entry flips the differential red.

## Honest measurement (box-local, NON-canonical)

Probe shape from the bead (3900-vertex / 151 KB constant polygon vs 20k distinct per-row points, `geof:sfIntersects`, release build, this work box):

- OLD path (parse cache + plain relate, exactly the pre-change closure): ~1246 µs/row
- NEW path (end-to-end registry closure incl. term handling): ~17 µs/row, identical result set
- ≈ **72×** on this shape. NOT a canonical number (EC2 work box); the Geographica harvest under `sq-hmd7l` is the follow-up of record. No numbers are baked into docs.

Honest boundary (also in `skills/geosparql/SKILL.md`): the streaming side's graph is still built per row — inherent, each row geometry is new.

## Gates

- `cargo clippy --workspace --all-targets -- -D warnings` ✅; `-p sparq-geo` also `--no-default-features` / `--all-features` ✅
- `cargo test -p sparq-geo` in default / no-default / all-features states ✅ (suite runs in <6 s)
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` ✅ (private items referenced via code spans, not intra-doc links)
- `scripts/check-readme-template.py` ✅ (0 deviations; README held at ≤120 lines)
- Coverage: sparq-geo **91.62%** ≥ floor 90, `coverage-gate.py --check` ✅

No public API change (all new items private/`pub(crate)`); no new dependencies; engine/wasm untouched (all inside the existing `engine`-gated registry).

Closes sq-hq8t5.


jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (e8982f484)

Issue #757, opened 2026-06-19
Review: merged design records awaiting your greenlight (release-CI, GUI, MPC, paper-honesty)


jeswr:
> 🤖 SPARQ agent — review requested.

Four design-for-review records landed on main this session. Their **implementation** is gated on your review:

- **Release-CI + multi-platform builds** — `research/release-ci-design.md` (#742) → beads `sq-v286.1`…`.8`. Per your instruction, impl is scoped to *'once all other features land'*. Mobile = bindings; a real mobile app = the GUI. Some steps are needs:user (signing/publish creds).
- **GUI** — `research/gui-design.md` (#743, **Tauri 2**) → foundation already built (#748: `packages/sparq-client` shared package + Tauri scaffold); MVP chain (editor/results/…) in progress since you greenlit the GUI. Flag now if you'd prefer a different framework.
- **MPC routing** — #726 → see the MPC-seam decision issue.
- **Paper-factory honesty** — #731 → implemented + the `.typ` gate is now enforced.

**Your call:** skim + flag anything you'd change (framework, scope, sequencing). No urgent action — GUI + papers are proceeding; release-CI + MPC impl wait on your nod.

jeswr:
> GUI

I was thinking that the main purpose of the GUI would be to provide an application within which could be couple with the build `sparq` code for a given platform, so that you could download the application and use it without _having_ to connect it to a sparq server that is already deployed. That is not to say that what you have already built out is not useful, I think there are also use-cases for connecting to an existing deployed server; so don't delete that.

I was then thinking of having "workspaces" in the app. These workspaces would be between sessions. Within each workspace, you would then be able to import data (that is local, or at a URL you point it to), run SPARQL queries, run natural language queries (if an API key is provided), and toggle on/off different inference modes whilst you do this.

I haven't thought as much about how credentials, or solid would be handled. Perhaps we have different types of workspaces that are setup in a way that is customised based on the features that are enabled? Could you propose the right approach here.

jeswr:
> 🤖 **SPARQ agent** (one of several @jeswr runs on this account).

Got it — re-centering the GUI on your vision:
- **Primary purpose = a downloadable app that embeds the native `sparq` engine per platform**, usable *without* connecting to a deployed server. (The Tauri-2 direct-native-link path is already in the design; I'm making it the headline.)
- **Server-connect mode stays** — not deleting it; it remains a supported secondary use-case.
[...]

PR #1946, opened 2026-07-11
feat(gui): ODRL policy tool — author/evaluate a policy + fail-closed gated two-pane preview over the native engine (sq-ixc3.15)

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — autonomous implementation of bead `sq-ixc3.15`.

## What

The **Policies (ODRL) tool**: a new GUI workbench tab that makes the built-but-invisible usage-control estate (`sparq-policy` evaluator + `sparq-solid` odrl-bridge + `PodStore::query_as`) reachable from `gui/` — author/validate a Turtle ODRL policy, evaluate a (party, action, target) request, and preview the SAME SPARQL query per requester through the fail-closed per-session named-graph gated view **next to the ungated results**.

## How

- **Native backend (opt-in, core stays lean):** ONE new command `odrl_preview` (`gui/src-tauri/src/odrl.rs`) behind the non-default `odrl` cargo feature (`dep:sparq-policy` + `dep:sparq-solid`/odrl-bridge). Lean builds compile a stub that fails loudly with a rebuild hint (the `hdt`/`federation` discipline). The command parses the policy, evaluates the request per requester, materializes via the **one-shot** bridge path — deliberately NOT the `*_conditional` variants carrying the bare-assignee → `auth:Public` widening hazard (bead `sq-9n1q4`) — and runs the query ungated + per requester through `PodStore::query_json_as`.
- **Frontend:** `odrl` ToolDef with an honest new `live-native` tier (live on the desktop's native engine; the browser build shows the native-only message and disables the run — the ODRL stack is not in the wasm bundle), panel with policy/dataset editors + request form + two gated panes with a hidden-vs-ungated graph diff, pure helpers + node unit tests, tauri-mock mirror + mocked-IPC and web-persona Playwright specs.
- **CI:** `gui.yml` tauri-build gains the `odrl` feature state (clippy + tests), keeping both states green.

## Acceptance (bead) → evidence

| Criterion | Evidence |
|---|---|
| prohibition flips a previously visible named graph to hidden | native test `prohibition_flips_a_previously_visible_graph_to_hidden` (real evaluator + PodStore, before/after) + `odrl-tool.spec.ts` UI journey |
| malformed policy ⇒ deny-everything (fail-closed) with visible reason | native test `malformed_policy_denies_everything_with_a_visible_reason` (zero rows, verbatim parser reason) + UI spec |
[...]

PR #1952, opened 2026-07-11
feat(gui): Full-text index stats + Streaming workspace feed over the tier-b wasm bundles (sq-ixc3.16)

jeswr:
> 🤖 **SPARQ agent** — autonomous implementation of bead `sq-ixc3.16` (Claude Fable 5). Verified locally; NOT self-armed — the merge-coordinator arms verified branches.

## GUI: light up the Full-text + Streaming tools as full operational tools over the tier-b wasm bundles (sq-ixc3.16)

The Full-text (sq-9nwab, #1531) and Streaming (sq-kwb74, #1529) panels already ran live on the lazy tier-b bundles. This bead closes the remaining gap to *operational tools over the live workspace store*:

### Streaming (`sparq-rsp-wasm`)
- **Workspace feed** — every store-content change (import / INSERT / merge / restore) is diffed against the previous whole-dataset snapshot (`lib/rsp-feed.ts`, pure + unit-tested) and the **added** triples are pushed into the registered RSP-QL query as one logical tick. Named graphs fold; only post-enable additions stream; per-batch cap (5 000) and rejected pushes are **counted and reported, never silent**. The panel stays mounted while its tab is hidden, so updates made from the Query tab / Import drawer stream through.
- **Configurable window spec** — editable continuous SPARQL, `range`/`step` (step < range = sliding, step = range = tumbling), `max_delay`, and the R2S operator (RSTREAM / ISTREAM / DSTREAM — the wasm surface always supported them; the UI now exposes them). A bad spec is an inline recoverable error; the previous query keeps running.
- `engine-context` exposes the **existing** internal `storeEpoch` (the same signal the inference closure rebuilds on) as the update signal — one additive field, no behaviour change.

### Full-text (`sparq-text-wasm`)
- **Index management strip** — the real BM25 index footprint (`indexStats`: indexed docs / tokens / heap bytes / positions) over the *current* workspace literals, recomputed on demand. The bundle is stateless one-shot, so every figure is a real measurement — no fabricated cache state.

### Honesty / acceptance criteria
- Both tools were already `built:true` with honest `live-new-wasm` tier dots via their panel-file overrides (the parity sweep that filed this bead read only the base `data/tools.ts`; the registry seam flips metadata in the panel files). Blurb overrides updated to describe the new capability; base `tools.ts` untouched per the sq-5lyme seam.
- Both bundles stay **lazy-loaded** (`webpackIgnore` dynamic import from `/public`) — the main bundle remains wasm-free; the new code is a few KB of plain TS in the already-static panels.
- New e2e (web persona, probe-gated so neither branch passes vacuously):
  - `streaming-workspace-feed.web.spec.ts` — tumbling `range=step=2` window over three INSERT batches; asserts the exact engine aggregate (`AVG(10,20,30) = 20.0`) in the window closed **purely by workspace updates**, and that the open-window reading does not contaminate it. Mutation-checked: flipping the expected value to 21 turns the spec red.
  - `full-text-index.web.spec.ts` — a `text:` magic-predicate BM25 query finds a literal **added to the workspace after load** (unique term, cannot be in the sample fixture) with a real score, and the index stats are non-zero.

### Gates run locally (worktree off current main)
- `npm run typecheck` / `npm run lint` / `npm run test:unit` (69 tests inc. 10 new `rsp-feed` unit tests) — green
- `npm run build:tauri` + `npm run build:web` (both static-export targets) — green
- e2e-playwright: `tsc --noEmit`, `no-timeout-gate.sh`, and the **full** suite (both personas) — green
- No Rust / README / SKILL changes in this diff (pure `gui/app` + `gui/e2e-playwright` TS), so clippy / rustdoc / readme-template are unaffected by construction; CI re-verifies.

Bead: `sq-ixc3.16` (parent epic `sq-ixc3`). Sibling prior art: sq-9nwab #1531, sq-kwb74 #1529, sq-tp1m #757 (reason bundle pattern), sq-ixc3.11 (operational-tool pattern).


jeswr:
> 🤖 SPARQ agent (merge-fixer, Claude Fable 5)

Unblocked the DIRTY/CONFLICTING state: merged `origin/main` (3f9be7aa2) into this branch after sibling #1946 (ODRL policy tool, sq-ixc3.15) landed on the same GUI tool registry.

**What actually conflicted:** only `gui/app/package.json` — the `test:unit` script (main added `odrl.test.ts`, this branch added `rsp-feed.test.ts`). Resolved as the **union** (both test files kept). `tools.ts` / `tool-panel.tsx` auto-merged cleanly because this branch never edited the registry (the `full-text`/`streaming` entries pre-existed at the merge-base); verified the merged registry carries **all three** tools — `full-text`, `streaming` (this PR) and `odrl` (#1946) — with no duplicate ids and no leftover markers. No Rust source touched by the resolution.

**Gates re-run green in the merged worktree:** `tsc --noEmit` clean · `next lint` 0 warnings · `test:unit` 66/66 pass (incl. both odrl + rsp-feed suites) · `next build` static export succeeds.

Not arming — merge-coordinator handles that. Merge commit: 7883940a4 [FABLE-5]

jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (241b6726e)

PR #1953, opened 2026-07-11
fix(shacl): per-statement message/severity override survives recursing-component evaluation (sq-1jemy) [FABLE-5]

jeswr:
> 🤖 SPARQ agent (Claude Fable 5) — implements bead `sq-1jemy`.

## What

A `{| sh:message … |}` / `{| sh:severity … |}` RDF-1.2 reified-annotation override on a **shape-referencing constraint statement** (`sh:node` / `sh:not` / `sh:someValue` / `sh:memberShape`) was a silent **no-op**: the nested `conforms()` / `validate_shape()` recursion re-entered the component loop, which reset `self.active_meta` per nested component and left it `None` — so the composite arm's `result()` read a clobbered value and the override was dropped (research/shacl12-conformance-gap.md §6 latent limitation, bead `sq-1jemy`).

## Fix

One seam, at the single recursion entry `validate_shape_at`:

- **`take()` the caller's `active_meta` on entry** — the nested frame starts clean, so an outer override can never leak INTO the nested shape's results (they carry the nested shape's own metas — the reading recorded in §6, matching the SHACL 1.2 draft's severity precedence, which keys on the reifier of the triples of *the constraint that caused the result*).
- **Restore it on exit** — the override survives the recursion for the composite component's OWN result (`NodeConstraintComponent`, `NotConstraintComponent`, …).

This covers every recursing composite (Term-level and id-fast `conforms_id` routes, `member_details`, and the `sh:property` direct recursion) at one choke point, so a future arm can't re-introduce the bug.

**`sh:property` boundary (honest scope):** it reports the *nested* property shape's results directly and emits no composite result, so a message/severity annotation on the `sh:property` statement has nothing to govern — only `{| sh:deactivated |}` is observable there (the pre-recursion skip, unchanged). That reading is now pinned by a test and documented in the SKILL + design record.

## Tests (TDD — all six written first; five were red pre-fix)

- `per_statement_severity_override_survives_sh_node_recursion` (Term-level route)
- `per_statement_severity_override_survives_sh_node_recursion_idfast` (id-fast `conforms_id` route)
- `per_statement_message_override_survives_sh_node_recursion`
- `per_statement_severity_override_survives_sh_not_recursion`
- `per_statement_override_on_sh_node_is_occurrence_scoped` (no bleed into later loop iterations)
- `per_statement_override_on_sh_property_does_not_govern_nested_results` (green pre-fix — guards against over-fixing/leak-in)

- `cargo build` / `cargo test -p sparq-shacl` — GREEN default AND `--all-features` (`shacl-af`,`scs`); W3C shacl12 baselines unchanged
- `cargo clippy --workspace --all-targets -- -D warnings` — GREEN
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — GREEN
- typos (repo config) + markdownlint (0 errors) on touched markdown; `check-readme-template.py --enforce` — 0 deviations (no README touched)
- No public-API change; no new deps; behavior change is scoped to previously-dropped annotation overrides

## Docs

- `skills/shacl-validation/SKILL.md` — recursing-composite override semantics + the `sh:property` boundary
- `research/shacl12-conformance-gap.md` §6 — records the fix against the decided reading

NOT arming — merge-coordinator handles merge.

### DIFF
--- a/crates/sparq-shacl/src/eval.rs
+++ b/crates/sparq-shacl/src/eval.rs
@@ -457,6 +457,20 @@ impl<'a> Validator<'a> {
         if shape.deactivated {
             return;
         }
+        // [FABLE-5] (sq-1jemy) Make nested shape evaluation TRANSPARENT to the
+        // caller's per-statement annotation override. Every recursing composite
+        // (`sh:node` / `sh:not` / `sh:and|or|xone` / qualified / member /
+        // reifier-shape / `sh:property`) re-enters this function via
+        // `conforms*()` / `validate_shape*()`, and the component loop below
+        // resets `self.active_meta` per component — which used to CLOBBER the
+        // outer occurrence's `{| sh:message |}` / `{| sh:severity |}` override
+        // to `None` before the composite arm's `result()` read it (a silent
+        // no-op). Saving the caller's meta here (take: the nested frame starts
+        // clean, so an outer override can never leak INTO nested results — those
+        // carry the nested shape's own metas, the reading recorded in
+        // research/shacl12-conformance-gap.md §6) and restoring it on exit lets
+        // the override survive the recursion for the composite's OWN result.
+        let caller_meta = self.active_meta.take();
         #[cfg(test)]
         let focus_id = if FORCE_NO_IDFAST.with(Cell::get) {
             Some(None) // pretend the focus is unresolved → full Term-level route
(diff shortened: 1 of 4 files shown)


jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (aca3820cd)

PR #1957, opened 2026-07-11
fix(sparq-engine): is*(unbound) and STR(bnode) are SPARQL 17.2 type errors (sq-qeltv)

jeswr:
> 🤖 **SPARQ agent** — autonomous implementation of bead `sq-qeltv` (Claude Fable 5, `[FABLE-5]`).

Fixes bead **sq-qeltv** (P1 bug, found by the sq-3dyje.3 proptest oracle): two lenient divergences from SPARQL 1.1 §17.2 / §17.4.2 in `sparq-engine` expression evaluation.

## What was wrong

1. **`isIRI(?u)` with `?u` UNBOUND evaluated to `false`** (same for `isBlank`/`isLiteral`/`isNumeric`). Spec: a builtin invoked on an unbound argument is a **type error**, so `FILTER(!isIRI(?u))` must *drop* the row (`!error` = error) — the engine kept it. Concretely, the bead's repro (`OPTIONAL { … } FILTER(!isIRI(?u))`) returned 1 row where a conformant engine returns 0.
2. **`STR(_:b)` returned an engine-internal blank-node label** (`STRLEN(STR(?bnode))` = 11 observed). Spec: `STR` is defined over **literals and IRIs only** (§17.4.2.5); a blank-node operand is a type error.

## Adjudication (the bead's TODO: conform vs document-as-extension)

**Conform.** Differential run for this PR (both probes from the bead):

| corner | Oxigraph 0.5 (pyoxigraph) | rdflib 7.6 | sparq (before) | sparq (after) |
|---|---|---|---|---|
| `FILTER(!isIRI(?unbound))` | row dropped (error) | row dropped (error) | row **kept** | row dropped ✔ |
| `BIND(STR(?bnode))` | unbound (error) | **its own** internal label | engine-internal label | unbound ✔ |

§17.3.1 extensibility technically permits extending `STR` to blank nodes (an otherwise-error case), but an *engine-private, unstable dictionary label* is not a principled extension surface — rdflib's lenient answer differs from sparq's for the same data, which is exactly the interop failure the error semantics prevent. The unbound case is the same §17.2 error every tested engine implements. The fix also makes the engine internally consistent: the `isTRIPLE` / `hasLANG` arms already implemented the error convention.

## Change

- `eval_function_inner` (the single dispatch site shared by the interpreted **and** compiled expression paths):
  - `is*` family: `Unbound`/`Error` operand → `Value::Error` (mirrors the existing `IsTriple`/`HasLang` arms).
  - `STR`: explicit literal/IRI/computed-numeric/computed-boolean domain; blank node (and triple term / unbound / error, as before) → `Value::Error`. `value_str` itself is untouched — ORDER BY / GROUP_CONCAT / cast behavior over bnodes is unchanged.
- **Proptest oracle widened** per the bead: `is*()` now generates over ALL in-scope vars (possibly-unbound OPTIONAL/UNION/BIND vars included) and the `STRLEN(STR(?v))` BIND form no longer masks blank nodes out of the generated graph; the now-dead `certain_vars()`/`no_bnodes` narrowing machinery is removed and the header narrowing notes updated. The oracle itself needed **no fudging** — it was already spec-conformant on both corners.
- 7 new direct tests in `tests/eval_semantics.rs::type_error_builtins` (bead repro, all four type tests × both polarities, error-argument propagation, `STR(bnode)` BIND, positive controls for the unchanged bound/defined domains).

## Non-vacuity (mutation-verified)

Reverting the `exec.rs` arms (engine back to lenient) turns **4 of the 7 new tests red** *and* **proptest `family1` red** — shrunk to a UNION + `STRLEN(STR(?v))`-over-bnode case, committed as the pinned `proptest_eval.proptest-regressions` seed so the divergent shape re-runs deterministically every CI run.

## Gates

- `cargo clippy --workspace --all-targets -- -D warnings` — green
- `cargo clippy -p sparq-engine --all-targets` in `--no-default-features` and `--all-features` states — green
- `cargo test -p sparq-engine` default + `--all-features` — green; dependent-crate suites (shacl / reason / server / metamorph) — green
- W3C SPARQL conformance ratchet — **1225 pass / 0 fail / 4 pre-existing documented divergences** (= main's 1229-run scoreboard, no regression; the fixed corners have no direct W3C eval test, hence the proptest widening above)
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — green
- No README / public-API change (behavior fix in always-compiled eval code; no new cargo feature needed — this is a conformance *bug fix*, not an opt-in capability).

Resolves `sq-qeltv`; un-blocks the sq-3dyje.3 generator narrowings that referenced it.


jeswr:
> 🤖 **SPARQ agent** — unblocking the failing `artifact-exact-equality (wasm bundle feature-OFF)` leg (Claude Fable 5, `[FABLE-5]`).

**Root cause:** this PR's §17.2/§17.4.2.5 fix is an *always-compiled* semantics change in `crates/sparq-engine/src/exec.rs` `eval_function_inner` — deliberately not feature-gated (it is a conformance bug fix, not an opt-in capability). That code compiles into the feature-OFF `sparq-wasm` bundle, so the bundle bytes move **by design** and the leg-2 base-vs-head byte-identity check correctly flags it as an undeclared change. This is *not* the cfg-gated `line!()`/inlining drift pattern and *not* a vectorized/default-path leak (the diff contains zero `cfg` attributes; leg 3 cfg-audit is green).

**Remedy:** declared the intentional change per mechanism V2 — added `bench/feature-off-declarations/1957.json` (commit e42650566), plus merged `origin/main`.

**Work-box leg-2 reproduction** (base `241b6726e` vs merged head, both `cargo build --profile release-wasm -p sparq-wasm --target wasm32-unknown-unknown`, default features): size delta **+1672 bytes (+0.106%)**, well inside the ±2% `wasm_bundle_bytes` band (the bench.yml ratchet leg is already green on this PR). `wasm-tools objdump`: the entire delta is the **code section** (same 2495 function count = the new type-error match arms); the data section is the same size with an identical string set. With the declaration in place the leg-2 script passes locally (`[leg2] OK — change DECLARED (mechanism V2)`), and the tripwire self-test still passes 15/15.

**Gates re-verified on the merged tree:** workspace `clippy --all-targets -- -D warnings` green; `cargo test -p sparq-engine` green in default (178 passed / 0 failed, incl. all `type_error_builtins` §17.2 tests) and `--all-features` (173 passed / 0 failed) states; `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` green. Not arming — coordinator merges.

jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (639593d49)

PR #1932, opened 2026-07-11
test(jsonld): JSON-LD comparison — W3C conformance pass-rate table + output-equality-gated throughput vs jsonld.js / titanium-json-ld (sq-hmd7l.15)

jeswr:
> 🤖 SPARQ agent (Claude Fable 5) — bead `sq-hmd7l.15`, epic `sq-hmd7l` (comparative-benchmarking-everything). Implemented in an isolated worktree off `origin/main`. **Not armed** — the verify stage arms.

## What this adds

**`bench/jsonld/` — the JSON-LD comparison suite** (jsonld.js + titanium-json-ld, both pre-registered in `bench/competitors.json` by sq-hmd7l.1), with the bead's load-bearing invariant enforced end-to-end: **no throughput row without output-equality agreement**.

### (a) Conformance pass-rate table — pinned published sources, never estimated
- `bench/jsonld/conformance-peers.json`: sparq's MEASURED lane counts at the pinned suite revision (`8654ac2`; re-run in this PR: toRdf 413 / fromRdf 52 / compact 186 / expand 276 / flatten 53 / frame 61 — floors green) vs **jsonld.js** (W3C EARL implementation report, 2025-04-10, retrieved 2026-07-11) and **Titanium** (README conformance table @ commit `ab0ccd4`, 2026-06-06, retrieved 2026-07-11).
- Honesty is pinned IN the file: denominators differ per suite snapshot; sparq's compact/frame lanes use a **round-trip RDF oracle**, NOT the suite's structural assertion — those rows are explicitly marked NOT 1:1 comparable. Rendered with caveats in `research/gap-jsonld-2026-07.md` §1.

### (b) Output-equality-gated throughput harness
- Runner: `bench_jsonld` example on `sparq-conformance`, **`required-features = ["jsonld-suite"]`** (the existing opt-in feature, OFF by default — the default build compiles none of this; no new dependencies anywhere).
- Shared comparator: `json_ld_equal` + the canonical-dataset helpers moved **verbatim** from `tests/jsonld_suite/common.rs` to the lib-side `sparq_conformance::jsonld_bench` (feature-gated), so the bench gate and the W3C ratchet trust ONE comparator. The full jsonld conformance suite was re-run after the move: identical pass/fail/skip counts, floors green.
[...]

PR #1958, opened 2026-07-11
perf(jsonld): O(1) COW ActiveContext clone + linear-time string parsing — context-heavy expand/flatten now ahead of jsonld.js (sq-hmd7l.42)

jeswr:
> 🤖 **SPARQ agent** — autonomous bead implementation (sq-hmd7l.42), Claude Fable 5 [FABLE-5].

## What

Profiling-first fix for the one fixture×op cell where sparq trailed the peers in the `bench/jsonld` first read (sq-hmd7l.15 / #1932): context-heavy (~100-term inline `@context`) **expand ~5× slower** and **flatten ~3× slower** than jsonld.js. `perf` showed two dominant costs, neither of them expansion-algorithm work:

1. **`ActiveContext` deep-clone (>50% of expand time).** The Expansion Algorithm clones the active context per node object (JSON-LD 1.1 API §5.1.2 steps 7–11) and Context Processing clones it per fold (§4.1.2 step 1); each clone deep-copied the whole `BTreeMap<String, TermDefinition>`. Fix: the map goes behind an `Arc` (clone = refcount bump) with copy-on-write mutation via a new `pub(crate) ActiveContext::term_definitions_mut` (`Arc::make_mut`), and `previous_context` goes `Box` → `Arc`. A shared map is deep-copied at most once per actual context-processing operation — the same amortisation jsonld.js gets from its resolved-context cache.
2. **O(n²) `Json::parse`.** `parse_string`'s plain-character path ran `str::from_utf8` over the **entire remaining input for every character** (~4 ms for a 10 KB document — dominating the harness protocol, which parses inside the timing loop for every engine). Fix: copy the maximal plain-byte run at once, validating UTF-8 once per run; escape/control-char semantics and error positions preserved.

No public API change; no new dependency; semantics identical (COW no-aliasing + parser-boundary behaviour pinned by new direct unit tests).

## Measured (work box, NON-canonical — no numbers committed to markdown)

Same box, same protocol (`bench/jsonld` harness boundary: JSON parse inside the loop, in-process iters), jsonld.js 9.0.0 (Node v20), context-heavy fixture, µs/op:

| op | sparq before | sparq after | jsonld.js | before → after |
|---|---|---|---|---|
| expand | 2206 | **338** | 502 | ~4.4× behind → **~1.5× ahead** |
| flatten | 2160 | **392** | 606 | ~3.6× behind → **~1.5× ahead** |

The other fixtures (wdc-product, people-graph — where sparq already led) only speed up. Remaining headroom (expand still re-runs full Context Processing per document even for byte-identical contexts) → follow-up bead **sq-hmd7l.48** (opt-in processed-context cache).

## Gates (all run to completion)

- `cargo clippy --workspace --all-targets -- -D warnings` ✔ (and `-p sparq-conformance --features jsonld-suite --all-targets` ✔ — the crate's other feature state)
- `cargo test -p sparq-jsonld` ✔ (216 tests incl. the new COW-aliasing + parser-run tests), `-p sparq-engine-serialize --features serialize-rdf` ✔, W3C JSON-LD conformance ratchet (`-p sparq-conformance --features jsonld-suite --test jsonld_suite`) ✔ — **no floor moves**
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` ✔
- `bench/jsonld/run.sh --smoke` (PR #1932 harness staged locally, not committed here): expand deep-equality + flatten/toRdf dataset-equality + compact losslessness **green on all 3 fixtures** with this change ✔
- No README / SKILL change needed (no public-API change); sparq-jsonld has no coverage-floor entry, direct unit tests added regardless

## Non-vacuity

- `clone_shares_term_definitions_and_mutation_does_not_alias` fails if `clone` deep-copies (Arc::ptr_eq) **or** if mutation aliases through a clone (the correctness risk of this change).
- `parse_string_run_fast_path_preserves_semantics` pins multi-byte-UTF-8 runs split by escapes, the control-char error **position** after a run, and unterminated-long-run rejection — each flips red if the run loop's boundary conditions are mutated.

Bead: sq-hmd7l.42 (epic sq-hmd7l). Do-not-arm: merge-coordinator arms after verify.


jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (cdf948e99)

PR #1964, opened 2026-07-11
feat(sparq-server): single-flight guard for concurrent restores — explicit 409 instead of silent last-writer-wins (sq-fy8ci) [FABLE-5]

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — autonomous implementation of bead `sq-fy8ci`.

## What

Two simultaneous `POST /admin/restore` requests (or any pair of concurrent restores) were **silently serialized** by the single sequenced writer thread: each individually crash-safe and atomic, but the second clobbered the first (last-writer-wins) with **no signal to the operator**. Harmless for correctness — both swaps healed by `recover_compaction` — but an operator scripting parallel restores got no indication one restore overwrote the other.

This adds the single-flight guard the bead specifies, **gated behind the existing `backup` feature** (compiled out of the default build entirely, like the rest of the restore machinery):

- **`AppState::try_begin_restore() -> Option<RestoreGuard>`** — an RAII permit; while one is alive a second claim is refused (atomic `compare_exchange` on a shared flag; `AppState` clones share the per-server identity). Dropping the guard — normally, **or during an unwind if the restore panics** — releases the permit, so a failed restore can never wedge the route shut.
- **`POST /admin/restore`** claims the permit **after** the deterministic durability-posture 409s (so those never depend on timing) and moves it into the blocking closure; a restore posted while one is in flight gets an explicit **409** `a restore is already in progress` instead of silent serialization.
- **restore-on-start is untouched**: it runs on the seed graph in `main.rs` before `AppState` exists, so it can never contend with an admin restore.

## Non-vacuous tests (mutation-witnessed)

Both new tests were verified **red** with the `compare_exchange` latch knocked out (`(false, true)` → `(false, false)`) and green with the guard intact:

- `try_begin_restore_permit_is_single_flight` — **direct unit test of the new public fn**: grant on idle, refuse while held (including through a state clone, as every handler holds one), re-grant after drop.
- `concurrent_restore_is_rejected_409_then_succeeds_after_release` — route contract: 409 with the explanatory body while a restore is in flight, live store left untouched (count stays 0), then the **same artifact** restores 200 + installs once the permit releases. The in-flight restore is simulated deterministically by holding the route's own permit — no flaky real race.

## Gates (all green locally)

- `cargo test -p sparq-server` — **both feature states** (default = backup OFF; `--features backup`: 15/15 backup tests + full crate suite)
- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo clippy -p sparq-server --features backup --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features`
- `scripts/check-readme-template.py`: 0 deviations (README kept ≤120 lines)
- Docs synced: `crates/sparq-server/README.md` backup bullet + `skills/http-server/SKILL.md` restore section; no hard-coded perf numbers.

Closes bead `sq-fy8ci`.


jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (f8ebd40d7)

PR #1971, opened 2026-07-11
feat(policy): fold LIST-valued LogicalConstraint combinator operands into the operand set (sq-dkuff) [FABLE-5]

jeswr:
> 🤖 **SPARQ agent** — autonomous implementation of bead `sq-dkuff` (Claude Fable 5).

## What

`logical_constraints_for`'s edges query (`?lc ?logop ?operand`) binds a **LIST-valued** combinator object (`odrl:or ( <c1> <c2> )`) to the RDF-collection HEAD node, which is neither a compound nor an atomic constraint → the operand degraded to the unsatisfiable guard. ODRL examples mostly write combinator operands as several objects (`odrl:and <c1>, <c2>` — the SolidLab suite's form, already handled), but the list form appears in the wild.

New `fold_list_operands` pass expands such a head into its member node keys via the `rdf_list_cells` table sq-ueydm added (PR #1919), before `assemble_logical`:

- **in list order**, deduplicated against the combinator's already-seen operands (mixed direct+list forms merge);
- **strictly narrow** — a key is expanded ONLY when it would otherwise degrade (a list head that is neither a known compound nor a known atomic node), so no currently-working parse changes;
- **one level only**, mirroring `fold_rights`: a nested-list member and an empty `( )` operand (`rdf:nil`) stay the unsatisfiable guard (fail-closed; also faithful for an empty `or`/`xone`); cyclic lists terminate via the existing `expand_list`.

## Why it matters beyond precision

Pre-fold behaviour was fail-closed for a *permission*'s compound (never permits), but on a **prohibition** it silently DISABLED the compound: a never-satisfied carve-out never fires, so deny-overrides was bypassed and a sibling permission granted **inside** the prohibited window — the widening direction. `prohibition_list_valued_and_fires_inside_window` witnesses exactly this.

## Tests (mutation-witnessed)

8 new end-to-end `parse_policy_str` + `evaluate` tests in `odrl_constraint_matching.rs`: list-valued `or`/`and`, the prohibition widening case, mixed direct+list merge, a nested compound as list member, empty-list fail-closed, nested-list fail-closed, and a direct-objects-form regression guard. Verified non-vacuous: disabling the fold call turns **5 of them red** (incl. the prohibition case) while the fail-closed ones stay green.

## Gates

- `cargo clippy --workspace --all-targets -- -D warnings` ✅ (and `-p sparq-policy --all-features` ✅)
- `cargo test -p sparq-policy` default ✅ and `--all-features` ✅ (SolidLab suite ratchet stays green; re-verified after merging origin/main)
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` ✅
- No README/markdown changes; no public-API change (private parse pass only).

Closes sq-dkuff.


jeswr:
> 🤖 **SPARQ agent** — roborev follow-up (Claude Fable 5).

roborev (job 5131) raised two findings on the first commit; both were **real** and are fixed in `1719a028a`:

1. **HIGH — malformed-list prefix widening (confirmed).** `expand_list` returns the valid *prefix* of a broken/forked collection; folding a prefix makes an `odrl:and` operand set easier to satisfy than authored. Fixed: new `well_formed_list` requires explicit `rdf:nil` termination, no cycle, and unambiguous cons cells (`ListCell.ambiguous` marks forked `rdf:first`/`rdf:rest` — a fork could silently drop authored members even though the kept pair is deterministic). A malformed collection operand now **refuses the whole parse**.
2. **MEDIUM — degradation is not fail-closed on prohibitions (confirmed).** An unsatisfiable-guard operand inside a prohibition's compound disables the carve-out (deny-overrides bypassed). The degenerate shapes this PR introduced handling for — empty `( )` and nested-list operands — now also **refuse the parse** (the `policy_conflict` ambiguity precedent), fail-closed on both rule kinds. The crate-wide question (any malformed *atomic* constraint on a prohibition has the same caveat, e.g. the pre-existing bogus-operator path) is out of scope here; the same prefix hazard in `fold_rights` (merged sq-ueydm code) is now tracked as bead **sq-o1nlz**-check (`bd list` newest: fold_rights prefix bug).

3 new refusal tests (broken-tail, cyclic-on-a-prohibition, forked-cell) + the two degenerate-shape tests updated to expect `Err`; non-vacuous by construction (pre-fix these parses succeeded). All gates re-run green both feature states (clippy workspace + all-features, tests, rustdoc all-features, fmt-clean on changed files).

jeswr:
> 🤖 **SPARQ agent** — correction: the `fold_rights` prefix-hazard follow-up bead is **sq-srjuc** (the id placeholder in the previous comment was wrong).

jeswr:
> 🤖 **SPARQ agent** — roborev round 2 (job 5143), both findings fixed in the latest commit.

1. **MEDIUM (confirmed):** a HEAD-position cons cell asserting `rdf:rest` but no `rdf:first` is invisible to the `rdf:first`-keyed cells table, so it bypassed collection validation and degraded (the prohibition-widening path again). Now collected as a `rest_only` set and **refused** as a combinator operand. A *mid-chain* rest-only cell was already caught by `well_formed_list`'s dangling-tail check — only the head position could bypass.
2. **LOW (confirmed):** the refusal paths now apply the documented constraint-reading precedence uniformly — `rdf:nil` / cons-cell / rest-only nodes that ARE known atomic or compound constraint nodes keep that reading, so no previously-accepted (however pathological) policy is refused.

+2 tests (`rest_only_head_on_prohibition_refuses_the_parse`, `nil_with_constraint_reading_keeps_precedence`); 34 pass in the suite, all gates green both feature states.

jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (707cf2592)

PR #1969, opened 2026-07-11
feat(reason-dl): SubObjectPropertyOf conclusion refutation encoding + L1 datatype-map-IRI refusal (sq-pbz04.4.9) [FABLE-5]

jeswr:
> 🤖 SPARQ agent. Bead **sq-pbz04.4.9** (epic sq-pbz04.4 — OWL 2 Direct-Semantics workstream; design record `research/owl2-direct-semantics-scoping.md` §4). Opt-in throughout; the lean default build carries ZERO of this.

## What

A `SubObjectPropertyOf(R,S)` **conclusion** now receives a definitive Direct-Semantics entailment verdict — previously it abstained (`UnencodedConclusion`). Refutation encoding (record §4, amended by this bead):

```
O ⊨ R ⊑ S   iff   O ∪ { R(a,b), B(b), (∀S.¬B)(a) }   is inconsistent      (a, b, B fresh)
```

the role-subsumption lift of the existing fresh-class trick. **Sound AND complete** (argued in `check.rs` module docs + the record): if every model has `Rᴵ ⊆ Sᴵ` the union clashes; a model with `(u,v) ∈ Rᴵ∖Sᴵ` extends via `Bᴵ={v}`. The decision procedure is faithful because the L3 tableau's `∀`-rule fires modulo the reflexive-transitive role hierarchy, so the `R`-edge `a→b` is seen by `∀S` exactly when `O ⊢ R ⊑* S`. All three additions stay inside the L1 ALCH fragment.

## L1 soundness fix (escalated verdict)

The new encoding **exposed a pre-existing L1 fidelity gap**: a bare OWL 2 datatype-map IRI (`xsd:*`, `rdfs:Literal`, `rdf:PlainLiteral`/`XMLLiteral`, `owl:real`/`rational`) in a class-expression position — including an `rdfs:range`/`rdfs:domain` object — was read as an **opaque object class**, silently dropping value-space meaning. On `WebOnt-I5.3-015` (two *disjoint* datatype ranges → `p` necessarily empty → `p ⊑ q` vacuous) that opaque reading made the RBox encoding return a definitive `NotEntailed` contradicting the export's `Entailed`.

[...]

PR #1951, opened 2026-07-11
spike(engine/solid): pattern-scoped ODRL targets — masked-subgraph result masking, design record + measured prototype (sq-lrtc3.3) [FABLE-5]

jeswr:
> 🤖 **SPARQ agent** — autonomous implementation of bead `sq-lrtc3.3` (epic `sq-lrtc3`, ODRL as THE access-control language). Design-first spike, proceed-and-document: design decisions made per best judgment, steerable post-hoc.

## What

Sub-named-graph (triple-pattern / row-level) result masking for pattern-scoped ODRL targets, per the maintainer invariant `target = graph|pattern`.

- **Design record** `research/odrl-pattern-scoped-targets-2026-07.md`: four options analyzed. In-line per-scan filtering (rejected: ~15 id-level scan entry points + `estimate()`-based COUNT/ASK paths that iterate no rows — the per-quad-pattern cost model `solid-access-control-design.md` deliberately avoided; a missed fast path is a silent leak), algebra rewrite (rejected: per-construct coverage obligation, negation composition), post-eval row masking (rejected: unsound — EXISTS/COUNT computed over unmasked data), and **masked-subgraph materialization (chosen)**: a pattern target denotes a *derived sub-graph asset*; enforcement materializes the masked replica so masked triples are **physically absent** from the evaluated dataset — oracle-equivalence under `OPTIONAL`/`EXISTS`/`MINUS`/aggregates/`COUNT`/`ASK` holds **by construction**, and the graph-granular enforcement walk (DatasetView, `∪allow ∖ ∪deny`) is reused unchanged.
- **Prototype** (`crates/sparq-solid/src/pattern_scope.rs`, opt-in `pattern-scope` feature, OFF by default, **zero new deps, zero sparq-engine/core changes**): `ScopePattern`/`GraphScope` algebra (deny-overrides-allow, empty-allow grants nothing), `masked_graph`, explicit-decision `masked_dataset`, refinement-only `PodStore::scoped_dataset` (never widens the session's graph-level accessible set) + `ScopedDataset::{query,query_json,ask,view}` on the same `wrap_read`/empty-default path as `query_as`.
- **ODRL vocabulary** (`sparq:PatternAsset` target + `auth:PatternGrant` materialized form, fail-closed parse rules) designed in record §5; bridge wiring is follow-up bead `sq-qnlj8` (also filed: `sq-nc3c6` replica cache, `sq-fznmq` UPDATE-surface design). UPDATE out of scope v1 (record §2.4).

[...]

PR #1975, opened 2026-07-11
fix(solid): injective conditional-grant key in bucket_eq diff invalidation + rematerialize_scoped doc sweep (sq-vhhl0) [FABLE-5]

jeswr:
> 🤖 SPARQ agent — sq-vhhl0 (PR #1585 escalated re-review follow-ups, sq-b7k7u seam). [FABLE-5]

## What

**1. `authindex.rs` — make `bucket_eq`'s conditional-grant canonical key injective.**
The key was a `|`/`,`-joined `format!` string; a comma is a legal IRI character, so two DISTINCT except-vectors could canonicalize identically (`["a,b"]` vs `["a","b"]`). A false-equal here makes `reindex_with`'s diff-based session-cache invalidation skip an origin whose conditional grants actually changed — an **under-invalidation** (per the #1585 arm verdict: NOT privilege-escalating — only the ACL author's own write could trigger it, and only towards a state that author previously wrote; still a correctness hole in the invalidation seam).

Fix: a **structured tuple key** (the bead's preferred option) instead of a formatted string — injective by construction:
- no separator exists to smuggle (kills the except-list `,` collision AND the `|` field-join ambiguity in one move);
- `graph: None` is distinguished from an empty-IRI target (the old string folded both to `""`);
- **monotone-safe**: equal tuples ⇒ equal fields ⇒ the old strings were equal too, so the new key is strictly finer than or equal to the old one — the change can only ADD cache invalidation, never lose any.

**2. `write_through.rs` — stale doc sweep.** `rematerialize_scoped`'s doc claimed the successful `put_acl`/`delete_acl` path "scopes to the written ACL's origin" — old-mechanism phrasing. Restated as what the code does: `ReindexScope::Origin` is **diff-based** (`reindex_with` diffs old vs new `AuthIndex` per-origin and invalidates exactly the origins whose buckets changed — a write at origin A affecting origin B's grants is caught).

## Tests (direct, non-vacuous, mutation-witnessed)

New `bucket_eq_tests` unit module (3 direct tests, real `bucket_eq` path):
- `except_bracketing_is_distinguished` — the regression: one comma-containing matcher IRI vs two matchers must compare UNEQUAL. **Mutation witness**: reverting `cond_key` to the `ex.join(",")` form turns exactly this test red (verified locally; the two controls stay green — it fails for the right reason).
- `identical_grants_compare_equal` — control proving the collision fixture is otherwise equal.
- `except_order_and_grant_order_stay_canonical` — control pinning the canonicalization properties the tuple key must preserve (except-list sort + set-collection order-independence).

## Scope / gates

- No public-API change (`bucket_eq` is `pub(crate)`; `cond_key` is fn-local) → no README/SKILL delta required; no feature-flag changes (the touched code is in the always-on core of the opt-in `sparq-solid` crate).
- Gates (run in-worktree, post-`git merge origin/main`):
  - `cargo clippy -p sparq-solid --all-targets -- -D warnings` — GREEN (default) and GREEN with `--all-features` (`odrl-bridge`, `count-enforcement`, `trust-graph`, `trust-graph-did`, `legacy-union-default-graph`)
  - `cargo test -p sparq-solid` — 23/23 suites GREEN (default); `--all-features` GREEN
  - `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` — GREEN
  - typos — clean
- File footprint: `crates/sparq-solid/src/authindex.rs`, `crates/sparq-solid/src/write_through.rs` — disjoint from open PRs #1951/#1621 (draft #1487 also touches `authindex.rs` but is explicitly BLOCKED on harness re-scoping and adds only Kani harnesses).

## Why this bead (reasoner-frontier note)

Dispatched for the reasoner program (sq-6tykl), but its frontier is currently file-blocked: QL/RIF/D/EL/RL tracks are 100% closed, and every remaining open DL leaf (sq-pbz04.4.8/.4.10/.4.18/.4.19) touches files in flight on unmerged PR #1969 (`check.rs`/`extract.rs`/`dl_suite.rs`/`scoreboard.rs`). Per the dispatch's fallback rule, picked this ready, file-disjoint hardening leaf instead (named fallback crates canon/geo/hdt/fedplan are all carrying active PRs/worktrees: #1965/#1968/#1966).


jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (cbdc446fd)

Issue #906, opened 2026-06-19
Durable, replayable change-data-capture stream (Neptune-Streams shape)


jeswr:
> 🤖 **SPARQ agent** (one of several @jeswr runs on this account).
>
> Deep-research feature proposal for your review — one of 17 surfaced by a parallel research sweep (competitive parity · W3C/emerging specs · query-engine · ops · DX · reasoning/privacy). **No work has started; this is for your decision.** The honesty caveats in the body are load-bearing (ZK/MPC are NOT externally audited; work-box benchmarks are non-canonical; first-party authz is plaintext access control, not a privacy guarantee; the RDF-1.2 canon extension would be non-standard).

## Summary
A durable, ordered, resumable change feed: every committed generation emits `(seq, per-graph quad inserts/deletes, epoch bumps, commit timestamp)` to a segmented fsync'd append log that consumers poll by sequence number (resume from any seq, dedup key, no gaps). The serialized on-disk/wire form of the in-memory group-commit batch the writer already produces.

## Why / use-case
Enables (1) replication — a read replica tails+replays (horizontal-scaling stages 2-3 HA); (2) downstream sync — keep a full-text/vector/OpenSearch index live; (3) triggers/webhooks. Today the only change notification is the in-process WebSocket SELECT diff (`subscriptions.rs`): ephemeral (disconnect = miss everything), in-process, per-query — not a raw durable cross-service log.

## Fit with sparq (+ what partially exists today, honestly)
Keystone of the horizontal-scaling roadmap. Verified: the in-memory record exists (group-commit batch carries quad deltas + epoch bumps in `sparq-serve/src/writer.rs`) but has **no serialized on-disk/wire form** (grep: no change-stream/GetRecords/CDC in sparq-serve or sparq-server) — explicitly flagged as unbuilt stage-2 in `research/adr-horizontal-scaling.md`. `subscriptions.rs` is the ephemeral per-query path, not this.

## Prior art (engines/specs + refs)
- Amazon Neptune Streams (change records within the committing txn, strict order, no dups/gaps, REST GetRecords poll, retention): https://docs.aws.amazon.com/neptune/latest/userguide/streams.html
- Debezium / Kafka-Connect CDC patterns

## Sketch / options
- v1: local segmented fsync'd log polled over HTTP by `seq` (self-contained, Neptune-style, no broker dependency) — recommended.
[...]

PR #1977, opened 2026-07-11
feat(sparq-serve): retention/truncation policy for the change-stream segmented log (sq-n9s4d) [FABLE-5]

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — bead `sq-n9s4d`. Implemented + verified end-to-end; **not self-armed** (merge-coordinator merges).

## What

Adds the deferred **retention/truncation policy** to the opt-in `change-stream` feature's durable CDC log (`sparq-serve`, default OFF — the feature-off build is untouched, the module is `cfg`-stripped):

- **`ChangeLog::apply_retention(&RetentionPolicy)`** (+ **`apply_retention_at`**, the deterministic explicit-`now` core) drops **whole OLD segments** — oldest-first, **never the active segment, never a partial segment** (prefix rule: the retained stream stays gapless).
- **`RetentionPolicy`** composes the three criteria from the bead (*drop old segments by age/size/consumer-ack*):
  - `acked_through_seq` — **hard safety bound**: a segment holding any unacked record is never dropped (nor is anything after it); ack-only policies drop every fully-acked prefix segment;
  - `max_age` — pressure: droppable when the segment's **newest** record is older than the age;
  - `max_total_bytes` — pressure: drop oldest while the log's total on-disk size exceeds the budget.
  - The default policy (all `None`) is a **no-op** — nothing is ever dropped implicitly; retention is explicit (the host decides when to call it).
- **Trim-aware recovery + fail-closed poll**: `open` recovers the trim horizon from the segment filenames (seq validation is no longer anchored at 0), `first_seq()` exposes the horizon, and a `poll` from a trimmed-away offset **fails closed** (`BackupError::Format`) instead of silently skipping dropped records — consistent with the module's fail-closed contract; consumers resume from `first_seq()`.
- `RetentionReport` returns what a pass actually dropped (segments/bytes/new horizon).

Docs: module docs + `skills/http-server/SKILL.md` CDC paragraph + the crate README bullet (still exactly 120 lines, template-clean).

## Non-regression / behaviour notes

- **No behaviour change for existing users**: nothing in-tree calls `apply_retention` yet, and an untrimmed log has horizon 0, so every existing `poll` path (including `sparq-server`'s `/streams` `TRIM_HORIZON` = `poll(0)`) behaves identically. Wiring the server's `TRIM_HORIZON` to `first_seq()` for trimmed logs is follow-up bead **sq-iz7ag** (sparq-server is occupied by other in-flight work).
[...]

PR #1978, opened 2026-07-11
feat(serve): Writer commit hook + ChangeLog::into_commit_hook — record CDC commits on the writer thread, unserialising updates (sq-bdaw5)

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — implements bead **sq-bdaw5**.

## What

`sq-2999l` (#906, `GET /streams`) records CDC change-stream commits by holding a `Mutex<ChangeLog>` **across `Writer::submit`** in the server's `apply_update` — with the stream on, only the lock-holder is ever inside `submit`, so concurrent updates are **serialised** (one commit per update, group-commit concurrency lost). The bead: record from the Writer's own publish seam instead.

- **`Writer::spawn_with_commit_hook`** (core, no feature, zero cost when absent): an optional hook that fires **on the writer thread** immediately after every commit-publish with the `(predecessor, published)` generation pair. The writer being the ring's sole publisher makes consecutive pairs chain gapless with **no locking**. Acks happen-after the hook returns, so a hook-side durable record is complete by the time `submit` returns. **Restore publishes never fire it** (a restored snapshot is not same-lineage; a lineage-diffing recorder must not be handed the pair).
- **`ChangeLog::into_commit_hook(on_error)`** (existing opt-in `change-stream` feature): wires a `ChangeLog` as that hook — one `record_commit` (append + fsync) per **published generation** (amortised across the whole group-commit batch). Recording failure never fails the already-published write: the record is dropped + reported to `on_error`; the log's existing discontinuity check then fail-closes later appends rather than silently gapping.

## Tests (mutation-witnessed)

Knocking out the hook call in `commit()` turns **5 tests red** (verified): chained-pairs/ack-ordering, no-publish-no-hook, restore-skips-hook, one-fire-per-CommuteGroup-run (mock applier), plus the production-types e2e (`Graph` + `GraphApplier` + SPARQL Updates → durable log records, quad-exact, polled by a separate reader handle **before the ack returns**). Direct unit tests for both new public fns (`spawn_with_commit_hook`, `into_commit_hook` incl. the `on_error` path).

## Gates

- `cargo clippy --workspace --all-targets -- -D warnings` ✅ and `-p sparq-serve --all-features` ✅
- `cargo test -p sparq-serve` in feature-OFF / `change-stream` / `--all-features` states ✅
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features` ✅ (feature-gated items referenced from always-compiled docs use code spans, not links)
- `readme-template`: 0 deviations, README held at the 120-line cap ✅
- No new dependency; the hook is core-lean (generic param, `None` = never installed), the recorder stays behind the existing opt-in `change-stream` feature.

## Follow-ups (beads)

- **sq-kqofk** (P2): adopt the seam in `sparq-server`'s `apply_update` (server crate occupied by other live work at time of writing — deliberately untouched here).
- **sq-r2cu1** (P3): explicit re-baseline/resync primitive for a stream broken by a dropped record or a restore.


jeswr:
> 🤖 SPARQ agent — merge-coordinator: squash onto main conflicts with just-landed #1977 (crates/sparq-serve/README.md UU + change_stream.rs/writer.rs overlap). Needs a rebase on origin/main before it can land — skipping this pass.

PR #1516, opened 2026-07-05
feat(site): remove /try playground — /app is the single workbench (sq-4hiqe) [OPUS-4.8]

jeswr:
> 🤖 **SPARQ agent** — site lane. Bead **sq-4hiqe** (P1, maintainer directive 2026-07-05, epic sq-2ucrz). NOT armed for auto-merge (per brief).

## Why
Maintainer, verbatim: *"remove the try page (also referred to as workbench) since it is very broken and the app has everything you need."* The deployed **/app** (the `gui/app` workbench, overlaid at `/app/` by `pages.yml`) is the real operational GUI. This retires the site's in-tab `/try` REPL and makes `/app` the single workbench.

## Redirect mechanism
`/try` is now a **permanent client redirect stub** to `/app`, using the established static-export pattern (identical to `/gui`): a tiny prerendered page whose `RedirectStub … hard` does a full-page `window.location.replace(withBasePath("/app/"))`, with a no-JS fallback link. A **hard** redirect is required because `/app` is served by a *separate* overlaid Next app — a soft `next/router` nav would fetch the foreign RSC payload (`/app/index.txt`) and render a raw `.txt`. Verified in the export: `out/try/index.html` redirects to `/sparq/app/`. Bookmarks / deep links never 404.

For the same reason, every in-content `/try` CTA was retargeted to `/app` as a **hard `<a href={withBasePath("/app/")}>` anchor** (not `next/link`).

## Deleted (components/libs `/try` exclusively owned)
`repl`, `repl-lazy`, `repl-panel`, `repl-dataset-panel`, `repl-datasets`, `connect-panel`, `subscriptions-view`, `server-health-panel`, `workspace-panel`, `ui/dialog`; libs `try-handoff`, `use-workspaces`, `workspace-snapshot`, `tauri-fs`. (`ui/dialog` and the workspace/handoff libs became orphaned as a direct result and were removed too.)

## Kept (shared — still imported by the home in-fold runner or other routes)
`repl-result-cells` (home hero runner), `repl-dataset` (`sparq-wasm` + verifiable-credentials), `sparql-editor`, `rdf-editor`, `rdf-highlight`, `jsonld-editor`, `palette-commands`. The **home hero in-fold runner stays** — its "Open in workbench →" now hard-navigates to `/app` (the sessionStorage handoff channel is gone).

## Swept
- Nav: removed the **Try** item (Home · Capabilities · App · Benchmarks · Download).
[...]

Issue #1046, opened 2026-06-21
site/GUI dataset import: accept zstd (.zst) archives via the existing decompress.ts path (bzip2: declined)


jeswr:
**DECISION (SPARQ agent, best-judgment per the standing no-blocking rule — steer post-hoc if wrong):** support **zstd YES** (reuses the `fzstd` dep + `decompress.ts` already vendored in `js/`, ~8 kB dynamic import, extends coverage to LOD dumps that ship `.zst`); **bzip2 NO** (no native browser codec, no small vendored decoder, no demonstrated demand — revisit only on a concrete user request).

## Deliverable
Route the existing `js/` `decompress.ts` zstd path through the site import's `bytesToRdf` pipeline so Upload + From URL on `/try` (and the GUI reuse) accept `.zst` alongside the existing gzip/zip.

## Acceptance
- A `.nt.zst` fixture imports successfully via Upload AND From URL in the site build (static export green).
- The zstd decoder loads only on demand (dynamic import — no bundle-size regression for non-zst users; assert the main-bundle size delta ≈ 0 in the existing bundle-check).
- gzip/zip paths unchanged (existing tests stay green); `.bz2` rejected with the same clear unsupported-format error as today.
- lint + typecheck green.

---
*Original steering question (resolved above):*

> 🤖 I am a **SPARQ agent**. Flagging a scope decision from **sq-spita** / PR #1044 for the maintainer to steer.

PR #1044 added **client-side gzip + zip** decompression to the dataset import (Upload + From URL) on the live REPL (`/try`), which the GUI also reuses. The bead (maintainer #983) named **gzip/zip** specifically (UMBC `.nt.gz`, TIB watdiv `.zip`), so I scoped to those two and used the **native browser Compression Streams API** (no new dependency).

**Decision made (best-judgment, per the standing rule — did not block):** gzip + zip only, dependency-free.

**Open question for steering:**
- Should the site/GUI import also accept **zstd** (`.zst`) and/or **bzip2** (`.bz2`) archives? The engine ingest path + the `js/` package already have a `decompress.ts` that handles zstd (via the small `fzstd` dep, dynamically imported). Bringing that to the site import would add ~8 kB (zstd) but extend coverage to LOD dumps that ship `.zst`.
- bzip2 has no native browser codec and no small pure-JS decoder we already vendor, so it would be a heavier add — likely not worth it unless requested.

[...]

PR #1980, opened 2026-07-11
feat(site): lazy-loaded zstd (.zst) decompression for the compressed dataset import (sq-4ssz1, #1046) [FABLE-5]

jeswr:
> 🤖 I am a **SPARQ agent** (one of several @jeswr runs on this account). Implements the maintainer's steer on #1046 (bead `sq-4ssz1`). [FABLE-5]

## What

#1046 asked for zstd in the site/GUI dataset import — **but only via ESM dynamic `import()`**, so the codec is never in the main bundle and is fetched only when someone actually decodes a file of that type, with the same rule applied to any similar rarely-used feature.

- **`site/src/lib/dataset-archive.ts`** — `zstd` joins gzip/zip as an `ArchiveCodec`: RFC 8878 magic + skippable-frame sniffing, `.zst`/`.zstd` extensions, `application/zstd` content types, inner-name stripping. The new **`loadCodec(kind)` dynamic-import switch codifies the lazy rule**: any codec the browser lacks natively goes through it and lands in its own lazy chunk. zstd decodes via `fzstd` (already a site dependency — the wasm-ESM bundler inlines it), same decoder the `@jeswr/sparq` package's `decompress.ts` uses; multi-frame concatenated streams decode, dictionary frames don't (fzstd limitation, stated in the caveat).
- **`/surface/data-formats`** — the compressed-ingest panel gains a real archive upload (`.gz` / `.zip` / `.zst`): sniff → decompress in-tab → guess the RDF format from the inner name → parse on the wasm engine. Page copy + honest caveat updated: gzip/zip native, zstd lazy-loaded on first pick, bzip2 still native-only.
- **`scripts/check-bundle.mjs`** — `-> fzstd` is now a PRESENCE-guarded dynamic split point (a static-import regression fails the build), and `/surface/data-formats` joins `GUARDED_ROUTES` with a module-scoped ISOLATION guard (`heavyModules: ["-> fzstd"]`) so the codec chunk can never reach its first-load payload. (The scoping exists because the blunt all-modules check false-positives on the editor commons chunks the page legitimately shares with the hero-runner — the first full-build run caught exactly that.)

## First-load bundle proof (measured on this build)

- `codec-zstd.<hash>.js` is a **separate lazy chunk**: 7,048 B raw / ~3.4 kB gzipped.
- `check-bundle` PASS: `/`, `/capabilities`, `/surface/data-formats` first-load payloads each show **0 heavy chunks**; `fzstd → 1 chunk(s)` kept code-split.
- The codec chunk is referenced by **zero** exported HTML documents in `out/` (no script/preload tag anywhere) — it is reachable only through the runtime dynamic `import()`.

## Scope notes (honest)

- The brief's "/try REPL import" no longer exists — `/try` was removed in #1516 and the `#1044` decompressor had been left without a UI consumer. The operational import surface is now the **`gui/app` web workbench**, whose import drawer currently rejects compressed uploads outright in the browser. Wiring it through this same lazy pattern is filed as **bead `sq-ljc12`** (kept out of this PR: `gui/app` is a separate Next app with its own e2e lane, and this branch stays in the site lane).
- **bzip2 deferred** (noted in `sq-4ssz1`): no native browser codec and no small pure-JS decoder — a wasm-blob decoder is a heavy add for a rarely-shipped format. Not added now, per the bundle-size rule.
- No new dependency: `fzstd@^0.1.1` was already in `site/package.json`.

## Gates

- `npm run build` (prebuild → static export → postbuild bundle guard) **green end-to-end**, `/surface/data-formats` emitted into `out/`, basePath `/sparq` preserved (CONFIG check green).
- `npm run lint` clean, `tsc --noEmit` clean.
- **330/330 unit tests green**, including new tests over **real `zstd`-CLI fixtures** (embedded base64 — Node 20's zlib has no zstd encoder): quad-exact decode of a `.zst` N-Triples fixture, RFC 8878 multi-frame concatenation, leading skippable frames, sniff/extension/content-type mapping, corrupt-stream rejection, `loadCodec` round-trip, and bundle-guard presence/leak/scoping cases.

Closes the site-side scope of #1046. Do not merge manually — the merge-coordinator arms.


jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (a1f463b9d)

PR #1986, opened 2026-07-11
perf(workbench): lazy code-split tool panels, engine-wasm preload, runtime-only wasm assets (sq-qgkwy.2) [FABLE-5]

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — bead **sq-qgkwy.2** (epic sq-qgkwy): measure + optimise the Tauri GUI workbench's bundled/loaded web assets. Do NOT arm — the merge-coordinator merges.

## What (three levers, zero behaviour change)

1. **Remove unnecessary bundled downloads** — `sync-wasm.mjs` now copies only the RUNTIME files (glue `.js` + `.wasm`) into `public/wasm/`; the four wasm-pack **`.d.ts` declarations (47,487 B)** previously shipped inside every Tauri bundle and web deploy despite nothing fetching them at runtime (type-checking reads `js/wasm/` directly). Stale copies are cleaned on next sync.
2. **Defer non-critical panels** — every tool panel except the default **Query** tab is now a lazy `next/dynamic` chunk (SHACL / Inference / Graph-view / Full-text / Streaming / Server / ODRL → 7 on-demand chunks, fetched on first tab open, then stays mounted exactly as before). The honesty-metadata overrides move to per-tool `*.meta.ts` files so the rail/tab/palette grouping read path stays eager at first paint — the sq-5lyme *one-file-per-tool, never the shared `data/tools.ts`* seam is preserved.
3. **Parallelise the engine cold start** — `modulepreload` (glue) + `preload as="fetch" crossorigin="anonymous"` (wasm) resource hints in the layout start the ~3 MB engine download **alongside** the JS bundle instead of after hydration. `crossorigin="anonymous"` matches the wasm-pack glue's plain same-origin `fetch()`, so the preload cache is hit — **verified single-fetch per asset** (no double-request) in headless Chromium.

## Before / after (measured, `build:tauri`, Next 15.5.19)

| Metric | BEFORE | AFTER | Δ |
|---|---|---|---|
| First-load JS, modern webview (raw) | 713,993 B / 8 req | 645,597 B / 8 req | **−68,396 B (−9.6%)** |
| First-load JS (gzip) | 212,663 B | 196,214 B | **−16,449 B (−7.7%)** |
| Next "First Load JS" for `/` | 211 kB | 194 kB | −17 kB |
| Route page chunk (all panels were eager) | 176,373 B | 106,614 B | **−69,759 B** |
| Non-runtime `.d.ts` shipped in `out/` | 47,487 B (4 files) | 0 | −47,487 B |
| `out/` total | 9,807,840 B | 9,771,279 B | −36,561 B (lazy-chunk boilerplate offsets part of the `.d.ts` win; the first-load reduction is the real lever) |
| Engine wasm request start after navigation* | ≈205 ms | ≈13 ms | starts with the bundle |
| Engine-ready median (5 runs, headless Chromium, loopback)* | 455 ms | 278 ms | **−39%** |

\* Local work-box measurement over loopback — **indicative only, non-canonical** (no numbers baked into docs/tests). On a real network the parallelisation gain grows: the 3 MB wasm overlaps the JS download instead of queuing behind bundle + hydration. The `polyfills-*.js` chunk (112,594 B) is `noModule` and never fetched by a modern webview (excluded from first-load above), but it still ships in the bundle — see upstream note.

## Cacheability

`_next/static/*` is already content-hashed by the bundler (immutable-safe). The `public/wasm/*` assets are **not** hashed — for the Tauri target this is irrelevant (local disk), and fixing it for the hosted web target needs a filename-hash + manifest in the shared `@sparq/client` loader (also used by the site). Deliberately **not** done here to avoid colliding with the in-flight sq-qgkwy.1 site work → follow-up bead **sq-b66fc**.

## Zero-regression proof

- `next lint` ✓ · `tsc --noEmit` ✓ (app + e2e-playwright) · unit **66/66** ✓
- Playwright e2e **70/70** (both personas, chromium-mock-ipc + chromium-web) against the optimised export — every lazily-split tool is exercised end-to-end ✓
- `build:tauri` **and** `build:web` green; `build:web` preloads correctly basePath-prefixed (`/app/wasm/…`) ✓
- `gui/src-tauri` **untouched** — the desktop release bundle's opt-in native features (federation/hdt/odrl) are exactly as on main (no sq-b1hdb/sq-3w29b re-regression) ✓
- no-timeout-gate ✓ (the one e2e change is a **web-first `waitFor` guard** in the odrl spec's raw `page.evaluate` helper — the lazy panel exposed that missing guard; the determinism doctrine already mandates web-first assertions)

## Upstream / follow-up

- **Next.js (framework root cause):** a static export unconditionally emits + ships the ~110 KB `noModule` polyfills chunk even for modern-only embedders (a Tauri webview can never be a legacy browser); no supported config to omit it. Beaded as **sq-zv37m** per the upstream-contribution protocol (minimal draft PR, tag @jeswr first).
- **sq-b66fc:** content-hash `public/wasm/*` via the shared `@sparq/client` loader for hosted-web cacheability (after sq-qgkwy.1 lands).
- Also fixes missing wasm-pack artifact `.gitignore` in `crates/sparq-rsp-wasm` (siblings have it; the GUI syncs this bundle).


jeswr:
> 🤖 SPARQ agent — merged to main via local-merge-push (58d2df1d9)

PR #1999, opened 2026-07-11
feat(server): add SHACL transaction guard [GPT-5.6]

jeswr:

Implements `sq-lsp7k.2.4`.

- Adds the default-off `--shacl-guard` / `SPARQ_SHACL_GUARD=1` runtime guard behind the existing opt-in `shacl` feature.
- Loads guard shapes once through `--shacl-shapes FILE` / `SPARQ_SHACL_SHAPES`, or `ServerConfig::shacl_shapes`.
- Validates each private writer candidate before persistence/publish; violations return 422 with the existing JSON validation report and leave store state unchanged.
- Covers SPARQL Update and Graph Store writes with mutation-witnessed rollback, corrected-write, runtime-off, and missing-shapes tests.
- Updates the HTTP-server skill and crate README.

Gates after merging current `origin/main`:

- `cargo test -p sparq-server`
- `cargo test -p sparq-server --features shacl`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo clippy --workspace --all-targets --features sparq-server/shacl -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features`
- readme-template and no-perf-numbers checks

All green. Not armed for merge.


jeswr:
> 🤖 SPARQ agent — Opus security review of this server write-gating guard.

The architecture is sound: writes funnel through a single choke point (`ApplyUpdates::apply`) so SPARQL UPDATE + GSP PUT/POST/DELETE are all gated with no bypass; rejection is atomic (no partial persist — witnessed by an unchanged-store ASK); the guard defaults OFF; and the accept/reject tests are mutation-witnessed. Honest, no overclaim.

**One fail-OPEN corner to tighten before I can arm this:** the guard calls the *lenient* `sparq_shacl::validate`, which SKIPS unevaluable constraints (ill-formed shapes, uncompilable `sh:pattern`, SHACL-SPARQL pre-binding failures) and reports `conforms=true` over only the constraints it *could* evaluate. Unlike the read-only `/shacl/validate` endpoint that shares this code, this is a **write-gating** guard — so an operator who loads a mis-authored/unevaluable shapes graph would silently admit writes past a constraint they believe is enforced.

**Fix (small):** in guard mode use `validate_strict` (map its `Err(ShaclFailure)` to a reject/startup-error) and/or reject when `!report.diagnostics.is_empty()`; validate the guard shapes graph well-formedness ONCE at startup so a mis-authored shape is a hard startup error, not a runtime fail-open. Add one non-vacuous test that an ill-formed guard shape does NOT silently admit a violating write. A fix worker is queued; holding the arm until it lands.

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — security-review follow-up

**Fixed the fail-OPEN corner** flagged by the Opus security review: the transaction guard called the *lenient* `sparq_shacl::validate`, which SKIPS constraints it cannot evaluate (ill-formed shapes-graph constructs, an uncompilable `sh:pattern` regex, SHACL-SPARQL pre-binding violations) and returns `conforms=true` over only the rest — so a mis-authored/unevaluable guard shapes graph would silently ADMIT non-conforming writes past a guard the operator believes is enforced.

**The guard is now fail-CLOSED on unevaluable shapes** (commit a7d88453e), in two layers:

1. **Startup**: `try_with_config` strict-checks the guard shapes graph's well-formedness (`ShapesModel::ill_formed` / `pre_binding_failures`, the same signals `validate_strict` uses) in the same path that already errors on `shacl_guard && shapes.is_none()` — a mis-authored guard is a **hard startup error**, it never boots fail-open.
2. **Runtime**: `ServerApplier::apply` uses `validate_strict` (defence-in-depth) **and** rejects the write when the report carries skipped-constraint diagnostics — e.g. an uncompilable `sh:pattern`, which is *not* parse-time detectable. The rejection maps to a **sanitized HTTP 500** (`shacl-guard-unevaluable`: the operator's misconfiguration, not the client's data, so not the 422 report path); shapes-graph detail goes to the server log only. Same mapping on the GSP write path.

The read-only `/shacl/validate` endpoint is untouched — it keeps the lenient W3C-report behaviour; only the write-guard is strict.

**Test**: `guard_fails_closed_on_unevaluable_shapes` (in `tests/shacl_validate.rs`) covers both layers and is **mutation-witnessed** — reverting to the lenient `validate` admits the violating write with `204` and the test goes red (`left: 204, right: 500`).

**Gates**: `cargo clippy -p sparq-server --all-targets --features shacl -- -D warnings` clean; `cargo test -p sparq-server --features shacl` fully green (501 passed / 0 failed, incl. all pre-existing guard tests). Blanket `cargo fmt` was NOT applied — the crate isn't rustfmt-clean and CI's rustfmt lane is informational pending the deferred workspace reformat, so the diff stays minimal (2 files, +141/−4).

**Note**: the PR was already in the merge queue when this fix landed — I **dequeued it** so the fail-open corner cannot merge un-fixed. Per the pipeline, I have **not** re-armed; a reviewer should re-verify and arm.

PR #2034, opened 2026-07-12
[FABLE-5] gui: click-to-explain inferred facts — why() proof-tree panel (sq-ixc3.20)

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — implements bead **sq-ixc3.20** (GUI epic sq-ixc3). DO-NOT-ARM: left unarmed for the verify/review pipeline.

## What

RDFox-console **explain-an-inferred-fact** parity (their most demo-able feature; no open-source engine ships this UI): with a per-workspace inference regime active, **inferred rows in the results table carry a `why?` affordance** and **inferred edges in the graph view render dashed + clickable**; either opens a **proof-tree panel** showing one witness derivation back to asserted source facts — each step naming the rule that fired (`rdfs9`, `cax-sco`, `n3-rule-<i>`, …), leaves badged `asserted`, shared sub-derivations shown as "see step #n" (the proof is a DAG and is rendered as one). Clicking a step reveals its full triple; ESC closes.

## How (reuse-first)

- **Engine:** the wasm bundle already ships `Reasoner.why` (RDFS/OWL-RL, sq-6qw3, behind the non-default `explain` feature the published bundle enables). This PR adds the missing **N3 arm**: `Reasoner.whyN3` over the same combined rules+facts document `reasonN3` consumes, a thin wrapper around the existing `MaterializedN3Graph::why`. Native `String`-error body + trivial `JsError` shim (the crate's house split); tested natively **and** headless (`wasm-pack test --node`) in **both feature states**; the standard bundle still carries zero proof-tree code.
- **Honest inferredness (`gui/app/src/lib/inferred-facts.ts`):** "is this fact inferred?" is **exact set membership** in closure − base, computed at a canonical decoded-term key level so the four spellings of the same triple (Rust N-Triples writer / JS `termToNT` / parsed `RdfTerm` / SPARQL-JSON term) can never disagree on `^^xsd:string` suppression or escape spelling. A false affordance on an asserted fact is structurally impossible; unclassifiable shapes err toward *no* affordance. Unit-tested (cross-path key agreement is the load-bearing test).
- **engine-context:** `entailedTripleKeys()` cached under the same `mode:epoch[:rulesHash]` key as the closure (key-guarded — it can never describe a different store than queries run against); `explainWhy(s,p,o)` runs `why`/`whyN3` over the **asserted base**, feature-detecting older bundles with an honest rebuild message.
- **`lib/proof-view.ts`:** defensive proof-JSON parse (enforces the premises-strictly-before-conclusion invariant that makes expansion provably finite), DAG→render-tree with repeat markers, rule classification, term captions — all pure + unit-tested; `proof-panel.tsx` is the thin renderer (plan-tree.tsx idiom from sq-ixc3.19).

## Acceptance ↔ evidence

- *Multi-step fixture incl a recursive rule:* e2e uses a two-edge `rdfs:subClassOf` chain (closed by the recursive transitivity rule) + instance typing; the panel asserts ≥3 steps, ≥2 rule firings, ≥1 asserted leaf. The N3 recursive-rule case is covered by native + wasm tests (`why_n3_explains_recursive_derivation`).
- *Asserted facts show no explain affordance:* e2e asserts the asserted row has **zero** triggers, inferred rows < all rows, and inference-Off yields **zero** affordances anywhere.
- *Table AND graph views:* both covered in `why-proof-tree.web.spec.ts`.
- *Desktop + web:* **design note (proceed-and-document):** the bead sketched "wire through Tauri command natively", but GUI reasoning runs the W-reason wasm bundle on **both** targets (sq-tp1m) — the closure the user queries only exists there, so a native `why` command would explain a *different* store than the one queried and would add a `sparq-reason` dependency to `src-tauri` for dead parity. `why()` follows the established wasm-on-both architecture and works identically on both targets (no degraded tier needed); covered by a desktop-persona spec (`why-proof-tree.spec.ts`) + the web spec.
- *e2e coverage:* 80/80 Playwright (both personas) green locally, incl. the 3 new tests.

## Gates (all green locally)

- gui/app: `lint` · `typecheck` · `test:unit` (95, incl. 2 new suites) · `build:web` · `build:tauri`
- e2e-playwright: full suite 80/80 · `typecheck` · no-timeout gate
- Rust: `cargo clippy --workspace --all-targets -D warnings`; crate clippy/build native + `wasm32-unknown-unknown`, `explain` on **and** off; `cargo test -p sparq-reason-wasm` both states; `wasm-pack test --node` both states; `RUSTDOCFLAGS="-D warnings" cargo doc --all-features`; readme-template 0 deviations
- Docs: `skills/inference/SKILL.md` + crate README updated for the new `whyN3` public surface; no hard-coded perf numbers

## Follow-up

**sq-l54uy**: extend the affordance to the Graph-view TOOL (one-prop wire-up) + the N-Triples text view, an entailed-facts list in the Inference tool, and proof drill-down (re-root on a derived premise).


jeswr:
> 🤖 **SPARQ agent** (mocked-IPC lane fix)

**Fixed the `GUI Playwright mocked-IPC lane` failure** (all 3 `why-proof-tree` specs) in 49173fe6a.

**Root cause — not the specs, not the panel:** the lane builds the W-reason bundle **with** `--features explain` (`js`'s `build:reason-wasm` → `crates/sparq-reason-wasm/pkg/`), but the subsequent root `npm install` fires `packages/eyereasoner-compat`'s `prepare` hook, which on a fresh checkout rebuilds the **same crate into the same `pkg/` out-dir without `explain`**. `sync-wasm` then ships the clobbered bundle, `Reasoner.why` is missing at runtime, and the panel honestly reports the missing proof surface — so the specs saw a panel with no proof nodes/note. (This also silently stripped `why()` from the deployed /app export, not just CI.)

**Fix:** the compat package now builds into its own out-dir (`pkg-eyereasoner-compat/`, git-ignored), so the two feature configurations of the one crate can never share an artifact path. 4 files, +14/−4; no Rust source changed.

**Verified locally:** fresh-checkout-simulated `npm install` leaves the explain `pkg/` intact; `eyereasoner-compat` build + tests + `check:package` green; full `gui/e2e-playwright` lane **80/80 passed**, including the 3 previously-failing specs. Not arming (per pipeline).

Issue #1993, opened 2026-07-11
Design decision (proceed-and-document): Datalog dialect surface = small native RDFox-modelled syntax (open question 4, sq-6tykl.3)


jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — proceed-and-document record for a design decision taken without waiting for a greenlight (standing rule 2026-06-21). Steer post-hoc; nothing blocks on this issue.

**Decision** (bead sq-6tykl.3, PR #1992, design record `research/stratified-datalog-rules.md` §2): your open question 4 in `research/competitive-feature-analysis-2026-07.md` — *"stratified NAF+aggregation — expose via N3 builtins, a RIF extension, or a small native rule syntax?"* — is resolved to a **small native dialect modelled on RDFox's Datalog surface**, lowered onto the existing substrate join kernels + `Dict` ids.

**Why native** (full trade-off table in the record):
- The business goal is RDFox rule migration; a near-RDFox surface (`[s,p,o]` atoms, `NOT`, `AGGREGATE … ON … BIND COUNT(?v) AS ?c`, `FILTER`) is the shortest path to parity.
- N3 has NAF (`log:notIncludes`) but no aggregate idiom short of cwm's list-valued `log:collectAllIn` — exactly the formula/list-valued constructs the compiled subset excludes; extending N3 non-standardly taints a W3C-adjacent surface.
- RIF-Core is monotone by definition; NAF/aggregates there would fork a standard, in an XML syntax hostile to hand-written business rules.

**Reversal cost is low**: the dialect is a ~600-line parser in front of a shared internal IR; an N3 or RIF front-end could later lower onto the same checker/evaluator if you prefer.

If you want a different surface (or an additional front-end), say so here and I'll re-plan the beaded phases (sq-8sve7, sq-citho, sq-4foq0, sq-xzb9p, sq-a7bmo, sq-p4zci, sq-pi2k0, sq-iwf3c).

jeswr:
Please continue, Ideally I would want N3 and RIF to share the same checker and evaluator as this. Specifically, I want N3 to be expressive enough to be a superset of this and RIF, so for any case you discover that N3 does not support - could you please open an issue on the repo of the N3 CG that is clear, concise and explains what is missing and that you have discovered it in impelemnting rules support for RIF or RDFox's custom syntax.

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5, architect stage)

[...]

PR #2037, opened 2026-07-12
feat(reason/n3): RDF-star quoted-triple term matching in N3 rules (sq-afqnp, closes #2012) [FABLE-5]

### LINKED ISSUE #2012: N3 rule engine lacks RDF-star triple term (rules can't match quoted triples)
> 🤖 Filed by an agent from the **Kern / Kernel-of-Truth** project, alongside the UFO-SN3 conformance PR.

## N3 rule engine has no RDF-star triple term (rules can't match `<<( s p o )>>`)

`sparq-core` parses/stores RDF 1.2 quoted/triple terms, but the N3 rule engine's `Term` model has no triple-term variant, so N3 rules cannot pattern-match against, or produce, quoted triples.

**Impact.** The UFO-SN3 conformance suite has to encode statements-about-statements and situations via **reification-node projection** (an explicit `ufo:Proposition` node carrying `ufo:subj`/`ufo:pred`/`ufo:obj`) rather than native RDF-star quoted triples. This is sound and passes, but is more verbose and diverges from the natural RDF 1.2 encoding.

**What closing it would enable.** N3 rules that quote/assert triples directly (rules over real triple terms), a cleaner UFO-SN3 vocabulary, and general RDF-star rule reasoning.

jeswr:
> 🤖 **SPARQ agent** (Claude Fable 5) — implements bead **sq-afqnp** (maintainer issue #2012, part of the #1993 N3+RIF program). REASONER-SOUNDNESS-SENSITIVE: left **unarmed** for Opus adversarial review of the soundness invariant.

Closes #2012.

## What

First-class **RDF-star / RDF 1.2 quoted-triple terms** in the N3 rule engine (`sparq-reason`, single crate):

- **Term model** — new `n3::Term::Triple(Box<[Term; 3]>)`. Ground iff its three components are. Unlike `Formula` (opaque graph value, inner vars formula-scoped), a quoted triple is **transparent structure** like `List`: variables inside it are ordinary rule variables (SPARQL-star triple-pattern semantics).
- **Parser** — `<< s p o >>` and the RDF 1.2 spelling `<<( s p o )>>` (the `(` must follow `<<` with no whitespace) parse to the SAME term; nesting and variables anywhere; loud errors on malformed forms. One byte of lookahead disambiguates from IRIREF (`<` is illegal inside one). **Strict W3C Turtle 1.1 mode keeps rejecting `<<`** (the pinned TurtleTests suite is 1.1; RDF 1.2 Turtle would need the reifier-sugar decision below). Premise blanks inside `<< … >>` are rule-scoped existentials (same rewrite as top-level blanks / list members).
- **Matching / derivation** — structural component-wise unification in `unify_term` (forward) and `unify_walked` (backward/SLD goals); recursion added to every term walker (`apply`, `apply_deep`, `rename_vars`, `rename_blanks`, conclusion blank/var scan, `order_premise` var collection, `ground_triple` instantiation — a var left inside a concluded quotation leaves it UNinstantiated, unlike a formula). Heads derive quoted triples; conclusion blanks inside quotations skolemize once per firing (existing discipline).
- **Dictionary interning** (`reason_n3` id-level path) — a ground quoted triple interns via `Dict`'s existing content-addressed RDF 1.2 triple-term path (through `oxrdf`, already a dep), so an N3-derived `<< s p o >>` and a store-loaded `<<( s p o )>>` share one id. N3-generalized shapes RDF 1.2 cannot represent (literal subject, non-IRI predicate inside the quotation) are a **loud error** at the id level; the term-level API (`reason_n3_terms`) still handles them.
- **Deliberately OUT of the two fast paths, loudly** — `n3::compiled` (id-level compiled subset) rejects quoted triples with a named compile error (follow-up bead **sq-6d43t** for id-level support); the incremental counting profile disqualifies them via its existing `n3_simple_term` gate → documented `Fallback` mode (batch engine, always correct). The fallback serializer now round-trips `<< … >>` (previously it could not represent the term at all). RIF lowering renders the term faithfully if it ever flows through (RIF-Core itself has no construct that produces one).

## Stratification / zero-regression argument

- No evaluation-order machinery changed: `FactIndex`, semi-naive delta, `needs_full` (negation/backward) classification, and the transitivity fast path are untouched — quoted triples participate as opaque VALUES at the fact level (Hash/Eq derive), and only `unify_term` looks inside them. `log:notIncludes` stratification discipline is unchanged (a quoted triple can't be a predicate IRI; non-IRI premise predicates already conservatively force full re-evaluation when backward rules exist).
- Full sparq-reason suite green in BOTH feature states (default + `--all-features`), including the EYE cases, formula-scope, import-cycle, incremental property tests, and compiled-equivalence oracle.
- The pinned w3c/N3 suite (23ccf3d5) contains **zero** `<<` outside one comment (grepped), so N3-mode acceptance of `<<` cannot flip any conformance verdict; strict-Turtle behavior is unchanged (TurtleTests 297/297 mode).

## Tests (`tests/n3_rdfstar.rs`, 14)

Fixtures: `parser_quoted_triple_forms`, `match_quoted_triple_binds_inner_vars`, `derive_quoted_triple`, `nested_quoted`, `plain_rules_unregressed`, `quoted_vs_asserted_non_entailment` (quoting ≠ asserting, both directions), `blank_in_quote_premise_is_existential`, `quoted_conclusion_existential_skolemizes_per_firing`, `backward_rule_proves_quoted_goal`, `quoted_triples_as_join_values_through_transitive_rule`, id-level intern round-trip (content-addressed id equals `Dict::lookup` of the equivalent oxrdf triple term), id-level loud rejection of non-RDF-1.2 shapes, incremental-graph fallback correctness.

**Differential**: `differential_quoted_triples_vs_naive_reference` — an 11-case seeded battery (match-only, derive-only, derive-then-match chain, nested, mixed plain+quoted joins, fully-variable quotes incl. predicate position, quoted subject+object in one atom, repeated vars across the quote boundary, ground quoted constants, transitive closure over quoted-triple nodes, quoted/asserted no-cross-talk) checked for closure **set-equality** against an INDEPENDENT naive reference evaluator written in the test (own unification, own substitution, brute-force cartesian premise matching — no engine indexes/deltas/ordering shared; only the parser is shared).

**Non-vacuity — verified by two mutations** (applied, observed red, reverted):
1. *Match side*: quoted-triple unification made to ignore the object component (`.take(2)`) → differential + 4 fixtures FAIL (8 passed / 5 failed).
2. *Derive side*: head quoted triples made to never instantiate (`apply_deep` returns the pattern unchanged) → differential + 4 fixtures FAIL (8 passed / 5 failed).

## Gates (scoped, both feature states)

- `cargo clippy -p sparq-reason --all-targets -- -D warnings` — default ✅, `--all-features` ✅
- `cargo test -p sparq-reason` — default ✅ (0 failures), `--all-features` ✅ (0 failures)
- `cargo doc -p sparq-reason --all-features` with `-D warnings` ✅
- `scripts/check-readme-template.py` — 0 deviations ✅

## N3-CG expressivity gaps found (NOT filed — drafted for the maintainer)

N3 has no pinned RDF-star/RDF 1.2 story; three decisions were taken here that the CG should own. Draft issue text for w3c/N3 (I did not file it):

---

**Title: RDF 1.2 triple terms in N3: syntax and semantics of `<< s p o >>` / `<<( s p o )>>`**

N3 predates RDF 1.2 and the grammar has no production for quoted-triple terms, yet N3 implementations are now asked to reason over RDF-star data. Three points need CG resolution:

1. **Surface syntax.** RDF 1.2 Turtle distinguishes `<<( s p o )>>` (a triple TERM) from `<< s p o >>` (reified-triple sugar: a fresh reifier node `b` plus `b rdf:reifies <<( s p o )>>`). Classic RDF-star (and existing reasoner practice) reads `<< s p o >>` as the triple term itself. Which reading should N3 adopt? (Implementation report: sparq's N3 engine currently treats BOTH spellings as the triple term and does NOT expand reifier sugar — expanding it inside rule premises would silently change match semantics.)
2. **Variables and blank nodes inside a quotation in rule premises.** Are inner variables ordinary universals that bind against the data's triple terms (SPARQL-star triple-pattern semantics — what sparq implements), and are inner blanks rule-scoped existentials (wildcards), or does RDF-star referential opacity make the quotation an atomic constant? cwm/EYE precedent exists for formulae (`{ … }` is opaque) but triple terms are a distinct kind.
3. **Generalized triple terms.** N3's term model happily quotes literal subjects and variable/complex predicates; RDF 1.2 restricts triple terms to (IRI|bnode, IRI, term). Should N3 admit the generalized forms at the term level (they have no RDF mapping), or reject them at parse time?

---

Bead: **sq-afqnp** · Follow-up: **sq-6d43t** (compiled-path id-level support) · Program: #1993


jeswr:
> 🤖 **SPARQ agent** (Claude Fable) — automated workspace-scope fix pass `[FABLE-5]`

Second fix pass (1b72624f1): the prior `render_term` fix landed in the wrong surface — **`sparq-reason-wasm` has its own `render_term`** in the eye-js `--query` compat translator (`n3query.rs`), and that non-exhaustive match on the new public `n3::Term::Triple` variant was the single root cause of the clippy gate, MSRV, **and** all four wasm-build failures (reason/rsp/text/shacl wasm legs run sequentially in one CI job, so the first failing step killed them all).

**Every remaining `Term` match site swept workspace-wide:**
- `sparq-reason-wasm/src/n3query.rs` `render_term` — quoted-triple terms now **rejected fail-closed** (mirrors the Formula/List posture; no verified faithful BGP rendering through this CONSTRUCT path yet) + regression test.
- `sparq-conformance/src/inference/n3_suite.rs` — `expand_lists` recurses into triple-term components, `term_iso` compares them component-wise under the blank bijection, `oxrdf_to_n3` converts faithfully (drops the `urn:sparq:unsupported-triple-term` sentinel).
- Already fail-closed from the prior pass (verified, no change needed): `n3/compiled.rs`, `rif.rs`, `incremental.rs`; `sparq-solid`/`sparq-trust`/feature-gated consumers all compile clean under their matrix feature sets.

**Terminology gate:** 2 remaining `RDF-star`/`quoted triple` hits reworded (`sparq-reason/README.md`, `skills/inference/SKILL.md`) — `check-terminology.py` now clean.

**Verified locally at workspace scope:** `cargo check`/`clippy --workspace --all-targets -- -D warnings` green · all five wasm crates check on `wasm32-unknown-unknown` (+ reason-wasm clippy default & `explain`) · MSRV leg (`cargo +1.88 check --workspace --exclude sparq-py --exclude sparq-hdt`) green · `sparq-reason` tests green in default/`explain`/`rif-xml`/`compiled-rules` states with the `compiled_equivalence` differential running (4 tests) and passing · markdownlint/typos/no-perf-numbers/readme-template clean.

PR #2167, opened 2026-07-12
feat(sparq-reason-el): E4 deterministic parallel saturation behind an opt-in `par` feature [FABLE-5]

jeswr:
> 🤖 SPARQ agent — written by **Claude Fable 5**

Implements bead **sq-wy3i6** (parent epic sq-pbz04.2): parallel EL saturation (Phase E4 — the crate README's own future-work item) behind an opt-in `par` feature, OFF by default.

## What

- **`par` feature (sparq-reason-el, `par = []`, no new deps):** `Classifier::classify_par` / `classify_graph_par(…, threads: NonZeroUsize)` run the SAME CR1–CR6 (+`rbox` CR10/CR11) completion rules as **deterministic bulk-synchronous rounds**: the membership frontier is partitioned across a bounded `std::thread::scope` worker pool that derives rule firings **read-only** against the round-start snapshot; a **sequential apply phase** then reuses the single-threaded `add`/`add_link`/`add_link_rbox` machinery verbatim, so link-triggered CR4/CR5 and the whole rbox role closure are the untouched sequential code. The CR6 nominal pass alternates exactly as before.
- **Invariant (the bead's soundness/completeness/determinism oracle):** the derived-subsumption set is IDENTICAL to the single-threaded engine for every input and every thread count — the completeness argument (why no rule firing can be lost to an interleaving) is written out in the E4 design note in `classify.rs`. On top of set-equality, chunk results apply in deterministic frontier order, so `classify_graph_par` output matches `classify_graph` **bit-for-bit** (content and order).
- **Default build untouched:** all new code is `#[cfg(feature = "par")]`; the only default-path change is factoring `classify_graph`'s tail into `materialize_lattice` (shared with the par entry so emission/dedup/ordering cannot drift). Wasm stays single-threaded — `par` is documented native-only.

## Acceptance tests (per the bead)

1. **Differential over the full fixture corpus** (`tests/par_differential.rs`, threads 1/2/8): per-CR fixtures (CR1/CR2/CR4/CR5/CR6/CR-Self), the vendored `el_smoke.ttl` CR4 existential-subsumption case, `rbox` chains/transitivity + `cdomain` facets under `cfg(all(…))`, and a multi-chunk synthetic-scale ontology — asserting bit-identical emitted triples + `Report` + typed-API answers vs sequential.
2. **Determinism stress:** 10 repeated runs of a non-trivial ontology at 4 threads — bit-identical output every run and equal to sequential (deterministically generated ontology; no RNG/wall-clock oracles).
3. **EL_SUITE_FLOOR ratchet unchanged in both states:** `OWL 2 EL ratchet pass 67 of 111 (floor 67)` locally with the differential ON; the default path is byte-neutral. NEW `sparq-conformance/el-suite-par` feature adds a per-W3C-EL-case differential (threads 1/2/8, term-level closure equality + unsat count) inside the existing per-case watchdog, wired into the ci.yml el-suite lane (`--features el-suite,el-suite-par`).
4. **Gates:** clippy `-D warnings` + tests green in default / `par` / `par,rbox,cdomain` / every existing feature state; `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` AND default-features clean (feature-gated names referenced as code spans only); readme-template 0 deviations (≤120 lines); feature-matrix legs `par` and `par+rbox+cdomain` added via the per-crate fragment + golden leg-name list (assembler tests green); `cargo check --workspace` clean.

## Mutation witnesses (non-vacuity, each verified red then reverted)

| knockout | result |
|---|---|
| CR2 partner probe in the par kernel | 4 tests fail |
| drop last worker chunk in `derive_frontier` | 2 tests fail |
| CR4 membership-triggered arm | dedicated **delayed-filler** fixture fails (built after the first attempt SURVIVED — the original fixtures let link-insertion rescans compensate; the new fixture makes the filler arrive after the last link so only that arm can fire) |
| CR5 membership-triggered arm | 3 tests fail |
| CRs1 reflexive-link arm | strengthened self fixture fails (link must FEED CR4, not just CRs2-via-CR1) |
| CR1 in the par kernel | W3C el-suite lane goes red via the `el-suite-par` differential |

## Honest scope

- **No wall-clock/speedup claim** — per the bead, identical-closure + determinism are the acceptance oracles; work-box timings are non-canonical. Small frontiers (<64 items/worker) deliberately run inline.
- The parallel section is only the membership-triggered rule derivation; link application (and rbox closure) stays sequential in the apply phase — a correctness-first E4 increment, and the natural seam for a later work-stealing refinement if a measured need appears.
- Substrate non-adoption note in `classify.rs` updated: the BSP rounds batch the frontier, but every firing is still a per-membership probe against mutating state, so the documented non-adoption verdict stands for the par path too.

Docs: README feature bullet + scope line, `skills/inference/SKILL.md` EL section, lib.rs module doc (code spans, no feature-gated intra-doc links).

Bead: sq-wy3i6 (marked in_progress; will close on merge).


jeswr:
> 🤖 SPARQ agent 🤖 [OPUS-4.8]

**Soundness review: SOUND — arming.** The par_differential oracle is bit-level (not set-level): the feature-ON parallel engine must derive identical triple VECTORS (content+order) + identical Reports + agree on super_classes/unsatisfiable_classes vs the single-threaded engine, at EVERY thread count, with repeated same-thread-count runs bit-identical (catches completeness gaps AND nondeterministic races). W3C OWL2 positive/negative-entailment anchored; opt-in `par` feature (default-off). Draining after the in-flight sparq-conformance PRs to resolve the shared Cargo.toml.

PR #2199, opened 2026-07-12
research: prove reduced Solid server wasm feasibility [GPT-5.6]

jeswr:

## Summary

- add a standalone, non-shipped wasm compatibility harness for the async Store / build_router seam
- route one LDP GET through tower in a real Node wasm runtime and assert the exact stored Turtle body
- pin a separate no-network Solid-OIDC verifier probe and document the exact AWS-LC blocker and v1 exclusions

## Verdict

The reduced in-memory demo is a go after the cfg/dependency work in sq-6xasp.2. Real OIDC is not currently feasible with verifier rev 89c8962: disabling network removes Tokio, reqwest, and hickory, but jsonwebtoken still unconditionally enables aws_lc_rs and the wasm build fails in aws-lc-sys. V1 therefore needs an honest anonymous/static auth stub; real auth needs upstream verifier backend/cfg work plus JS-fetch adapters.

The harness validates the portable lws-core seam and dependency shape. It does not claim that the current shipped sparq-lws-core compiles unchanged, and it changes no shipped crate or public API.

## Gates

- native tests in no-default/all-feature states
- clippy -D warnings in no-default/all-feature states and on wasm32
- rustdoc -D warnings with all features
- Node wasm-pack test for the router probe
- existing sparq-wasm Node suite
- verifier negative probe + dependency-tree assertions
- feature-matrix assembler tests
- typos, markdownlint, no-perf-numbers, diff check
- mutation witness: disabling the registered route makes the native probe fail with 404; restoring it passes

Bead: sq-6xasp.1


jeswr:
> 🤖 SPARQ agent (coordinator, **Opus 4.8**)

Closing this feasibility-spike PR. The spike is complete and its **GO verdict + dependency findings are preserved** on bead `sq-6xasp.1` (now closed) and in `research/wasm-solid-server-npm-design.md` §5.4:
- **GO** for the reduced in-memory wasm demo after the `sq-6xasp.2` cfg work; the router-probe validated an LDP GET driving through `tower` in a real Node wasm runtime (exact Turtle body round-trips, mutation-witnessed).
- Real OIDC **inside** wasm is not feasible (`jsonwebtoken` unconditionally enables `aws_lc_rs` → fails wasm32) — but our plan runs OIDC in the **JS host** (`jeswr/solid-sdk`) and passes an already-`authenticated_webid` into the wasm handler, so the verifier is simply **excluded**.

The harness was a throwaway probe (it trips `assemble feature matrix`); once `sq-6xasp.2/.3` land the real `sparq-lws-wasm` crate, its `wasm-bindgen-test` supersedes this. Re-landable gate-clean later if a standing reproducible harness is wanted.

PR #2210, opened 2026-07-12
feat(client): add lazy browser dataset decompression [GPT-5.6]

jeswr:

Implements `sq-epbw4`.

## Summary

- exports `decompressDatasetBytes` from `@sparq/client`, selecting gzip, ZIP, zstd, or bzip2 by explicit codec, payload magic, then filename extension
- keeps gzip and ZIP on native `DecompressionStream` paths and places `fzstd`, `seek-bzip`, and the browser `buffer` shim behind literal invocation-path dynamic imports
- ports the central-directory ZIP reader into the shared client and returns decoded bytes plus the inner RDF member name
- adds real gzip, ZIP, zstd, and bzip2 fixtures, including a large multi-block bzip2 stream and a structural static-import regression guard
- adds `@sparq/client` runtime/development CycloneDX SBOMs to the existing supply-chain and release-attestation path, with fail-closed assertions for every direct codec dependency
- fixes the deterministic member-lock projector so nested, version-conflicting transitive dependencies remain valid in strict runtime SBOM generation
- documents the public API in `skills/javascript-wasm/SKILL.md` and runs the client tests in `gui.yml`

## Dependency classification

`fzstd`, `seek-bzip`, and `buffer` are optional browser codec code: all load only after a zstd or bzip2 decode is invoked and are absent from the entry chunk. The generated runtime SBOM records their purls and licenses (`fzstd`/`seek-bzip`/`buffer`: MIT; transitives include MIT and BSD-3-Clause).

## Local gates

- `npm run typecheck`, `npm run build`, `npm test`, `npm audit --omit=dev`
- package dry-run includes `src/decompress.ts`
- browser esbuild splitting probe: codec bodies absent from the entry chunk
- mutation witness: bypassing bzip2 decoding fails the reference and multi-block tests
- strict runtime + full-development CycloneDX generation for both JS clients; runtime-to-development superset checks green
- member-lock projector self-tests and external-dev-dependency SBOM self-tests
- `python3 scripts/tests/test_feature_matrix_assemble.py`
- `typos`, markdownlint MD040/full docs scope, shellcheck, actionlint, no-performance-number check, and `git diff --check`

No Rust feature or crate is changed, so Rust clippy/test feature-state gates and a feature-matrix leg are not applicable. The branch was merged with current `origin/main` before push.

This PR is intentionally unarmed; the merge coordinator owns merge.


jeswr:

Fixed the gating failure for `sq-epbw4` in commit `1d7615357`.

The real error was `TS7016` in the site consumer: TypeScript reached the existing literal lazy `import("seek-bzip")` through `@sparq/client/src/decompress.ts`, but the consumer program did not automatically include the package-local ambient declaration. Both codec loaders were already invocation-path ESM dynamic imports, and the existing test confirms there are no static `fzstd` or `seek-bzip` imports.

The fix binds `seek-bzip.d.ts` from the internal decompression module with a triple-slash path reference. It is not re-exported, and the public API remains client-owned plain types (`Promise<DecompressedDatasetBytes>` containing `Uint8Array`), so consumers do not need codec-library types.

Verified locally:
- `@sparq/client`: `npm run build` and all 8 unit tests pass
- exact site consumer lane: lint, `npx tsc --noEmit`, static export build, and bundle guard pass
- fresh wasm type identity check passes
- `typos` passes
- no `site/` source change is included

The PR was not armed.

PR #2256, opened 2026-07-13
feat(vectors): recall-gated ANN index for concept dedup + k-NN (#2251)

### LINKED ISSUE #2251: sparq-vectors: recall-gated ANN index for concept-vector dedup + k-NN at 100k→millions scale (Kernel-of-Truth scale prereq)
## Recall-gated ANN index for concept-vector dedup + nearest-neighbour, at 100k→millions scale

**Context (Kernel-of-Truth scale track).** The kernel builds a large set of grounded, deterministically-vectorised concepts sourced from WordNet ∪ OBO ∪ SUMO ∪ Wikidata. A domain-balanced ~100k rung is now built (191,301 typed-structured concepts, biology share 49.77%). Two vector operations become the scaling bottleneck **past ~100k, toward millions**:

1. **Type-level deduplication** — cross-source near-duplicate concepts (e.g. the same class in WordNet and Wikidata, or overlapping OBO/SUMO terms) must be merged. Today this is an **exact O(m²) nearest-neighbour scan** (~3.2 CPU-h at 100k — still fine at this rung, measured), but it **dies between the rungs**: O(m²) is intractable past a few hundred k.
2. **Nearest-neighbour queries** over the concept-vector matrix — for the deterministic vectoriser's crosstalk/collision checks and (later) composition of novel concept vectors.

**What's needed:** an **ANN index over the structure-aware concept vectors** supporting (a) recall-gated dedup and (b) k-NN queries, with a **pre-registered recall gate: ≥0.99 recall vs a one-off exact O(m²) ground truth built at the 100k rung**, then metrics reported under the frozen dedup policy at each larger rung. Standard ANN (HNSW/FAISS/ScaNN-class) is fine mechanically; the ask is where it should *live*.

**Why sparq (the question @jeswr raised):** the kernel already builds on `sparq-vectors` (structure-aware vectorisation, the concept-hash→embedding seam). An ANN index + recall-gated dedup over those vectors is a natural, reusable **`sparq-vectors` capability** that the kernel and any other vector consumer would use — cleaner than the kernel bolting on its own ANN infra. The alternative is the kernel uses an off-the-shelf ANN library privately; that works but doesn't give the estate a shared, recall-gated primitive.

**Status — NOT a current blocker.** The 100k rung is built with exact dedup; this is a **forward prerequisite** for the millions-scale (the level at which the kernel becomes an *impactful* result rather than a feasibility demo). No urgency, and no maintainer input needed on the ANN mechanics — the open question is purely whether this becomes a `sparq-vectors` feature (preferred, reusable) or stays kernel-side.

**Concretely, if sparq-vectors adopts it:** a `build_ann(vectors, policy) → index` + `dedup(index, recall_gate=0.99, ground_truth) → merged_ids` + `knn(index, q, k)` surface, with the recall gate + frozen-policy metrics as the acceptance test. Happy to contribute the recall-gate harness (we already have the exact-O(m²) ground-truth builder at 100k).

jeswr:

Closes #2251.

## Why

The Kernel-of-Truth scale track (the consumer named in #2251) needs cross-source concept dedup + k-NN past the ~100k rung, where its current **exact O(m²)** nearest-neighbour scan becomes intractable. ANN scales but is approximate — so the issue pre-registers the honest contract: **an ANN-driven merge pass may only apply merges after proving ≥ 0.99 recall against a one-off exact O(m²) ground truth**. This lands that as a reusable `sparq-vectors` capability (the issue's preferred home) instead of kernel-side private ANN infra.

## What

Exactly the surface #2251 asks for, in `crates/sparq-vectors` behind the **existing opt-in `approx-ann` feature** (the crate's only third-party-ANN dependency; the default build carries none of this and gains **no new dependency**):

- **`build_ann(vectors, policy) -> ConceptAnnIndex`** — HNSW (the crate's existing `instant-distance` backend, same Arc-backed normalised point type as `VectorIndex`, so scores stay directly comparable) over **raw `(id, vector)` pairs** — no `Graph`/`VectorStore` needed, which is the shape the concept-vector consumer has. `HnswConfig` is the build policy; deterministic per seed. Fail-closed input validation (empty/zero-dim/dim-mismatch/non-finite/all-zero/duplicate-id ⇒ `Err`).
- **`knn(index, q, k)`** (free fn + method, plus `knn_of(id, k)` by indexed id) — cosine top-k, best first; zero query ⇒ empty, same contract as the rest of the crate. **APPROXIMATE, recall < 1.0.**
- **`dedup(index, policy, exact_ground_truth) -> DedupReport`** — **the recall gate runs FIRST**: set-based recall@k over the whole exact oracle; below `policy.recall_gate` (default **0.99**, the pre-registration) it returns `Err` and **no merge is computed** (fail-closed — an under-recalling index silently *misses* true duplicates, a correctness loss). Only after the gate: above-`merge_threshold` neighbour pairs → union-find transitive closure → `merges: Vec<(duplicate, canonical)>` (canonical = smallest id, deterministic) + `groups`.
- **`exact_ground_truth(vectors, k)`** — the exact O(m²) oracle builder (build once at the tractable rung); `GroundTruth::new` wraps an externally-computed oracle (the kernel already has one at 100k), with malformed-oracle validation so a bad oracle can't silently weaken the gate.
- Misconfigurations are hard errors, never silent degradation: a `k` the build `ef_search` beam can't serve, an oracle id absent from the index, out-of-range gate/threshold.

## Recall gate as the acceptance test

`tests/dedup.rs` (runs in CI — the nextest archive carries `approx-ann`):

- **`ann_recall_gate_passes_vs_exact_ground_truth`** — 3 000×32 corpus, exact O(m²) ground truth built in-test, asserts `dedup`'s measured recall **≥ 0.99** and that a passing gate over a duplicate-free corpus emits **zero** merges (gate and merge phases are independent).
- **`dedup_refuses_merges_below_the_recall_gate`** — a deliberately-wrong oracle drives measured recall far below the gate ⇒ `Err`, no merges (the fail-closed half).
- Planted near-duplicate clusters merge transitively to the smallest-id canonical, and nothing else merges; `knn` ordering/zero-query contracts; the full fail-closed validation matrix.

**Honest scope** (stated in the rustdoc + SKILL + README): the gate is measured evidence **for this index over this ground truth's corpus**, not a universal recall guarantee — re-gate after any re-vectorisation; `merge_threshold` has a plain (non-canonical) default and the policy is to be **frozen per scale track** with per-rung metrics reported under it, per the #2251 protocol.

## Docs (MAINTENANCE RULE, same change)

- `skills/vector-search/SKILL.md` — frontmatter description, Key-APIs block, new **recipe 22**, `approx-ann` gotcha bullet.
- `crates/sparq-vectors/README.md` — dedup surface noted in the Features bullet (kept at the 120-line template cap).

## Gates run locally

- `cargo test -p sparq-vectors` — default features **and** `--features approx-ann` (all suites green).
- `cargo clippy -p sparq-vectors --all-targets -- -D warnings` — both feature states.
- `RUSTDOCFLAGS="-D warnings" cargo doc -p sparq-vectors --no-deps --all-features`.
- `scripts/check-privacy-claims.sh`, `check-no-perf-numbers.py`, `check-readme-template.py` — clean. (`typos` binary not on this box — CI's typos lane is the check of record.)
- Full-workspace clippy/test are CI's authoritative gate (`ci-summary`).

## Notes for the sparq agent

- `bd` could not reach the beads Dolt DB from this box (`no beads database found`; only the exported JSONL is present), so **no bead id is attached** — please create/close the tracking bead on merge.
- Proceed-and-document: surface placement (a `dedup` module behind the existing `approx-ann` feature, raw-matrix index rather than a `VectorStore`-coupled one) follows the issue's own framing; no new feature flag and no new dependency were added.


jeswr:

Handing this over to the sparq agent to finalise, review, and merge.

PR #2320, opened 2026-07-15
fix(deploy): harden PaaS config after IaC review (sq-dwame) [GPT-5.6]

jeswr:
> 🤖 SPARQ agent

Post-hoc GPT-5.6 review of the Sonnet-authored PaaS IaC from PR #2314 found material correctness and security defects.

Fixed:
- Fly HTTP services defaulted to two Machines, which would split the uncoordinated in-memory SPARQL dataset and the LWS in-memory DPoP replay state. The documented launch/deploy flow now disables HA and scales to exactly one Machine.
- Railway selected no image, used an empty Dockerfile path and empty start command, omitted the fixed health-check port, and did not pin replicas. Small Dockerfiles now consume the canonical GHCR images, current schema keys are used, PORT is installed before deployment, and numReplicas is one.
- LWS omitted SOLID_SERVER_BIND, leaving the binary loopback-only default unreachable from PaaS ingress. All three platform paths now bind 0.0.0.0:3000.
- sparq-server inherited SPARQ_ALLOW_REMOTE=1 and could still boot open when a secret was missed. Every platform now overrides it to zero, enables read auth, and injects SPARQ_AUTH_TOKEN from Fly secrets, a Render-generated environment-group secret, or Railway variables. Missing auth now fails the public bind closed.
- Render now pins one instance, explicitly declares the image port, and generates the auth token in an environment group.
- The advertised generic deploy buttons did not safely select the nested configs or guarantee secret-before-ingress ordering. The README now gives provider-valid setup flows and correct current domain names.
- Fly LWS now checks both /livez and /readyz; all other health paths and ports are asserted per server.

Validation:
- Current official Render JSON schema: both Blueprints pass.
- Current official Railway JSON schema: both TOMLs pass.
- Current flyctl local parser: both Fly configs pass.
- Secret, port, image, health, fail-closed auth, and single-instance assertions pass; repository doc/privacy/performance gates pass.

Honest limits: no live provider deployment was performed because provider credentials are unavailable. The canonical sparq-org GHCR refs were not publicly pullable during review, so deployment remains externally blocked until those packages/tags are published; LWS publication is already tracked by sq-lmz40. I did not substitute the legacy jeswr image because the architecture contract requires sparq-org.

This PR is intentionally unarmed.

### DIFF
--- a/deploy/paas/README.md
+++ b/deploy/paas/README.md
@@ -1,138 +1,170 @@
 <!-- [SONNET-4.6] sq-dwame — PaaS one-click deploy configs for sparq-server + sparq-lws-core. -->
+<!-- [GPT-5.6] Post-hoc correctness/security review and hardening after PR #2314. -->
 
-# PaaS one-click deploy
+# PaaS deployment
 
-One-click deploy configs for the sparq SPARQL server (`sparq-server`) and the Solid/LWS
-server (`sparq-lws-core`) on three fast PaaS providers: **Fly.io**, **Render**, and
-**Railway**.
+Deployment configs for the SPARQL server (`sparq-server`) and Solid/LWS server
+(`sparq-lws-core`) on Fly.io, Render, and Railway.
 
-## Secure-defaults notice (R1/R2/R4/R9)
+These are intentionally explicit setup flows, not unsafe deploy buttons. A generic button cannot
+both select these nested config files and guarantee that secrets are installed before public
+ingress. Follow the provider steps below in order.
 
-> **sparq-server is open-by-default at the image layer** (it bakes `SPARQ_ALLOW_REMOTE=1`
-> and ships with no auth). Anyone who can reach the published port can read AND write the
-> dataset. **These templates enforce auth ON** at the template layer via `SPARQ_AUTH_TOKEN`
-> — do not remove that wiring.
+## Secure defaults
 
-Key rules applied in every config here:
+> **sparq-server is open by default at the image layer.** It bakes
+> `SPARQ_ALLOW_REMOTE=1` and has no token unless the deployment supplies one. These templates
+> inject `SPARQ_AUTH_TOKEN` from the provider's secret store and set
+> `SPARQ_AUTH_TOKEN_READ=1`, so both reads and writes require the token. Do not remove that
+> wiring. They also override the baked `SPARQ_ALLOW_REMOTE=1` to `0`, so a missing token makes
+> the public bind fail closed instead of starting an unauthenticated server.
 
-- **R1 (auth ON):** `SPARQ_AUTH_TOKEN` is a required secret for sparq-server. The token is
-  stored only in the PaaS secret store — never in the config files.
-- **R2 (no anonymous write):** Unauthenticated write requests return 401/403. The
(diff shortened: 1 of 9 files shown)


jeswr:
> 🤖 SPARQ agent

**Post-hoc IaC review of the already-merged PaaS config (#2314), per maintainer directive (IaC = Fable/Sol, not Sonnet).** Sol (GPT-5.6) reviewed + fixed; verified sound by Opus 4.8.

**Real defects fixed (all verified):**
- **Single-writer (the systemic bug):** Fly.io spins up **2 Machines by default** for an HTTP service → two writable, divergent in-memory datasets. Now `--ha=false` + `fly scale count 1`; Render `numInstances: 1`; Railway `numReplicas: 1`. (LWS same, for its in-memory DPoP replay store.)
- **Fail-closed bind:** sets `SPARQ_ALLOW_REMOTE=0` + `SPARQ_AUTH_TOKEN_READ=1`. Verified against `crates/sparq-server/src/main.rs:63-66`: a non-loopback bind is allowed only when the surface is fully authenticated — so a missing token makes startup **fail closed** instead of booting an open unauthenticated server (stronger than the Sonnet config, which leaned on the image’s baked-open bind).
- **Railway couldn’t select the image:** config-as-code can’t pull a GHCR image directly → added a one-line `Dockerfile` (`FROM ghcr.io/sparq-org/…`) + `builder = "DOCKERFILE"`.
- **LWS bound to loopback** (unreachable behind PaaS ingress) → `SOLID_SERVER_BIND=0.0.0.0:3000`; `PORT` set for Railway/Render probe routing.
- Render token now auto-generated into an env-group (`generateValue: true`) — no plaintext, no manual step. No inlined secrets anywhere (final scan clean).

**⚠️ Decision for the maintainer — deploy buttons removed.** The epic (sq-3vjdr) asked for 1-click deploy *buttons*; this PR **replaces them with explicit CLI flows**, arguing a generic button can’t guarantee the secret is set before public ingress. With the new fail-closed bind, a button is now actually **safe** (no token ⇒ no open server). If you want the 1-click buttons back, say so and I’ll restore them in a quick follow-up (additive). Flagging rather than silently dropping a stated goal.

Armed `--auto` (holds until CI green) for the correctness/security fixes. Fable capped until 22:00 UTC; its pass queued as a follow-up.

jeswr:
> 🤖 SPARQ agent

**Buttons restored (per maintainer decision: buttons + CLI flows).** Sol re-added the 1-click Fly/Render/Railway buttons for both servers alongside the explicit CLI flows; verified by Opus:
- **Render** button auto-generates `SPARQ_AUTH_TOKEN` (env-group `generateValue`) → works out of the box.
- **Railway** button prompts for `SPARQ_AUTH_TOKEN` + carries `SPARQ_AUTH_TOKEN_READ=1`, `PORT=3030` via `envs=…`.
- **Fly** button launches from `fly.toml`; the adjacent flow documents `fly secrets set` + `fly scale count 1` (fails closed until the token is set — intended safe behavior).
- **No literal secret in any button URL** (scanned); single-writer pins + `SPARQ_ALLOW_REMOTE=0` fail-closed wiring intact (not weakened).

Re-armed `--auto`.

PR #2330, opened 2026-07-15
fix(solid): close ODRL bare-assignee widening in conditional bridge (sq-9n1q4) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent (Opus tier) — a P1 access-control **widening** fix in the opt-in `sparq-solid` `odrl-bridge` feature (OFF by default). Bead **sq-9n1q4**. **UNARMED** — the orchestrator adversarially verifies security fixes before arming.

## The widening (probe-verified on origin/main)

An ODRL permission `[action read; target n1; assignee alice]` with **ZERO constraints**, materialized through `PodStore::materialize_odrl_permission_conditional`, emitted a `ConditionalGrant` headed by **`auth:Public`**. So **bob** AND an **anonymous** session then passed `accessible()` / `query_as` for `n1` — a permission scoped to one assignee granted *everyone*.

## Root cause

The conditional bridge computes its grant head via `condition_agents(_rule, recipients)`, where `recipients` is the recipient/assignee **CONSTRAINT** set produced by `map_constraints_to_agents` (which iterates only `rule.constraints`). With zero constraints that set is empty, and `condition_agents` defaulted an empty set to a single `auth:Public` head — its doc claimed "the rule had NO agent restriction", which is **false** when the `rule.assignee` **PROPERTY** (a distinct field on `Rule`) is set. `rule_action_target_match` deliberately defers the assignee "to the persisted condition", but no condition was ever persisted for the bare property, so the assignee restriction silently vanished.

## The fix (minimal, single-site)

Fold `rule.assignee` into the condition head when the recipient constraint set is empty: a bare-assignee rule now heads the grant with **only that assignee**, never `auth:Public`. The assignee is normalised + reserved-encoding-filtered exactly like a recipient (an all-reserved assignee → empty head → the callers already fail closed). **Preserved:** a rule with NO recipient constraint AND NO assignee still maps to `auth:Public` (legitimately public — action/target/duties were already satisfied at materialization).

## Prohibition dual — same root cause, one fix

`materialize_prohibition_conditional` shares the same `condition_agents` slot, so a bare-assignee prohibition previously became a **PUBLIC deny** (over-deny: everyone denied, not just the assignee — less dangerous but wrong). The single-site fix scopes that deny to the assignee too. No separate change was needed — the dual is fixed by the same edit.

## Before/after test evidence (both fail RED on the unfixed code, pass after)

Verified by running each test against the pre-fix `condition_agents` (both asserted the `auth:Public` head count was 1, expected 0) and again after the fix.

| Test | Before (unfixed) | After (fixed) |
|---|---|---|
| `bare_assignee_permission_conditional_scopes_to_assignee_not_public` | RED — grant head `auth:Public` (bob + anonymous read `n1`) | GREEN — grant head scoped to alice; bob + anonymous **denied** `accessible()` / `query_as` |
| `bare_assignee_prohibition_conditional_scopes_to_assignee_not_public` | RED — deny head `auth:Public` (over-deny) | GREEN — deny head scoped to alice; bob keeps the public allow |

Both tests exercise the **real enforcement path** (`PodStore::accessible` / `query_as`), not a mock. Existing test 19 (`no_constraint_conditional_grants_public` — the legitimate no-constraint AND no-assignee → `auth:Public` case) still passes.

Feature flag: **`odrl-bridge`** (OFF by default; the whole `sparq-solid` ODRL surface + its test file are gated behind it).

- `cargo clippy -p sparq-solid --all-targets -- -D warnings` — clean (feature OFF)
- `cargo clippy -p sparq-solid --all-targets --features odrl-bridge -- -D warnings` — clean (feature ON)
- `cargo test -p sparq-solid --features odrl-bridge` — 68/68 integration + doctests + lib unit tests pass
- `RUSTDOCFLAGS="-D warnings" cargo doc -p sparq-solid --no-deps --all-features` — clean
- `scripts/check-readme-template.py --enforce` — 0 deviations (README unchanged)
- `sparq-solid` line coverage **94.32%** (floor 89) — the modified `condition_agents` branch is directly exercised by both new tests

## Docs

Updated `skills/access-control/SKILL.md` (the public-API surface) and the `condition_agents` / `materialize_permission_conditional` rustdoc to state the bare-assignee scoping and the deny dual.

## Discovered work (out of scope — captured, not fixed here)

The conditional bridge ignores `rule.logical_constraints` entirely (`map_constraints_to_agents` iterates only `rule.constraints`). A permission carrying an `odrl:and`/`odrl:or`/`odrl:xone` **compound** constraint but no atomic constraints would map `Faithful` with an empty recipient set — the compound constraint is silently dropped from the conditional path, a distinct pre-existing widening hazard. Not introduced by this change and not the assignee root cause; flagged for a follow-up bead.

### DIFF
--- a/crates/sparq-solid/src/odrl_bridge.rs
+++ b/crates/sparq-solid/src/odrl_bridge.rs
@@ -872,8 +872,11 @@ fn recipient_principal_allowed(p: &str) -> bool {
 /// re-checked by [`crate::AuthIndex::accessible`]: a session whose agent is the
 /// recipient is granted; any other agent (or anonymous) is denied — **without**
 /// re-running the ODRL evaluator. A recipient *set* emits one grant per member (the
-/// auth view unions the allows). When the rule has NO recipient constraint, the grant
-/// head is `auth:Public` (any session) — only valid because the action/target/duties
+/// auth view unions the allows). When the rule has NO recipient constraint but carries
+/// an `odrl:assignee` PROPERTY, the grant head is scoped to that one assignee ([OPUS-4.8]
+/// sq-9n1q4 — a bare-assignee rule grants ONLY the assignee, never `auth:Public`). Only
+/// when the rule has NO recipient constraint AND NO assignee is the grant head
+/// `auth:Public` (any session) — legitimately public because the action/target/duties
 /// were already satisfied at materialization.
 ///
 /// # Fail-closed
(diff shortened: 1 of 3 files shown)


jeswr:
> 🤖 SPARQ agent

**Adversarially verified by Opus 4.8 — arming.** This closes a real P1 access-control **widening** (opt-in `odrl-bridge` feature).

- **Root cause confirmed:** `condition_agents` ignored the `rule.assignee` PROPERTY and defaulted an empty recipient set to `auth:Public`. Fix folds a bare `odrl:assignee` into the grant head (scoped to that party), keeps no-assignee-no-constraint → `auth:Public` (legit), and an all-reserved assignee → empty head → **fail-closed** (never widened to Public).
- **Non-vacuous proof (verified):** the closure test asserts via the REAL `PodStore::accessible` + `query_as` paths that alice reads n1 while **bob AND an anonymous session are denied** (rows 1/0/0) — RED on unfixed code, GREEN after. Prohibition dual fixed by the same single-site edit + its own test. Existing no-assignee→Public case unregressed.
- Gates green in BOTH feature states (clippy off/on, tests, rustdoc all-features, readme-template, coverage 94.32%).

Armed `--auto`.

jeswr:
> 🤖 SPARQ agent (merge-fixer, Opus 4.8)

**Conflict resolved — PR is now MERGEABLE** (was CONFLICTING/DIRTY; `gh pr update-branch` had failed).

**What was blocking:** a merge conflict against `origin/main`. The `crates/sparq-solid/src/odrl_bridge.rs` source auto-merged cleanly — the ONLY conflict was in `crates/sparq-solid/tests/odrl_bridge.rs`, where git interleaved this branch's two new tests (sq-9n1q4) with the already-merged #2331 tests (sq-izzak) because both test pairs share identical boilerplate tails (the anonymous `accessible()`/`query_as` end-to-end block, and the baseline public-allow setup).

**Resolution — BOTH fail-closed behaviours preserved (neither regresses to fail-open):**
- **sq-9n1q4 (this branch):** `condition_agents()` folds a bare `odrl:assignee` PROPERTY into the condition head — a bare-assignee rule scopes to ONE assignee, never `auth:Public` (an all-reserved assignee → empty head → callers fail closed). Regression tests 30/31.
- **sq-izzak (already on main, #2331):** `map_constraints_to_agents()` returns `Unmappable` on any compound `odrl:LogicalConstraint` → the one-shot fallback enforces the compound instead of folding an empty recipient set to `auth:Public`. Regression tests 32/33.

I kept all four tests intact. `skills/access-control/SKILL.md` auto-merged and retains both the sq-izzak and sq-9n1q4 paragraphs.

**Gates re-run GREEN in-worktree, both feature states:**
- `clippy -p sparq-solid --all-targets [--features odrl-bridge] -- -D warnings` — clean (OFF + ON)
- `test -p sparq-solid` — OFF: 50 lib pass; ON `--features odrl-bridge`: 70 integration (incl. all 4 ODRL tests passing **together**) + 35 doctests, exit 0
- `cargo doc -p sparq-solid --no-deps --all-features` with `RUSTDOCFLAGS="-D warnings"` — clean (the merged `condition_agents` doc-comment)

No new content added — pure conflict resolution. Not arming; leaving the security verify + arm to the orchestrator. `mergeStateStatus: BLOCKED` is the expected async ruleset/CI drain on the fresh merge commit.

PR #2331, opened 2026-07-15
fix(solid): close compound odrl:LogicalConstraint widening in conditional bridge (sq-izzak) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent (Opus tier). This closes P1 access-control widening bead **sq-izzak** — the sibling of sq-9n1q4 (#2330). Opt-in `odrl-bridge` feature (OFF by default). **Unarmed** pending review.

## Hazard (access-control WIDENING + over-deny)

An ODRL permission whose ONLY restriction is a **compound** `odrl:and`/`odrl:or`/`odrl:xone` constraint — with ZERO atomic `rule.constraints` and no bare `odrl:assignee` — mapped through the conditional bridge to `Faithful` with an **empty recipient set**, which folded to an `auth:Public` grant head. The compound restriction was **silently DROPPED**, so unlisted/anonymous agents were **granted**. The prohibition dual materialised a **public deny** (over-denied everyone). A distinct pre-existing widening hazard, NOT introduced by #2330.

## Root cause

`map_constraints_to_agents` (the Faithful/Unmappable classifier feeding both `materialize_permission_conditional` and `materialize_prohibition_conditional`) iterated only `rule.constraints` and **never** examined `rule.logical_constraints`. A rule carrying only a compound constraint therefore looked constraint-free → `Faithful { agents: [] }` → empty head → public.

## Fix: fail-closed FALL-BACK (not faithful fold) — and why

A rule carrying any `logical_constraints` now classifies **`Unmappable`**, routing both conditional entry points to the one-shot `materialize_permission`/`materialize_prohibition` path, whose evaluator **DOES** enforce the compound constraint (frozen — verified in `sparq-policy` eval, `logical_constraint_status` folds into the same three-valued classify).

Faithful folding was **deliberately rejected** as itself a widening risk:
- An ACP grant head is a **UNION** of `auth:agent` allows. An `odrl:and` of recipient constraints is an **INTERSECTION** → folding it as a union widens in the **wrong direction**.
- An `odrl:or` may mix a recipient operand with a **non-recipient** dimension (time/purpose) that has no head analogue.
- `odrl:xone` (exactly-one) has **no ACP analogue** at all.

Over-approximating any compound agent-restriction into the head would grant unlisted agents — the very widening this classification exists to prevent. Fall-back is provably correct and minimal (the one-shot path never drops the bound). This matches the crate's existing doctrine ("a rule mixing mappable + unmappable constraints falls back entirely to one-shot — never drops a bound").

The refresh path is consistent too: a compound-constraint prohibition now returns `false` from `prohibition_maps_faithfully`, so refresh routes through the deny-retraction-aware `refresh_prohibition` (fail-closed), not a fail-open one-shot.

## Test evidence (non-vacuous — RED before / GREEN after)

Both exercise the REAL enforcement path (`PodStore::accessible` + `query_as`), mirroring #2330's regression test. A `recipient eq alice` sub-constraint makes the one-shot fallback grant/deny EXACTLY alice while bob and anonymous are structurally excluded.

| test | before fix | after fix |
|---|---|---|
| `compound_only_permission_conditional_scopes_not_public` (`odrl:and`) | **RED** — auth:Public head=1, bob + anonymous read n1 | **GREEN** — only alice granted; bob + anonymous DENIED |
| `compound_only_prohibition_conditional_scopes_not_public` (`odrl:xone`) | **RED** — auth:Public deny head=1 (over-deny) | **GREEN** — only alice denied; bob + anonymous keep the public allow |

Confirmed RED on pre-fix code by stashing the src change:
```
test compound_only_permission_conditional_scopes_not_public ... FAILED
  assertion `left == right` failed: NO auth:Public head — the compound restriction is NOT dropped
  left: 1  right: 0
test compound_only_prohibition_conditional_scopes_not_public ... FAILED
  assertion `left == right` failed: NO auth:Public deny — the compound restriction is NOT dropped
  left: 1  right: 0
```

**Prohibition-dual verdict:** audited and closed. The pre-fix over-deny (public deny head dropping the compound recipient scope) is proven by test 33's RED→GREEN; the fix scopes the frozen deny to exactly the intended party via the one-shot fallback, so bob + anonymous keep their allow.

- `cargo clippy -p sparq-solid --all-targets -- -D warnings` — GREEN (feature **OFF**)
- `cargo clippy -p sparq-solid --all-targets --features odrl-bridge -- -D warnings` — GREEN (feature **ON**)
- `cargo test -p sparq-solid --features odrl-bridge` — GREEN (68/68 `odrl_bridge` integration tests + 35 lib/doctests)
- `cargo test -p sparq-solid` (feature OFF) — the test file is `#![cfg(feature = "odrl-bridge")]`, so it no-ops OFF; the crate builds + tests clean OFF
- `RUSTDOCFLAGS="-D warnings" cargo doc -p sparq-solid --no-deps --all-features` — GREEN (all doc-links intra-crate to `pub` items)
- The legit no-constraint→`auth:Public` case (`no_constraint_conditional_grants_public`) still passes — the fix is scoped to rules carrying `logical_constraints`.

## Coordination with #2330 (still OPEN)

Branched off current `origin/main`. My src edit is in `map_constraints_to_agents` — a **different function** from #2330's `condition_agents` edit. The SKILL.md edit lands in the "STAY one-shot" clause, not #2330's "fail open → Mapping" region. Test additions are appended at file end (both PRs append; trivially mergeable). No line overlap.

Feature flags: **`odrl-bridge`** (sparq-solid, OFF by default).

### DIFF
--- a/crates/sparq-solid/src/odrl_bridge.rs
+++ b/crates/sparq-solid/src/odrl_bridge.rs
@@ -714,7 +714,29 @@ enum AgentMapping {
 /// `isNoneOf`, or a numeric/dateTime operand under `isNoneOf` (the evaluator never
 /// satisfies those — `set_negation_representable` — so persisting an exception for
 /// them would widen access). Any one such constraint forces the whole rule one-shot.
+///
+/// **Unmappable (→ stay one-shot): a COMPOUND `odrl:LogicalConstraint`** ([OPUS-4.8]
+/// sq-izzak — WIDENING FIX). A rule that carries ANY `odrl:and`/`odrl:or`/`odrl:xone`
+/// (`rule.logical_constraints`) has no faithful single-head ACP analogue and MUST stay
+/// one-shot, even when it has ZERO atomic constraints. A grant head is a UNION of
+/// `auth:agent` allows, so: an `odrl:and` of recipient constraints is an INTERSECTION
+/// (folding it as a union widens — wrong direction); an `odrl:or` may mix a recipient
+/// operand with a non-recipient dimension (time/purpose) that has no head analogue; and
+/// `odrl:xone` (exactly-one) has no ACP analogue at all. Over-approximating any compound
+/// agent-restriction into the head would grant unlisted/anonymous agents — the very
+/// widening this classification prevents. Before this fix the loop below examined only
+/// `rule.constraints`, so a rule whose ONLY restriction was a compound constraint mapped
+/// `Faithful` with an EMPTY recipient set → an `auth:Public` head (the compound
+/// restriction silently DROPPED). Fail-closed: any `logical_constraints` forces the whole
+/// rule to the one-shot path ([`materialize_permission`]/[`materialize_prohibition`]),
+/// whose evaluator DOES enforce the compound constraint (frozen) — never dropping it.
 fn map_constraints_to_agents(rule: &Rule) -> AgentMapping {
+    // [OPUS-4.8] sq-izzak: a rule carrying any compound `odrl:LogicalConstraint` has no
+    // faithful ACP-condition head — stay one-shot so the compound restriction is enforced
+    // (frozen) rather than silently dropped by folding an empty recipient set to public.
+    if !rule.logical_constraints.is_empty() {
+        return AgentMapping::Unmappable;
+    }
     let mut agents: Vec<String> = Vec::new();
     let mut except: Vec<String> = Vec::new();
     let mut window = TimeWindow::default();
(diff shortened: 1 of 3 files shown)


jeswr:
> 🤖 SPARQ agent

**Adversarially verified by Opus 4.8 — arming.** Closes the compound-`odrl:LogicalConstraint` widening sibling (sq-izzak), coordinated + non-overlapping with #2330 (different function, EOF-appended tests).

- **Fix (fail-closed fallback):** `map_constraints_to_agents` returns `Unmappable` for any rule carrying `rule.logical_constraints`, forcing it to the one-shot path where the compound restriction is enforced — instead of the conditional path that dropped it to an `auth:Public` head. Correctly prefers fail-closed over unsound faithful-folding of and/or/xone.
- **Non-vacuous test (verified):** `compound_only_permission_conditional_scopes_not_public` isolates the compound-only case (0 atomic, 1 logical, no assignee), asserts NO `auth:Public` head, and via the real `PodStore` path grants only alice (bob + anon denied) — RED on unfixed code → GREEN. Prohibition dual covered by its own test.

Armed `--auto`.

PR #2332, opened 2026-07-15
feat(server): record streams on writer commit hook (sq-kqofk) [GPT-5.6]

jeswr:
> 🤖 SPARQ agent

Adopts `Writer::spawn_with_commit_hook` + `ChangeLog::into_commit_hook` in `sparq-server`, moving durable `/streams` recording onto the sequenced writer thread. This removes the caller-side `Mutex<ChangeLog>` held across `submit`, so concurrent updates can group-commit while each published generation remains durably recorded before acknowledgement. Polling uses a separate read handle, the `LATEST` tail advances only after successful appends, and online-restore writer swaps retain the same guarded hook. The public HTTP-server skill and change-stream docs now describe generation-level recording.

Gate status:
- `cargo clippy -p sparq-server --all-targets -- -D warnings` — passed
- `cargo test -p sparq-server` — passed
- `cargo clippy -p sparq-server --all-targets --features change-stream -- -D warnings` — passed
- `cargo test -p sparq-server --features change-stream` — passed (includes the non-vacuous concurrent group-commit hook test and all `/streams` integrations)
- `cargo clippy -p sparq-server --all-targets --features change-stream,backup -- -D warnings` — passed (additional feature-composition check)
- `git diff --check` — passed

Local whole-workspace `cargo fmt --check` remains red on extensive pre-existing `origin/main` formatting outside this change; no unrelated formatting was staged.

### DIFF
--- a/crates/sparq-server/Cargo.toml
+++ b/crates/sparq-server/Cargo.toml
@@ -311,9 +311,10 @@ terse = ["server", "dep:sparq-terse"]
 # change-stream (sq-b4fns / #1223, in sparq-serve). The endpoint reads the segmented, fsync'd
 # append-only `ChangeLog` and returns the ordered change records (op / quads / seq / generation /
 # timestamp) as JSON from a given offset, with a continuation token, driven by `iteratorType` /
-# `at` / `after` / `limit` parameters. The same feature also turns on RECORDING: when the durable
-# log directory is configured, every committed SPARQL Update appends one ordered change record to
-# the log (the in-server publish path), so the endpoint serves a live, resumable feed. It turns on
+# `at` / `after` / `limit` parameters. The same feature also turns on RECORDING:
+# [GPT-5.6] (sq-kqofk) with a configured log directory, every published group-commit generation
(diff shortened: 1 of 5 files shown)


jeswr:
> 🤖 SPARQ agent

**Verified — arming (Opus 4.8).** Moves durable `/streams` recording onto the sequenced writer thread (`Writer::spawn_with_commit_hook` + `ChangeLog::into_commit_hook`), removing the caller-side `Mutex<ChangeLog>` held across `submit` so concurrent updates group-commit while each generation stays durably recorded before ack.

- **Non-vacuity:** the load-bearing test commits 3 real updates over HTTP, polls `GET /streams` from offset 1, and asserts the EXACT ordered change sequence (op+quad+seq+generation) + durability across a `ChangeLog::open` restart — red if the hook dropped/reordered records.
- **Gates green** in all feature states: clippy `-D warnings` + tests for default / `change-stream` / `change-stream,backup`; diff scoped to `sparq-server` + the http-server skill.

Armed `--auto` (CI gate confirms the full suite incl. formatting).

Issue #2345, opened 2026-07-16
Restore workspace cargo fmt check on main


jeswr:
> 🤖 SPARQ agent — `cargo fmt --all -- --check` reports extensive pre-existing diffs across workspace files on current `origin/main`, so the documented formatting gate cannot pass cleanly.

jeswr:
> 🤖 **SPARQ agent** — duplicate of #2360, consolidated there with a full diagnosis, a single decision point (pin strategy), and a proceed-and-document offer to land the one-time reformat in the current clean (no-in-flight-branch) window. Closing to reduce tracker clutter; the work is tracked on #2360.

PR #2346, opened 2026-07-16
test(sparq-algos): pin graph-shape edge cases (sq-bif.29) [GPT-5.6]

jeswr:
> 🤖 SPARQ agent

Adds default-feature integration coverage for empty graph algorithm outputs, canonical two-node degree values, deterministic `top_k` tie-breaking, and disconnected weak-component counting. This pins the public graph-shape floors without source, dependency, manifest, or feature changes.

Gates: `cargo clippy -p sparq-algos --all-targets -- -D warnings` ✅; `cargo test -p sparq-algos` ✅; `cargo clippy --workspace --exclude sparq-py --all-targets -- -D warnings` ✅. Mutation spot-check (`2` components changed to `1`) failed as required, then the correct value was restored and the full crate suite passed. The new file passes `rustfmt --check`; workspace-wide formatting has unrelated pre-existing drift tracked in #2345.

### DIFF
--- a/crates/sparq-algos/tests/edge_graph_shapes.rs
+++ b/crates/sparq-algos/tests/edge_graph_shapes.rs
@@ -1,0 +1,59 @@
+// [GPT-5.6] sq-bif.29: pin the default-feature graph-shape seams through the public API.
+
+use sparq_algos::{
+    degree_centrality, degree_centrality_normalized, label_propagation, num_communities, pagerank,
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent

**Verified — arming (Opus 4.8).** Fable-specced (source-verified, self-loop-degree-corrected) sparq-algos graph-shape floor coverage; gates passed in-worker; test-only, non-vacuous. Armed `--auto`.

PR #2351, opened 2026-07-16
test(sparq-sim): pin absent-term and excluded-predicate contracts (sq-bif.37) [GPT-5.6]

jeswr:
> 🤖 SPARQ agent

Adds default-feature integration contracts for absent-term signatures, symmetric zero similarity, empty absent-term rankings, the zero-k guard, and present-but-empty signatures after excluding a subject's sole predicate. This pins the documented edge behavior without changing source, dependencies, or features.

Gates green:
- cargo clippy -p sparq-sim --all-targets -- -D warnings
- cargo test -p sparq-sim
- cargo test -p sparq-sim --test absent_and_excluded
- cargo clippy --workspace --exclude sparq-py --all-targets -- -D warnings
- mutation spot-check: changing the absent-term expected similarity from 0.0 to 1.0 failed at the intended assertion; restored before final gates.

Per the bead, validation used default features only and did not enable multi-hop.

### DIFF
--- a/crates/sparq-sim/tests/absent_and_excluded.rs
+++ b/crates/sparq-sim/tests/absent_and_excluded.rs
@@ -1,0 +1,63 @@
+// [GPT-5.6] sq-bif.37: pin absent-term and excluded-predicate signature contracts.
+
+use oxrdf::{NamedNode, Term};
+use sparq_core::Graph;
+use sparq_sim::{Sim, SimConfig};
+
+const NS: &str = "http://example.com/";
+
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent

**Verified — arming (Opus 4.8).** Fable-specced (source-verified) sparq-sim absent-term/excluded-predicate floor coverage; gates passed in-worker; test-only, non-vacuous. Armed `--auto`.

PR #2354, opened 2026-07-16
test(sparq-py): cover term rendering precedence edges (sq-bif.40) [GPT-5.6]

jeswr:
> 🤖 SPARQ agent

Adds pure-Rust sparq-py integration coverage for literal rendering precedence and catch-all edges: language overrides datatype, direction requires language, empty literals render correctly, and unknown kinds use literal rendering. This directly pins the Python repr renderer without PyO3 or a Python runtime.

Gates:
- PASS: cargo test -p sparq-py
- PASS: cargo clippy -p sparq-py --all-targets -- -D warnings
- PASS: cargo clippy --workspace --exclude sparq-py --all-targets -- -D warnings
- PASS: mutation spot-check (altered expected output failed the focused test)
- PASS: rustfmt --check crates/sparq-py/tests/term_render_edges.rs

Additional honesty: full-workspace cargo fmt --check reports pre-existing formatting drift outside this bead; it is already tracked in #2345.

### DIFF
--- a/crates/sparq-py/tests/term_render_edges.rs
+++ b/crates/sparq-py/tests/term_render_edges.rs
@@ -1,0 +1,37 @@
+// The sparq-py cdylib disables its lib test harness because linking a PyO3
+// extension-module harness fails on macOS. Compile the pure renderer directly so
+// these tests remain ordinary Rust integration tests with no Python dependency.
+#[path = "../src/term_render.rs"]
+mod term_render;
+
+use term_render::render_n3;
+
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent

**Verified — arming (Opus 4.8).** Fable-specced (source-verified) sparq-py term-render precedence coverage (language>datatype; direction needs language); gates passed in-worker; test-only, non-vacuous. Armed `--auto`.

Issue #910, opened 2026-06-19
Apache Arrow / columnar result export (Python binding first)


jeswr:
> 🤖 **SPARQ agent** (one of several @jeswr runs on this account).
>
> Deep-research feature proposal for your review — one of 17 surfaced by a parallel research sweep (competitive parity · W3C/emerging specs · query-engine · ops · DX · reasoning/privacy). **No work has started; this is for your decision.** The honesty caveats in the body are load-bearing (ZK/MPC are NOT externally audited; work-box benchmarks are non-canonical; first-party authz is plaintext access control, not a privacy guarantee; the RDF-1.2 canon extension would be non-standard).

## Summary
Project query-result `DataChunk`s (already columnar u64 ids internally) into Apache Arrow record batches for zero-copy transfer into pandas/Polars/DuckDB/BI, wired into the Python binding (`sparq-py`) first.

## Why / use-case
Arrow is the lingua-franca for columnar transfer into the dataframe/analytics/ML ecosystem; feeding SPARQL results straight into a DataFrame/ML pipeline without a CSV round-trip is the natural GenAI/analytics on-ramp. No mainstream RDF engine ships native Arrow, so it is a genuine differentiator, and the payoff is largest at the `sparq-py` boundary (pyoxigraph users reach for exactly this).

## Fit with sparq (+ what partially exists today, honestly)
Good, mid priority. sparq executes on columnar `DataChunk`s of u64 ids, so Arrow is a natural projection. Verified absent: grep for arrow/parquet/RecordBatch across crates hits only the SHACL parser's `Tok::Arrow` and dependency-arrow prose — **no Arrow dependency or export**. **Honest caveat:** RDF term -> Arrow type mapping is lossy/awkward (IRIs + typed literals are not native Arrow types — widen to strings or carry a datatype side-channel); the schema must be designed before committing code.

## Prior art (engines/specs + refs)
- Apache Arrow as result-transfer standard: https://arrow.apache.org/
- pyoxigraph; DuckDB/Polars/pandas Arrow interop

## Sketch / options
- Define the RDF-term -> Arrow schema first (struct-per-term {kind, lexical, datatype, lang} vs widened string + side-channel).
- Expose `Graph.query_arrow(...) -> pyarrow.Table` (or RecordBatchReader for streaming).
- Keep behind an `arrow` extra / feature so the lean wheel pays nothing.

## Decision needed from maintainer
[...]

PR #2376, opened 2026-07-16
feat(arrow): CSV connector round-trip (sq-lsp7k) [FABLE-5]

jeswr:
> 🤖 SPARQ agent (Claude Fable, salvaged + Opus-verified)

Adds a CSV container layer to sparq-arrow (issue #910 — Polars/DuckDB/pandas interop), mirroring the existing ipc/parquet opt-in layers over the shared term-struct RecordBatch projection.

- **Opt-in `csv` feature** (implies arrow, adds arrow-csv dep; Cargo.lock updated); default build byte-unchanged.
- **Fidelity**: the 5-field term struct (kind/value/datatype/language/direction) round-trips losslessly; the `kind` column distinguishes an unbound variable from a bound empty-string literal. The ONE shape CSV cannot carry — a zero-variable result — is documented + returns an honest error (not silent corruption).
- **Gates (Opus-verified)**: clippy both states -D warnings; test ON (proptest + unit round-trip) + default byte-unchanged; C1 guard OK; assemble self-test 38 passed; rustdoc --all-features green; feature-matrix leg + golden added.

Fable implemented this (incl. the Cargo.lock update); the impl agent stalled in its final gate sequence, so the orchestrator re-ran the gates, fixed a manual_is_multiple_of lint + added the missing feature-matrix leg, and Opus-reviewed. Unarmed.

### DIFF
--- a/.github/feature-matrix.d/sparq-arrow.yml
+++ b/.github/feature-matrix.d/sparq-arrow.yml
@@ -46,3 +46,8 @@
   crate: "sparq-arrow"
   features: "ipc"
   test: true
+
+- name: "sparq-arrow (csv — QueryResult <-> CSV byte round-trip)"
+  crate: "sparq-arrow"
+  features: "csv"
+  test: true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -272,42 +272,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arrow-cast"
-version = "59.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c82c236c3caf8df5664284f3f1fbe89938852163998c3fdbf37e84ac220445e9"
-dependencies = [
- "arrow-array",
- "arrow-buffer",
- "arrow-data",
- "arrow-ord",
- "arrow-schema",
- "arrow-select",
- "atoi",
- "base64",
- "chrono",
- "half",
- "lexical-core",
- "num-traits",
- "ryu",
-]
-
-[[package]]
-name = "arrow-csv"
-version = "59.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "12714e5fb7954159af1e26d4e0d37108bcf1a2ad5ee5c5bf02a944d564d588b7"
-dependencies = [
- "arrow-array",
- "arrow-cast",
- "arrow-schema",
- "chrono",
- "csv",
- "csv-core",
- "regex",
-]
-
-[[package]]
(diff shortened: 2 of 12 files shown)


jeswr:
> 🤖 SPARQ agent — Opus review: DISARMED pending two follow-ups (surfacing a supply-chain decision).

The CSV feature is implemented + gated correctly (clippy both states, proptest round-trip, rustdoc, feature-matrix leg — all green locally), but CI surfaces two blockers the local gates don't cover:

1. **flow-on-gates G2 (public-api-skill)**: the new `to_csv_bytes`/`from_csv_bytes` public surface needs a matching `skills/<surface>/SKILL.md` edit in the same PR (or a `skill-not-needed` label). Easy fix.
2. **supply-chain cargo-vet (GATING)**: `arrow-csv` pulls in **12 new unvetted transitive deps** (arrow-cast:59.0.0 etc.). These are apache/arrow-rs-family crates — the SAME trusted source as the already-vetted arrow-array/arrow-ipc/arrow-schema this crate uses for ipc/parquet — so exemptions are low-risk, but it's a real **supply-chain-surface expansion (12 deps) for a CSV convenience layer**.

**Decision to surface**: land it by cargo-vet-exempting the 12 arrow-family deps (mirrors how arrow-ipc/parquet were vetted), OR reconsider whether a CSV round-trip warrants arrow-csv vs a small manual serialiser of the 5-field term struct (zero new deps). I'll take the exemption path (low-risk, same-source, feature is opt-in) unless you'd prefer the dep-free rewrite. Disarmed so it doesn't merge before that's settled.

jeswr:
> 🤖 SPARQ agent — RESOLVED: rolled our own (per maintainer decision).

Measured hand-rolled vs arrow-csv (indicative, this box): **from_csv ~1.5x faster at scale** (100k: 253→166ms; **1M: 2607→1671ms = 1.56x**), roughly equal on tiny inputs. The hand-rolled serialiser (QueryResult→CSV directly, no RecordBatch intermediate, RFC-4180 quoting) is **byte-compatible** with the prior format (same round-trip + proptest tests pass) and **removes the entire arrow-csv/arrow-cast/lexical vet closure — 12 transitive deps → ZERO new deps** for the `csv` feature.

So this replaces the arrow-csv impl: faster at scale + no supply-chain expansion + no cargo-vet exemptions needed. Both blockers from before (the 12-dep vet + G2 SKILL) are gone — the SKILL edit is included, and there are no new deps to vet. Re-armed. [Opus-verified RFC-4180 quoting + term fidelity]

PR #3425, opened 2026-07-17
fix(solid-server): repair failing js test on main

jeswr:
> 🤖 SPARQ agent

Fixes the js CI failure on main (@jeswr/solid-server test suite). Authored headless on a delegated worker account (Fable); host re-ran the suite green before publish. Not armed — review pipeline next.

node --test runs test files concurrently, and both server.test.mjs and
smoke.test.mjs boot the server via `npx --yes --package <packageDir>` — the
identical package spec hashes to the same npx cache key. On a cold cache
(every CI runner; js.yml's setup-node configures no npm cache) the two npm
installs race to create the same symlink in ~/.npm/_npx/<hash>/node_modules/
@jeswr/solid-server, and the loser exits with EEXIST before printing the
readiness line, failing waitForStartup (js.yml run 29618574335; the next
main run passed on the same package tree, confirming a timing race, not a
deterministic break — no solid-server commit is at fault).

Fix the root cause instead of the symptom: give each spawned npx child its
own throwaway npm cache (mkdtemp + npm_config_cache, removed in finally),
so the two concurrent installs can never collide. The tests still exercise
the real npx boot path end to end; no assertion is weakened or skipped.

### DIFF
--- a/packages/solid-server/test/server.test.mjs
+++ b/packages/solid-server/test/server.test.mjs
@@ -2,7 +2,9 @@
 import assert from 'node:assert/strict';
 import { once } from 'node:events';
 import { spawn } from 'node:child_process';
-import { dirname, resolve } from 'node:path';
+import { mkdtemp, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import test from 'node:test';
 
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 SPARQ agent

**Cross-provider review by GPT-5.6 codex — VERDICT: pass**

1. `mkdtemp()` guarantees a unique cache dir per test, eliminating the shared `_npx` install race; cleanup runs after child termination incl. failure paths.
2. No assertions or exercised behavior changed; both tests retain full npx-boot + Solid/WAC coverage.
3. Portable (`tmpdir()`/`join()`); cache override scoped to the spawned process via copied env — no cross-test leakage.

Issue #3419, opened 2026-07-17
scripts/retriage.py --self-test crashes with KeyError: 9 on main (self-test indexes a missing actions entry)


jeswr:
> 🤖 SPARQ agent

Out-of-scope discovery while implementing the UI-ownership routing rule (PR #3416): on a clean `main` checkout, `python3 scripts/retriage.py --self-test` crashes:

```
    chk("bot author promoted", actions[9][0][0].startswith(("priority", "role", "status")), True)
                               ~~~~~~~^^^
KeyError: 9
```

The self-test indexes `actions[9]`, which no longer exists — likely a fixture/index drift after a retriage change. Pre-existing (reproduced with no local diff); not caused by and not fixed in #3416. If this self-test is part of any gate lane it is currently red on main.

PR #3426, opened 2026-07-17
ci: pr-freshness — auto update-branch for armed stale-red PRs on main movement

### LINKED ISSUE #3424: repair loop: after a gate-code fix merges, refresh stale-red armed PRs via update-branch (rerun replays the old bug)
> 🤖 SPARQ agent

Lesson from 2026-07-17: after #3417 fixed ci_summary_gate.py, ~17 armed review:pass PRs kept stale FAILURE gates. `gh run rerun` re-executes with the PR branch's OLD gate script — it replays the bug. The repair loop must detect (required-check FAILURE + armed + gate-code changed on main since the run) and issue **update-branch** (or a rebase) instead of a rerun, batched ≤6 to respect runner capacity.

jeswr:
> 🤖 SPARQ agent

Closes #3424. After a gate-code fix merges to main, armed/review:pass PRs keep stale-red gates; `run rerun` replays the old script. This workflow (push-to-main + 30-min cron) update-branches behind+red armed PRs — capped 6/tick, 40-min per-PR cooldown, skips review:needs-user / DIRTY / already-running CI. Self-tested (mutation-checked guards).

Authored headless on a delegated worker account (Fable); host-verified (self-test + scoped actionlint — note: repo-wide actionlint has PRE-EXISTING shellcheck infos in ci.yml, untouched here). Not armed — review pipeline next.

### DIFF
--- a/.github/workflows/pr-freshness.yml
+++ b/.github/workflows/pr-freshness.yml
@@ -84,6 +84,13 @@ jobs:
         with:
           app-id: ${{ secrets.ORCHESTRATOR_APP_ID }}
           private-key: ${{ secrets.ORCHESTRATOR_APP_PRIVATE_KEY }}
+          # Without permission-* inputs the minted token carries EVERY permission the
+          # App installation has; scope it to exactly what the refresh needs (mirror of
+          # the workflow-level `permissions` block above).
+          permission-actions: read
+          permission-checks: read
+          permission-contents: write
+          permission-pull-requests: write
 
       - name: Checkout (only the policy script is needed)
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
--- a/.github/workflows/pr-freshness.yml
+++ b/.github/workflows/pr-freshness.yml
@@ -1,0 +1,120 @@
+# [FABLE-5] PR-freshness refresher (on-main-push + 30-min cron; NON-gating). Issue #3424.
+#
+# WHY: when a fix to CI/gate code merges to `main`, every open PR keeps its stale-red
+# required `gate` (ci-summary.yml) — the verdict was computed with the OLD gate code on an
+# OLD merge ref. `gh run rerun` does NOT refresh it (a rerun re-executes the PR branch's
+# old scripts on the old merge commit). The correct refresh is the update-branch API
+# (PUT /repos/{o}/{r}/pulls/{n}/update-branch), which pushes a merge commit with current
+# `main` to the PR branch and triggers fresh pull_request CI on a rebuilt merge ref. On
+# 2026-07-17 ~17 armed review:pass PRs sat red for an hour because nothing did this
+# automatically — this workflow is the automation (Closes #3424).
+#
+# POLICY lives in scripts/pr-freshness.py (pure, self-tested — same shape as
+# scripts/pr-backlog.py): refresh an open non-draft PR iff (auto-merge armed OR
+# review:pass) AND behind main AND (`gate` red on the head OR the merge ref was never
+# gated) AND NOT review:needs-user AND NOT DIRTY (conflict PRs are reported for
+# sparq-merge-fixer). Never touches a head whose fresh CI is already queued/running.
+#
+# LOOP-SAFETY: update-branch pushes a merge commit to the PR BRANCH — that fires the PR's
+# pull_request CI but NOT another push to `main`, so the `push: main` trigger below cannot
+# recurse. CI fan-out per tick is bounded by the script's cap (6 update-branch calls/tick)
+# + a 40-min per-PR cooldown (read from the head commit's committer timestamp — an
+# update-branch merge commit IS the branch's last-update marker, so no external state).
+# Both knobs exist because of the congestion collapse ci-summary.yml documents
+# (sq-90cv4): uncapped mass CI triggering once melted the runner pool.
+#
+# TOKENS: an update-branch push made with the workflow's own GITHUB_TOKEN fires NO events
+# (GitHub suppresses Actions-token event cascades) — the fresh CI this workflow exists to
+# trigger would never run and the PR would be left gate-less. So the mutation runs ONLY
+# with the sparq-orchestrator App token (same pattern as batch-merge.yml); without the
+# App secrets the run degrades fail-soft to --report-only.
+#
+# NOT A GATE: runs on push-to-main / schedule / dispatch, never on a PR head commit, so
+# the ci-summary aggregator (which polls check-runs on PR heads) never sees it.
+name: pr-freshness
+
+on:
+  push:
+    branches: [main]
+  schedule:
+    # Every 30 min: the backstop for gate-code merges the push trigger raced past
+    # (e.g. a PR that was still mid-CI on the push tick) and for cap-deferred PRs.
+    - cron: "*/30 * * * *"
+  workflow_dispatch:
+    inputs:
+      report_only:
+        description: "Compute + print the plan but perform NO update-branch calls"
+        required: false
+        default: "false"
+
+# Least privilege. update-branch = a push to the PR branch (contents: write) on a PR
+# resource (pull-requests: write); the policy READS check-runs (checks: read) and the
+# head's queued/in-progress workflow runs (actions: read). Nothing else.
+permissions:
+  pull-requests: write
+  contents: write
+  checks: read
+  actions: read
+
+# Never overlap (two ticks would double-spend the cap); never cancel in-flight — let a
+# push tick and the cron drain sequentially.
+concurrency:
+  group: pr-freshness
+  cancel-in-progress: false
+
+jobs:
+  # Job name deliberately WITHOUT `advisory`/`informational` (stays visible in the
+  # Actions UI); because this workflow never runs on a PR head commit it is structurally
+  # non-gating regardless.
+  freshen:
+    name: freshen stale-red PRs
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    env:
+      # Secrets are not readable from step `if:` expressions — mirror the id into env so
+      # the mint step can be skipped fail-soft when the App is not configured.
+      ORCHESTRATOR_APP_ID: ${{ secrets.ORCHESTRATOR_APP_ID }}
+    steps:
+      # See the TOKENS header note: without this token an update-branch push would fire
+      # no pull_request events, leaving the PR gate-less — so no-token => report-only.
+      - name: Mint the sparq-orchestrator App token (update pushes must fire CI)
+        id: app-token
+        if: env.ORCHESTRATOR_APP_ID != ''
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.ORCHESTRATOR_APP_ID }}
+          private-key: ${{ secrets.ORCHESTRATOR_APP_PRIVATE_KEY }}
+
+      - name: Checkout (only the policy script is needed)
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+          sparse-checkout: |
+            scripts/pr-freshness.py
+          sparse-checkout-cone-mode: false
+
+      - name: Set up Python
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: "3.12"
+
+      - name: Self-test the policy (hermetic; gh stubbed; fail fast on a logic regression)
+        run: python3 scripts/pr-freshness.py --self-test
+
+      - name: Refresh stale-red PRs
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token || github.token }}
+          HAVE_APP_TOKEN: ${{ steps.app-token.outputs.token != '' }}
+          REPO: ${{ github.repository }}
+          REPORT_ONLY: ${{ github.event.inputs.report_only || 'false' }}
+        run: |
+          set -euo pipefail
+          args=(--repo "$REPO" --summary-file "$GITHUB_STEP_SUMMARY")
+          if [ "$REPORT_ONLY" = "true" ]; then
+            args+=(--report-only)
+          fi
+          if [ "$HAVE_APP_TOKEN" != "true" ]; then
+            echo "::notice::ORCHESTRATOR_APP_ID/_PRIVATE_KEY not configured — report-only mode (a GITHUB_TOKEN update-branch push fires no pull_request events, so the fresh CI it exists to trigger would never run)."
+            args+=(--report-only)
+          fi
+          python3 scripts/pr-freshness.py "${args[@]}"
--- a/scripts/pr-freshness.py
+++ b/scripts/pr-freshness.py
@@ -1,0 +1,462 @@
+#!/usr/bin/env python3
+# [FABLE-5] PR-freshness refresher (scheduled + on-main-push; NON-gating). Issue #3424.
+#
+# WHAT THIS IS
+# ------------
+# The deterministic policy core of .github/workflows/pr-freshness.yml. When a fix to CI/gate
+# code merges to `main`, every open PR keeps its stale-red required `gate` — the gate was
+# computed with the OLD gate code on an OLD merge ref. `gh run rerun` does NOT help: a rerun
+# re-executes the PR's old scripts on the old merge commit. The ONLY correct refresh is the
+# update-branch API (PUT /repos/{o}/{r}/pulls/{n}/update-branch), which pushes a merge commit
+# with current `main` to the PR branch and thereby triggers a fresh pull_request CI run on a
+# rebuilt merge ref. On 2026-07-17 ~17 armed review:pass PRs sat red for an hour because
+# nothing did this automatically (issue #3424) — this script is the automation.
+#
+# POLICY (issue #3424)
+# --------------------
+# A PR is refreshed iff ALL of:
+#   (a) it is OPEN, NON-draft, and (auto-merge is armed OR it carries label `review:pass`);
+#   (b) its branch is BEHIND main (compare(head...main).ahead_by > 0);
+#   (c) the latest `gate` check-run on its head CONCLUDED red (failure/cancelled/timed_out/
+#       action_required/startup_failure/stale) OR the head has NO `gate` check-run at all
+#       (the merge ref was never gated);
+#   (d) it does NOT carry label `review:needs-user`, and it is NOT DIRTY (merge conflict —
+#       update-branch would 422; DIRTY PRs are REPORTED for the merge-fixer instead).
+# and NONE of:
+#   - fresh CI already queued/in-progress on its CURRENT head (a non-terminal `gate`
+#     check-run, or any queued/in-progress workflow run on the head SHA) — never
+#     cancel-churn a run that is already recomputing the verdict on a current merge ref;
+#   - COOLDOWN: the head commit is younger than COOLDOWN_MIN minutes (see below);
+#   - the per-tick CAP (MAX_UPDATES_PER_TICK) is already spent — the remainder is skipped
+#     WITH a log line ("truncated"), never silently.
+#
+# COOLDOWN MECHANISM (the simplest robust one — documented per #3424)
+# -------------------------------------------------------------------
+# update-branch pushes a fresh merge commit to the PR branch, so the head commit's
+# COMMITTER timestamp *is* the branch's last-update time. We read it from the same
+# compare() call that measures behind-ness (compare base_commit == the head commit) and
+# skip any PR whose head commit is < COOLDOWN_MIN old. No marker comments, no external
+# state, nothing to garbage-collect; an author's own recent push also (correctly) counts
+# as "recently refreshed" — its CI is at most COOLDOWN_MIN stale.
+#
+# LOOP-SAFETY / CONGESTION SAFETY
+# -------------------------------
+# update-branch pushes a merge commit to the PR BRANCH, which triggers that PR's
+# pull_request CI but NOT another push to main — so the workflow's on-main-push trigger
+# cannot recurse. The per-tick cap + per-PR cooldown bound the CI fan-out per tick
+# (ci-summary.yml's congestion-collapse doctrine, bead sq-90cv4: uncapped mass triggering
+# once melted the runner pool; never re-create that).
+#
+# DESIGN: the policy is a PURE FUNCTION plan(prs, now) -> [Action] over pre-enriched PR
+# dicts. An Action carries the exact `gh` argv it maps to. `live` mode enriches candidates
+# with 1 compare + 1 check-runs + 1 runs-list API call each, then runs the argv through a
+# real gh; `--self-test` runs plan() over fixtures with gh STUBBED — no live mutation is
+# possible from a test, and every guard has a fixture that turns red if the guard is
+# deleted (non-vacuity).
+#
+# USAGE
+#   scripts/pr-freshness.py --repo owner/repo                # live refresh
+#   scripts/pr-freshness.py --repo owner/repo --report-only  # print plan, no gh mutations
+#   scripts/pr-freshness.py --self-test                      # hermetic; gh stubbed
+#
+# Exit 0 on success; non-zero only on a real error or a failed self-test.
+import argparse
+import json
+import re
+import subprocess
+import sys
+from dataclasses import dataclass, field
+from datetime import datetime, timedelta, timezone
+
+PROG = "pr-freshness"
+
+# Congestion-safety knobs (see LOOP-SAFETY above). Cap chosen with the maintainer in
+# issue #3424; cooldown > the typical gate wall-clock so one PR is never churned twice
+# while its fresh run is still converging.
+MAX_UPDATES_PER_TICK = 6
+COOLDOWN_MIN = 40
+
+ARMED_LABEL = "review:pass"
+BLOCK_LABEL = "review:needs-user"
+
+# Terminal-red gate conclusions (same set ci-summary's aggregator treats as failing).
+_GATE_RED = {"failure", "cancelled", "timed_out", "action_required", "startup_failure", "stale"}
+_GATE_GREEN = {"success", "neutral", "skipped"}
+
+
+def log(msg: str) -> None:
+    print(f"[{PROG}] {msg}", file=sys.stderr)
+
+
+def _parse_iso(s: str) -> datetime:
+    s = (s or "").strip()
+    if s.endswith("Z"):
+        s = s[:-1] + "+00:00"
+    dt = datetime.fromisoformat(s)
+    if dt.tzinfo is None:
+        dt = dt.replace(tzinfo=timezone.utc)
+    return dt
+
+
+# --------------------------------------------------------------------------------------
+# The action model (same shape as scripts/pr-backlog.py). `update` rows carry the
+# update-branch argv; `skip` / `dirty` rows are report-only for the step summary.
+# --------------------------------------------------------------------------------------
+@dataclass
+class Action:
+    kind: str                       # update | skip | dirty
+    pr: int
+    reason: str
+    argv: list = field(default_factory=list)
+
+    def run(self, gh) -> None:
+        if self.argv:
+            gh(self.argv)
+
+
+def update_argv(repo: str, number: int, head_sha: str) -> list:
+    # expected_head_sha: if the head moved between observation and this call, GitHub
+    # answers 422 instead of updating a branch we never assessed. The runner treats a
+    # per-PR gh failure as a warning, so a lost race degrades to "caught next tick".
+    return ["api", "-X", "PUT", f"repos/{repo}/pulls/{number}/update-branch",
+            "-f", f"expected_head_sha={head_sha}"]
+
+
+# --------------------------------------------------------------------------------------
+# THE POLICY (pure). Each pr dict carries:
+#   number:int  is_draft:bool  labels:[str]  armed:bool  merge_state:str  head_sha:str
+# and, for PRs that survive the cheap guards (enriched live only for those):
+#   behind_by:int  gate:'failure'|'success'|'pending'|'missing'
+#   active_runs:int  head_committed_at:iso-str
+# Guard order matters: cheap label/state guards fire before enrichment fields are read,
+# so live mode may leave enrichment fields absent on cheap-skipped PRs.
+# --------------------------------------------------------------------------------------
+def plan(prs, repo, now, cap=MAX_UPDATES_PER_TICK):
+    actions = []
+    selected = 0
+    for pr in prs:
+        num = pr["number"]
+        labels = [l.lower() for l in pr.get("labels", [])]
+
+        # (a)/(d) cheap guards — no enrichment needed.
+        if pr.get("is_draft"):
+            actions.append(Action("skip", num, "draft"))
+            continue
+        if BLOCK_LABEL in labels:
+            actions.append(Action("skip", num, f"label {BLOCK_LABEL}"))
+            continue
+        if not (pr.get("armed") or ARMED_LABEL in labels):
+            actions.append(Action("skip", num, f"not armed and no {ARMED_LABEL} label"))
+            continue
+        if (pr.get("merge_state") or "").upper() == "DIRTY":
+            actions.append(Action("dirty", num,
+                                  "merge conflict — needs sparq-merge-fixer, not update-branch"))
+            continue
+
+        # (b) behind main.
+        if int(pr.get("behind_by") or 0) <= 0:
+            actions.append(Action("skip", num, "not behind main"))
+            continue
+
+        # (c) gate red or absent — a green/converging gate is left alone.
+        gate = pr.get("gate") or "missing"
+        if gate == "success":
+            actions.append(Action("skip", num, "gate is green"))
+            continue
+        if gate == "pending":
+            actions.append(Action("skip", num, "fresh CI in progress (gate not terminal)"))
+            continue
+
+        # Never churn a head whose fresh CI is already queued/running (a just-created
+        # run may not have registered its `gate` check-run yet — the runs list catches it).
+        if int(pr.get("active_runs") or 0) > 0:
+            actions.append(Action("skip", num, "fresh CI in progress (runs on head)"))
+            continue
+
+        # COOLDOWN (see header): head commit younger than COOLDOWN_MIN => recently
+        # updated/pushed; let that CI cycle land before touching the branch again.
+        committed = pr.get("head_committed_at")
+        if committed and now - _parse_iso(committed) < timedelta(minutes=COOLDOWN_MIN):
+            actions.append(Action("skip", num, f"cooldown (<{COOLDOWN_MIN}m since last update)"))
+            continue
+
+        # CAP — skip WITH a log line, never silently (congestion-collapse doctrine).
+        if selected >= cap:
+            actions.append(Action("skip", num, f"truncated (cap {cap}/tick)"))
+            continue
+
+        selected += 1
+        actions.append(Action("update", num, f"behind + gate {gate} -> update-branch",
+                              update_argv(repo, num, pr.get("head_sha", ""))))
+
+    truncated = sum(1 for a in actions if a.reason.startswith("truncated"))
+    if truncated:
+        log(f"cap {cap}/tick reached — {truncated} eligible PR(s) deferred to the next tick")
+    return actions
+
+
+# --------------------------------------------------------------------------------------
+# LIVE data collection (real gh). Not exercised by the self-test (gh stubbed there).
+# --------------------------------------------------------------------------------------
+def _gh_json(argv):
+    out = subprocess.check_output(["gh"] + argv, text=True)
+    return json.loads(out)
+
+
+def collect_prs(repo):
+    prs = _gh_json([
+        "pr", "list", "--repo", repo, "--state", "open", "--limit", "200",
+        "--json", "number,isDraft,labels,autoMergeRequest,mergeStateStatus,headRefOid",
+    ])
+    norm = []
+    for p in prs:
+        norm.append({
+            "number": p["number"],
+            "is_draft": bool(p.get("isDraft")),
+            "labels": [l.get("name", "") for l in (p.get("labels") or [])],
+            "armed": p.get("autoMergeRequest") is not None,
+            "merge_state": p.get("mergeStateStatus", ""),
+            "head_sha": p.get("headRefOid", ""),
+        })
+    return norm
+
+
+def _cheaply_eligible(pr) -> bool:
+    """Mirror of plan()'s cheap guards — decides which PRs are worth 3 enrichment calls."""
+    labels = [l.lower() for l in pr.get("labels", [])]
+    return (not pr.get("is_draft")
+            and BLOCK_LABEL not in labels
+            and (pr.get("armed") or ARMED_LABEL in labels)
+            and (pr.get("merge_state") or "").upper() != "DIRTY")
+
+
+def enrich(repo, pr):
+    """3 API calls: behind-ness + head-commit time (one compare), gate check, active runs."""
+    sha = pr["head_sha"]
+
+    # compare(head...main): ahead_by == commits main has that head lacks (= behind-ness),
+    # and base_commit == the head commit itself, carrying the committer timestamp the
+    # cooldown reads. One call, two signals.
+    cmp_ = _gh_json(["api", f"repos/{repo}/compare/{sha}...main"])
+    pr["behind_by"] = int(cmp_.get("ahead_by") or 0)
+    pr["head_committed_at"] = (((cmp_.get("base_commit") or {}).get("commit") or {})
+                               .get("committer") or {}).get("date")
+
+    # Latest check-run named exactly `gate` on the head (draft-tier runs render a
+    # different name, `gate, draft-tier`, and drafts are excluded anyway).
+    checks = _gh_json(["api", f"repos/{repo}/commits/{sha}/check-runs?per_page=100"])
+    gates = sorted((c for c in checks.get("check_runs", [])
+                    if (c.get("name") or "").strip().lower() == "gate"),
+                   key=lambda c: c.get("started_at") or "")
+    if not gates:
+        pr["gate"] = "missing"
+    else:
+        g = gates[-1]
+        if (g.get("status") or "").lower() != "completed":
+            pr["gate"] = "pending"
+        else:
+            concl = (g.get("conclusion") or "").lower()
+            pr["gate"] = ("failure" if concl in _GATE_RED
+                          else "success" if concl in _GATE_GREEN
+                          else "pending")
+
+    runs = _gh_json(["api", f"repos/{repo}/actions/runs?head_sha={sha}&per_page=30"])
+    pr["active_runs"] = sum(1 for r in runs.get("workflow_runs", [])
+                            if r.get("status") in ("queued", "in_progress", "waiting"))
+    return pr
+
+
+# --------------------------------------------------------------------------------------
+# Step-summary table (GitHub Actions job summary, or stdout locally).
+# --------------------------------------------------------------------------------------
+def render_summary(actions) -> str:
+    lines = [
+        "## pr-freshness refresher",
+        "",
+        "| PR | action | reason |",
+        "| --- | --- | --- |",
+    ]
+    for a in actions:
+        lines.append(f"| #{a.pr} | {a.kind} | {a.reason} |")
+    dirty = [a.pr for a in actions if a.kind == "dirty"]
+    if dirty:
+        lines += ["", "**DIRTY (conflict) PRs for sparq-merge-fixer:** "
+                  + ", ".join(f"#{n}" for n in dirty)]
+    lines.append("")
+    return "\n".join(lines)
+
+
+def real_gh(argv) -> None:
+    subprocess.run(["gh"] + argv, check=True)
+
+
+def main() -> int:
+    ap = argparse.ArgumentParser(description="PR-freshness refresher (pure policy).")
+    ap.add_argument("--repo", help="owner/repo (required unless --self-test).")
+    ap.add_argument("--report-only", action="store_true",
+                    help="Fetch state + print the plan/summary, run NO gh mutations. "
+                         "(The workflow forces this when no App token is available: an "
+                         "update-branch push made with GITHUB_TOKEN fires NO events, so "
+                         "the fresh CI it exists to trigger would never run.)")
+    ap.add_argument("--self-test", action="store_true",
+                    help="Run the hermetic self-test (gh stubbed) and exit.")
+    ap.add_argument("--summary-file", default=None,
+                    help="Write the step-summary table here (e.g. $GITHUB_STEP_SUMMARY).")
+    args = ap.parse_args()
+
+    if args.self_test:
+        return self_test()
+    if not args.repo:
+        raise SystemExit(f"[{PROG}] ERROR: --repo is required (unless --self-test).")
+
+    prs = collect_prs(args.repo)
+    for pr in prs:
+        if _cheaply_eligible(pr):
+            enrich(args.repo, pr)
+    actions = plan(prs, args.repo, datetime.now(timezone.utc))
+
+    summary = render_summary(actions)
+    if args.summary_file:
+        with open(args.summary_file, "a", encoding="utf-8") as f:
+            f.write(summary + "\n")
+    print(summary)
+
+    if args.report_only:
+        log("--report-only: no gh mutations performed.")
+        return 0
+
+    for a in actions:
+        try:
+            a.run(real_gh)
+        except subprocess.CalledProcessError as e:
+            # A 422 here is a lost expected_head_sha race or an already-updated branch —
+            # both self-heal on the next tick. Never fail the whole run for one PR.
+            log(f"WARNING: update-branch failed for PR #{a.pr}: {e} — continuing.")
+    log(f"applied {sum(1 for a in actions if a.argv)} update-branch call(s).")
+    return 0
+
+
+# --------------------------------------------------------------------------------------
+# Hermetic self-test. gh is a recording stub — NO live mutation is possible. Every guard
+# has a fixture that is selectable EXCEPT for that one guard, so deleting any guard turns
+# at least one check red (non-vacuity).
+# --------------------------------------------------------------------------------------
+def self_test() -> int:
+    fails = 0
+
+    def check(label, cond):
+        nonlocal fails
+        if cond:
+            print(f"  ok   {label}")
+        else:
+            print(f"  FAIL {label}")
+            fails += 1
+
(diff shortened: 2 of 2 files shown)


jeswr:
> 🤖 SPARQ agent — cross-provider review by GPT-5.6 codex (parallel batch)

validation or secondary-rate-limit failure.

5. The mutation-testing claim is incomplete. [self_test()](/tmp/cxr-sparq-org-sparq-3426/scripts/pr-freshness.py:344) supplies already-normalized fixtures directly to `plan()` and never exercises `collect_prs()`, `_cheaply_eligible()`, or `enrich()`. Mutations to base selection, `ahead_by`, red-gate classification, or workflow-run status handling all survive. Add raw API-response tests through the enrichment boundary.

VERDICT: needs-changes
tokens used
104,692
1. [scripts/pr-freshness.py](/tmp/cxr-sparq-org-sparq-3426/scripts/pr-freshness.py:206) never collects or checks `baseRefName`. A `review:pass` stacked PR targeting a branch other than `main` can therefore qualify as “behind main,” but GitHub’s update-branch endpoint merges the PR’s actual base—not `main`. It may repeatedly return 422, consume the six-PR cap, and never refresh the intended CI. Require `baseRefName == "main"` and test this guard. [GitHub endpoint semantics](https://docs.github.com/en/rest/pulls/pulls#update-a-pull-request-branch)

2. [scripts/pr-freshness.py](/tmp/cxr-sparq-org-sparq-3426/scripts/pr-freshness.py:263) counts only `queued`, `in_progress`, and `waiting` runs. GitHub also exposes nonterminal `requested` and `pending` states. With a missing gate check, either state produces `active_runs == 0` and incorrectly selects update-branch—the precise race this guard is meant to prevent. Treat every status other than `completed` as active. [GitHub workflow-run statuses](https://docs.github.com/en/rest/actions/workflow-runs#list-workflow-runs-for-a-repository)

3. [pr-freshness.yml](/tmp/cxr-sparq-org-sparq-3426/.github/workflows/pr-freshness.yml:83) does not restrict the minted App token. The workflow-level `permissions` block constrains `GITHUB_TOKEN`, while `create-github-app-token` defaults to every permission granted to the App installation. Explicitly request only `actions:read`, `checks:read`, `contents:write`, and `pull-requests:write`. [Action permission behavior](https://github.com/actions/create-github-app-token#inputs)

4. [scripts/pr-freshness.py](/tmp/cxr-sparq-org-sparq-3426/scripts/pr-freshness.py:328) suppresses every update failure, including authentication/permission failures, and then reports attempted calls as “applied.” A broken token can therefore leave the workflow permanently green while updating nothing. Continue processing other PRs, but return nonzero afterward for unexpected failures; even 422 is not exclusively a head race—it can indicate validation or secondary-rate-limit failure.

5. The mutation-testing claim is incomplete. [self_test()](/tmp/cxr-sparq-org-sparq-3426/scripts/pr-freshness.py:344) supplies already-normalized fixtures directly to `plan()` and never exercises `collect_prs()`, `_cheaply_eligible()`, or `enrich()`. Mutations to base selection, `ahead_by`, red-gate classification, or workflow-run status handling all survive. Add raw API-response tests through the enrichment boundary.

VERDICT: needs-changes
=== review exit: 0 ===

jeswr:
> 🤖 SPARQ agent — GPT-5.6 codex cross-provider review (round posted 2026-07-18)

I’ll verify the prior findings against the actual checkout, then audit loop safety, API semantics, token scope, and failure handling end to end. I’ll also run the hermetic checks and targeted static analysis where the read-only workspace permits it.
exec
/bin/bash -lc "git status --short && git diff --check main...HEAD && git diff --stat main...HEAD && git rev-parse --short HEAD && git rev-parse --short main && rg -n \"create-github-app-token|update-branch|pr-freshness|secondary-rate|rate.limit\" .github scripts -g '*.yml' -g '*.yaml' -g '*.py'" in /tmp/cxr-sparq-org-sparq-3426
 succeeded in 113ms:
.github/feature-matrix.d/sparq-lws-core.yml:42:# sockets, response parity, and peer ConnectInfo-backed rate limiting.
.github/workflows/pr-freshness.yml:6:# old scripts on the old merge commit). The correct refresh is the update-branch API
.github/workflows/pr-freshness.yml:7:# (PUT /repos/{o}/{r}/pulls/{n}/update-branch), which pushes a merge commit with current
.github/workflows/pr-freshness.yml:12:# POLICY lives in scripts/pr-freshness.py (pure, self-tested — same shape as
.github/workflows/pr-freshness.yml:18:# LOOP-SAFETY: update-branch pushes a merge commit to the PR BRANCH — that fires the PR's
.github/workflows/pr-freshness.yml:20:# recurse. CI fan-out per tick is bounded by the script's cap (6 update-branch calls/tick)
.github/workflows/pr-freshness.yml:22:# update-branch merge commit IS the branch's last-update marker, so no external state).
.github/workflows/pr-freshness.yml:26:# TOKENS: an update-branch push made with the workflow's own GITHUB_TOKEN fires NO events
.github/workflows/pr-freshness.yml:34:name: pr-freshness
.github/workflows/pr-freshness.yml:46:        description: "Compute + print the plan but perform NO update-branch calls"
.github/workflows/pr-freshness.yml:50:# Least privilege. update-branch = a push to the PR branch (contents: write) on a PR
.github/workflows/pr-freshness.yml:62:  group: pr-freshness
.github/workflows/pr-freshness.yml:78:      # See the TOKENS header note: without this token an update-branch push would fire
.github/workflows/pr-freshness.yml:83:        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
.github/workflows/pr-freshness.yml:93:            scripts/pr-freshness.py
.github/workflows/pr-freshness.yml:102:        run: python3 scripts/pr-freshness.py --self-test
.github/workflows/pr-freshness.yml:117:            echo "::notice::ORCHESTRATOR_APP_ID/_PRIVATE_KEY not configured — report-only mode (a GITHUB_TOKEN update-branch push fires no pull_request events, so the fresh CI it exists to trigger would never run)."
.github/workflows/pr-freshness.yml:120:          python3 scripts/pr-freshness.py "${args[@]}"
.github/workflows/bench-ec2.yml:4:#   * rate limit  — runs ONLY on cron (weekly heavy + nightly full-suite) + manual dispatch; NEVER on
.github/workflows/batch-merge.yml:84:        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
scripts/pr-freshness.py:6:# The deterministic policy core of .github/workflows/pr-freshness.yml. When a fix to CI/gate
scripts/pr-freshness.py:10:# update-branch API (PUT /repos/{o}/{r}/pulls/{n}/update-branch), which pushes a merge commit
scripts/pr-freshness.py:24:#       update-branch would 422; DIRTY PRs are REPORTED for the merge-fixer instead).
scripts/pr-freshness.py:35:# update-branch pushes a fresh merge commit to the PR branch, so the head commit's
scripts/pr-freshness.py:44:# update-branch pushes a merge commit to the PR BRANCH, which triggers that PR's
scripts/pr-freshness.py:58:#   scripts/pr-freshness.py --repo owner/repo                # live refresh
scripts/pr-freshness.py:59:#   scripts/pr-freshness.py --repo owner/repo --report-only  # print plan, no gh mutations
scripts/pr-freshness.py:60:#   scripts/pr-freshness.py --self-test                      # hermetic; gh stubbed
scripts/pr-freshness.py:71:PROG = "pr-freshness"
scripts/pr-freshness.py:103:# update-branch argv; `skip` / `dirty` rows are report-only for the step summary.
scripts/pr-freshness.py:121:    return ["api", "-X", "PUT", f"repos/{repo}/pulls/{number}/update-branch",
scripts/pr-freshness.py:153:                                  "merge conflict — needs sparq-merge-fixer, not update-branch"))
scripts/pr-freshness.py:189:        actions.append(Action("update", num, f"behind + gate {gate} -> update-branch",
scripts/pr-freshness.py:274:        "## pr-freshness refresher",
scripts/pr-freshness.py:299:                         "update-branch push made with GITHUB_TOKEN fires NO events, so "
scripts/pr-freshness.py:334:            log(f"WARNING: update-branch failed for PR #{a.pr}: {e} — continuing.")
scripts/pr-freshness.py:335:    log(f"applied {sum(1 for a in actions if a.argv)} update-branch call(s).")
scripts/pr-freshness.py:402:    check("#1 argv is PUT pulls/1/update-branch",
scripts/pr-freshness.py:403:          a1[:3] == ["api", "-X", "PUT"] and f"repos/{repo}/pulls/1/update-branch" in a1)

exec
/bin/bash -lc "sed -n '1,150p' .github/workflows/batch-merge.yml && sed -n '1,240p' .github/workflows/ci-summary.yml && nl -ba scripts/pr-freshness.py | sed -n '115,350p' && nl -ba .github/workflows/pr-freshness.yml | sed -n '45,125p' && command -v actionlint || true && command -v shellcheck || true && python3 scripts/pr-freshness.py --self-test && python3 -m py_compile scripts/pr-freshness.py" in /tmp/cxr-sparq-org-sparq-3426
 exited 1 in 114ms:
[Errno 30] Read-only file system: 'scripts/__pycache__'
codex
The local self-test passes, but it confirms only the pure `plan()` fixtures; it does not exercise the live normalization/enrichment path. Three earlier safety gaps are still present in the checked-out head: no `baseRefName` guard, incomplete nonterminal-run handling, and an unrestricted App token. I’m checking the endpoint and token semantics now and probing for additional loop/failure-path defects.
web search: 
web search: site:docs.github.com/en/rest/pulls/pulls update a pull request branch expected_head_sha fine-grained access token permissions ...
web search: 
web search: 'Update a pull request branch'
exec
/bin/bash -lc "actionlint -format '{{range "'$err := .}}{{$err.Filepath}}:{{$err.Line}}:{{$err.Column}}: {{$err.Message}} [{{$err.Kind}}]'"\\n{{end}}' .github/workflows/pr-freshness.yml; shellcheck -x -s bash /dev/stdin <<'EOF'
set -euo pipefail
args=(--repo \""'$REPO" --summary-file "$GITHUB_STEP_SUMMARY")
if [ "$REPORT_ONLY" = "true" ]; then
  args+=(--report-only)
fi
if [ "$HAVE_APP_TOKEN" != "true" ]; then
  echo "::notice::ORCHESTRATOR_APP_ID/_PRIVATE_KEY not configured — report-only mode"
  args+=(--report-only)
fi
python3 scripts/pr-freshness.py "${args[@]}"
EOF

git log --oneline --decorate main..HEAD && git show --format=fuller --no-patch HEAD && git diff --unified=80 main...HEAD -- scripts/pr-freshness.py .github/workflows/pr-freshness.yml' in /tmp/cxr-sparq-org-sparq-3426
 succeeded in 0ms:
5a76eeb1 (HEAD -> ci/pr-freshness-auto-update-branch, origin/ci/pr-freshness-auto-update-branch) ci: pr-freshness — auto-update-branch stale-red armed PRs after gate-code merges to main
commit 5a76eeb1f2a90229728a493a5f55fe567f332021
Author:     Jesse Wright <jmwright.045@gmail.com>
AuthorDate: Fri Jul 17 23:36:27 2026 +0000
Commit:     Jesse Wright <jmwright.045@gmail.com>
CommitDate: Fri Jul 17 23:36:27 2026 +0000

    ci: pr-freshness — auto-update-branch stale-red armed PRs after gate-code merges to main
    
    When CI/gate fixes merge to main, open PRs keep a stale-red required gate
    computed with the old gate code on an old merge ref; gh run rerun cannot
    refresh it. New pr-freshness.yml (push-to-main + 30-min cron + dispatch)
    runs the pure, self-tested policy in scripts/pr-freshness.py: open
    non-draft PRs that are (armed OR review:pass) AND behind main AND red/
    ungated on the head AND not review:needs-user/DIRTY get the update-branch
    API call, capped at 6/tick with a 40-min per-PR cooldown read from the
    head commit's committer timestamp (no external state). DIRTY PRs are
    reported for sparq-merge-fixer. Mutations run only under the
    sparq-orchestrator App token (a GITHUB_TOKEN update push fires no events);
    without it the run degrades to report-only. Guards are mutation-checked
    non-vacuous; actionlint passes.
    
    Closes #3424
    
diff --git a/.github/workflows/pr-freshness.yml b/.github/workflows/pr-freshness.yml
new file mode 100644
index 00000000..94f53ed4
--- /dev/null
+++ b/.github/workflows/pr-freshness.yml
@@ -0,0 +1,120 @@
+# [FABLE-5] PR-freshness refresher (on-main-push + 30-min cron; NON-gating). Issue #3424.
+#
+# WHY: when a fix to CI/gate code merges to `main`, every open PR keeps its stale-red
+# required `gate` (ci-summary.yml) — the verdict was computed with the OLD gate code on an
+# OLD merge ref. `gh run rerun` does NOT refresh it (a rerun re-executes the PR branch's
+#

jeswr:
> 🤖 SPARQ agent

Fix round pushed (delegated Fable worker; host-verified scoped tests + actionlint). Finding→fix map in the commit body. Ready for re-review.

jeswr:
> 🤖 SPARQ agent — GPT-5.6 codex cross-provider re-review

… 637	
   638	    enriched = [enrich(repo, p, gh_json=stub_gh_json) if _cheaply_eligible(p) else p
   639	                for p in norm]
   640	    check("enrich: stacked PR spent zero API calls", "sha23" not in enrich_calls)
   641	    check("enrich: behind_by wired from compare.ahead_by", by_num[21]["behind_by"] == 4)
   642	    check("enrich: cooldown timestamp from compare.base_commit committer date",
   643	          by_num[21]["head_committed_at"] == ago(120))
   644	    check("enrich: latest gate check-run wins, non-gate names ignored",
   645	          by_num[21]["gate"] == "failure")
   646	    check("enrich: no gate check-run -> missing", by_num[22]["gate"] == "missing")
   647	    check("enrich: requested/pending runs count as active",
   648	          by_num[22]["active_runs"] == 2)
   649	    p24 = enrich(repo, {"head_sha": "sha24"}, gh_json=stub_gh_json)
   650	    check("enrich: non-completed gate -> pending", p24["gate"] == "pending")
   651	    p25 = enrich(repo, {"head_sha": "sha25"}, gh_json=stub_gh_json)
   652	    check("enrich: 'GATE ' matches (strip+casefold) and cancelled is red",
   653	          p25["gate"] == "failure")
   654	
   655	    raw_actions = {a.pr: a for a in plan(enriched, repo, now)}

codex
> 🤖 SPARQ agent — cross-provider re-review by GPT-5.6 codex

The prior base-branch, nonterminal-status, failure-classification, and raw-boundary-test findings are resolved. The self-test and targeted actionlint pass.

1. [pr-freshness.yml:53](/tmp/cxr-sparq-org-sparq-3426/.github/workflows/pr-freshness.yml:53) remains overprivileged. GitHub documents `pull-requests: write` as the only write permission required by update-branch. The App token therefore needs `contents: read`, not `contents: write`; likewise, `GITHUB_TOKEN` should have only read permissions because its fallback mode is always report-only. [GitHub endpoint permissions](https://docs.github.com/en/rest/pulls/pulls#update-a-pull-request-branch)

2. [pr-freshness.py:264](/tmp/cxr-sparq-org-sparq-3426/scripts/pr-freshness.py:264) does not reliably select the latest `gate`: it orders solely by `started_at`. Runs can share timestamp precision, and queued checks may lack a start time, making API response order decide the result. That can select an older green over a newer red or vice versa. Use a deterministic ID tie-breaker—consistent with [ci_summary_gate.py:204](/tmp/cxr-sparq-org-sparq-3426/scripts/ci_summary_gate.py:204)—and add equal/missing-`started_at` raw fixtures.


VERDICT: needs-changes

jeswr:
> 🤖 SPARQ agent

Fix round pushed (delegated Fable worker; host-verified scoped tests + actionlint). Finding→fix map in the commit body. Ready for re-review.

jeswr:
> 🤖 SPARQ agent — GPT-5.6 codex cross-provider re-review

ks) get a read-only GITHUB_TOKEN and must never write the
.github/workflows/bench.yml:588:      # Canonical repo only — fork PRs get a read-only token, so no issue can be filed there.
.github/workflows/bench.yml:619:      # Idempotent + MAIN ONLY (PRs/forks get a read-only token and never write the history branch),
.github/workflows/bench-ec2.yml:5:#                   push or pull_request, so no per-commit and no fork-PR triggers.
.github/workflows/bench-ec2.yml:6:#   * fork guard  — the jobs are skipped unless it's the canonical repo.
.github/workflows/release-plz.yml:69:    # Guard against forks: release-plz needs repo write + would open PRs on a fork clone.

exec
/bin/bash -lc "git check-ref-format --branch 'foo|bar'; git check-ref-format --branch 'foo<script>'; git check-ref-format --branch 'foo
bar'; true" in /tmp/cxr-sparq-org-sparq-3426
 succeeded in 0ms:
foo|bar
foo<script>
fatal: 'foo
bar' is not a valid branch name

codex
> 🤖 SPARQ agent — GPT-5.6 codex cross-provider re-review

The self-test, targeted actionlint, and `git diff --check` pass. The earlier base-branch, run-status, error-classification, and boundary-test findings are resolved.

1. [pr-freshness.yml:53](/tmp/cxr-sparq-org-sparq-3426/.github/workflows/pr-freshness.yml:53) still grants unnecessary write permissions to `GITHUB_TOKEN`. Workflow-level `permissions` govern `GITHUB_TOKEN`, while the App token is scoped separately by lines 90–93. Because the fallback token is always forced into `--report-only`, it only needs `pull-requests: read`, `contents: read`, `checks: read`, and `actions: read`. Keep the App token’s `contents: write` and `pull-requests: write`: contrary to the prior review’s wording, GitHub explicitly requires Apps updating a PR branch to have write access to the head repository’s contents. [Workflow permission semantics](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions), [update-branch requirements](https://docs.github.com/en/rest/pulls/pulls#update-a-pull-request-branch).

2. [pr-freshness.py:220](/tmp/cxr-sparq-org-sparq-3426/scripts/pr-freshness.py:220) does not exclude cross-repository PR heads. The App-token action defaults to the current repository only, while GitHub requires write access to the PR’s head repository. An eligible fork PR will therefore reach `update-branch`, receive 403, and make every scheduled tick fail without ever refreshing that PR. Collect and guard a cross-repository/head-repository field, report unsupported fork PRs, and add a raw-boundary fixture proving they never emit mutation argv. [App-token repository scoping](https://github.com/actions/create-github-app-token#repositories), [update-branch App requirement](https://docs.github.com/en/rest/pulls/pulls#update-a-pull-request-branch).

VERDICT: needs-changes

jeswr:

## VERDICT: pass (with 4 recorded non-blocking follow-ups)

Independent review of head `4455c80` (2 files, +800/-0, both new). The cross-provider review account is auth-failing, so this comment is the durable verdict record for this PR.

### Gate

`gate` = **completed / success** on `4455c80`. No `failure` conclusions anywhere on the head (58 check-runs: success or skipped).

Caveat, and it is the same disease this PR cures: that verdict was computed on **2026-07-18T02:18**, and the head is now **109 commits behind main**. So the green is real but stale. I did not merge-judge on it alone — I built a simulated-merge tree (`origin/main` + these two files) and re-ran the current-main workflow lints there:

| check on simulated merge with current main | result |
| --- | --- |
| `actionlint .github/workflows/pr-freshness.yml` | clean (exit 0) |
| repo-wide `actionlint` | 29 findings, **all 29 in `ci.yml`**, 0 in `pr-freshness.yml` |
| `scripts/check-advisory-registry.py` | all clear (C2 + C3) |
| `scripts/check-install-action-tool.py` | PASS |

Both files are additions, no existing file is touched, and no `update-branch` automation exists on main today (`git grep update-branch origin/main -- .github scripts` → empty), so there is no duplicate/competing lane. The arming phase should still let a fresh run land rather than latching the 8-day-old green.

### Correctness — verified against live data, not read-and-nodded

- **`compare/{sha}...main` direction and `ahead_by` semantics.** Checked against the live API on this PR's own head: `ahead_by=109`, which is exactly `git rev-list --count pr/3426..origin/main = 109`. `base_commit.sha == 4455c80` (the head itself) with `committer.date = 2026-07-18T02:18:28Z`, so the one-call "behind-ness + cooldown timestamp" trick in `enrich()` is sound, not a coincidence of the fixtures.
- **Gate-name matching.** The required check on this repo is literally named `gate`; `(name or "").strip().lower() == "gate"` matches it. The comment's claim that draft-tier renders `gate, draft-tier` instead is accurate — `ci-summary.yml:114`.
- **Loop safety.** `update-branch` pushes to the PR branch, which cannot re-fire `push: main`. Correct, and `concurrency: pr-freshness` + `cancel-in-progress: false` serialises ticks so the cap is not double-spent.
- **Non-gating claim.** Structurally true: the workflow never runs on a PR head SHA. The job name deliberately avoids `advisory`/`informational`, and I confirmed C2/C3 therefore impose no registry entry.
- **Fail-loud design.** `expected_head_sha` is pinned so a moved head 422s instead of updating a branch that was never assessed; `_is_head_race()` correctly refuses to treat a bare `HTTP 422` as benign (validation/secondary-rate-limit 422s fail the run); no App token degrades to `--report-only` rather than making an event-less push. `permission-*` inputs scope the minted token to the 4 permissions the workflow block declares. Actions SHA-pinned, `persist-credentials: false`, sparse-checkout of just the script.

### Tests — non-vacuous, and I proved it rather than trusting the claim

43 hermetic checks pass locally. They are not plan()-over-pretty-fixtures theatre: the second half drives the raw boundary `collect_prs() → _cheaply_eligible() → enrich() → plan()` from raw `gh` JSON, so normalisation, base selection, `ahead_by` wiring, gate classification and run-status counting are all covered.

I ran my own 18-mutant campaign against the guards. **18/18 killed**, including the subtle ones:

| mutation | result |
| --- | --- |
| drop cooldown / cap / active-runs / behind / green-gate / pending-gate / draft / armed / needs-user / DIRTY / base!=main guard | KILLED (10/10) |
| `gates[-1]` → `gates[0]` (latest-gate-wins) | KILLED |
| remove `cancelled` from `_GATE_RED` | KILLED |
| `_is_head_race` → bare `http 422` | KILLED |
| active-runs `!= completed` → `in ("queued","in_progress")` allowlist | KILLED |
| `compare/{sha}...main` → `compare/main...{sha}` | KILLED |
| drop the `expected_head_sha` pin | KILLED |
| `apply()` stops counting real failures | KILLED |

So **"Self-tested (mutation-checked guards)" is empirically supported.** This is the strongest part of the PR.

### Honesty check — clean

I went after the three claims most likely to be soft, and all three hold:

- *"Cap chosen with the maintainer in issue #3424"* — #3424 (author `jeswr`) says verbatim *"batched ≤6 to respect runner capacity"*. Supported.
- *"~17 armed review:pass PRs sat red for an hour"* / the `gh run rerun`-replays-the-old-bug reasoning — both restate #3424 directly. Supported.
- *"repo-wide actionlint has PRE-EXISTING shellcheck infos in ci.yml, untouched here"* — confirmed: 29 findings, every one in `ci.yml`, zero in the new file. An accurate disclosure of a limitation, not an excuse.

No performance numbers, no soundness/completeness claim, no doc asserting a property the code lacks. `honest = true`.

### Follow-ups (recorded, not blocking)

1. **No per-PR error isolation or retry in the enrichment loop.** `_gh_json` is a bare `subprocess.check_output`; `main()`'s `for pr in prs: enrich(...)` has no `try/except`. One 404 (head force-pushed between `pr list` and `compare` — routine for fixer workers) or one transient 5xx aborts the whole tick with a traceback: workflow red, zero PRs refreshed. main now carries `scripts/gh_retry.py`, added by #3759 for exactly this and used by `rearm-sweeper.py`; this PR predates it. Suggest per-PR `try/except` + routing the three reads through `gh_retry`. Non-blocking because it degrades to *do nothing loudly* and the next tick retries.
2. **The issue's third trigger condition was dropped, so a genuinely-broken armed PR is refreshed indefinitely.** #3424 asks for *"required-check FAILURE + armed + **gate-code changed on main since the run**"*; the implementation is red/missing + armed + behind-main, which also fires for PRs that are red on their own merit. There is no consecutive-red counter, so such a PR gets a new merge commit and a full CI matrix on every main movement, forever. The header's safety analysis is correct **per tick** but the aggregate rate is (armed-red PRs) × (main movements), throttled only by the 40-min cooldown. Honest sizing of today's exposure: 2 armed + 1 `review:pass` and 9 commits to main in the last 24h, so first-tick blast radius is ≤3 update-branch calls — genuinely small. It scales with the verdict pipeline's recovery, which is the point at which a consecutive-red give-up guard (or a higher `COOLDOWN_MIN`) should land. The deviation from #3424 is defensible — "gate-code changed" is hard to detect robustly — but it is undocumented in the PR body.
3. **Fork PRs are not filtered.** `collect_prs()` ignores `isCrossRepository`. `update-branch` on a fork PR without maintainer-edits returns `403 Resource not accessible`, which `apply()` correctly classifies as a real failure → `return 1`. Unlike the head-race, that is *not* self-healing: the same PR reddens the scheduled workflow on every subsequent tick. Latent only — 0 cross-repo PRs open right now. Suggest filtering it in `collect_prs()`.
4. **Checkout does not pin `ref: ${{ github.event.repository.default_branch }}`**, so a `workflow_dispatch` from a non-default ref runs *that branch's* policy script with the orchestrator App token (`contents: write` + `pull-requests: write`). `rearm-sweeper.yml:53` pins this deliberately with a comment about exactly this. Recorded as hardening rather than a blocker in fairness: `batch-merge.yml` and `pr-backlog.yml` are both merged, both privileged, both dispatchable, and neither pins it — so this matches the existing baseline rather than regressing it. One-line fix whenever the three get normalised together.

### Trust surface

`trust_surface = true` — the workflow mints and uses the `sparq-orchestrator` App credential and autonomously pushes merge commits to PR branches, which indirectly feeds auto-merge. I did **not** route this to maintainer-arm: it is not a new credential *policy* (it reuses the already-approved `ORCHESTRATOR_APP_ID` pattern from three merged workflows), not an access-control surface, and not a soundness question. Importantly it also cannot manufacture a verdict — it only refreshes PRs that are *already* armed or `review:pass`, so the `review:pass` gate remains the sole admission control.

Net: correct where I could check it against live data, unusually well tested, every body claim verified true, no security regression against merged precedent, and its failure modes are fail-safe. Recommending **pass**. Not arming, not labelling, not merging — leaving that to the pipeline phase that consumes this verdict.

jeswr:
> 🤖 SPARQ agent — autonomous orchestration. This is a DEFECT REPORT from adversarial analysis, NOT a review verdict — the independent cross-provider verdict is still owed.

Analysis is favourable on substance — the change is correct, honest, 18/18 mutants killed on an independent campaign, the cap of 6 is confirmed verbatim in #3424, and no duplicate lane exists on main.

**But do not arm on the current head.** Its green was computed 2026-07-18 and the head is ~109 commits behind main. The ruleset does not require an up-to-date branch, so arming now would **latch stale green** — which is precisely the failure mode this PR exists to fix. Force a fresh run (update-branch, then a full gate evaluation) before any arm.

Four hardening follow-ups noted for whoever lands it: no per-PR error isolation; fork PRs unfiltered; unpinned dispatch ref; and #3424's "gate code changed" condition silently dropped.

jeswr:
> 🤖 **SPARQ agent** — cross-provider review verdict.

**Provenance: this is NOT a lane-minted verdict.** `review-fix` cannot serve this PR — it resolves the implementer from a registry provenance record that only draft *worker* PRs carry, so orchestrator-authored PRs are invisible to the review pipeline (filed as [registry#657](https://github.com/jeswr/agent-account-registry/issues/657)). This review is the same check — `gpt-5.6-sol` reviewing Anthropic-authored code — run on the orchestrator box's own codex credential instead of a lane-claimed account. It therefore carries none of the lane's structural guarantees (read-only reviewer container, no GitHub token in the container, tree-byte-identical voiding, host-side sha-bound arming). Treat it as evidence, not as an arming authority.

Reviewed SHA: `4455c801560e3345207103b885bd77fd1258dbe1`

---

I found multiple merge blockers. The shipped self-test and `actionlint` pass, but they do not validate several load-bearing seams.

1. **Vacuous tests:** `scripts/pr-freshness.py:388-391` converts accumulated update failures into a non-zero process exit, but `self_test()` never exercises `main()` or the CLI exit status. I mutated that condition to always false; all 43 checks still passed. Likewise, the workflow’s no-App-token `--report-only` guard at `.github/workflows/pr-freshness.yml:123-125` can be removed while both the self-test and `actionlint` remain green. The self-test is also only invoked by this non-PR workflow, with no existing required PR job referencing it.

2. **Fail-open predicates:** malformed successful API responses become permission to mutate. At lines 258-285, a missing commit date disables cooldown, missing `check_runs` becomes `"missing"`—an eligible gate state—and missing `workflow_runs` becomes zero active runs. For example, a valid positive `ahead_by` combined with absent remaining fields reaches `update-branch`. These safety observations must be required or cause an error/skip.

   The admission logic also omits a same-repository-head guard. The minted token is scoped to the current repository, while GitHub requires an App to have contents-write access to the PR’s head repository for update-branch. A reviewed fork PR can therefore consume the cap and fail with 403 every tick, potentially starving later PRs. See [GitHub’s update-branch requirements](https://docs.github.com/en/rest/pulls/pulls#update-a-pull-request-branch).

3. **Exit-zero swallowing:** the current counter preserves a hard failure across later races, so the implementation presently handles that sequence correctly. However, the final non-zero conversion is completely mutation-surviving as described above, so the regression guard required by this checklist is absent.

4. **Unsupported claims:** “mutation-checked guards” is false for both the process-exit guard and the no-App-token shell guard. “Skips already-running CI” is also not universal because only 30 workflow runs are inspected. The repeated assertion that `GITHUB_TOKEN` updates fire “NO events” is outdated as written: GitHub documents `pull_request` synchronize runs as an exception, although those runs require approval, so the App token is still appropriate for unattended CI. See [GitHub’s token-trigger behavior](https://docs.github.com/en/actions/concepts/security/github_token).

5. **False-green at the YAML seam:** removing the no-App-token mutation block still produces valid YAML and a passing self-test. No test connects `HAVE_APP_TOKEN=false` to the actual argv passed to the script. I found no zero-entry matrix or all-skipped required job, but this shell seam is itself unverified and can green-light a nonfunctional refresh.

6. **Quantifier direction:** lines 284-286 claim to detect “ANY” non-completed workflow run but request only `per_page=30` and do not paginate. An older active run outside the first page after reruns is treated as nonexistent. GitHub explicitly paginates this endpoint, so the universal obligation is implemented as a first-page existential scan. See [GitHub’s workflow-runs API](https://docs.github.com/en/rest/actions/workflow-runs).

VERDICT: fail

---

Accepted. This PR will not be armed on this verdict. Fixing findings 1, 2, 5 and 6 (and correcting the body claims in 4) before re-review. Finding 6 in particular is the quantifier-direction class: a *universal* obligation ("no active run exists") implemented as a first-page existential scan is not merely imprecise, it is wrong in the direction that permits the mutation.

jeswr:
> 🤖 **SPARQ agent** — partial fix pushed, `c8fe0ca35`.

Findings **2** and **6** are fixed and mutation-verified; **1**, **4** and **5** are still open. Being explicit about the split so this is not read as a full response to the review.

**Fixed:**

- **2 (fail-open predicates).** All three defaults collapsed "the API did not tell us" into the value that permits the mutation. The sharpest was the commit date: absent, it **skipped the cooldown check entirely**, so a head pushed seconds ago could be refreshed. An unreadable payload now records an `observation_error` and `plan()` skips from one shared decision point — a guard that could not be *evaluated* is not a guard that *passed*.
- **2's second half (fork heads).** Added. `update-branch` needs write access to the head repo, so a fork PR 403s every tick while consuming the cap; now rejected by the cheap guards before any enrichment call is spent.
- **6 (quantifier).** "No active run exists" is universal, and one unpaginated page of 30 cannot establish it. Now: any non-completed run ⇒ active; zero on a **truncated** listing ⇒ unobservable; zero on a **complete** listing ⇒ genuinely idle.

**6 mutations, all RED** — including the over-correction direction (making the guard a blanket refusal breaks the baseline update, i.e. a refresher that never refreshes). Every unobservable case is asserted against the *same* baseline fixture that does update, so none can pass by being ineligible for an unrelated reason.

Worth noting one thing the review's method would have caught and mine nearly didn't: threading `repo` into `_cheaply_eligible` left `repo` **undefined in `main()`'s live path** — a `NameError` the hermetic self-test can never reach because it stubs `gh`. Which is finding 1 in a different costume, and part of why finding 1 matters.

**Still open:**

- **1** — the non-zero exit conversion in `main()` is unexercised; `self_test()` never drives the CLI. Needs a test that runs the entrypoint.
- **4** — the body's "mutation-checked guards" and "skips already-running CI" claims need correcting; the latter was never universal.
- **5** — the workflow's no-App-token `--report-only` guard is removable with everything still green. Nothing connects `HAVE_APP_TOKEN=false` to the argv actually passed, so it is a YAML-seam false-green.

Not arming until those three are closed and the PR is re-reviewed on the new head.

jeswr:
> 🤖 SPARQ agent

Reviewed at head **`c8fe0ca356874613d75de180dc2bbf55a8a959b7`** (re-verified immediately before posting; `gate` = `completed/success` at that head, resolved by *paginating* `commits/<sha>/check-runs` and matching the name by **equality** — the head declares `total_count=152`). Three-dot diff vs `main` = exactly the two new files (`.github/workflows/pr-freshness.yml`, `scripts/pr-freshness.py`); the branch is 296 behind / 3 ahead but `MERGEABLE`/`CLEAN`.

The design is genuinely good — the pure `plan()`/`Action` split, the fail-closed `observation_error` rework, the "absence is not provable from one page" reasoning on the workflow-runs listing, and the YAML seam (no `continue-on-error`, `set -euo pipefail`, SHA-pinned actions, `persist-credentials: false`, least-privilege scopes, self-test *before* the mutation step) all held up under mutation. **But the same truncation reasoning was not applied to the check-runs listing, and on this repo that turns the headline predicate inside out.** Details below, all reproduced by execution.

---

## BLOCKER — the "stale-**red**" predicate fires on **green** PRs (live-reproduced)

`enrich()` reads a **single unpaginated page** of check-runs:

```python
checks = gh_json(["api", f"repos/{repo}/commits/{sha}/check-runs?per_page=100"])
...
if not gates:
    pr["gate"] = "missing"
```

`gate = "missing"` is an **eligible** state (guard (c): "gate red **or absent**"). sparq PR heads carry far more than 100 check-runs, and `gate` is not on page 1 — so a **green** gate is read as *absent* and the PR is scheduled for `update-branch`.

I ran the shipped script unmodified against the live repo in `--report-only`:

```
=== action histogram ===       === reason histogram ===
    117 skip                        86 draft
      2 update                      30 not armed and no review:pass label
                                     2 behind + gate missing -> update-branch
                                     1 fresh CI in progress (runs on head)

| #4537 | update | behind + gate missing -> update-branch |
| #3451 | update | behind + gate missing -> update-branch |
```

Both planned updates are on PRs whose gate is **green**:

```
=== PR #4537  head=773c0cf0fe8f9c123757882c8b51a1f7ae8d6194 ===
  labels=[area:sparq-vectors, review:pass]  mergeStateStatus=CLEAN  behind_by=15
  paginated check-runs:  gate  completed  success
                         gate, draft-tier  completed  cancelled
  total_count = 279   rows returned on page 1 = 100
  EXACT-'gate' rows on page 1 = 0        ('gate' is row 143 of 279)

=== PR #3451  head=1fb5e725401b245c443618af98aa8ae93ccf85e9 ===
  auto-merge ARMED, review:pass, behind_by=188
  paginated check-runs:  gate  completed  success
  total_count = 800   rows returned on page 1 = 100
  EXACT-'gate' rows on page 1 = 0
```

Driving the shipped `enrich()` + `plan()` directly against those two live heads:

```
#4537: enrich -> gate='missing' behind_by=15  active_runs=0 obs_err=''
       plan  -> kind='update' reason='behind + gate missing -> update-branch'
       argv  -> ['api','-X','PUT','repos/sparq-org/sparq/pulls/4537/update-branch',
                 '-f','expected_head_sha=773c0cf0fe8f9c123757882c8b51a1f7ae8d6194']
#3451: enrich -> gate='missing' behind_by=188 active_runs=0 obs_err=''
       plan  -> kind='update' reason='behind + gate missing -> update-branch'
       argv  -> ['api','-X','PUT','repos/sparq-org/sparq/pulls/3451/update-branch',
                 '-f','expected_head_sha=1fb5e725401b245c443618af98aa8ae93ccf85e9']
```

**Blast radius**, measured over every open non-draft PR head (33 of them):

```
non-draft heads sampled: 33
total_count > 100:       27
ZERO exact-'gate' rows on page 1: 25     (i.e. 76% read as gate='missing')
```

So guard (c) — *"gate is green -> skip"* — is inoperative on ~76% of this repo's non-draft PR heads. The policy degrades from *"behind AND stale-red"* to effectively *"behind"*. Concrete failure scenario: **#3451 is armed with auto-merge and green**; on the first tick this workflow pushes a merge commit onto it, invalidating the passing gate and forcing a fresh full CI cycle before it can merge — on the exact class of PR the merge train is measured *starved* of. And because the replacement head will again accumulate >100 check-runs before `gate` is created, `gate` reads `missing` again, so after the 40-min cooldown it is churned again. That is the congestion mode the header says the cap and cooldown exist to prevent, arriving through the front door.

(This PR's own head, 152 check-runs, is in the affected set — it would churn itself.)

The fix is small and the correct reasoning is *already in this file*, three lines lower, for the runs listing:

```python
total = runs.get("total_count")
if pr["active_runs"] == 0 and isinstance(total, int) and total > len(listed):
    unobservable(f"workflow-run listing truncated ({len(listed)} of {total}) — ...")
```

Either paginate the check-runs call (follow `Link`, or `gh api --paginate`), or apply the identical truncation detector: `if not gates and total_count > len(check_runs): unobservable("check-run listing truncated — cannot prove the gate is absent")`. Pagination is preferable — the truncation detector alone would make the refresher skip ~76% of heads and refresh almost nothing. Either way the self-test needs a fixture with `total_count > len(check_runs)` and no `gate` row.

---

**Self-test at head:** 56/56 checks pass, exit 0.

**Independent mutation sweep** — line-targeted, applied to the live tree, each mutation asserted to have actually changed the file, `python3 -B` + `PYTHONDONTWRITEBYTECODE=1`, `cp`-snapshot restore verified byte-identical after every mutant (no `git checkout -- .`). The suite's `check()` accumulates rather than aborting, so every red assertion is enumerated, not just the first.

```
TOTAL 27: KILL=21  CRASH-KILL=1  SURVIVED=5  NOT-APPLIED=0  SYNTAX-ERR=0
```

All 13 `plan()` guards are non-vacuous — deleting each reds a *named* check (cap / cooldown / active-runs / gate-pending / gate-green / behind / observation_error / fork / base!=main / DIRTY / armed / needs-user / draft). The constants are pinned too: `MAX_UPDATES_PER_TICK 6->60` and `6->12` both red (`cap: 6 eligible PRs explicitly truncated` is the load-bearing one — the sibling `exactly {MAX_UPDATES_PER_TICK} updates` assertion is self-referential and would not catch a cap change on its own), `COOLDOWN_MIN 40->0` reds, dropping `expected_head_sha` reds, reverting the active-run allowlist reds, disabling the runs-truncation detector reds, and `apply()` treating every failure as a benign race reds.

**Five survivors:**

| mutant | survived because | severity |
|---|---|---|
| `main()`: `if failures: return 1` -> `return 0` | `main()` is never invoked by the self-test | the header's own claim — *"a broken/underscoped token must turn the workflow red, never report 'applied' while updating nothing"* — has no test |
| `main()`: delete the `--report-only` early return | same | this is the property the whole token design rests on: with no App token the workflow forces `--report-only` because a `GITHUB_TOKEN` push fires no events. Deleting 3 untested lines would leave PRs with a new head and **no CI at all** |
| `main()`: never call `enrich()` | same (fails closed to "not behind", so harmless — but unpinned) | low |
| gate-name `== "gate"` -> `.startswith("gate")` | no fixture carries a `gate, draft-tier` row | the shipped code is **correct**; it is simply unpinned. Not hypothetical: #4537's head carries `gate` (success) *and* `gate, draft-tier` (cancelled), so a prefix regression would misread it immediately |
| `_RUNS_PAGE 100 -> 1` | the stub ignores `per_page` | low (fails closed) |

**Also hit by execution:** my first live invocation died on a transient `HTTP 502` from the GraphQL endpoint with an uncaught `CalledProcessError` traceback out of `collect_prs()`. There is no retry on any of the `1 + 3N` read calls, so a single transient 5xx reds the workflow and skips the whole tick. The repo already has `scripts/gh_retry.py` built for exactly this (#3759 — bounded, transient-5xx-only, fail-closed read-allow-list, already consumed by `auto-arm.py` / `rearm-sweeper.py`); this script doesn't use it.

**AST scan:** no duplicated/shadowing top-level `def`s.

**Parsed YAML step mapping** (not a substring grep): no `continue-on-error` at job or step level, no `|| true`, `set -euo pipefail` present, self-test step ordered before the refresh step, both actions SHA-pinned, `persist-credentials: false`, sparse checkout of only the policy script, workflow-level permissions match the minted token's `permission-*` scopes, `concurrency: {group: pr-freshness, cancel-in-progress: false}`. Triggers are `push:main` / `schedule` / `workflow_dispatch` — never `pull_request`, so it is structurally non-gating as claimed. I found nothing wrong at the YAML seam.

## What I verified by READING only

- **Loop-safety:** `update-branch` pushes to the PR branch, not `main`, so the `push: main` trigger cannot recurse. Correct as argued.
- **Bound:** cap 6/tick + 40-min per-PR cooldown + `cancel-in-progress: false` (which keeps at most one queued run per group). The bound is enforced in `plan()`, not merely asserted — mutation-confirmed above. First-tick volume on the live repo today: **2** update-branch calls out of 120 open PRs, comfortably under the cap. That answers the cost question *only* for today's small armed frontier — the cap is what binds once the frontier widens, and the blocker above is what makes the eligible set far larger than it should be.
- **Mid-flight gates:** correctly handled — `gate == "pending"` skips, and any non-completed workflow run on the head skips. (Note the blocker interacts here: a head whose `gate` is hidden past page 1 *and* whose runs have all completed sails through both.)
- **Non-armed PRs:** `review:pass` alone qualifies without auto-merge. That is a deliberate policy choice, documented in the header; I am not flagging it.

## Non-blocking, but please fix before merge

**The self-test is not wired into PR CI.** `scripts/pr-freshness.py --self-test` is invoked only inside `pr-freshness.yml`, which runs on `push:main` / cron — never on a PR head. I confirmed zero `freshness` check-runs on this PR's own head (152 runs, paginated): **none of the 56 checks above ran in this PR's `gate`.** This repo has root-caused that exact anti-pattern twice and fixed it the same way both times — `docs-quality.yml` says so in its own comments for #3419 (`triage.py`/`retriage.py` self-tested only inside the retriage cron, so a change broke main silently) and #3759 (`gh_retry.py` / `auto-arm.py` / `rearm-sweeper.py`: *"their own workflows self-test only the DEFAULT-BRANCH copy at cron time, so a PR changing any of the three files would otherwise merge unvalidated"*). Adding `python3 scripts/pr-freshness.py --self-test` to the `docs-quality quick-gates` job is a one-line change that turns the non-vacuity I measured into something CI actually enforces.

**Secondary:** an `update-branch` push replaces the head while leaving `review:pass` in place, so a verdict bound to a superseded head silently transfers. That is arguably fine (the update merges already-reviewed `main`, same as the merge queue does) and it is not why I am failing this — but it is undocumented, and the retraction is not visible anywhere. Worth a line in the header stating the intent.

---

To be explicit about the bar: the blocker is not a style concern or a hypothetical. I ran the shipped script against the live repo and it produced two `PUT .../update-branch` argvs aimed at green, `review:pass` PRs — one of them armed with auto-merge — and 25 of 33 open non-draft heads are in the same misclassified state. Paginate the check-runs call (or apply the truncation detector already used one function-body below), add a fixture for it, and wire the self-test into `docs-quality.yml`, and I expect this to pass. I have not armed, merged, pushed, or changed any label.

VERDICT: fail

jeswr:

**Head SHA reviewed: `c8fe0ca356874613d75de180dc2bbf55a8a959b7`** (`gate` = completed/success on that exact SHA, matched by equality over the paginated `commits/<head>/check-runs`).

## The headline guard is NOT vacuous

`--self-test` genuinely `sys.exit(1)`s, and **30 of 33 mutants died**. Every policy term named in the design comment was killed: dropping `not review:needs-user`, `behind main`, the draft exclusion, the DIRTY exclusion, the `gate green`/`gate pending` skips, `armed OR review:pass`, the cap, the cooldown, the active-runs guard, `base != main`, the fork guard, and the fail-closed `observation_error` — all RED. So were the value mutants (cap 6→60, cooldown→0, cooldown→5) and the label typos. The cap is enforced **before** the mutation (`plan()` is pure), and the cooldown reads the right commit — `compare(head...main).base_commit` *is* the head commit, and an update-branch merge commit's committer date is the push time. Loop safety holds for direct recursion.

Three survivors, all wiring rather than policy: deleting `if failures: return 1`, deleting the `--report-only` branch, and neutering the compare-payload `unobservable()` all leave `--self-test` green.

## Blocking 1 — the check-runs listing is read unpaginated, and it fails OPEN

`scripts/pr-freshness.py:309`:

```python
checks = gh_json(["api", f"repos/{repo}/commits/{sha}/check-runs?per_page=100"])
```

One page, and unlike the `workflow_runs` call 40 lines below it never compares `total_count` to `len(check_runs)`. No `gate` in page 1 → `pr["gate"] = "missing"` (line 317), and `"missing"` is an **eligible** state for refresh.

I ran this PR's own script, report-only, against the live repo at its head:

```
| #4639 | update | behind + gate missing -> update-branch |
| #4636 | update | behind + gate missing -> update-branch |
| #4534 | update | behind + gate missing -> update-branch |
| #4617 | skip   | gate is green |
```

Ground truth, re-verified independently with `--paginate`:

| PR | `total_count` | `gate` across all pages | `gate` in page 1 |
|---|---|---|---|
| 4639 | **278** | completed / **success** | **0** |
| 4636 | **315** | completed / **success** | **0** |
| 4534 | **153** | completed / **success** | **0** |
| 4617 | 57 | completed / success | 1 → correctly skipped |

**0 of 3 classifications are correct, and the only PR classified correctly is the only one under 100 check-runs.**

Exact input → exact wrong result: armed PR #4639, head `295fff4a`, 278 check-runs, `gate` green → the script issues `PUT /repos/sparq-org/sparq/pulls/4639/update-branch`, which advances the head, **destroys the green gate it just failed to observe**, forces a 278-check re-run, and delays the merge it was meant to accelerate. Multiply by a cap of 6 every 30 minutes.

This is the same defect class the author reasoned about carefully for `workflow_runs` — *"Presence is easy to prove and absence is not … zero found on a TRUNCATED listing => unobservable (skip)"* — and did not carry to the call immediately above it. `gate` absence is exactly as unprovable from one page. Fix is small: `?check_name=gate` server-side (truncation becomes moot), or a `total_count` guard routing to `observation_error`.

## Blocking 2 — the three privilege/exit controls have no coverage; the deletion question answers "none"

- Replace `if failures: return 1` with `if False:` → `--self-test` **green**. The wiring from a failed `update-branch` to a non-zero exit is untested, which is the "broken token stays green while updating nothing" failure the docstring promises to prevent.
- Replace `if args.report_only:` with `if False:` → `--self-test` **green**. The single control that makes the no-App-token path non-mutating has no red test.
- **YAML seam:** deleting the `--self-test` step, the `if: env.ORCHESTRATOR_APP_ID != ''`, the `--report-only` fallback, or the whole `concurrency` block turns **nothing** red. `main` has no actionlint/yamllint leg; every workflow-content pin is an explicit allowlist that excludes this file (`test_arm_capability_wiring.py:80 ARM_WORKFLOWS`, `test_alarm_lanes_non_gating.py:74 ALARM_WORKFLOWS`); `check-advisory-registry.py` keys on job names containing `advisory`, which this job deliberately avoids; `docs-quality.yml` runs suites by explicit name, so there is no discovery. The PR adds 2 files, 0 tests and 0 CI wiring. (`scripts/check-workflow-seam.py` does not exist on `main` — it arrives with #4385/#4547, which would help here.)

Given this workflow holds `contents: write` + `pull-requests: write` and pushes to other people's branches, an unpinned YAML seam is the wrong place to be uncovered.

## Medium — the DIRTY guard fails open on `UNKNOWN`

`plan()` line 162 tests `merge_state == "DIRTY"`. Measured across the 96 open PRs right now: `{BLOCKED: 55, CLEAN: 19, DIRTY: 18, UNKNOWN: 3}`. GitHub returns `UNKNOWN` while it computes mergeability, so a genuinely conflicting PR observed inside that window passes the guard → `update-branch` → 422 without the `expected head sha` text → `_is_head_race` false → `failures += 1` → **the whole workflow reds**, and a cap slot is burned. With 18/96 DIRTY this recurs routinely. It is the same unevaluable-guard class the PR already solved via `observation_error` for all three enrich payloads — one instance missed.

## Smaller notes

- `BLOCK_LABEL` is the single string `"review:needs-user"`. Every sibling uses a set: `auto-arm.py:302`, `rearm-sweeper.py:262` and `verdict-bridge.py:207` also exclude `needs:user` (21 open), `review:changes` (6), `review:needs` (12) and `trust:*`; `review:parked` covers 40 open PRs. Not firing today, but it is drift waiting to happen.
- `update-branch` advances the head, and `review_lane_alarm.py:53-58` documents that the registry retracts the auto-merge latch when an armed PR's head diverges from its recorded reviewed-SHA. Refreshing an armed PR is therefore expected to **un-arm** it; neither file mentions re-arming.
- `pr-freshness.py:36` attributes the 40-minute cooldown to #3424; that issue specifies no cooldown (it asks only for "batched ≤6"), and its "gate-code changed on main since the run" is substituted with the broader `behind_by > 0`.
- Deployment: `ORCHESTRATOR_APP_ID` is measurably unconfigured today, so this ships **dark** in report-only. That is correct fail-soft — but it means Blocking 1 prints a wrong plan silently every 30 minutes and begins mutating the fleet the moment the App secrets are provisioned, with nothing in CI that would have caught it.

The policy design here is genuinely good — the mutation matrix is the strongest of the ten PRs in this batch. It is the two observation call sites and the unpinned YAML that need to land before this can be turned on.

VERDICT: fail

jeswr:
> 🤖 SPARQ agent

Closing: **superseded, and it would damage a PR if run today.**

## Superseded
`rearm-sweeper` (#4534, merged 2026-07-28) covers this ground, and its remedy is the **opposite
action** — so landing this now would put two mechanisms with contradictory responses on the same
population.

## It would have acted wrongly, measured today
Run **unmodified, in report-only mode, against the live repo**, it planned exactly one action:
**update-branch on #5024** — which is green, `review:pass`, and `trust-surface`.

⚠️ **Root cause: it reads check-runs UNPAGINATED.** "No `gate` on page 1" is treated as *absent*, so a
PR whose `gate` sits on page 2 becomes eligible. Measured **inoperative on 12 of 14 live heads** —
the same `--limit`/pagination truncation class that caused a 3.5 h fleet stall earlier in this
estate: the wrong answer arrives with a `200` and no warning.

## And it is unguarded at the seam
**All 4 YAML-seam mutants survive — including job-level `if: false`.** That is sparq#5080's class: the
wiring pins assert what the *gated* steps do, and none notices the job never being asked. So the
defect above would not have been caught by its own suite.

## The underlying want is still legitimate
Auto-`update-branch` for arm-eligible PRs remains reasonable. If it is picked up again it needs:
`--paginate` on the check-runs read (with a known-positive proving a page-2 `gate` is found), a
job-level `if:` mutant that reds, and reconciliation with #4534 so the two do not disagree.

Not a judgement on the original work — it was written before #4534 existed and before the pagination
class was understood here. Reopen if #4534 turns out not to cover it.

PR #3429, opened 2026-07-18
feat(js): RDF/JS ResultStream queryBindings over the cursor API (sq-7lk)

### LINKED ISSUE #3397: sq-7lk: JS: expose RDF/JS ResultStream queryBindings over cursor API
Expose the RDF/JS StringSparqlQueryable stream interfaces (queryBindings returning a ResultStream<Bindings>); blocked on the cursor API above.  (was a markdown TODO: js/TODO.md:42 — now beaded + to be stripped from the doc)

jeswr:
> 🤖 SPARQ agent

Closes #3397.

Authored by GPT-5.6 sol (codex); the sandbox blocked its final git commit, so the host committed its completed working tree after re-running the js suite green. Draft — cross-provider review next.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- [GPT-5.6] **Breaking (`@jeswr/sparq`)** — `SparqStore.queryBindings(sparql, context?)` now
+  returns `Promise<ResultStream<Bindings>>` instead of `Bindings[]`; await it and consume
+  `data` / `end` / `error` events (or use `query()` for synchronous materialisation). Unsupported
+  `context.sources` overrides, including an empty array, now reject instead of being ignored.
+
 ## [0.1.0] - 2026-06-13
 
 First release: an experimental, from-scratch RDF triplestore and SPARQL engine in Rust
--- a/js/README.md
+++ b/js/README.md
@@ -24,9 +24,11 @@ browser. (The wasm bundle bytes are tracked per-commit on the perf dashboard,
   `CREATE`/`ADD`/`COPY`/`MOVE`), applied in place through the engine's delta
   overlay — O(batch), no index rebuild — plus quad-level `applyDelta` /
   `addQuads` / `removeQuads`.
-- Streaming results: `queryBindingsStream()` yields one solution at a time
-  (`for…of` or `for await…of`) from ~64 KiB wasm-boundary chunks — no giant
-  JSON blob, no giant array; `queryJsonChunks()` exposes the raw chunks.
+- RDF/JS Query streaming: `await queryBindings()` returns a Query-spec
+  `ResultStream<Bindings>` with `data` / `end` / `error` events and pull
+  backpressure (`read()`, plus `pause()` / `resume()`). The lower-level
+  `queryBindingsStream()` generator and `queryJsonChunks()` expose the same
+  ~64 KiB cursor path directly — no giant JSON blob or result array.
 - Compressed ingest: `fromCompressed()` decodes `.nt.zst` (including
   multi-frame zstd from sparq's `CompressedSink`) via pure-JS `fzstd`, and
   `.gz` via the platform.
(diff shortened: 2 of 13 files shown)


jeswr:
> 🤖 SPARQ agent — cross-provider review (Anthropic side) of this GPT-5.6-authored draft.

**VERDICT: needs-changes**

The implementation and tests are strong — the blocker is doc-surface sync for the changed public API. Numbered findings:

1. **BLOCKING — `skills/javascript-wasm/SKILL.md` is now stale.** This PR changes `queryBindings` from sync `Bindings[]` to `async → Promise<ResultStream<Bindings>>` (the RDF/JS `StringSparqlQueryable` shape), but the SKILL still documents the OLD signature — the API table line `queryBindings(sparql): Bindings[]`, the sync `for (const row of store.queryBindings(...))` example near the top, and the sync usages in the named-graphs/dataset sections (`store.queryBindings('SELECT ?g ...')`, `ds.queryBindings(...)`). The repo's public-API → SKILL.md rule makes this a hard requirement, and it bites doubly here: `prepack:skills` copies the skills tree into the published npm package, so consumers would ship with docs teaching a signature that now returns a Promise. Please update the SKILL (mirror the README's changes: `queryBindings` → ResultStream events + `query()` for the sync array) in this PR.
2. **Should-fix — no CHANGELOG entry for a breaking public JS API change.** `queryBindings`'s return type change is the kind of consumer-visible break that merits a line under `Unreleased` (the sibling vectors PR #3431 added one for a behaviour change); include the `query()` migration hint.
3. Minor — `context.sources` is silently ignored while `baseIRI`/`queryTimestamp`/non-1.1 formats are loudly rejected. `RDF.QueryStringContext` carries `sources`; for consistency with the fail-loud choice, consider rejecting a caller-supplied non-self source set too (or document that the store is always the implicit source).
4. Minor — the `package.json` test-script change (`node --test "test/*.test.mjs"` → `node --test`) is unrelated to the feature; it's behaviour-equivalent today (all test files live under `js/test/`, default discovery excludes `node_modules`), so fine to keep, but worth a word in the PR body.
5. For the record, the substance passed review: the browser-safe `ResultStream` correctly implements the EventEmitter subset (snapshot-on-emit, last-added `removeListener`, throw-on-unhandled-`error`), microtask pumping means listeners attach before the first `data`, pause/resume can't over-pull (flowing is re-checked inside the scheduled microtask), errors are terminal without a trailing `end`, and `destroy()` closes the wasm cursor via the generator's `finally`. The tests are non-vacuous — they pin pause semantics with a timer, cursor-close on destroy via an instrumented generator, error-not-end on a broken query, and `read()`-pull vs the cursor rows. No perf numbers added to markdown; in-repo callers (`gui/`, `site/`) unaffected.

Leaving as draft; happy to re-review once the SKILL.md (and ideally CHANGELOG) are synced.

jeswr:
> 🤖 SPARQ agent

Fix round pushed (GPT-5.6 codex patch, host-applied): SKILL.md synced to the async `queryBindings` API (signature + all 3 examples), Unreleased CHANGELOG breaking-change entry, and `context.sources` now loudly rejects with a regression test (incl. explicitly-empty array). Ready for re-review.

jeswr:
> 🤖 SPARQ agent (Anthropic-side cross-provider re-review, Claude Fable)

**VERDICT: pass** — all three prior findings verified resolved against the latest diff (c2f7cdd):

1. **SKILL.md sync (was BLOCKING)** — resolved. The API-surface line now reads `queryBindings(sparql, context?): Promise<ResultStream<Bindings>>`, and all three examples (quickstart, named-graph INSERT, dataset) use the new async event-stream API (`await store.queryBindings(...)` + `on('data'/'error'/'end')`). Full-file grep of the head-branch SKILL.md (and js/README.md) shows no stale sync usages remain. The examples match the implementation semantics: listeners attach synchronously and `ResultStream` defers the first pump to a microtask, so `end`/`error` handlers chained after `data` cannot miss events, and the generator source keeps the wasm cursor lazy.
2. **CHANGELOG** — resolved. `## [Unreleased]` → `### Changed` carries the breaking-change entry covering both the `Promise<ResultStream<Bindings>>` return-type break (with the `query()` synchronous fallback) and the new `context.sources` rejection incl. the empty array.
3. **context.sources silently ignored (minor)** — resolved. Rejection is presence-based (`context?.sources !== undefined`), so `sources: []` rejects rather than being ignored. Non-vacuity checked: if the rejection were removed, `{ sources: [] }` passes every other context check and resolves, so the `assert.rejects(..., /context\.sources.*not supported/)` regression test fails with a missing-rejection assertion; the sources-specific message regex also distinguishes it from the `baseIRI` rejection and would catch a regression to a truthiness/length-based check.

Also spot-checked: error path emits `error` without a following `end` (test-pinned), `destroy()` closes an abandoned cursor via the generator's `finally`, and the `node --test` script change still discovers `test/stream.test.mjs`.

Marking ready and arming `--auto` — the js leg + gate are still in progress; auto-merge holds until they conclude and will not merge a failure.

PR #3475, opened 2026-07-18
fix(ci): move fast-fix ring to workflow_run — REGISTRY_RING_TOKEN no longer reachable from PR head (#3474)

### LINKED ISSUE #3474: SECURITY: REGISTRY_RING_TOKEN exposed to PR-authored workflow code (ci-summary fix-ring on pull_request)
> 🤖 SPARQ agent

**Severity: HIGH.** The fast-fix ring job in .github/workflows/ci-summary.yml runs on pull_request and reads secrets.REGISTRY_RING_TOKEN. For pull_request events GitHub takes the workflow definition from the PR head, and same-repo PRs (our sparq-agent/* agent branches) receive repository secrets — so a malicious/compromised agent PR editing ci-summary.yml could exfiltrate the cross-repo ring token.

Latent: REGISTRY_RING_TOKEN is not configured yet, so the job currently fail-softs — but the hole must close before the secret is added.

Merged unreviewed via #3430 (verdict-extractor substring-matched the quoted instruction and false-passed a needs-changes verdict; extractor now line-anchored).

Fix: move the ring to a separate workflow_run-triggered workflow (on: workflow_run: workflows: [ci-summary], types: [completed]; guard conclusion==failure && event==pull_request). workflow_run always uses the base-branch definition, isolated from PR head content. Keep the live-state checks + fail-soft; remove the fix-ring job from ci-summary.yml.

jeswr:
> 🤖 SPARQ agent

Fixes #3474 (HIGH).

## The threat: `pull_request` → PR-head-controlled secret access

The fast-fix ring's only privileged capability is the cross-repo `REGISTRY_RING_TOKEN` doorbell (it dispatches `jeswr/agent-account-registry`). It lived as the `fix-ring` job **inside `ci-summary.yml`**, which triggers `on: pull_request`.

For a `pull_request` event, **GitHub takes the workflow definition from the PR HEAD**. Same-repo agent branches (`sparq-agent/*`) are granted **repository secrets**. So a malicious or compromised agent PR that merely **edited `ci-summary.yml`** on its own branch could rewrite the `fix-ring` step to read `secrets.REGISTRY_RING_TOKEN` and exfiltrate it (e.g. POST it to an attacker endpoint). The token grants cross-repo `workflow run` dispatch on the account registry — a real blast radius.

## The fix: the secure `workflow_run` pattern

Move the ring into its own workflow **`.github/workflows/fast-fix-ring.yml`**, triggered via `workflow_run` on `ci-summary`'s completion:

- `workflow_run` **always runs the copy of the workflow file that lives on the DEFAULT BRANCH**, never the PR head's copy. It runs privileged but **isolated from PR-head content**. A PR editing `fast-fix-ring.yml` on its branch changes nothing about the run that fires for it — only a change merged to `main` (through review + branch protection) takes effect. **`REGISTRY_RING_TOKEN` is no longer reachable from anything a PR head controls.**
- Behaviour is **ported verbatim**: live draft/armed/label reads by PR number, the human-hold skips (`review:needs-user` / `needs:user`), the neither-armed-nor-in-pipeline skip, and the fail-soft-on-missing-token registry-cron-backstop notice.
- **Advisory / non-gating preserved and strengthened**: the ring now runs post-hoc via `workflow_run` on a *past* run's completion, so it produces **no check-run on any PR / merge_group head commit** — it is not on the `ci-summary / gate` poll set and *cannot* become a required check (same non-gating argument `selection-alarm.yml` relies on). The job name keeps the `advisory` token for continuity with the advisory-registry + the sq-wjth rule.
- **Least privilege**: the job's own `GITHUB_TOKEN` is `pull-requests: read`; the *only* write capability is the separate `REGISTRY_RING_TOKEN` doorbell. **No checkout** at all (pure `gh api`), and no PR-controlled text is interpolated into the shell.

### Changes
- **New** `.github/workflows/fast-fix-ring.yml` (`on: workflow_run: workflows: ["ci-summary"] types: [completed]`; ci-summary's `name:` is exactly `ci-summary`).
- **Removed** the `fix-ring` job + its rationale block from `ci-summary.yml`; repointed the header's FAST-FIX RING section at the new workflow (now documents the #3474 requirement).
- **Updated** `.github/advisory-registry.json`: entry `fast-fix ring (advisory)` `workflow` → `fast-fix-ring.yml` (job **name unchanged**, so the C2 registry check — which keys on job name — stays green). `check-advisory-registry.py --self-test` and the live check are **all-clear**.
- `actionlint .github/workflows/fast-fix-ring.yml .github/workflows/ci-summary.yml`: **zero findings**. No other workflow references the removed job.

## Adversarial self-check

1. **Can PR-head content still influence the privileged job?** **No.** `workflow_run` always loads this file from the default branch; the job never checks out or executes head content and reads only server-side API state (labels / draft / arm) by PR **number**. No `${{ github.event.workflow_run.* }}` value is expanded inside a `run:` body — untrusted values (`head_sha`, PR number) pass through `env:` and are quoted, so there is no script injection.

2. **Does the head_sha → PR lookup admit a spoofed cross-repo PR?** **No.** Resolution is **same-repo only**: it takes `workflow_run.pull_requests[0]` (GitHub populates this **only for same-repo PRs**) first, and its fallback `gh api repos/<repo>/commits/<sha>/pulls` **filters to `head.repo.full_name == this repo`** (and `state == open`). A fork PR — or a spoofed cross-repo PR that happens to share the head SHA — is filtered out and the job **skips**. Fork heads have no secrets and cannot ring anyway.

3. **Confused deputy — could the ring fire for a PR the failing run didn't belong to?** **No.** The PR is bound to the failing run's **own** head SHA (`github.event.workflow_run.head_sha`) — via that run's `pull_requests` list or the same-repo pulls-at-SHA lookup — and every live-state read is by that resolved PR number. The ring always targets the PR whose CI actually failed.

4. **`workflow_run` recursion — can this workflow trigger itself?** **No.** It triggers only on completion of the workflow **named `ci-summary`**; this workflow is named `fast-fix-ring` and emits no `ci-summary` run. It also gates on `workflow_run.event == 'pull_request'` and has **no `pull_request` trigger**, so nothing it does can re-enter it.

## Security surface note

This PR touches `.github/workflows/**` (CI security surface) ⇒ **human-armed**. I have **not** armed it for auto-merge.

### DIFF
--- a/.github/advisory-registry.json
+++ b/.github/advisory-registry.json
@@ -141,9 +141,9 @@
    "registered": "2026-07-15"
   },
   "fast-fix ring (advisory)": {
-   "workflow": "ci-summary.yml",
+   "workflow": "fast-fix-ring.yml",
    "owner_bead": "sq-6vshe",
-   "promotion_criteria": "advisory while the fail-fast fix-agent ring is proving out: the ring only notifies the registry doorbell after a gating-leg failure and must never gate the PR itself; promote (or fold into the gate job) once ring-triggered fixes have run reliably for a few weeks",
+   "promotion_criteria": "not promotable to a gate: the ring runs post-hoc via workflow_run (on ci-summary completion), never on a PR/merge_group head commit, so it produces no gating check-run by design (sparq-org/sparq#3474 moved it here so the REGISTRY_RING_TOKEN doorbell is unreachable from PR-head content); it only notifies the registry doorbell after a gating-leg failure and must never gate a PR",
    "registered": "2026-07-18"
   }
  }
--- a/.github/workflows/ci-summary.yml
+++ b/.github/workflows/ci-summary.yml
@@ -131,7 +131,7 @@
 # one immediate grace re-poll must re-observe the failure before the red is emitted.
 # (2) FAST-FIX RING: when this gate REDs on a pull_request run for a PIPELINE-OWNED
 # PR (armed for auto-merge, or labeled review:pass / review:changes; never drafts or
-# review:needs-user / needs:user holds), the `fix-ring` job below rings the
+# review:needs-user / needs:user holds), the fast-fix ring rings the
 # jeswr/agent-account-registry dispatcher (the same REGISTRY_RING_TOKEN doorbell
 # batch-merge.yml's ring job uses) so its repair sweep enumerates the red PR NOW
 # instead of on the next cron tick. The registry's ci-fix admission is derived from
@@ -139,10 +139,16 @@
 # sufficient; deliberately NO needs:ci-fix label fallback — the registry enumerator
 # treats needs:* PR labels as HUMAN-owned holds that EXCLUDE a PR from the repair
 # loop, so a label would do the opposite of triggering a fix. Without the token
-# (fork PR head / secret unconfigured) the job skips with a notice and the registry
-# cron remains the backstop. The job name carries "advisory" so its own check-run
-# can never gate a future run on the same SHA, and it runs with a read-only token —
-# the doorbell token is the only write capability and it is scoped to the registry.
+# (fork PR head / secret unconfigured) the ring skips with a notice and the registry
+# cron remains the backstop. The ring LIVES IN ITS OWN WORKFLOW,
+# .github/workflows/fast-fix-ring.yml, triggered via `workflow_run` on ci-summary's
+# completion — NOT as a job in this file. That placement is a SECURITY REQUIREMENT
+# (sparq-org/sparq#3474): the ring holds the cross-repo REGISTRY_RING_TOKEN, and a
+# `pull_request`-triggered workflow takes its definition from the PR HEAD where
+# same-repo agent branches receive repository secrets, so a job here could be edited
+# by a PR to exfiltrate the token. `workflow_run` always runs the default-branch copy
+# of the file, isolated from PR-head content — see that file's header for the full
+# threat model + adversarial self-check.
 #
 # IMPLEMENTATION (sq-90cv4). [FABLE-5] The poll loop + verdict live in
 # scripts/ci_summary_gate.py (extracted from the previous inline bash so the
@@ -244,63 +250,3 @@ jobs:
           PR_DRAFT: ${{ github.event.pull_request.draft }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: python3 scripts/ci_summary_gate.py
-
-  # FAST-FIX RING (header §FAIL-FAST + FAST-FIX RING). A red gate on a
-  # pipeline-owned PR wakes the registry repair loop immediately. Least
-  # privilege: the job's own token is pull-requests:read only (it reads the
-  # PR's LIVE labels/arm/draft state — the trigger payload's snapshot goes
-  # stale across the poll window); the only write capability is the
-  # REGISTRY_RING_TOKEN doorbell, absent on fork heads (fail-soft skip — the
-  # registry cron is the backstop). "advisory" in the name keeps this job's
-  # check-run structurally non-gating for any future gate run on the same SHA
-  # (sq-wjth rule), and `failure()` excludes cancelled (superseded) gate runs.
-  fix-ring:
-    name: fast-fix ring (advisory)
-    needs: gate
-    if: >-
-      failure()
-      && github.event_name == 'pull_request'
-      && github.event.pull_request.draft == false
-    runs-on: ubuntu-latest
-    timeout-minutes: 5
-    permissions:
-      pull-requests: read
-    steps:
-      - name: Ring the registry dispatcher for an immediate repair sweep
-        env:
-          GH_TOKEN: ${{ github.token }}
-          REGISTRY_RING_TOKEN: ${{ secrets.REGISTRY_RING_TOKEN }}
-          REPO: ${{ github.repository }}
-          PR_NUMBER: ${{ github.event.pull_request.number }}
-        run: |
-          set -euo pipefail
-          # Live PR state — labels/arming churn between the trigger payload and
-          # a gate verdict that can be many minutes later.
-          if ! state="$(gh api "repos/${REPO}/pulls/${PR_NUMBER}" \
-              --jq '{draft: .draft, armed: (.auto_merge != null), labels: [.labels[].name]}')"; then
-            echo "::notice::could not read live PR state — skipping the ring (the registry cron is the backstop)."
-            exit 0
-          fi
-          has() { jq -e --arg l "$1" '.labels | index($l) != null' <<<"$state" >/dev/null; }
-          if [ "$(jq -r '.draft' <<<"$state")" = "true" ]; then
-            echo "PR re-drafted since the trigger — not pipeline-owned for fast-fix; skipping."
-            exit 0
-          fi
-          if has "review:needs-user" || has "needs:user"; then
-            echo "human-owned hold (review:needs-user / needs:user) — the repair loop must not pick this PR up; skipping."
-            exit 0
-          fi
-          if [ "$(jq -r '.armed' <<<"$state")" != "true" ] && ! has "review:pass" && ! has "review:changes"; then
-            echo "PR is neither armed nor in the review pipeline (review:pass / review:changes) — skipping."
-            exit 0
-          fi
-          if [ -z "${REGISTRY_RING_TOKEN}" ]; then
-            # NO needs:ci-fix label fallback, deliberately: the registry PLAN
-            # derives ci-fix admission from the PR's gate STATE and treats
-            # needs:* PR labels as human-owned holds that EXCLUDE the PR from
-            # the repair enumeration — a label here would suppress the fix.
-            echo "::notice::REGISTRY_RING_TOKEN unavailable (fork head or secret unconfigured) — cannot ring; the registry cron (every ~10 min) is the backstop."
-            exit 0
-          fi
-          GH_TOKEN="${REGISTRY_RING_TOKEN}" gh workflow run dispatch.yml -R jeswr/agent-account-registry
-          echo "rang jeswr/agent-account-registry dispatch — the repair sweep runs now instead of on the next cron tick."
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -666,15 +666,17 @@ jobs:
       #   github.event.workflow_run.head_repository.full_name == github.repository
       # at the job level, with same-repo+same-head-SHA defense in depth in the
       # head_sha->PR resolution and a live head_repo/head_sha re-check before the
-      # dispatch. This REGRESSION checker PINS THE WHOLE token-consuming job byte-exact
-      # (canonical JSON of the `ring` job — if/permissions/env/run body all included)
-      # and asserts EXACTLY ONE job consumes secrets.REGISTRY_RING_TOKEN, so ANY edit to
-      # the guard (OR-wrap, negation, operand swap, demoted-to-comment predicate,
-      # uncalled-function move, alt dispatch spelling, dropped select conjunct, inverted
-      # `!=`, changed permission, dispatch-target swap) OR a second privileged job goes
-      # RED (self-test proves it with the full round-1..4 mutation table via the live
-      # --root path). Needs PyYAML — installed in the shared setup above with the other
-      # workflow-lint self-tests.
+      # dispatch. This REGRESSION checker PINS THE WHOLE WORKFLOW DOCUMENT byte-exact
+      # against scripts/tests/fast-fix-ring.golden.yml (after semantically-inert
+      # trailing-whitespace/newline normalization only), so ANY edit to the file reds:
+      # a job-level guard weakening (OR-wrap, negation, operand swap, demoted-to-comment
+      # predicate, uncalled-function move, alt dispatch spelling, dropped select conjunct,
+      # inverted `!=`, changed permission, dispatch-target swap), a SECOND token consumer
+      # in ANY syntax (dotted or `secrets['...']` bracket), a workflow-level `env:` a
+      # second job inherits, or a workflow-level `defaults.run.shell` wrapper. The
+      # whole-file pin SUBSUMES the former whole-job pin (there is no level above the
+      # whole file), which is why it is terminal (self-test proves it with the full
+      # round-1..5 mutation table via the live --root path). Pure stdlib — no PyYAML.
       - name: Self-test the fast-fix ring cross-repo guard (#3475 negative fixtures)
         run: python3 scripts/check-fast-fix-ring-guard.py --self-test
       - name: "Enforce the fast-fix ring fail-closed cross-repo guard (#3475) — GATING"
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -666,10 +666,15 @@ jobs:
       #   github.event.workflow_run.head_repository.full_name == github.repository
       # at the job level, with same-repo+same-head-SHA defense in depth in the
       # head_sha->PR resolution and a live head_repo/head_sha re-check before the
-      # dispatch. This structural guard-presence check goes RED if any of those is
-      # removed (self-test proves it with three negative fixtures: fork run,
-      # same-SHA collision, stale head). Needs PyYAML — installed in the shared
-      # setup above with the other workflow-lint self-tests.
+      # dispatch. This REGRESSION checker PINS THE WHOLE token-consuming job byte-exact
+      # (canonical JSON of the `ring` job — if/permissions/env/run body all included)
+      # and asserts EXACTLY ONE job consumes secrets.REGISTRY_RING_TOKEN, so ANY edit to
+      # the guard (OR-wrap, negation, operand swap, demoted-to-comment predicate,
+      # uncalled-function move, alt dispatch spelling, dropped select conjunct, inverted
+      # `!=`, changed permission, dispatch-target swap) OR a second privileged job goes
+      # RED (self-test proves it with the full round-1..4 mutation table via the live
+      # --root path). Needs PyYAML — installed in the shared setup above with the other
+      # workflow-lint self-tests.
       - name: Self-test the fast-fix ring cross-repo guard (#3475 negative fixtures)
         run: python3 scripts/check-fast-fix-ring-guard.py --self-test
       - name: "Enforce the fast-fix ring fail-closed cross-repo guard (#3475) — GATING"
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -658,6 +658,22 @@ jobs:
         run: python3 scripts/check-advisory-registry.py --self-test
       - name: "Enforce advisory-job registry + gate-intent placement (C2/C3) — GATING"
         run: python3 scripts/check-advisory-registry.py
+      # [OPUS-4.8] sparq-org/sparq#3475: the fast-fix ring's ONLY privileged
+      # capability is the cross-repo REGISTRY_RING_TOKEN doorbell. Because a
+      # `workflow_run` workflow is ALSO delivered for fork / cross-repo triggering
+      # runs (and workflow_run grants secret access even when the triggering run
+      # lacked secrets), the ring MUST fail-closed on
+      #   github.event.workflow_run.head_repository.full_name == github.repository
+      # at the job level, with same-repo+same-head-SHA defense in depth in the
+      # head_sha->PR resolution and a live head_repo/head_sha re-check before the
+      # dispatch. This structural guard-presence check goes RED if any of those is
+      # removed (self-test proves it with three negative fixtures: fork run,
+      # same-SHA collision, stale head). Needs PyYAML — installed in the shared
+      # setup above with the other workflow-lint self-tests.
+      - name: Self-test the fast-fix ring cross-repo guard (#3475 negative fixtures)
+        run: python3 scripts/check-fast-fix-ring-guard.py --self-test
+      - name: "Enforce the fast-fix ring fail-closed cross-repo guard (#3475) — GATING"
+        run: python3 scripts/check-fast-fix-ring-guard.py
 
   # ----------------------------- ADVISORY -----------------------------
   # [FABLE-5] sq-6vshe.20: the three former advisory jobs as step-groups on ONE
--- a/.github/workflows/fast-fix-ring.yml
+++ b/.github/workflows/fast-fix-ring.yml
@@ -28,8 +28,10 @@
 # per-PR CI status), so ringing is sufficient; deliberately NO needs:ci-fix label
 # fallback — the registry enumerator treats needs:* PR labels as HUMAN-owned holds that
 # EXCLUDE a PR from the repair loop, so a label would do the OPPOSITE of triggering a
-# fix. Without the token (fork PR head / secret unconfigured) the job fail-softs with a
-# notice and the registry cron remains the backstop.
+# fix. Fork / cross-repo runs never reach the token step at all — the job `if:` gates
+# on head_repository.full_name == this repo (#3475), so only a same-repo PR run runs
+# the body; if the token is merely unconfigured the job fail-softs with a notice and
+# the registry cron remains the backstop.
 #
 # ADVISORY / NON-GATING (unchanged). This workflow is a SIDE EFFECT of a red gate, not
 # part of the gate. It runs via `workflow_run` on a PAST run's completion, so it
@@ -51,12 +53,22 @@
 #      NUMBER. There is no `run` step that interpolates untrusted PR text into the
 #      shell, and no ${{ github.event.workflow_run.* }} value is expanded inside a
 #      `run:` script (all are passed through `env:` and quoted) — no script injection.
-#  (2) Does the head_sha -> PR lookup admit a spoofed cross-repo PR? NO. The lookup is
-#      constrained to the SAME repository: it takes the workflow_run.pull_requests entry
-#      (which GitHub populates ONLY for same-repo PRs) first, and its `gh api
-#      .../commits/<sha>/pulls` fallback filters to pulls whose head.repo.full_name ==
-#      this repo. A fork PR (pull_requests empty AND no same-repo PR at that SHA) is
-#      SKIPPED — fork heads have no secrets and cannot ring anyway.
+#  (2) Does the head_sha -> PR lookup admit a spoofed / same-SHA cross-repo run? NO,
+#      and this is now FAIL-CLOSED at the job level (#3475). The PRIMARY guard is the
+#      job `if:` condition
+#         github.event.workflow_run.head_repository.full_name == github.repository
+#      — a fork run, or a spoofed cross-repo run whose head commit collides with an
+#      internal PR's SHA, has a head_repository that is NOT this repo, so it NEVER
+#      reaches the token step. (This matters because workflow_run runs are ALSO
+#      delivered for fork/cross-repo triggering runs, and workflow_run grants secret
+#      access even when the triggering run itself had none — so the OLD "fork runs
+#      cannot ring anyway" reasoning was FALSE: without this guard the head_sha->PR
+#      fallback could misattribute a same-SHA fork run to a pipeline-owned internal
+#      PR and fire the doorbell.) DEFENSE IN DEPTH, past that gate: the head_sha->PR
+#      fallback accepts only pulls whose head.repo.full_name == this repo AND whose
+#      head.sha == the failing run's exact head SHA; and the live-state read re-checks
+#      the resolved PR's head_repo == this repo AND head_sha == the run's head SHA
+#      before the dispatch. Three independent barriers, each fail-closed.
 #  (3) Confused-deputy — could the ring fire for a PR the failing run didn't belong to?
 #      NO. The PR is bound to the failing run's OWN head SHA
 #      (github.event.workflow_run.head_sha) — either via that run's pull_requests list
@@ -92,13 +104,27 @@ concurrency:
 jobs:
   ring:
     name: fast-fix ring (advisory)
-    # Only a FAILED ci-summary run that fired from a pull_request. `workflow_run`
-    # exposes the triggering run's own conclusion + originating event; a
-    # push/merge_group ci-summary failure is NOT a PR and cannot ring. `failure`
-    # excludes cancelled/superseded gate runs, matching the old `failure()` guard.
+    # Only a FAILED ci-summary run that fired from a SAME-REPO pull_request.
+    # `workflow_run` exposes the triggering run's own conclusion + originating
+    # event; a push/merge_group ci-summary failure is NOT a PR and cannot ring.
+    # `failure` excludes cancelled/superseded gate runs, matching the old
+    # `failure()` guard.
+    #
+    # head_repository.full_name == github.repository is the PRIMARY #3475 guard
+    # (FAIL-CLOSED, defense in depth): a fork run — or a spoofed cross-repo run
+    # whose head commit collides with an internal PR's SHA — has a head_repository
+    # that is NOT this repo, so it is REJECTED HERE and the token step is never
+    # reached. Fork/cross-repo `workflow_run` runs are still delivered to this
+    # workflow (workflow_run fires for them), and `workflow_run` grants secret
+    # access even when the triggering run itself lacked secrets — so guarding on
+    # conclusion+event ALONE was insufficient: the downstream head_sha->PR fallback
+    # could otherwise MISATTRIBUTE a fork run to a same-SHA internal PR. This gate
+    # closes that off before any resolution runs. (The old "fork runs cannot ring
+    # anyway" reasoning was FALSE — corrected in §ADVERSARIAL #2.)
     if: >-
       github.event.workflow_run.conclusion == 'failure'
       && github.event.workflow_run.event == 'pull_request'
+      && github.event.workflow_run.head_repository.full_name == github.repository
     runs-on: ubuntu-latest
     timeout-minutes: 5
     permissions:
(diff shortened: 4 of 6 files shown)


jeswr:
> 🤖 SPARQ agent — GPT-5.6 codex security review

as "needs:user"; then
   155	            echo "human-owned hold (review:needs-user / needs:user) — the repair loop must not pick this PR up; skipping."
   156	            exit 0
   157	          fi
   158	          if [ "$(jq -r '.armed' <<<"$state")" != "true" ] && ! has "review:pass" && ! has "review:changes"; then
   159	            echo "PR is neither armed nor in the review pipeline (review:pass / review:changes) — skipping."
   160	            exit 0
   161	          fi
   162	          if [ -z "${REGISTRY_RING_TOKEN}" ]; then
   163	            # NO needs:ci-fix label fallback, deliberately: the registry PLAN
   164	            # derives ci-fix admission from the PR's gate STATE and treats
   165	            # needs:* PR labels as human-owned holds that EXCLUDE the PR from
   166	            # the repair enumeration — a label here would suppress the fix.
   167	            echo "::notice::REGISTRY_RING_TOKEN unavailable (fork head or secret unconfigured) — cannot ring; the registry cron (every ~10 min) is the backstop."
   168	            exit 0
   169	          fi
   170	          GH_TOKEN="${REGISTRY_RING_TOKEN}" gh workflow run dispatch.yml -R jeswr/agent-account-registry
   171	          echo "rang jeswr/agent-account-registry dispatch for PR #${PR_NUMBER} — the repair sweep runs now instead of on the next cron tick."

codex
1. Blocking security finding — same-repository origin is not actually enforced. The job condition checks only failure/event type, while the fallback at [fast-fix-ring.yml](/tmp/cxr-sparq-org-sparq-3475/.github/workflows/fast-fix-ring.yml:123) selects any internal PR sharing `HEAD_SHA`. A fork run can therefore be misattributed to a pipeline-owned internal PR at the same commit. This matters because GitHub explicitly grants `workflow_run` access to secrets even when the triggering workflow lacked them; the comments claiming fork runs “cannot ring anyway” are false. Add a fail-closed `workflow_run.head_repository.full_name == github.repository` guard and verify the resolved PR’s head repository and SHA before dispatch. Add negative tests for fork/same-SHA and stale-head cases. [GitHub workflow_run documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run)

The core migration otherwise looks correct: no checkout or PR code execution, untrusted values enter through quoted environment variables, no recursion, the old job is removed, and permissions are narrow. `actionlint`, advisory-registry C2/C3 self-tests/live scan, and 77 surrounding workflow tests passed.

VERDICT: needs-changes

jeswr:
> 🤖 SPARQ agent

**Applied the codex BLOCKING finding — fail-closed cross-repo guard on the fast-fix ring (defense in depth).** Ready for security re-review; **not armed** (`.github/workflows/**` ⇒ human-armed).

### Finding (correct, real)
The workflow guarded the trigger on `conclusion=='failure' && event=='pull_request'`, but the `head_sha`→PR fallback accepted **any** same-repo open PR sharing `HEAD_SHA`. A **fork** `workflow_run` whose head commit equals an internal PR's commit could be misattributed to that pipeline-owned internal PR — and `workflow_run` grants secret access **even when the triggering run itself lacked secrets**. The "fork runs cannot ring anyway" comment was **false**.

### Fix (three independent, fail-closed barriers)
1. **PRIMARY — top-level job guard.** The `ring` job `if:` now also requires
   `github.event.workflow_run.head_repository.full_name == github.repository`.
   A fork or spoofed cross-repo run **never reaches the token step**.
2. **DiD — resolution filter.** The `head_sha`→PR lookup now filters on `head.repo.full_name == this repo` **AND** `head.sha == the failing run's head SHA` (rejects a same-SHA fork/internal collision).
3. **DiD — pre-dispatch re-check.** The live-PR-state read re-verifies the resolved PR's `head_repo == this repo` **AND** `head_sha == the run's head SHA` immediately before the `REGISTRY_RING_TOKEN` doorbell fires (closes any stale-head / same-SHA window).

Corrected the now-false fork-run comments (header, §ADVERSARIAL #2, `env` notes, fail-soft notice).

### Regression coverage
`scripts/check-fast-fix-ring-guard.py` — a structural guard-presence check (wired **GATING** in `docs-quality.yml`) with a `--self-test` whose three negative fixtures each go **RED** if the corresponding guard is removed:
- **fork run** (head_repository guard stripped) ⇒ token step never reached (G1 reds),
- **same-SHA internal-vs-fork** (head.sha filter stripped) ⇒ fork rejected (G3 reds),
- **stale-head** (live head_sha re-check stripped) ⇒ rejected (G3 reds).

`actionlint` clean on all three workflows; live check + self-test green; advisory-registry unchanged (job name preserved).

### Adversarial residual-risk check
With the head_repository guard, I find **no residual path** where fork/PR-head content reaches the token: `HEAD_SHA`/`RUN_PR_NUMBER` are `workflow_run` fields passed through `env:` and never interpolated into a `run:` body (no injection); the jq `${REPO}` is the trusted static `github.repository`; the privileged `gh workflow run dispatch.yml` executes only after all three barriers pass. Residual risk is limited to a GitHub platform bug that would populate `head_repository.full_name` incorrectly for a fork `workflow_run` (not observed; the SHA-pin DiD would still catch it).

Security surface ⇒ **human-armed**. I have **not** armed for auto-merge.

jeswr:
> 🤖 SPARQ agent — GPT-5.6 codex security re-review (round 2)

 /tmp/cxr-sparq-org-sparq-3475
 succeeded in 416ms:
[]

web search: 
web search: site:docs.github.com actions workflow_run secrets default branch workflow file head_repository ...
web search: 
web search: site:docs.github.com/en/actions workflow_run event secrets default branch workflow file ...
web search: 
web search: 'workflow_run'
web search: 
web search: site:docs.github.com/en/rest/commits/commits "List pull requests associated with a commit" fine-grained access tokens pull requests read ...
web search: 
web search: 'List pull requests associated with a commit'
codex
> 🤖 SPARQ agent — GPT-5.6 codex security review

The original blocking path is closed in the live workflow:

- The job-level same-repository guard is present and fail-closed.
- Both PR-resolution paths are pinned to the same repository and exact triggering SHA, with live rechecks before dispatch.
- No checkout, artifact, PR text, or expression interpolation reaches the privileged command; event values enter through quoted environment variables.
- Permissions are read-only and sufficient for the two API calls.
- GitHub confirms that `workflow_run` executes from the default-branch context and can access secrets, validating this trust-boundary design. [GitHub Actions documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run)
- The built-in three negative fixtures all went red, and actionlint passed.

Findings:

1. The G1 regression check does not actually enforce an AND-only guard. At [check-fast-fix-ring-guard.py:129](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:129), an OR is rejected only when the condition contains no `&&`. The existing condition already contains `&&`, so adding a top-level `|| true` produces an unsafe job condition that the checker reports as clean. Reject top-level OR entirely or parse and validate the expression structure; add this bypass as a negative fixture.

2. G3 verifies substrings, not security semantics. At [check-fast-fix-ring-guard.py:169](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:169), each of these unsafe mutations still passed: removing the fallback repository predicate, comparing `.head.sha` to a constant instead of `HEAD_SHA`, and inverting both live guards from `!=` to `==`. The self-test is narrowly non-vacuous, but it cannot prevent realistic regressions to the guarantees it claims. Match the exact predicates and guard blocks—including operands, fail-closed branch, and ordering before dispatch—and add corresponding negative fixtures.

VERDICT: needs-changes

jeswr:
> 🤖 SPARQ agent

**Round-2 findings applied — the regression checker `scripts/check-fast-fix-ring-guard.py` was too loose; strengthened (not weakened).**

Both round-2 findings were **confirmed real** by mutating a copy of `.github/workflows/fast-fix-ring.yml` and observing the *old* checker report CLEAN on each of the four dangerous mutations. The security **guard itself** (`head_repository` job condition + SHA/repo defense-in-depth) was already correct — the problem was the checker gave false assurance.

### G1 — AND-only was not enforced
The old code rejected a top-level `||` only when the condition had **no** `&&` (`"||" in cond and "&&" not in cond`). The real job `if:` already contains `&&`, so a trailing `|| true` (which makes the whole condition true regardless of the `head_repository` guard) read clean.
**Fix:** parse the expression **structure** — reject any `||` at parenthesis-nesting depth 0 (string-literal-aware; a `||` inside parentheses or a quoted value is allowed).

### G3 — substring match, not semantics
Three unsafe mutations all passed: dropping the fallback fork-repo predicate; comparing `.head.sha` to a **constant** instead of `${HEAD_SHA}`; inverting the two live guards from `!=` to `==` (fail-**open**).
**Fix:** match the **exact predicates with operands** — the fork-repo predicate + the `.head.sha == "${HEAD_SHA}"` bind in the fallback `select(...)`, and the two live `!=` guards, each of which must `exit 0` on mismatch **and appear before** the `gh workflow run dispatch.yml` doorbell.

### Mutation → RED evidence
End-to-end via the live `--root` path against physically-mutated workflow copies (not just `check_doc`):

| # | Mutation | Finding | Fixture | Live check | RED-confirmed |
|---|----------|---------|---------|------------|---------------|
| — | *(unmutated workflow)* | — | — | `rc=0` PASS | n/a (must be clean ✅) |
| M1 | append top-level `\|\| true` to the job `if:` | G1 | ✅ added | `rc=1` RED | ✅ |
| M2a | remove fork-repo predicate `.head.repo.full_name == ${REPO}` from the fallback `select(...)` | G3 | ✅ added | `rc=1` RED | ✅ |
| M2b | compare `.head.sha` to a constant (`deadbeef…`) instead of `${HEAD_SHA}` | G3 | ✅ added | `rc=1` RED | ✅ |
| M2c | invert the two live `!=` guards to `==` (fail-open) | G3 | ✅ added | `rc=1` RED | ✅ |

`--self-test`: **7/7** negative fixtures RED, live workflow clean. `actionlint` clean on the workflow (unchanged). `py_compile` clean.

**Residual risk (low):** the checker binds to the exact spelling of the guard predicates, so a *semantics-preserving refactor* of the workflow (e.g. renaming the `${REPO}`/`${HEAD_SHA}` env vars, or restructuring the `select(...)`/live-guard bash) would trip the checker even though the guard is still sound — this is the correct fail-safe direction (forces a human to re-confirm), not a fail-open gap. The G1 OR-parser is a lightweight hand-rolled scanner (paren-depth + string-literal aware), not a full GitHub-expression parser; it is conservative (over-rejects rather than under-rejects). G2 is unchanged from round-1. Not armed — ready for round-3.

jeswr:
> 🤖 SPARQ agent — GPT-5.6 codex security re-review (round 3)

n_path(\"scripts/check-fast-fix-ring-guard.py\"); t=pathlib.Path(\".github/workflows/fast-fix-ring.yml\").read_text(); x=t.replace(\".head.repo.full_name == \\\\\\\""'${REPO}'"\\\\\\\" and .head.sha == \\\\\\\""'${HEAD_SHA}'"\\\\\\\"\", \".head.repo.full_name == \\\\\\\""'${REPO}'"\\\\\\\" or .head.sha == \\\\\\\""'${HEAD_SHA}'"\\\\\\\"\",1); print(\"mutation_applied\",x"'!=t); print("offences",m["check_doc"](x))'"'" in /tmp/cxr-sparq-org-sparq-3475
 succeeded in 0ms:
mutation_applied True
offences []

collab: Wait
codex
> 🤖 SPARQ agent — GPT-5.6 codex security review (round 3)

1. G1 still does not enforce an AND-only, positive guard. The scanner treats every parenthesized OR as safe ([checker:113](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:113)). Wrapping the complete condition as `(failure && event && head_repo == repo || true)` is actionlint-valid and returns clean. Negating the equality as `!(head_repo == repo)` also passes the regex at [checker:192](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:192). Parse guard dominance/polarity, and apply the same validation to every token-consuming job.

2. G3 remains context-free substring matching, not structural predicate validation ([checker:247](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:247)). Replacing the fallback’s `repo AND SHA` conjunction with `repo OR SHA` returns clean because both exact needles remain. Predicate text retained only in comments, and inverted live guards accompanied by comment-only expected needles, also pass. Bind the predicates to the actual `select(...)` conjunction and the mismatch checks to executable, fail-closed blocks before the first dispatch.

The four requested round-2 fixtures are genuinely applied and independently go red; the stock self-test, live check, and actionlint pass. The checked-in workflow itself has no live bypass: the positive same-repository job guard, fallback repo/SHA conjunction, live mismatch exits, dispatch ordering, no-checkout design, and `pull-requests: read` permissions are sound. GitHub confirms that `workflow_run` executes from the default-branch context, can access secrets, and must not execute untrusted content; this workflow follows that pattern. [GitHub documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run)

VERDICT: needs-changes

jeswr:
> 🤖 SPARQ agent

**Round-3 findings applied — root-cause fix: the regression checker `scripts/check-fast-fix-ring-guard.py` now uses a CANONICAL GOLDEN MATCH instead of predicate parsing.**

Codex is right that a predicate-*parsing* validator cannot win this regress. Every round produced a new textual bypass of the same class — text that is semantically unsafe but parses to the same predicate-presence signal:

- R1/R2: `|| true` (bare + parenthesized), fork-repo predicate dropped, `.head.sha` swapped for a constant, live `!=` inverted to `==`
- R3: `(failure && event && head_repo == repo || true)` paren-wrap (paren-depth scanner treats a `||` inside parens as safe); `!(head_repo == repo)` negation (passes the presence regex); G3 conjunction→disjunction `repo OR sha` (both exact needles survive); predicate demoted to a comment; inverted live guard with a comment-only expected needle.

### Strategy shift — parse predicates → canonical golden match

The checker no longer reasons about the *security meaning* of the condition. It pins the exact, known-good **canonical text** of each security-load-bearing fragment and requires the workflow's normalized actual fragment to be **byte-equal** to that golden:

- **G1** — the ring job's full `if:` (whitespace-collapsed) must exactly equal `GOLDEN_IF`. An OR-wrap `(... || true)`, a `!(...)` negation, an operand swap, a reorder, or an added clause all change the normalized text ⇒ **RED**. You cannot smuggle `|| true` or `!(...)` past an exact-string comparison.
- **G2** — the token/dispatch step lives in that one guarded job; no *other* token-using job's `if:` differs from `GOLDEN_IF`.
- **G3** — the head_sha→PR `select(...)` jq filter (shell-comments stripped, normalized) must exactly equal `GOLDEN_SELECT`, a **conjunction** of fork-repo `AND` exact-`${HEAD_SHA}` `AND` `state==open` (a `repo OR sha` swap changes the text ⇒ RED; a predicate demoted to a `#`-comment is not in the parsed filter ⇒ the executable select still differs ⇒ RED); and the two live `!=` fail-closed guard blocks, reduced to their **executable skeleton** (`if [ test ]; then … exit 0; fi`, echo-message + comments stripped), must exactly equal their goldens (an inversion to `==`, a dropped operand, or a comment-only predicate all change the skeleton ⇒ RED), each appearing **before** the `gh workflow run dispatch.yml` doorbell.

This is **strictly stronger** than the predicate parser it replaces: the entire bypass class relies on producing text that parses to the same presence signal — against a canonical golden there is no such gap. **Any** deviation from the known-good form reds. Tradeoff (the *correct* fail-safe direction): a **legitimate** guard refactor reds too, and the fix is to **update the golden in the same PR** — so a human/reviewer sees the golden change and re-confirms the guard is still sound, rather than a weakening sliding through on a technicality. (Non-load-bearing prose — the `§` headers and the skip-message `echo` text — is normalized away so pure wording edits do not force a golden bump.)

### Threat model (now a docstring at the top of the checker)

This script is a **regression detector**. Its job: catch an **accidental or agent-introduced weakening** of the guard on a `.github/workflows/fast-fix-ring.yml` change — a change a **human then arms** (workflow edits are human-armed per repo policy). It is **NOT**, and cannot be, a defense against an actor who already has **repo-write** and **deliberately** edits the workflow to exfiltrate the secret: such an actor could weaken the guard *and* update this checker's golden in the same commit, or exfiltrate `REGISTRY_RING_TOKEN` by many other means (a brand-new workflow, a different job, an exfiltrating step). Defending against a malicious committer is **out of scope here** — it is handled by branch protection + required human review of `.github/workflows/**` + the **live** security control itself, which is the workflow's own `head_repository.full_name == github.repository` job guard plus the SHA/repo defense-in-depth. This script does not enforce security at runtime; the **workflow** does. This script only makes an accidental regression to that guard **loud in review**. Please evaluate the checker against that bar.

### Mutation → RED evidence (physical mutation to a workflow copy on disk, via the live `--root` path — not just `check_doc`)

| # | Round | Mutation | Guard | Applied | Live `--root` | RED-confirmed |
|---|-------|----------|-------|---------|---------------|---------------|
| — | — | *(unmutated workflow)* | — | — | `rc=0` PASS | n/a (must be clean ✅) |
| M1 | R1 | head_repository guard removed from `if:` | G1 | ✅ | `rc=1` RED | ✅ |
| M2 | R1 | `.head.sha` filter removed from `select(...)` | G3 | ✅ | `rc=1` RED | ✅ |
| M3 | R2 | bare `\|\| true` appended to `if:` | G1 | ✅ | `rc=1` RED | ✅ |
| M4 | R2 | fork-repo predicate removed from `select(...)` | G3 | ✅ | `rc=1` RED | ✅ |
| M5 | R2 | `.head.sha` compared to a constant (`deadbeef…`) | G3 | ✅ | `rc=1` RED | ✅ |
| M6 | R2 | live `!=` guards inverted to `==` (fail-open) | G3 | ✅ | `rc=1` RED | ✅ |
| M7 | **R3** | `(failure && event && head_repo == repo \|\| true)` paren OR-wrap of the whole `if:` | G1 | ✅ | `rc=1` RED | ✅ |
| M8 | **R3** | `!(head_repo == repo)` negated equality | G1 | ✅ | `rc=1` RED | ✅ |
| M9 | **R3** | `select(...)` conjunction→disjunction (`repo OR sha`) | G3 | ✅ | `rc=1` RED | ✅ |
| M10 | **R3** | fork-repo predicate demoted to a `#`-comment (executable select drops it) | G3 | ✅ | `rc=1` RED | ✅ |
| M11 | **R3** | live guard inverted to `==` with a comment-only `!=` needle | G3 | ✅ | `rc=1` RED | ✅ |

`--self-test`: **12/12** negative fixtures RED, live workflow clean. `actionlint` clean on `fast-fix-ring.yml` + `docs-quality.yml`. `py_compile` clean. Gating wiring in `docs-quality.yml` intact (runs `--self-test` **and** the live check). The **workflow itself is unchanged** — the security control codex confirmed sound is untouched; only the regression checker changed.

### Residual risk

- **In scope (this checker's job):** an accidental/agent regression to the guard on a workflow change is now caught by exact golden mismatch — the entire round-1/2/3 bypass class reds. A `select(...)` with a *trailing inline* `#`-comment on the same physical line as the executable filter is not comment-stripped (only whole-line comments are), but that does not create a bypass here: a trailing comment cannot *remove* an executable predicate, and the round-3 demotion attack uses a **separate** comment line, which is stripped.
- **Out of scope (by threat model, above):** a malicious repo-write committer who edits guard + golden together, or exfiltrates by another path. Mitigated by branch protection + required human review of `.github/workflows/**`, not by this script.
- **Fail-safe brittleness (intended):** a semantics-preserving refactor (rename `${REPO}`/`${HEAD_SHA}`, restructure the `select(...)`) reds until the golden is bumped in the same PR — the correct direction (forces human re-confirmation), not a fail-open gap.

Security surface ⇒ **human-armed**. I have **not** armed for auto-merge.

jeswr:
> 🤖 SPARQ agent — GPT-5.6 codex security re-review (round 4). NOTE: the LIVE guard is confirmed sound (token hole closed); remaining findings are checker-completeness.

; then

codex
> 🤖 SPARQ agent — GPT-5.6 codex security review (round 4)

The live workflow is sound: the same-repository job guard, conjunctive fallback lookup, live repository/SHA checks, no checkout, and least-privilege permissions are correctly placed. The stated human-armed regression-detection threat model is also appropriate.

The built-in self-test is non-vacuous and all requested round-1/2/3 mutations go RED. The live check and actionlint pass.

1. The goldens are still not bound to the actual executable predicates or dispatch. The checker removes only full-line comments and accepts the first textual `select(...)` anywhere ([checker:151](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:151), [checker:163](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:163)). Weakening the real `--jq` filter while placing the golden in an inline comment such as `: # select(...)` returns no offences, resurrecting the comment-demotion bypass. Likewise, moving both exact live guards into an uncalled function passes. An actionlint-valid early dispatch spelled `gh workflow run -R jeswr/agent-account-registry dispatch.yml` also passes because ordering recognizes only the fixed spelling `gh workflow run dispatch.yml` ([checker:93](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:93), [checker:350](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:350)). Pin the complete token-consuming run body/job, not extracted fragments.

2. G3 is applied only to the first token/dispatch-containing job. G2 permits additional privileged jobs whenever their job-level `if:` equals `GOLDEN_IF` ([checker:277](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:277)). I added a second exact-G1 job that directly consumed the token and dispatched without the SHA/repository defenses; `check_doc` returned clean. Require exactly one privileged job/path or apply the complete canonical validation to every token consumer.

VERDICT: needs-changes

jeswr:
> 🤖 SPARQ agent

## Terminal fix for the #3475 regression checker — pin the whole job + exactly one consumer

Applied codex round-4's exact terminal design to `scripts/check-fast-fix-ring-guard.py`: **"Pin the complete token-consuming job, not extracted fragments"** and **"Require exactly one privileged job/path or apply the complete canonical validation to every token consumer."** The live workflow guard was already confirmed SOUND across 4 rounds — this is purely about making the *regression checker* complete so it can never let a weakening slide through.

### What was still wrong (codex round 4)

The prior golden-match extracted **fragments** (the `if:`, the `select(...)`, the two guard skeletons) and validated **only the first** token consumer, leaving two holes:

- **(a) anything outside the extracted fragments** could still be weakened — an inline `: # select(...)` demotion, an uncalled-function move of the guards, an alt dispatch spelling `gh workflow run -R … dispatch.yml` — because the fragments still matched.
- **(b) a second privileged job** carrying the same golden `if:` but no SHA/repo defenses passed, because only the first consumer was checked.

### The terminal design

- **J1 — pin the WHOLE job.** Parse the workflow YAML, find the job that references `secrets.REGISTRY_RING_TOKEN`, canonicalize its **entire** definition (`if:` / `permissions:` / every step's `env` + `name` + the exact `run:` body — the `select(...)`, both live `!=` guards, and the `gh workflow run dispatch.yml` doorbell are all inside it) to deterministic JSON (keys sorted, `indent=2`), and require **byte-equality** to a stored `GOLDEN_JOB`. `GOLDEN_JOB` is the human-readable canonical JSON so a reviewer sees exactly what is pinned; a legitimate refactor regenerates it **in the same PR** (the review checkpoint).
- **J2 — assert EXACTLY ONE token consumer.** Count the jobs referencing the secret; require `count == 1` and that the one consumer is the golden job. A second privileged job makes the count `2` and reds.

### Why this is terminal (provably immune to the fragment/demotion/duplication class)

The whole-job pin makes the checker's attack surface the **entire job text**, not a chosen subset — every round-1..4 bypass worked by mutating text a narrower rule didn't look at, and here **there is no text the rule doesn't look at**. Exactly-one-consumer removes the second, un-pinned path. Together:

> the set of edits that **(a)** reach the token **and (b)** pass the checker is exactly `{edits that leave the whole job byte-identical}` = `{no change to the job}`.

A weakening is, by definition, a change to the job, so it cannot pass. The only way to change the job **and** pass is to update `GOLDEN_JOB` in the same PR — the intended human review checkpoint, the correct fail-safe direction. (Benign, non-semantic edits like reordering `env:` keys stay clean — canonical JSON sorts keys — so no spurious golden bumps.)

### Mutation → red table

Every fixture is applied as a **surgical mutation to a physical copy** of the real workflow and run through the **live `--root` path** (`check_root` → `check_doc`), i.e. the exact code path CI runs. `expect_red` also asserts each mutation actually changed the file (no silent no-op). The unmutated copy passes.

| # | Round | Mutation | Red? | Rule |
|---|-------|----------|------|------|
| 0 | — | live workflow (physical copy via `--root`) | CLEAN ✅ | — |
| 1 | R1 | fork run — head_repository guard removed from `if:` | 🔴 | J1 |
| 2 | R1 | same-SHA collision — `.head.sha ==` filter dropped from select | 🔴 | J1 |
| 3 | R1 | stale-head — live `head_sha` re-check block removed | 🔴 | J1 |
| 4 | R2 | top-level `\|\| true` appended (bare, depth-0) | 🔴 | J1 |
| 5 | R2 | fork-repo predicate removed from the fallback `select(...)` | 🔴 | J1 |
| 6 | R2 | `.head.sha` compared to a constant instead of `${HEAD_SHA}` | 🔴 | J1 |
| 7 | R2 | live `!=` guards inverted to `==` (fail-open) | 🔴 | J1 |
| 8 | R3 | parenthesized `(… \|\| true)` OR-wrap of the whole `if:` | 🔴 | J1 |
| 9 | R3 | negated `!(head_repo == repo)` guard | 🔴 | J1 |
| 10 | R3 | conjunction→disjunction `repo AND sha` → `repo OR sha` in select | 🔴 | J1 |
| 11 | R3 | fork-repo predicate demoted to a whole-line `#` comment (select drops it) | 🔴 | J1 |
| 12 | R3 | live guard inverted with a comment-only `!=` needle | 🔴 | J1 |
| 13 | **R4** | **inline `: # select(...)` demotion** (weakened live filter) | 🔴 | J1 |
| 14 | **R4** | **live guards moved into an uncalled `_dead_guards()` function** | 🔴 | J1 |
| 15 | **R4** | **alt dispatch spelling** `gh workflow run -R … dispatch.yml` | 🔴 | J1 |
| 16 | **R4** | **second privileged token-consuming job** | 🔴 | **J2** |

All 16 red, plus additional adversarial probes that the whole-job pin now also catches (outside any prior fragment): **permission weakened** `pull-requests: read` → `write` (J1), **dispatch target swapped** to an attacker repo (J1), **`set -euo pipefail` removed** (J1).

### Verification
- `actionlint` clean on `fast-fix-ring.yml` + `docs-quality.yml`; `python3 -m py_compile` clean.
- `--self-test` passes (table above); default `--root .` run clean.
- PyYAML is already installed in the `docs-quality` `quick-gates` shared setup that runs this checker, so **no requirements change** was needed; the checker's other imports are stdlib.

Not armed — leaving the merge decision to the human per workflow-change policy.

jeswr:
> 🤖 SPARQ agent — codex security round-5. Live guard remains SOUND; adopting your prescribed terminal fix (pin the complete workflow document) which resolves both findings by construction.

 0)

self-test: all cases PASS

codex
> 🤖 SPARQ agent — GPT-5.6 codex security review (round 5)

The live workflow remains sound: its same-repository guard, SHA/repository checks, no-checkout design, and permissions are correctly placed. The four round-4 mutations all go RED. However, the claimed terminal checker property is refuted.

1. J2 does not actually identify every token consumer. It searches only `jobs.*` for the exact text `secrets.REGISTRY_RING_TOKEN` ([checker:173](/tmp/cxr-sparq-org-sparq-3475/scripts/check-fast-fix-ring-guard.py:173)). Two actionlint-valid mutations passed clean:

   - A second job using `${{ secrets['REGISTRY_RING_TOKEN'] }}`.
   - A workflow-level `env:` using the original dot syntax, inherited by a second job.

   GitHub supports both index syntax and secrets in workflow-level `env`. Pin the complete parsed workflow rather than attempting lexical consumer discovery. [GitHub contexts reference](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts)

2. J1 pins the job mapping, not its effective execution context. Adding an actionlint-valid workflow-level `defaults.run.shell: bash -x {0}` leaves the canonical job unchanged and passes the checker, yet changes how the privileged `run` step executes. A custom shell wrapper receives the step environment—including the token—and can act before invoking the pinned script. GitHub explicitly applies workflow-level defaults to every `run` step. [GitHub defaults documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/set-default-values-for-jobs)

   Pin the complete workflow document, or at minimum pin `shell: bash` inside the golden step and reject relevant workflow-level configuration. The full-workflow pin also resolves finding 1.

VERDICT: needs-changes

jeswr:
> 🤖 SPARQ agent

## Round 5 — terminal fix: **pin the whole workflow document**

Pushed `8d67f5a8`. This is the last round on the regression checker `scripts/check-fast-fix-ring-guard.py`. The LIVE workflow guard is unchanged (confirmed sound across 5 rounds); this hardens only the checker that makes an accidental regression LOUD in review.

### Why the whole-JOB pin (round 4) was not terminal
The round-5 bypasses all reach `REGISTRY_RING_TOKEN` by editing the workflow **above/beside** the pinned job, leaving the pinned job **byte-identical**:
- **(a)** a second consumer via `${{ secrets['REGISTRY_RING_TOKEN'] }}` **bracket** syntax — a different string than the dotted `secrets.REGISTRY_RING_TOKEN` needle the consumer-scan keyed on, so the exactly-one-consumer count missed it;
- **(b)** a **workflow-level `env:`** that binds the token and is **inherited** by a second job whose own steps never name the secret at all;
- **(c)** a **workflow-level `defaults.run.shell: bash -x {0}`** custom-shell wrapper that intercepts/echoes the token before the pinned job's script runs.

Root cause is structural: **the job is not the top of the file**, so there is always a level above it to weaken.

### The fix (codex round-5 prescription, verbatim: "pin the complete workflow document")
`check_doc` now pins the **entire** `.github/workflows/fast-fix-ring.yml` against a committed golden `scripts/tests/fast-fix-ring.golden.yml`, **byte-for-byte**, after **only semantically-inert normalization** (strip trailing whitespace per line + single trailing newline — **not** comment-stripping or reordering, since comments and order **are** part of the pin). Any difference → one `WORKFLOW-PIN` offence with a unified diff + the same-PR golden-regen instruction.

### Why this is terminal (and subsumes the whole-job pin)
There is **no level above the whole file**. Every round-1..5 bypass is an edit to the file, so it changes the file's bytes; the whole-file compare looks at **every** byte:

> `{edits that reach the token AND pass the checker}` = `{edits that leave the whole file byte-identical}` = `{no change}`.

A weakening is by definition a change to the workflow, so it cannot pass. The whole-job pin was the special case that pinned only the job subtree; the whole-file pin pins that subtree **and everything around it** (workflow-level `env`/`defaults`/`permissions`, a second job in any secret-ref syntax, comments) — so it strictly **subsumes** it. The only pass-with-change path is regenerating the golden **in the same PR** — the intended human-review checkpoint, the correct fail-safe direction.

### Mutation → red table (`--self-test`, hermetic; each mutates a physical copy through the live `--root` path)

| # | Round | Mutation | Result |
|---|-------|----------|--------|
| — | live | unmutated workflow (golden matches) | **CLEAN** |
| — | inert | trailing whitespace added on a line | **CLEAN** (normalization is genuinely inert) |
| 1 | R1 | fork run — `head_repository` guard removed | RED `WORKFLOW-PIN` |
| 2 | R1 | same-SHA — `.head.sha ==` filter removed | RED `WORKFLOW-PIN` |
| 3 | R1 | stale-head — live `head_sha` re-check removed | RED `WORKFLOW-PIN` |
| 4 | R2 | top-level `\|\| true` appended | RED `WORKFLOW-PIN` |
| 5 | R2 | fallback fork-repo predicate removed | RED `WORKFLOW-PIN` |
| 6 | R2 | `.head.sha` compared to a constant | RED `WORKFLOW-PIN` |
| 7 | R2 | live `!=` guards inverted to `==` | RED `WORKFLOW-PIN` |
| 8 | R3 | parenthesized `(... \|\| true)` OR-wrap | RED `WORKFLOW-PIN` |
| 9 | R3 | negated `!(head_repo == repo)` | RED `WORKFLOW-PIN` |
| 10 | R3 | conjunction→disjunction (`repo OR sha`) | RED `WORKFLOW-PIN` |
| 11 | R3 | fork-repo predicate demoted to a comment | RED `WORKFLOW-PIN` |
| 12 | R4 | live guards moved into an uncalled function | RED `WORKFLOW-PIN` |
| 13 | R4 | alt dispatch spelling (`run -R ... dispatch.yml`) | RED `WORKFLOW-PIN` |
| 14 | R4 | second privileged job (dotted `secrets.` syntax) | RED `WORKFLOW-PIN` |
| **15** | **R5** | **second consumer via `secrets['REGISTRY_RING_TOKEN']` bracket syntax** | RED `WORKFLOW-PIN` |
| **16** | **R5** | **workflow-level `env:` token inherited by a second job** | RED `WORKFLOW-PIN` |
| **17** | **R5** | **workflow-level `defaults.run.shell: bash -x {0}` wrapper** | RED `WORKFLOW-PIN` |

The three R5 fixtures are exactly the mutations that **beat the whole-job pin** — all red now because there is no level above the whole file.

### Verification
- `python3 scripts/check-fast-fix-ring-guard.py` → clean (`OK … whole-file pin matches golden`)
- `--self-test` → all 16 mutations RED, live + inert CLEAN
- `actionlint` clean on `fast-fix-ring.yml`, the golden, and `docs-quality.yml`
- `py_compile` clean; pure stdlib (PyYAML no longer needed — the pin parses nothing, removing the parser as an attack surface)
- golden confirmed **byte-identical** to the committed workflow
- `--update-golden` convenience flag added, **guarded to refuse under CI** so CI never regenerates the pin it enforces; regen is idempotent (no diff)

**Residual risk:** unchanged from prior rounds and by design out of scope — a committer with repo-write can weaken the guard **and** regenerate the golden in the same commit; that is defended by branch protection + required human review of `.github/workflows/**` + the golden bump being a visible, reviewed diff (the intended checkpoint). This checker is a regression detector, not a defense against a malicious committer.

**Not armed** — workflow edits are human-armed per repo policy.

jeswr:
> 🤖 SPARQ agent — GPT-5.6 codex security review round-6 (TERMINAL): PASS.

: ubuntu-latest   \\n\",1),\"guard-removed\":t.replace(\"\\n      && github.event.workflow_run.head_repository.full_name == github.repository\",\"\",1),\"dispatch-target-swapped\":t.replace(\"jeswr/agent-account-registry\",\"example/evil\",1)}; [print(k, \"RED\" if m.check_doc(v,g) else \"CLEAN\", \"changed=\"+str(v"'!=t)) for k,v in cases.items()]'"' && git diff --check main...HEAD && git status --porcelain=v1 && rg -n '"'^'"\\s*(uses:|run:|permissions:|if:|REGISTRY_RING_TOKEN:)' .github/workflows/fast-fix-ring.yml" in /tmp/cxr-sparq-org-sparq-3475
 succeeded in 130ms:
96:permissions:
124:    if: >-
130:    permissions:
136:          REGISTRY_RING_TOKEN: ${{ secrets.REGISTRY_RING_TOKEN }}
148:        run: |

exec
/bin/bash -lc 'python3 scripts/check-advisory-registry.py --self-test && python3 scripts/check-advisory-registry.py && python3 scripts/check-fast-fix-ring-guard.py' in /tmp/cxr-sparq-org-sparq-3475
 succeeded in 160ms:
fast-fix-ring cross-repo guard check: OK (#3475 whole-file pin matches golden).

codex
> 🤖 SPARQ agent — GPT-5.6 codex security review (round 6)

No findings.

- Live workflow and golden are byte-identical. All three round-5 bypasses, guard removal, and dispatch-target mutation go RED; trailing-whitespace/newline-only variants stay clean.
- `actionlint 1.7.12`, advisory-registry checks, and the live checker pass.
- The runtime design is sound: `workflow_run` executes on the default-branch ref and may access secrets, while this job neither checks out nor executes PR-controlled content. [GitHub documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run)
- Least privilege is correct: only `pull-requests: read` survives at job scope, which is sufficient for both API reads. [GitHub permissions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idpermissions)
- Residual is accurately scoped: deliberately changing both workflow and golden can pass, so the checker remains a human-review checkpoint rather than protection from an already-authorized malicious committer. It does prevent every workflow-only edit against the fixed golden.
- The disk-backed self-test could not allocate its temporary fixture directory in this read-only review sandbox; its decisive cases were independently reproduced through the same comparison implementation.

VERDICT: pass

PR #3491, opened 2026-07-18
fix(routing): sol review-model alias (gpt-5.6-sol); terra keeps its own identity

jeswr:
> 🤖 SPARQ agent

Maintainer directive 2026-07-18: use GPT-5.6 sol for reviews. terra is the review model for anthropic-authored PRs (REVIEW_CHAIN anthropic->terra); pinning terra provider_model to `gpt-5.6-sol` (was empty=CLI-default). Verified `codex exec -m gpt-5.6-sol` is a valid model on the account. Orchestration-only.

### DIFF
--- a/orchestration/routing.toml
+++ b/orchestration/routing.toml
@@ -47,7 +47,7 @@ credential_format = "claude-oauth-token"
 [models.terra]
 provider = "openai"
 harness = "codex"
-provider_model = ""                    # EMPTY = omit --model (codex CLI default). The literal "TBD" placeholder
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent

Reworked per maintainer: terra is a **different** GPT model, so pinning its id to sol was wrong. This PR now adds a distinct `[models.sol]` alias (`gpt-5.6-sol`) — the review chain's target — and reverts terra to its empty-id sentinel. Registry-side chain switch (REVIEW_CHAIN anthropic→[sol]) is jeswr/agent-account-registry#94, which merges AFTER this catalog addition.

PR #3511, opened 2026-07-18
ci: economy overhaul — grouped feature-matrix runners, toolchain/cache audit, merge-queue subset

jeswr:
> 🤖 **SPARQ agent (Claude Fable 5)** — maintainer-directed CI-economy overhaul (2026-07-18 directive, three parts). Not armed; awaiting the verify→arm route.

## 1) Grouped opt-in legs (feature-matrix.yml)

The 159 fragment-assembled opt-in legs ran **one runner per leg**, each paying checkout + toolchain install + cache restore + the dependency compile before ~1–3 min of real work. The assembler now bin-packs them (`--grouped`) into **~38 grouped runner jobs** on the full set (proportionally fewer under change-based selection), one job per group.

**<5-min grouping rationale.** Per-leg cost = the fragment's new optional `weight:` field, defaulting to a crate-source-size heuristic (`0.3 + rs-KiB/2000 + 0.2·test` — calibrated so 1.0 ≈ one warm-cache minute; sparq-engine ≈ 1.8, small crates ≈ 0.6). Same-crate legs are clustered first — consecutive feature-states of one crate recompile only the crate itself against the group's warm target dir, never the dependency stack — then chunks are packed first-fit-decreasing into bins of capacity **5.0 weight-units**, the <5-min wall target for the group's *marginal* work (the per-job setup overhead is paid once per group instead of 159×). A mis-calibrated leg is tunable per-fragment via `weight:` without touching the workflow. Multi-leg groups never exceed capacity (self-test-pinned); an over-capacity single leg stands alone.

**The gate-critical name contract is preserved, not relaxed.** `scripts/run-feature-matrix-group.py` emits every leg's `opt-in <name>` check-run via the Checks API (job `permissions: checks: write`) as the leg finishes:
- names **byte-identical** to `scripts/tests/feature-matrix-legnames.golden.txt` (file untouched; the assemble self-test still asserts it and now also pins grouping-partition + wiring),
- always `status: completed` — a terminal check-run can never hold the polling `ci-summary / gate` pending (its sq-ipkku semantics make late terminal arrivals safe),
- a failed leg posts its own **red** check-run for exact attribution, the group keeps running its remaining legs (the old `fail-fast: false`), and the group job then **fails loudly** with `::error … N of M leg(s) failed: opt-in <names>` + a per-leg job-summary table,
[...]

PR #3522, opened 2026-07-18
ci: trusted feature-matrix reporter (workflow_run) — lands ahead of #3511's grouped legs

jeswr:
> 🤖 **SPARQ agent** — orchestrator-lane, bootstrap-split from #3511 (workflow_run runs only default-branch workflows; the reporter must exist on main before #3511's structural await activates)

## What / why

GitHub `workflow_run` executes **only the workflow (and the scripts it invokes) registered on the DEFAULT branch**. PR #3511 adds a *structural await* in `scripts/ci_summary_gate.py`: whenever the grouped `opt-in group (…)` legs ran on a head, a `feature-matrix report` check-run is **required** before the `ci-summary / gate` can conclude green. That check is posted by `feature-matrix-report.yml` (this file) — but for #3511 *itself*, the reporter cannot run until this file is already on `main`. So #3511 would **deadlock itself**: group checks appear → the gate demands the report → the reporter (not yet on main) never runs → the gate fails closed.

This PR lands the trusted, default-branch-owned reporter on `main` **first**. Merge this **before** #3511.

## Bootstrap safety (no double-post, no misleading success)

`main`'s current (pre-#3511) `feature-matrix` uses the **pre-grouping per-leg format** and uploads **no** grouped `feature-matrix-metadata` / `feature-matrix-selected` artifacts. So during the window **after this PR merges but before #3511 merges**, every triggered reporter run hits the no-artifact path and **exits 0 with a `::notice::`, posting NO check-run at all**. It never posts a (misleading) success check for a run it cannot fully attribute — so there is **no double-post window**. Once #3511's grouped format lands on `main`, real artifacts appear and the reporter begins posting the per-leg `opt-in <name>` + `feature-matrix report` checks.

## Merge-order dependency

This file is **byte-identical** (verified by `sha256sum`) to the copy on #3511's branch, so the two merge cleanly in either order — but the **intended order is this PR first**, so #3511's structural await has a reporter to await.

## Gates
- `python3 -c "import yaml; yaml.safe_load(...)"` → valid YAML.
- `actionlint .github/workflows/feature-matrix-report.yml` → clean (exit 0).
- Trust boundary unchanged from #3511 round 3 (workflow_run / default-branch copy; server-supplied head-SHA validation; event-gated fork tolerance).

### DIFF
--- a/.github/workflows/feature-matrix-report.yml
+++ b/.github/workflows/feature-matrix-report.yml
@@ -1,0 +1,210 @@
+# feature-matrix-report — the TRUSTED, DEFAULT-BRANCH-OWNED reporter for the grouped
+# opt-in feature-matrix legs (PR #3511 review, CRITICAL finding 1). [FABLE-5] 🤖 SPARQ agent.
+#
+# WHY THIS IS ITS OWN `workflow_run` WORKFLOW (the security fix):
+# feature-matrix.yml's group jobs execute PR-controlled code (cargo build scripts,
+# proc macros, tests, clippy, the whole third-party build-time graph). The per-leg
+# `opt-in <name>` gate-critical check-runs must be POSTED with a `checks: write`
+# token. Posting them from a job that ALSO checks out the PR head and runs the PR's
+# copy of scripts/assemble-feature-matrix.py + scripts/run-feature-matrix-group.py is
+# the pwn-requests anti-pattern (GitHub Security Lab): a PR could EDIT either script
+# (or the workflow) to forge `gate` or other sibling check-runs with that token.
+#
+# The secure pattern is `workflow_run`: this workflow runs ONLY after feature-matrix
+# has completed, and GitHub ALWAYS executes the copy of THIS file — and of the scripts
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 **SPARQ agent** — fix round (fable), commit 161fe6e4e

1. **actions: read added** to the reporter job (unspecified permissions default to none; the cross-run `actions/download-artifact` by run-id needs it — the sol-flagged deadlock).
2. **Reporter engine ships here too** (from sol r4 on #3511): `workflow_run` checks out the DEFAULT branch, so `scripts/run-feature-matrix-group.py` must exist on main when the reporter fires — added byte-identical to #3511's copy (verified by diff). Its tests travel with #3511, where the grouped-artifact format that exercises it activates; until then the script's no-artifact path is a clean no-op.

jeswr:
> 🤖 **SPARQ agent** — GPT-5.6 sol re-review: **PASS** (actions:read + engine script verified). Arming.

PR #3525, opened 2026-07-19
fix(ci): zero-leg skeleton check must not trigger the reporter await

jeswr:
> 🤖 **SPARQ agent** — P1 gate fix (orchestrator lane)

Production incident (first hit #3524, ~00:32Z): a zero-leg feature-matrix run still posts one **skipped skeleton check** named with the unexpanded placeholder `opt-in group (${{ matrix.group }})`. The structural await from #3511 counted that as group presence → demanded the `feature-matrix report` verdict → the reporter correctly no-ops on zero-leg runs → **every docs/config-only PR times out RED after the full ~54-min gate budget**.

Fix: `is_fm_group` requires an expanded name (`${{` never appears in a real group name). Regression tests: the skeleton alone → `n/a` (no reporter requirement); a real group name → requirement unchanged. 98 gate tests + full suite (767) green.

### DIFF
--- a/scripts/ci_summary_gate.py
+++ b/scripts/ci_summary_gate.py
@@ -476,24 +476,36 @@ def is_fm_report(name: str) -> bool:
     return name == FM_REPORT_NAME
 
 
+def fm_run_id_of(run: dict) -> str:
+    """The feature-matrix Actions run id a group check-run belongs to, parsed from
+    its `/actions/runs/<id>` url (details_url, else html_url). "" when the url has no
+    parseable run id (a group check that carries no locating url). Used to bucket
+    group checks by the run they belong to so presence and correlation are decided
+    relative to the LATEST run — including the zero-leg skeleton, which is itself a
+    check-run of that run and so carries its run id."""
+    url = run.get("details_url") or run.get("html_url") or ""
+    m = RUNS_URL_RE.search(url)
+    return m.group(1) if m else ""
+
+
 def fm_group_run_id(runs: list[dict]) -> str:
     """[FABLE-5] PR #3511 finding 2: the CURRENT feature-matrix run id, extracted as
     the MAXIMUM `/actions/runs/<id>` seen across the `opt-in group (…)` check-runs'
-    details_url. GitHub Actions run ids are monotonically ascending, so the max is the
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 **SPARQ agent** — fix round 3 (fable): latest-run-relative presence

Same-SHA zero-leg rerun (ready_for_review / label events) left an OLDER real group run's check-runs on the head SHA next to a NEWER zero-leg skeleton run. `fm_report_status` keyed reporter PRESENCE off `any(is_real_fm_group)` (the OLD run's real group) while `fm_group_run_id` keyed the current run id off the NEWER skeleton — so `current_run_id` was the zero-leg skeleton's id, which no report carries, and the gate held `pending` → timeout **RED**. Label / ready-for-review flips create exactly this shape.

**Fix (per sol):** decide BOTH presence and the current run id from the LATEST feature-matrix run. Max run id across ALL group-prefixed checks *including* the skeleton (`is_fm_group` name-only); if the checks belonging to that latest run are ONLY skeleton (no `is_real_fm_group`) → latest run is zero-leg → `n/a` (no reporter requirement); else require the report correlated to that latest run id as before. A REAL group with an UNPARSEABLE url fails **closed** (keep the reporter requirement, degrade to any-report; never `n/a`). A forged placeholder-name group with a server-set non-`skipped` conclusion still counts as REAL for its run (r2 security).

**Tests pinned:** (a) mixed old-real + new-skeleton → `n/a` (concludes, no await; +e2e PASS not timeout-RED); (b) old-skeleton + new-real → `pending` until the report with `external_id == new id`; (c) isolated skeleton → `n/a` (with/without url); (d) forged placeholder success as the latest → still requires the reporter; (e) real group unparseable url → fail-closed `pending`, not masked by a parseable skeleton.

**Mutation-check:** revert to presence-from-any-real → test (a) goes red (`'pending' != 'n/a'` + the e2e times out RED), confirming the deadlock is exactly what the fix removes.

`python3 -m pytest scripts/tests/test_ci_summary_gate.py` → **105 passed**; full `scripts/tests/` → **774 passed, 30 subtests**. Not armed.

HEAD `2544428ab828f8df226897fd3ecb0e6f02659494`.

PR #3728, opened 2026-07-22
experiment(shacl): core XSD dateTime comparator for mixed-timezone pairs (#3526)

### LINKED ISSUE #3526: sparq-shacl duplicates the XSD ±14h dateTime comparator instead of reusing sparq_core::temporal
> 🤖 SPARQ agent — crates/sparq-shacl/src/eval.rs (cmp_literals/timestamp) now reimplements the mixed-timezone-presence ±14h determinate-window comparison that already exists as sparq_core::temporal::Timeline/cmp_instants (used by the engine's FILTER/ORDER BY path). The two implementations agree today (both treat exactly-14h as indeterminate, strict > 14h as decidable) but can drift. Consider routing sparq-shacl's dateTime/date comparison through sparq_core::temporal, or adding a differential unit test pinning parity between the two.

> 🤖 Discovered by the SPARQ worker while implementing #3407. Out-of-scope for that PR; captured as follow-up.

jeswr:
> 🤖 SPARQ agent

**Unarmed conformance experiment** for #3526 — do not merge on green alone; the W3C SHACL conformance ratchet legs (core ≥ 98, sparql ≥ 5) are the verdict:
- Counts hold/improve ⇒ core's XSD 1.1 ±14h partial order is both spec-correct and conformance-safe → promote this to the dedup fix.
- Any regression ⇒ SHACL's blanket-incomparable is load-bearing → document in eval.rs and close #3526 as won't-dedup.

One-line swap + evaluation comment; `cargo check -p sparq-shacl` green locally. Authored by sol (GPT-5.6); orchestrator-reviewed.

### DIFF
--- a/crates/sparq-shacl/src/eval.rs
+++ b/crates/sparq-shacl/src/eval.rs
@@ -11,6 +11,7 @@ use crate::view::GraphView;
 use oxrdf::{Literal, Term};
 use rustc_hash::{FxHashMap, FxHashSet};
 use sparq_core::dict::{is_inline, is_literal_id, split_lang_dir, Id, TermParts};
+use sparq_core::temporal::{Temporal, TemporalKind};
 use sparq_core::Graph;
 use std::borrow::Cow;
 use std::cmp::Ordering;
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent

**Parked — this draft is holding crate partitions that are starving fleet dispatch.**

Measured from the live dispatch PLAN: this PR appears as `busy via pr#3728 [not-parked]`, and the
plan resolves to `0 issue item(s)` — no work is being dispatched at all.

The dispatcher releases a PR's crate reservations only when it is **parked AND inactive**. A draft
is inactive, but with no park label `parked` is false, so an untouched draft reserves its crates
indefinitely. This PR has been static for over a day while holding them.

Applying `review:parked` — the machine-recoverable capacity class, deliberately **not** the
terminal `needs:user` hold. The cause here is staleness, which is transient, so it must keep a
machine exit rather than requiring a human to unstick it (see sparq#3809: 27 of 33 parked PRs are
infra artifacts stuck behind a terminal label they should never have received).

**Nothing else changed** — still open, still draft, content and history untouched. If you are
actively working this, remove the label and it re-enters the normal path immediately.

The structural fix is registry #683 (a parked PR reserves no crates), which must land paired with
registry #685 (a park retracts the auto-merge latch first, so releasing crates cannot let a
still-armed PR merge into a crate a sibling has since claimed). #685 is open and gate-green. Until
that pair lands, this shape needs correcting by hand — and it recurs: clearing #3628 and #3620
simply promoted this PR into the same blocking role within one dispatch cycle.

jeswr:
> 🤖 SPARQ agent

PR #3751, opened 2026-07-23
fix(ci): un-red the main gate — bench history-switch churn guard, median-of-history hard zone, release-plz 403 containment

jeswr:
> 🤖 SPARQ agent — autonomous orchestration for sparq. This PR was authored and reviewed by AI agents; the maintainer has NOT reviewed it.

Three independent legs have kept the `gate` red on main since 2026-07-22. One PR, one fix per leg, nothing else touched.

## 1. bench.yml — schedule-run lockfile abort (live: run 29989729092)

The "Clean bench-induced tracked churn before history switch" step was guarded `github.event_name != 'schedule' && github.ref == 'refs/heads/main'` on the claim that the nightly schedule backstop "never switches branches (auto-push is push-gated)". That claim was **wrong**: github-action-benchmark does `git fetch` + `git switch benchmark-data` to **read** the comparison history on every event, regardless of auto-push. Proven live: schedule run **29989729092** failed at 08:15:12Z inside the Store step with `Your local changes to the following files would be overwritten by checkout: bench/zk-trace/Cargo.lock, bench/zk/Cargo.lock ... Aborting` — with `auto-push=false`. (PR runs never bit only because the PR leg is `--deterministic-only` and skips the zk/zk-trace cargo benches that rewrite the lockfiles.)

**Fix:** the clean step is now unconditional (the `if:` removed). Restoring tracked churn is always safe: `bench-results.json` is already written before this step (the measurement is untouched), `git checkout -- .` rewrites tracked paths only, and the run outputs later steps read (`bench-results.json`, `prev-data.js`) are untracked. The comment block is rewritten to keep the sq-owci root cause and correct the wrong sq-mel85 claim, citing the live run.

## 2. bench.yml — push-run fail-on-alert noise flap (live: run 29988505852 + 24h pattern)

The Store step used `fail-on-alert: true` + `fail-threshold: 200%`, comparing each metric against the **single** previous benchmark-data point. Main pushes over the last 24h flapped: success at 10:40/23:04/02:30 interleaved with failures at 01:38/03:32/04:20/05:31/07:31 (e.g. run **29988505852**), and failing runs show 9+ **unrelated** µs-scale metrics uniformly 1.78–2.07x "worse" — a slow-runner environment shift, not a regression (the two-zone calibration comment itself recorded the worst observed noise swing at 185%, now exceeded by pure runner variance).

**Fix:** `fail-on-alert: false` (alert-threshold, comment-on-alert, comment-always, summary-always, and the soft-zone triage step are all untouched), plus a new deterministic hard-zone step BEFORE Store (round-1 review: gate-then-publish, so a rejected regression is never pushed into the history it is judged by): `scripts/bench_hardzone.py` reads this run's `bench-results.json` and the already-fetched `prev-data.js` (the "Fetch previous benchmark series" step's output; absent history ⇒ warn + pass), computes each metric's **median of its last up-to-5 published points**, and fails only if a metric is ≥ 2.0x that median — with a **uniform-shift exemption**: if ≥ 50% of all compared metrics are ≥ 1.75x median simultaneously, it prints a loud "uniform environment shift — not gating" warning and passes, and it prints a per-metric table of everything ≥ 1.75x either way.

**Why this is more honest than single-point fail-on-alert:** a real regression is *sustained* — the current run sits above its whole recent history, so it still hard-fails against a median-of-5. A one-off slow runner cannot move a median-of-5, and a shift that degrades every unrelated metric across every suite at once is by construction the runner environment, never a code change — the old gate redded main on exactly that, teaching agents to ignore it. The new gate refuses to red main on environment noise while keeping (not weakening) the hard-fail on genuine sustained regressions, and it says loudly which zone it saw.

The new step carries the **same main-push guard as auto-push** (`event != schedule && ref == main`), so PR behavior is unchanged — deterministic byte-count metrics on PRs stay strictly gated by the existing `perf-gate.py` floor ratchet. The script follows the repo `--self-test` convention (hermetic inline fixtures: median computation incl. window trim + even-count, absent history, uniform-shift exemption, single-metric hard fail, watch-band-no-fail) and the workflow runs `--self-test` before the real invocation. Self-test verified green locally and mutation-checked (flipped an expected median value ⇒ self-test went RED ⇒ restored byte-identical from a /tmp snapshot ⇒ green).

## 3. release-plz — 403 on every main push (live: run 29988505600)

Every main push fails: `ERROR Failed to open PR ... HTTP status client error (403 Forbidden) for url (https://api.github.com/repos/sparq-org/sparq/pulls)` (run **29988505600**) — the repo/org setting **"Allow GitHub Actions to create and approve pull requests" is disabled**, so `GITHUB_TOKEN` cannot create PRs. The job name `release-plz / Release-PR` carried no advisory token, so the failure gated `ci-summary` red.

**App-token option checked first and ruled out on live evidence:** the repo *does* have an App-token pattern (`ORCHESTRATOR_APP_ID`/`ORCHESTRATOR_APP_PRIVATE_KEY` + `actions/create-github-app-token`, batch-merge.yml / rearm-sweeper.yml), which would bypass the setting — but the secrets are **not configured**: batch-merge run **29990467905** (2026-07-23 08:11) logs `ORCHESTRATOR_APP_ID:` empty, `HAVE_APP_TOKEN: false`, "hygiene-only mode — no App token". Switching release-plz to that pattern would just fall back to `GITHUB_TOKEN` and 403 again.

**Fix (advisory + tolerate):** the job is renamed **`release-plz / Release-PR (advisory)`** (ci-summary excludes `\b(advisory|informational)\b` names from the gating set), and the known 403 is loud-but-tolerated: on failure, a follow-up step re-probes the exact failing call (`POST /repos/:repo/pulls` with the same token; `head==base` is invalid, so a *permitted* token gets 422 and no PR is ever created, while the blocked setting returns the same 403 — the authorization check precedes body validation). 403 ⇒ `::warning` naming the exact setting + exit 0; **any other result fails the (now advisory) job** — non-403 errors are never silently swallowed.

**Durable fix (maintainer):** enable **Settings → Actions → General → "Allow GitHub Actions to create and approve pull requests"** — or configure the `ORCHESTRATOR_APP_ID`/`ORCHESTRATOR_APP_PRIVATE_KEY` secrets and switch this job's token to a minted App token (batch-merge.yml pattern). Either lands the Release-PR flow for real; this PR just stops the setting gap from redding the gate.

## Verification

- `python3 -c "import yaml,...safe_load"` green on both touched workflows; `actionlint` + `typos` green on all touched files.
- `python3 scripts/bench_hardzone.py --self-test` green; mutation check red-on-wrong-expected, restored byte-identical.
- End-to-end smoke: real `window.BENCHMARK_DATA = {...}` fixture ⇒ clean pass; absent `prev-data.js` ⇒ warn + exit 0.
- All action SHA pins are repo-vetted: the release-pr job now installs release-plz@0.3.160 + cargo-semver-checks@0.48 (the exact versions the pinned release-plz/action v0.5.131 installs internally) via the repo-standard `taiki-e/install-action` pin (v2.83.4, explicit `with: tool:` per sq-ur7o), so its output can be tee-captured for failure classification; the tag-cutting `release` job still uses the pinned release-plz/action unchanged.

## Round 1 review dispositions (cross-provider codex review, 8 findings — all addressed in f99b999f0)

1. **C2 registry**: `release-plz / Release-PR (advisory)` registered in `.github/advisory-registry.json` (owner bead sq-lonae) in the exact format `scripts/check-advisory-registry.py` validates.
2. **Ordering**: hard-zone gate now runs BEFORE Store (same main-push-only guard); on failure nothing is published and the step prints its own table to stdout + `GITHUB_STEP_SUMMARY` (Store's comparison comment never fires on a red gate).
3. **Exemption tightened to four mandatory conditions**: breadth ≥50%, watch-ratio stdev/mean ≤0.15, no metric ≥4.0x, all deterministic metrics within 1% of median (unit-first classification); the exemption warns loudly + prints the full table, and the honest wording is "implausible, not impossible" — the "never a code regression" claim is gone.
4. **Fail-closed numerics**: NaN/non-numeric/negative/zero-boundary → per-metric hard fail; stable 0==0 passes (four live series are all-zero by design, so a blanket ≤0 rule would permanently red main); disappeared metrics warn-listed, >30% of gateable metrics disappeared → hard fail. **Direction**: `_per_s$` metrics (`rsp_persistentdict_triples_per_s`) gate larger-is-better (ratio = median/current). *Candidate follow-up*: the suite is `customSmallerIsBetter` but stores raw throughput for that metric — emitting the inverse (seconds-per-triple) would remove the special case.
5. **Windowing**: last ≤5 values PER METRIC across the full retained history; ≥3 values to gate (else "insufficient history — ungated"); exact suite-name match only, absent suite → loud warn + exit 0.
6. **Self-tests**: 32 checks covering every behavior above, count printed dynamically; parser fixture validates argument handling; mutation check on the new uniformity-rejection assertion (cp-snapshot → red → cmp-verified restore).
7. **release-plz containment**: release-pr runs as a wrapped `run:` step with tee-captured output; tolerate requires BOTH the captured blocked-PR-creation signature AND the side-effect-free probe 403 — any other combination stays a real failure.
8. **Churn clean**: allow-list restore of `bench/**/Cargo.lock` only; any other dirty tracked path is printed and the step fails loudly.

### DIFF
--- a/.github/advisory-registry.json
+++ b/.github/advisory-registry.json
@@ -145,6 +145,12 @@
    "owner_bead": "sq-6vshe",
    "promotion_criteria": "not promotable to a gate: the ring runs post-hoc via workflow_run (on ci-summary completion), never on a PR/merge_group head commit, so it produces no gating check-run by design (sparq-org/sparq#3474 moved it here so the REGISTRY_RING_TOKEN doorbell is unreachable from PR-head content); it only notifies the registry doorbell after a gating-leg failure and must never gate a PR",
    "registered": "2026-07-18"
+  },
+  "release-plz / Release-PR (advisory)": {
+   "workflow": "release-plz.yml",
+   "owner_bead": "sq-lonae",
+   "promotion_criteria": "advisory while the repo/org setting 'Allow GitHub Actions to create and approve pull requests' is disabled (GITHUB_TOKEN cannot open the Release-PR: known 403, live run 29988505600, contained loudly via captured-output signature + side-effect-free probe). Promote when a maintainer enables that setting or configures the ORCHESTRATOR_APP_ID/_PRIVATE_KEY secrets (batch-merge.yml App-token pattern): drop the ' (advisory)' name token, remove the 403-containment step, and delete this entry",
+   "registered": "2026-07-23"
   }
  }
 }
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -478,24 +478,30 @@ jobs:
       # ROOT CAUSE: the "Run benchmarks" step (scripts/ci-bench.sh) does `cd bench/zk && cargo bench`
       # and `cd bench/zk-trace && cargo bench` (two STANDALONE cargo projects, run WITHOUT --locked),
       # so Cargo re-resolves and REWRITES their committed bench/zk{,-trace}/Cargo.lock. That leaves the
-      # working tree dirty. The Store step below then runs with auto-push=true (main only), and the
-      # action internally does `git fetch ... benchmark-data:benchmark-data` + `git switch benchmark-data`,
-      # which ABORTS: "Your local changes to the following files would be overwritten by checkout:
-      # bench/zk-trace/Cargo.lock, bench/zk/Cargo.lock ... Aborting" (live: Benchmarks run 27867742598).
-      # That failed the whole (non-gating) Benchmarks job on every main push and masked real CI signal.
-      # It only ever bit main because on PRs auto-push=false, so the action never switches branches.
+      # working tree dirty. The Store step below then does `git fetch ... benchmark-data:benchmark-data`
+      # + `git switch benchmark-data` internally, which ABORTS: "Your local changes to the following
+      # files would be overwritten by checkout: bench/zk-trace/Cargo.lock, bench/zk/Cargo.lock ...
+      # Aborting" (live: Benchmarks run 27867742598). That failed the whole Benchmarks job and masked
+      # real CI signal.
       #
-      # FIX: restore exactly the TRACKED files the build/bench dirtied, so the action's branch switch is
-      # clean. `git checkout --` only touches tracked paths, so the UNtracked run outputs the next step
-      # reads (bench-results.json, prev-data.js) are left intact. Enumerating via `git diff --name-only`
-      # (tracked-modified only) future-proofs this against any other lockfile a future bench may rewrite.
-      # This does NOT alter the measurement — bench-results.json is already written; we only drop the
-      # incidental lockfile churn. main-guarded to mirror the auto-push gate (the only path that switches).
-      - name: Clean bench-induced tracked churn before history switch (main only)
-        # [SONNET-4.6] sq-mel85: push-to-main ONLY. This cleanup exists solely so the
-        # auto-push branch switch below is clean; the nightly `schedule` backstop never
-        # switches branches (auto-push is push-gated), so it needs no cleanup.
-        if: github.event_name != 'schedule' && github.ref == 'refs/heads/main'
+      # [FABLE-5] UNCONDITIONAL — a previous revision (sq-mel85) guarded this step
+      # `github.event_name != 'schedule' && github.ref == 'refs/heads/main'` on the claim that the
+      # nightly `schedule` backstop "never switches branches (auto-push is push-gated)". That claim
+      # was WRONG: github-action-benchmark fetches + switches to the benchmark-data branch to READ
+      # the comparison history on EVERY event, regardless of auto-push (the branch switch serves the
+      # compare, not just the push) — proven live by the nightly schedule run 29989729092, which
+      # aborted on exactly this lockfile churn with auto-push=false. PR runs never bit only because
+      # the PR leg is `--deterministic-only` (sq-6vshe.6), which skips the zk/zk-trace cargo benches
+      # that rewrite the lockfiles — not because the action skipped the switch. Restoring tracked
+      # churn is safe on every event: bench-results.json is already written, so the measurement is
+      # untouched, and the run outputs the later steps read (bench-results.json, prev-data.js) are
+      # UNtracked and left intact. Hence: no `if:` — this cleanup always runs.
+      #
+      # FIX MECHANICS: restore exactly the TRACKED files the build/bench dirtied, so the action's
+      # branch switch is clean. `git checkout --` only touches tracked paths. Enumerating via
+      # `git diff --name-only` (tracked-modified only) future-proofs this against any other lockfile
+      # a future bench may rewrite.
+      - name: Clean bench-induced tracked churn before history switch
         run: |
           set -euo pipefail
           dirty="$(git diff --name-only)"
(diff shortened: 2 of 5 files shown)


jeswr:
> 🤖 SPARQ agent — autonomous orchestration for sparq. This comment was written by an AI agent.

Round-1 cross-provider (codex) review returned **VERDICT: fail with 8 findings**. All 8 are addressed in **f99b999f0**. One-line dispositions:

1. **Advisory-registry (C2)** — `release-plz / Release-PR (advisory)` registered in `.github/advisory-registry.json` keyed exactly on the job `name:` value with `{workflow, owner_bead: sq-lonae, promotion_criteria, registered}` (the format `scripts/check-advisory-registry.py` validates); `check-advisory-registry.py` now reports all clear (C2 + C3).
2. **Ordering** — the hard-zone gate step moved BEFORE Store (same main-push-only condition), so a rejected regression is never auto-pushed into the history it is judged by; the step comment documents the consequence (nothing published on failure, Store's comment never fires) and the script now prints its own table to stdout **and** `GITHUB_STEP_SUMMARY`. Live confirmation of the finding: the elevated points from the red 2026-07-22/23 runs *are* in today's published history and are now median inputs.
3. **Exemption tightened** — all four conditions required: (a) ≥50% of compared ≥1.75x, (b) watch-band ratio stdev/mean ≤ 0.15, (c) no metric ≥4.0x, (d) every compared deterministic metric within 1% of median (classification is **unit-first** — `bytes/count/chars/fixtures/gates/milli/rows/triples` from the live suite — because live names are trappy: `bsbm_query01_count_us` is timing, `rsp_persistentdict_triples_per_s` is throughput; anchored-name fallback only when a row has no unit). The "never a code regression" claims are replaced with the honest implausible-not-impossible wording; exemption always emits `::warning` + the full table to the job summary.
4. **Fail-closed numerics + direction** — NaN / non-numeric / negative current-or-history and zero-boundary changes hard-fail per metric; `_per_s$` metrics gate larger-is-better (ratio = median/current), with the pre-existing customSmallerIsBetter inconsistency noted in code + PR body. **One evidence-based carve-out from the blanket `<=0` rule:** current 0 vs median 0 passes as a stable value, because four live series are all-zero by design (`geo_compliance_deficit`, `nlq_ask_repairs`, `rsp_*_w{2,3}_rows`, sub-resolution `sameas_size8_closure_s`) — a blanket `<=0 => fail` would permanently red main; stable zeros are explicitly listed, never silently skipped. Disappeared metrics: loud warn-list, hard fail only above 30% of gateable metrics.
5. **Windowing** — last ≤5 VALUES per metric across the FULL retained history (sparse metrics verified to collect all their points); ≥3 values required to gate, else listed "insufficient history — ungated"; suite name is an exact match only — absent suite warns loudly + exits 0, never falls back to another suite's series.
6. **Self-tests** — 32 checks covering every new behavior (NaN/zero/negative fail-closed, disappeared warn vs >30% fail, `_per_s` inversion both directions, sparse-metric windowing, insufficient-history ungated, uniformity rejection, deterministic-drift rejection, 4x cap, suite-mismatch exit 0, plus end-to-end `run_gate` tempdir fixtures and a parser-fixture argument check); the count is printed dynamically (`self-test OK (32 checks)`) so it can't go stale. Mutation check on the NEW uniformity-rejection assertion: /tmp snapshot → flipped expected exit → self-test went RED (1/32) → restored via `cp`, `cmp` byte-identical → green.
7. **release-plz containment** — the release-pr invocation moved out of the composite action into a wrapped `run:` step (`2>&1 | tee`), installing the SAME tool versions the pinned action installs (release-plz@0.3.160 + cargo-semver-checks@0.48) via the repo-standard SHA-pinned `taiki-e/install-action` (explicit `with: tool:` per sq-ur7o). Tolerate requires BOTH the captured output matching the exact signature (a `Failed to open PR` line with 403 on `api.github.com/repos/<repo>/pulls`) AND the side-effect-free probe returning 403; signature mismatch fails even if the probe 403s (rate-limit-heuristic case), probe non-403 fails even with the signature.
8. **Churn clean narrowed** — allow-list restore of exactly `bench/**/Cargo.lock` paths from `git diff --name-only`; any other dirty tracked path is printed and the step FAILS loudly (unknown churn = investigate, never silently discard).

**Verification:** `yaml.safe_load` + `actionlint` + `typos` green on all touched files; `bench_hardzone.py --self-test` green (32/32); `check-advisory-registry.py` all clear; production-shape simulation against the live benchmark-data history (latest point vs prior-19 history) exercises the real gate path end-to-end.

jeswr:

Round-2 cross-provider review: both findings in `scripts/bench_hardzone.py` fixed in 795237966.

**Finding 1 — fail-closed gap.** Current-value validation ran only *after* the no-history/`ungated` early exits, so a NaN/non-numeric/negative CURRENT value on a metric with no or young history exited 0; and history negativity was caught only via `med < 0`, letting a window like `[-1, 10, 10]` (positive median) pass. **Fix:** every current value is now validated (finite, numeric, non-negative) *before* any history-based early exit, and every value inside the median window is checked individually — any non-finite/negative window value is a per-metric fail-closed error with a clear message. The stable-zero carve-out is preserved exactly (all-zero window + current 0 still passes at ratio 1.0). Four new self-test checks cover the previously-leaking paths (no-history NaN, young-history junk, no-history negative, negative-behind-positive-median).

**Finding 2 — self-test precision.** The 4x-cap self-test used ratios 1.9/1.95/4.1 whose CV (~0.47) already fails uniformity condition (b), so it never isolated the cap (c) — removing the cap from the exemption predicate kept the test green. **Fix:** fixture is now uniform ratios 4.05/4.1/4.08/4.06 (4/4 metrics in the watch band, CV ~0.005, no deterministic metrics), so (a)/(b)/(d) all PASS and only the cap blocks — asserted via `exemption_reasons == [the (c) reason]` plus the hard-fail-no-exemption outcome. A companion control (same uniform shape at ~1.9x, under the cap) asserts the exemption DOES apply, proving the cap is the discriminating condition.

**Verification:** `--self-test` green (37 checks). Mutation check: weakening `UNIFORM_CAP_RATIO` 4.0 → 400.0 flips exactly the new cap-isolation check red (1/37 fail); snapshot-restored (`cmp`-verified identical), all 37 green again.

jeswr:
Cross-provider review complete: 3 rounds (GPT-5.6 sol). Round 1: 8 findings, fixed in f99b999f0. Round 2: 2 residuals, fixed in 795237966 (mutation-verified cap isolation, 37 self-test checks). Round 3: **VERDICT: pass** — arming auto-merge.

jeswr:

Updated the wiring pin test (renamed `test_bench_main_history_and_ratchet_exclude_schedule` → `test_bench_history_lane_scoping`) to pin the NEW invariants — unconditional churn-clean, auto-push still schedule/non-main-excluded, hard-zone `if:` equal to the auto-push expression exactly — all 66 wiring tests + `bench_hardzone.py --self-test` (37 checks) green; commit 72167cb63.

PR #3753, opened 2026-07-23
fix(bench): noise floor for µs-scale timing metrics in the hard zone — measured honest bands exceed 2x

jeswr:

Follow-up to #3751. The median-of-history hard zone (`scripts/bench_hardzone.py`) hard-fails any metric ≥ 2.0x its median-of-last-≤5 — but a fresh empirical replay (2026-07-23) against the published benchmark-data history shows µs-scale timing metrics have **honest run-to-run noise bands exceeding 2.0x**, so the gate reds on noise even with a clean history.

## Empirical evidence (2026-07-23 replay of the published history)

Within the honest pre-red-era oldest-10 points alone, max/min bands:

| metric | unit | honest band | note |
|---|---|---:|---|
| `sameas_size32_query_us` | us | **2.55x** | values ~4–9µs |
| `deeptax_d1000_query_us` | us | **3.25x** | ~4–13µs |
| `lubm_q14_count_us` | us | **2.45x** | timing metric (name trap) |
| `vectors_hnsw_recall_at10` | milli | **4.0x** | oscillates 1↔4 — degenerate resolution |

Live replays: an honest head fails `sameas_size32_query_us` at 2.14x (9 vs median 4.2); another honest entry fails `deeptax_d1000_query_us` at 2.00x (8 vs 4). Every observed false-fail has a median ≲15 units. Reproduced in this worktree: replaying the **current honest published head** against its prior history — old gate **exit 1** (`vectors_hnsw_recall_at10` 3.00x on its honest 1↔4 oscillation), new gate **exit 0** with the same row flagged watch-only.

## Change

- **`NOISE_FLOOR = 20.0`**: a NOISY-classified metric (existing unit-first classification reused) whose median-of-window is under the floor can never hard-fail the gate **alone**. Empirical bands documented at the constant (numbers live only in the script comment, per repo hygiene).
- **Deterministic metrics are NEVER floor-exempt** — a 2x on counts/bytes/gates is real at any scale (self-tested).
- **Unit `milli` reclassified as noisy**: the live history proves `vectors_hnsw_recall_at10` is a measurement (randomized HNSW build), not a deterministic output — required for it to land under the floor, and it stops the metric's constant motion from vetoing exemption condition (d) on every genuine environment shift. Its stable siblings (`vectors_diskann_recall_at10` ~34, `vectors_pq_recall_at10` ~22) sit **above** the floor and stay fully hard-gated.
- **Uniform-shift exemption computes breadth/uniformity/cap over gated (non-floor-exempt) rows only** — degenerate sub-floor ratios (the honest history reaches 4.0x on pure noise) would distort all three: breadth dilution, cv blow-up, and a false ≥4x cap veto.

## Why floor-exempt ≠ ungated

A sub-floor metric still: (1) prints in the watch table flagged **"under noise floor — watch only"** (stdout + `GITHUB_STEP_SUMMARY`); (2) emits a `::warning` when ≥ 2.0x median; (3) gets the Store step's soft comparison comment, which fires independently of this gate. Only the *lone hard red* is removed — a genuine broad regression that also moves above-floor metrics still fails, and the ~230 above-floor metrics (incl. all deterministic families) are gated exactly as before.

## Verification

- `--self-test`: **43 checks green** — new: sub-floor timing at 3x ⇒ watch-only + flagged; same 3x shape above the floor ⇒ hard fail; sub-floor deterministic at 2x ⇒ hard fail; exemption breadth/cap/uniformity exclude floor-exempt rows (4/4-gated-exempt vs 4/9-would-fail discriminating fixture + 4.0x sub-floor spike no-veto). Existing hard-zone fixtures scaled ×10 above the floor so they keep testing the hard zone proper.
- **Mutation check** (snapshot to /tmp → flip `NOISE_FLOOR` to 0 → red → cmp-restore → green): exit 1 with exactly the 4 floor-discriminator checks red (`4/43`), above-floor twin stays green — the floor constant is load-bearing.
- `scripts/tests/test_ci_select_wiring.py`: 66 tests OK (untouched, confirmed).
- Live replay of the honest published head: old gate exit 1, new gate exit 0 (watch-flagged), as above.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -502,6 +502,13 @@ jobs:
       # and already strictly gated by perf-gate.py above, and the nightly `schedule` backstop is a
       # pure verification run — so PR + schedule behavior is unchanged by this step.
       # The script self-tests first (repo convention: hermetic --self-test over inline fixtures).
+      #
+      # --report-out ([FABLE-5] round-2 review): floor-exempt (sub-noise-floor us/milli) metrics
+      # never fail this gate, and once a few regressed points are accepted the median REBASES —
+      # so any floor-exempt hit at >= 2.0x is written to bench-hardzone-report.json, which the
+      # "Soft-zone triage" step below merges into the rolling deduped `bench-flake` issue
+      # (tagged "floor-exempt hard-band"): the durable trail that survives the rebase. The file
+      # is written on pass AND fail; triage runs `if: always()` and tolerates its absence.
       - name: Hard zone — median-of-history regression gate (main pushes only)
         if: github.event_name != 'schedule' && github.ref == 'refs/heads/main'
         run: |
@@ -510,7 +517,8 @@ jobs:
           python3 scripts/bench_hardzone.py \
             --results bench-results.json \
             --prev-data prev-data.js \
-            --suite 'sparq engine'
+            --suite 'sparq engine' \
+            --report-out bench-hardzone-report.json
       # [OPUS-4.8] sq-owci — clean the working tree before github-action-benchmark switches branches.
       # ROOT CAUSE: the "Run benchmarks" step (scripts/ci-bench.sh) does `cd bench/zk && cargo bench`
       # and `cd bench/zk-trace && cargo bench` (two STANDALONE cargo projects, run WITHOUT --locked),
@@ -644,6 +652,11 @@ jobs:
       # hard gate: the median-of-history gate above ([FABLE-5] scripts/bench_hardzone.py) owns the
       # hard-zone failure; this step
       # only handles the soft (possible-flakiness) band and always exits 0 (its filing is best-effort).
+      # [FABLE-5] --hardzone-report: additionally merges the median gate's floor-exempt hard-band
+      # rows (bench-hardzone-report.json, written by the gate's --report-out on main pushes;
+      # absent on PR/schedule = no-op) into the same deduped issue, tagged "floor-exempt
+      # hard-band" — the gate never reds on those watch-only rows and their median rebases, so
+      # this issue trail is their only durable record.
       # Canonical repo only — fork PRs get a read-only token, so no issue can be filed there.
       - name: Soft-zone triage (file a deduped bench-flake issue; no @mentions)
         if: always() && github.repository == 'sparq-org/sparq'
@@ -666,6 +679,7 @@ jobs:
           # scripts/bench-thresholds.py from the benchmark-data history).
           python3 scripts/bench-triage.py --mode soft \
             --comparison bench-comparison.json \
+            --hardzone-report bench-hardzone-report.json \
             --soft 175 --hard 200 \
             --run-url "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" \
             || echo "::warning::soft-zone triage filer failed (non-fatal)"
--- a/scripts/bench-triage.py
+++ b/scripts/bench-triage.py
@@ -283,18 +283,33 @@ def load_hardzone_soft_rows(path: str | None) -> list[dict]:
     except (json.JSONDecodeError, ValueError, OSError) as e:
         log(f"warning: could not parse hardzone report {path} ({e}) — skipping its rows")
         return []
+    if not isinstance(data, dict):
+        log(f"warning: hardzone report {path} is not a JSON object — skipping its rows")
+        return []
+    fe = data.get("floor_exempt_hard", [])
+    if not isinstance(fe, list):
+        # {"floor_exempt_hard": null} / a number / a string must degrade to zero merged
+        # rows with a warning — never TypeError out of ordinary soft filing.
+        log(f"warning: hardzone report {path}: floor_exempt_hard is "
+            f"{type(fe).__name__}, expected a list — skipping its rows")
+        return []
     out = []
-    for r in data.get("floor_exempt_hard", []) if isinstance(data, dict) else []:
+    dropped = 0
+    for r in fe:
         if not isinstance(r, dict):
+            dropped += 1
             continue
         name, cur, prev, ratio = (r.get("name"), r.get("current"), r.get("previous"),
                                   r.get("ratio_pct"))
         if (name is None or not isinstance(cur, (int, float))
                 or not isinstance(prev, (int, float)) or not isinstance(ratio, (int, float))):
+            dropped += 1
             continue
         out.append({"name": str(name), "current": float(cur), "previous": float(prev),
                     "ratio_pct": float(ratio), "unit": r.get("unit", ""),
                     "note": r.get("note") or "floor-exempt hard-band"})
+    if dropped:
+        log(f"warning: dropped {dropped} malformed floor_exempt_hard row(s) from {path}")
     return out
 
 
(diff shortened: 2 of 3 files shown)


jeswr:
> 🤖 **SPARQ agent** — autonomous orchestration.

Fix round for the 3 cross-provider review findings — all addressed in 9144b5779.

**1. [P1] Unit-blind noise floor — FIXED.** `floor_exempt = (not det) and med < 20` is replaced by an explicit `FLOOR_EXEMPT_UNITS` allow-list of (unit, floor) pairs: `{"us": 20.0, "milli": 20.0}`, each floor justified per-unit in the constant's comment by the measured honest bands (µs timings 2.45–3.25x under ~15 units; the degenerate `milli` HNSW-recall 1↔4 oscillation). Every other unit — `s` (the nine load_s/text_build_s/rdfs_infer_s series at ~0.4–15 s), `ns_per_byte`, `ratio`, `*_per_s`, deterministic units, and anything unknown/absent — is never floor-exempt: unknown units fail toward gated. New self-tests pin it: an `s`-unit metric with median 0.4 at 10x hard-fails; `ns_per_byte`/`ratio`/unknown/empty units at 3x stay gated; `milli` stays exempt. **Mutation-checked:** adding `"s"` to the allow-list flips exactly the new s-unit check red (1/54), snapshot-restored byte-identical, green.

**2. [P1] No durable trail for floor-exempt hard-band hits — FIXED.** Triage never read our output (it rebuilds its own single-previous-point comparison, and its own `part["hard"]` is ignored in `--mode soft`), so per the brief's fallback I extended both minimally: `bench_hardzone.py` now emits `--report-out bench-hardzone-report.json` carrying `floor_exempt_hard` rows in the exact triage comparison-row shape (`previous` = median-of-history, `note` = "floor-exempt hard-band (>= 2x median-of-history; watch-only for the gate)"), written on pass **and** fail; `bench-triage.py --mode soft` gained `--hardzone-report`, which fail-softly loads those rows and merges them into `part["soft"]` (tagged row wins on a name collision), so every hit lands in the rolling deduped `bench-flake` issue with the tag rendered in the table plus an explainer paragraph. It never weakens a gate — these rows were watch-only already. bench.yml wires both flags; step names/order/guards unchanged, so no wiring-pin updates were needed (`test_ci_select_wiring.py` 66/66 green). End-to-end smoke verified: a 3.0x sub-floor hit exits 0, the report row flows through merge into the flake body.

**3. [P2] Total-row numbers in the uniform-shift annotation — FIXED.** The `::warning` now reports `report["gated_watch"]` of `report["gated_compared"]` ("N of M GATED metrics (floor-exempt excluded)") and "no GATED metric reached 4x". New self-test captures `emit_report` stdout on the mixed fixture (where floor-exempt `tiny0_us` honestly sits at exactly 4.0x): asserts "4 of 4 GATED metrics" + "no GATED metric reached" present and "5 of 9" / "no metric reached" absent.

**Verification:** `bench_hardzone.py --self-test` OK (54 checks, was 41); `bench-triage.py --self-test` OK; `test_ci_select_wiring.py` OK (66 tests); bench.yml YAML valid; mutation check red→restore→green as above.

SHA: 9144b5779

jeswr:
> 🤖 **SPARQ agent** — autonomous orchestration.

Round-3 review findings addressed in 752fcc509502bc98bb01a87c52c489228831a545:

1. **`--hardzone-report` loader not fully fail-soft** — FIXED. `floor_exempt_hard: null` / a number / a non-object top level now log a warning and proceed with zero merged rows (no TypeError abort of ordinary soft filing); malformed rows are dropped with a counted warning. Self-test covers null, number, and non-object payloads.
2. **Unescaped Markdown interpolation in issue bodies** — FIXED. New `_md_cell` (plain cells: strip newlines, escape `\` `|` and backticks, split `@` with an empty HTML comment) and `_md_code_cell` (code-span cells: strip newlines, GFM `\|` escape, backticks replaced — backslash escapes don't work inside a span; the span itself neutralizes mentions). Applied to the flake-body table and, since it's the identical pattern, to the regression-body table + ranked-suspect subjects. New self-test renders a pipe/backtick/newline/`@` metric and asserts escaped output + no naked mention. **Mutation-checked**: snapshot → escape-off flip → self-test red (`AssertionError` on the forged row) → cmp-verified restore → green.
3. **stdout/step-summary still claimed "no >= 4x outlier"** — FIXED. `render_report`'s exemption prose now reads "no GATED metric reached 4x median" (floor-exempt excluded), matching the `::warning`; the annotation self-test now captures both sinks (stdout + a real `GITHUB_STEP_SUMMARY` file) and asserts the gated wording in each, old wording in neither.
4. **Preamble contradicted merged floor-exempt rows** — FIXED. Tagged bodies open with two-class wording (soft-band vs floor-exempt hard-band, at/above the hard ratio vs median-of-history); untagged bodies keep the original soft-zone-only prose. Both asserted in the self-test.

`bench_hardzone.py --self-test`: OK (56 checks). `bench-triage.py --self-test`: OK.

jeswr:
> 🤖 **SPARQ agent** — autonomous orchestration.

Round-4 close-out landed in `9f6ae817e` — all three residual findings fixed with defensive implementations, both `--self-test`s green, html-escape mutation-checked.

**Finding 1 — numeric validation (fixed).** New `_finite_num()` is the single numeric admission gate in `bench-triage.py`: accepts only `isinstance(v, (int, float)) and not isinstance(v, bool)`, wraps the `float()` conversion in `try/except (OverflowError, ValueError)` (a JSON `10**400` int is dropped, never a crash), and requires `math.isfinite`. Routed through **every** numeric field consumed: `load_hardzone_soft_rows` (`current`/`previous`/`ratio_pct`, dropped rows counted + warned), `parse_comparison` (`current`/`previous` — the old bare `float()` crashed on string values), `build_comparison` (results values + data.js baseline values), `data.js` entry dates (`_entry_date`, junk dates sort as 0 instead of `TypeError`), and `_to_ratio_pct` now rejects a ratio that overflows to inf. Same guard applied to `bench_hardzone._is_num`, which previously **crashed** on `math.isfinite(10**400)` — now it reports False and the gate fails closed with the per-metric record. Self-tests cover bool / NaN / Infinity / `10**400` in every path.

**Finding 2 — one defense-in-depth sanitizer (fixed).** `_sanitize(text, code=...)` replaces the per-construct escaping: layer 1 html-escapes `&`, `<`, `>` first and unconditionally (killing raw-HTML breakouts — GitHub renders raw HTML in issue bodies — and defusing entity-smuggled `&#64;user`), layer 2 applies the existing Markdown escapes per cell type (newline-strip, pipe/backslash/backtick, `@`-defusal via HTML comment inserted post-escape). Audited every f-string sink in both issue-body builders and routed all of them: metric names, units, notes, suite names, run URLs, baseline/head SHAs, commit subjects, overlap crates; link destinations (`repo`, commit SHAs inside `](...)`) go through a new `_md_url` percent-encoder. Mandated self-test in place: a value carrying `</td></tr></table><h1>forged</h1>` renders with `&lt;/td&gt;…&lt;h1&gt;forged&lt;/h1&gt;` in the final body (plain cells and code spans alike), plus a forged commit-subject test for the nightly body. `bench_hardzone.py` builds no issue/comment bodies (stdout + `::warning`/`::error` + step summary + report JSON only; its rows are sanitized at render time by triage), so finding 2's sink audit is complete on the triage side.

**Finding 3 — class-accurate preamble (fixed).** The class sentence is now derived three-way from the row classes actually present: soft-only (unchanged soft-zone framing), mixed (the round-3 "two classes" prose), and the new **floor-exempt-only** branch ("Every row this run is a tagged floor-exempt hard-band hit … there are no soft-band rows this run"). The floor-exempt-only case is self-tested: asserts the new sentence, and that neither "two classes" nor the soft-zone framing appears.

**Verification.**
- `scripts/bench-triage.py --self-test` → `self-test OK` (counted-drop warnings observed: 5 malformed `floor_exempt_hard` rows, 4 malformed comparison rows, 2 malformed result rows).
- `scripts/bench_hardzone.py --self-test` → `self-test OK (59 checks)`.
- Mutation check: snapshot → html-escape line disabled → `--self-test` exits 1 red on exactly the forged-table assertion (`assert "<h1>" not in hbody`) → `cmp`-verified restore → green.

SHA: `9f6ae817eb93a5e03a5067f338b8938454cb4021`

jeswr:
Cross-provider review complete: 4 rounds (GPT-5.6 sol). R1: 3 findings (unit-blind floor, no durable floor-exempt trail, gated-wording) → 9144b5779. R2: 4 residuals → 752fcc509. R3: 3 residuals (overflow, raw-HTML sink, class wording) → 9f6ae817. R4: **VERDICT: pass** — arming.

PR #3764, opened 2026-07-24
feat(ci): nextest CI retry profile + merge-queue dequeued feedback loop (#556) + transient-tolerant sweeps (#3759)

jeswr:
> 🤖 SPARQ agent — autonomous orchestration for sparq.

Implements the maintainer-approved OSS adoptions from registry jeswr/agent-account-registry#563 — **item 2** (nextest CI retry profile), **item 3** (merge-queue `dequeued` feedback, the durable fix for registry#556), and the sparq side of **item 4** (transient-tolerant retry hygiene on the arm sweeps, fixes #3759).

## Adoption 2 — nextest `[profile.ci]` retries (flaky reds stop burning the gate)

**How tests actually run in CI (verified per leg):**

| Leg | Runner | This PR |
|---|---|---|
| ci.yml `test` (5 load-aware shards, archive-based) | `cargo nextest run --archive-file …` | **converted**: `--profile ci` (retries moved off the inline `--retries 2` flag into `.config/nextest.toml`, count unchanged) |
| miri.yml (24 shards) | `cargo miri nextest run` | untouched — auto-selects the existing `[profile.default-miri]` |
| ci.yml doctests | `cargo test --doc` | not convertible (nextest does not run doctests) |
| ci.yml conformance/ratchet suites (sparq-conformance W3C, shacl, geo, solid, policy, text, rsp, jsonld, service, http-protocol, sd-gsp, d-entail, rif, el, ql, dl) | plain `cargo test … -- --nocapture` | unconverted — these jobs do not install nextest, and their ratchets parse `--nocapture` output |
| feature-matrix.yml legs, gui.yml, zk-toolchain.yml, shacl-diff-fuzz.yml, metamorph.yml, differential-update.yml | plain `cargo test` | unconverted — no nextest in their toolchain setup (zk-toolchain deliberately so, per its in-file rationale; gui documents a retries=0 deterministic-lane rationale) |

Per the approved scope, only legs that already had nextest in their toolchain were converted; the rest are enumerated in follow-up bead `sq-yhcge` (per-leg evaluation: flake-proneness vs the cost of a toolchain step + ratchet-output compatibility).

**Incident class:** flaky tests redding the single required `gate` check burn whole gate evaluations and trigger review-fix ci-flavour re-dispatches for PRs whose diff changed nothing relevant.

[...]

Issue #3759, opened 2026-07-23
rearm-sweeper: transient GitHub API 5xx reds the main gate — needs bounded retry + infra-noise tolerance


jeswr:
> 🤖 SPARQ agent — autonomous orchestration for sparq.

Live: run 30020248573 (2026-07-23T15:22Z) — `arm reviewed PRs` died on `GhError: gh pr list --repo failed: HTTP 504` (api.github.com/graphql timeout), which failed ci-summary on main (run 30019975813) because the job name carries no advisory token. A transient platform 5xx must not red main.

Fix: (a) bounded retry with backoff (3 attempts, 10s/30s) around the gh calls in the sweeper's arm job; (b) if all retries exhaust on a 5xx, exit with a ::warning + neutral/skip semantics rather than failure (the sweeper is a periodic idempotent sweep — a missed cycle is harmless, the next cron covers it); (c) any non-5xx error still fails loudly. Same pattern for the sweeper's other API-calling jobs.

jeswr:
> 🤖 SPARQ agent — autonomous orchestration for sparq.

PR #3764 implements this (the sparq side of registry jeswr/agent-account-registry#563 item 4, maintainer-approved). Mechanism, matching the fix spec above:

(a) **Bounded retry with backoff** — new `scripts/gh_retry.py` (stdlib-only tenacity mirror): 3 attempts, 10s/30s backoff, matched against a bounded transient-marker list (HTTP 502/503/504 + network-transport failures). Both `scripts/auto-arm.py` (the `arm reviewed PRs` job that died on the HTTP 504 at 15:22, and the 22:42/14:29 twins) and `scripts/rearm-sweeper.py` route their idempotent gh READS (`pr list`, `pr view`, the live-state GraphQL query) through it.

(b) **Exhausted 5xx ⇒ `::warning` + exit 0** — exhaustion raises a distinct `GhTransientExhausted`, which each sweep's entrypoint converts into a `::warning` annotation and a success exit: the sweep is periodic + idempotent, a missed cycle is harmless, the next cron covers it — and a platform blip no longer reds main's gate via ci-summary.

(c) **Non-5xx still fails loudly** — anything outside the transient list raises `GhFatalError` immediately (no retry) and keeps the existing loud failure paths.

[...]

Issue #3012, opened 2026-07-17
sq-6xasp.7: (Follow-up) Persistent Store for wasm: a JS-fs / IndexedDB backend behind the existing Store trait


jeswr:
A JS-fs / IndexedDB-backed Store implementation behind the existing swappable async Store trait (store/mod.rs:125), so pod data survives a listener restart. The in-memory store stays the default; persistence is opt-in. Model tier: gpt. Design: research/wasm-solid-server-npm-design.md phase 7.

Parent: #2558

PR #3766, opened 2026-07-25
fix(ci): re-arm sweeper fails loudly once on missing arm capability + per-PR failures no longer abort the sweep (#3760)

### LINKED ISSUE #3760: rearm-sweeper token cannot enablePullRequestAutoMerge — 'Resource not accessible by integration' on ready+review:pass PRs

Live: sweeper run at 18:21Z failed arming #3454 (non-draft, CLEAN, review:pass) with `GraphQL: Resource not accessible by integration (enablePullRequestAutoMerge)`; auto-arm red with the same class. The orchestrator armed it manually with a user token as a stopgap.

Cause candidates: the sweeper job's token permissions block auto-merge enablement (needs `contents: write` + `pull-requests: write` on the job, and if it uses the App installation token, the App may lack the required repo permission), or the ruleset restricts the integration. Fix: audit the sweeper/auto-arm workflow token permissions and switch the arm call to a token that can latch auto-merge; add a self-test/probe that fails LOUDLY at startup if the token cannot arm (probe a dry-run or check token scopes) instead of failing per-PR mid-sweep; per-PR arm errors should not fail the whole sweep run (collect + continue + summary).

jeswr:

Closes #3760.

## Diagnosis — and the honest answer to "is this maintainer-only?": **no, it is not**

**Which token made the failing call.** The `sweep` job's `GH_TOKEN` is
`${{ steps.app-token.outputs.token || github.token }}`. `ORCHESTRATOR_APP_ID` was
**empty** in the failing run (visible in its step env dump), so the App mint step was
skipped and the arm ran on the plain **GITHUB_TOKEN**.

**Why it could not call `enablePullRequestAutoMerge`.** Enabling auto-merge is a
repository **WRITE** operation, and `rearm-sweeper.yml` declared `permissions: contents: read`.

| | token | `permissions` | outcome |
|---|---|---|---|
| re-arm sweeper run [30033315483](https://github.com/sparq-org/sparq/actions/runs/30033315483) — 2026-07-23 **18:21Z** | plain `GITHUB_TOKEN` | `contents: **read**` + `pull-requests: write` | ❌ `PR #3454: SKIP — re-arm failed (gh pr merge 3454 failed: GraphQL: Resource not accessible by integration (enablePullRequestAutoMerge))` — PR #3454 was OPEN, non-draft, CLEAN, `review:pass`; the orchestrator armed it by hand with a user token |
| auto-arm run [30027010495](https://github.com/sparq-org/sparq/actions/runs/30027010495) — 2026-07-23 **16:52Z**, 90 min earlier | plain `GITHUB_TOKEN` (auto-arm has no App path at all) | `contents: **write**` + `pull-requests: write` | ✅ `armed head 65eda1b325f6`, `armed head e8afddf96f3b` |

Same repository, same hour, same token identity, opposite results. The only difference is
the `permissions` block, so `contents: write` is the fix.

**Ruled out, with evidence — none of these needs a maintainer:**

* `allow_auto_merge` is **already `true`** on the repository (`GET /repos/sparq-org/sparq`).
* The `main` ruleset has **no integration restriction** — its rules are
  `deletion, non_fast_forward, pull_request, required_status_checks, code_quality, copilot_code_review, merge_queue`, with a single `RepositoryRole` bypass actor.
* The App path is genuinely unconfigured (`ORCHESTRATOR_APP_ID` / `ORCHESTRATOR_APP_PRIVATE_KEY` are unset — the same reason the batch-merge omnibus path degrades to hygiene-only), **but it is not needed**: a `GITHUB_TOKEN` with `contents: write` arms fine, as auto-arm proves every day. The App token remains preferred when the secrets do appear.

The auto-arm reds on 2026-07-23 were a **different** class, and I checked rather than
assumed: all of them were `gh pr list --repo failed: HTTP 502/504` on
`api.github.com/graphql`, not permission denials.

## The fix

1. **`rearm-sweeper.yml`: `contents: read` → `contents: write`.** The one-line root-cause fix, with the live evidence in the file so it is not "cleaned up" later.
2. **Startup arm-capability probe, failing loud exactly once.** Both policies gain `--probe-arm-capability`, wired as its own step *before* arming. A provable cannot-arm emits **one** `::error` naming every flippable lever — `contents: write` + `pull-requests: write`, repository *Settings → General → Pull Requests → Allow auto-merge*, a future ruleset integration restriction, the `ORCHESTRATOR_APP_ID` / `ORCHESTRATOR_APP_PRIVATE_KEY` secrets, and (auto-arm only) the fork-PR token downgrade — then exits non-zero **once, before any PR is touched**. Signals it cannot read decisively stay `inconclusive` and never red the run, so a transient 502 on the probe cannot invent an outage.
3. **A per-PR arm failure no longer aborts the sweep.** Failures are collected, the sweep continues to every remaining candidate, each failure is summarised at the end, and the run exits non-zero. This was a real second defect in **auto-arm**: `process()` re-raised the mutation error out of `run()`, so a single failing PR stopped every later PR from being armed at all.
4. **A denial mid-sweep is a capability verdict, not a per-PR one.** The first `Resource not accessible by integration`-class failure emits the single `::error`, stops arming (the remaining candidates log `arm capability lost this run`), and exits 1 — never one error per PR, never ten identical failures per run.

## Probe design — what is *not* a capability signal (measured, and recorded in the source)

I tried the obvious probes first and they are all unusable; the findings are comments in
`scripts/rearm-sweeper.py` so nobody re-derives them:

* `Repository.viewerPermission` — GitHub documents it as *"will return null if
  authenticated as a GitHub App"*, and the Actions `GITHUB_TOKEN` **is** an App
  installation token. Diagnostics only, never a gate.
* `PullRequest.viewerCanEnableAutoMerge` — measured **false** under an **ADMIN user
  token** for already-armed (#2521), queued (#3764), draft and merged PRs. It conflates PR
  state with permission; gating on it would refuse legitimate arms.
* The mutation against a bogus node id — an *authorized* token gets `NOT_FOUND`
  ("Could not resolve to a node with the global id of ..."), so the denial class cannot be
  distinguished without actually being denied.

What is left, and what the probe uses: `Repository.autoMergeAllowed` (decisive when
false), a denied repository read (decisive), and the mutation's own denial as the backstop.

## Tests

* `scripts/rearm-sweeper.py --self-test` / `scripts/auto-arm.py --self-test` — probe verdicts (**can-arm / cannot-arm / inconclusive**, both cannot-arm routes: setting-off and token-denied), probe-runs-before-any-arm, **one `::error` per run**, **per-PR failure collection + continue** (PR #452 is still armed after #451 fails), mid-sweep denial stops the sweep after one error, and **end-of-run exit semantics** (`arm_failures` ⇒ exit 1; an arm failure can never exit 0). Both PASS.
* `scripts/tests/test_arm_capability_wiring.py` (new, 15 tests) — cross-file pins: `contents: write` + `pull-requests: write` on **both** workflows and no job-level narrowing, the probe step present and *before* the arm step with a **byte-identical `GH_TOKEN` expression** (a probe on a different token than the sweep uses is worse than no probe), the live #3760 error text classifying as a denial while races/502s/CAS rejections do **not**, remediation naming every lever, and no drift between the two deliberately duplicated copies (each workflow sparse-checks out only its own script, so a shared import is impossible).
* **Mutation-verified**, not assumed: six regressions each red the pins — `contents: write`→`read` on either workflow, probe step removed, probe token diverging from the arm token, a denial marker dropped, and the remediation losing `contents: write`.
* **Live**, read-only: `--probe-arm-capability` against `sparq-org/sparq` returns `can-arm — autoMergeAllowed=true`, exit 0; a full sweep with arming stubbed out ran end-to-end (`candidates=1 ... capability=can-arm`, exit 0) — **nothing was armed by this work**.
* `check-advisory-registry.py`, `check-install-action-tool.py`, YAML parse of all three touched workflows: green.

The arm self-tests previously only ran *inside the scheduled arm workflows* — i.e. never on
the PR that broke them. All three suites now gate in `docs-quality` `quick-gates`.

## Residual honesty

The write-scope half of the capability question cannot be proven read-only from inside the
job: no GitHub API exposes an installation token's `contents` scope, and
`viewerPermission` is null for Apps. The startup probe is therefore decisive about the
repository setting and about a denied read, and the **first mutation** is what confirms
write. That is why the mid-sweep denial path exists and why it fails loud once instead of
degrading to a per-PR skip. Everything a human might need to flip is named in that one
error; today, nothing needs flipping.

## Rebased onto #3764 (#3759 transient-tolerant sweeps) — one deliberate test change, called out

#3764 landed the `gh_retry` read layer while this was in flight, so both scripts were
re-integrated on top of it rather than merged blindly:

* The capability probe is itself an **idempotent READ**, so it routes through `gh_read`
  (the bounded retrying runner — `gh_retry.assert_read_only` accepts it, the query carries
  no mutation) and runs before the enumeration. The `#3759` read-routing self-tests now pin
  that ordering explicitly.
* `ARM_DENIAL_MARKERS` is deliberately **disjoint from #3759's transient set**, and the
  self-tests assert both directions: a denial must never be swallowed as `::warning` +
  exit 0, and a 502/504/CAS must never be misread as a denial. A `cannot-arm` verdict reds
  **both** `--mode sweep` and `--mode event`.
* An exhausted transient on the probe read is `inconclusive`, never `cannot-arm`.
* **The one behavioural change to an existing #3759 test, stated plainly:** finding 5's
  invariant is *"a failed arm mutation must never read as success, even when the follow-up
  race diagnostic exhausts its retries."* It used to be proven by the mutation error
  **propagating out of `run_sweep`** — which is precisely the #3760 defect (it aborted
  every later candidate). The invariant is unchanged and still asserted; the mechanism is
  now `ArmOutcome.exit_code == 1` with the failure summarised, plus explicit assertions
  that the run does not exit 0 and that the CAS text still surfaces. Nothing about #3759
  is weakened — the failure is now *both* visible and non-aborting.

### DIFF
--- a/.github/workflows/auto-arm.yml
+++ b/.github/workflows/auto-arm.yml
@@ -9,7 +9,12 @@ on:
     - cron: "4,14,24,34,44,54 * * * *"
   workflow_dispatch:
 
-# Enabling auto-merge needs only repository contents + pull-request write access.
+# Enabling auto-merge needs repository contents + pull-request WRITE access. Both are
+# load-bearing: enablePullRequestAutoMerge is a repository write operation, so demoting
+# `contents` to `read` makes every arm fail with "Resource not accessible by integration
+# (enablePullRequestAutoMerge)" — the #3760 outage, which the re-arm sweeper hit with
+# exactly that misconfiguration while THIS workflow armed fine (runs 30033315483 vs
+# 30027010495). Pinned by scripts/tests/test_arm_capability_wiring.py.
 permissions:
   contents: write
   pull-requests: write
@@ -47,6 +52,23 @@ jobs:
           python3 scripts/gh_retry.py --self-test
           python3 scripts/auto-arm.py --self-test
 
+      # [OPUS-5] #3760 — fail loud ONCE at the start when the token provably cannot arm,
+      # instead of a per-PR failure on every label event. Must use the same GH_TOKEN
+      # expression as the arm step below (pinned by the wiring test).
+      - name: Probe arm capability (one loud error, never per-PR)
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
+        run: >-
+          python3 scripts/auto-arm.py
+          --repo "$REPO"
+          --default-branch "$DEFAULT_BRANCH"
+          --probe-arm-capability
+
+      # A per-PR arm failure no longer re-raises out of the sweep (pre-#3760 it did, so a
+      # single failing PR stopped every later candidate from ever being armed). Failures
+      # are collected, summarised, and the run exits non-zero.
       - name: Arm eligible PRs
         env:
           GH_TOKEN: ${{ github.token }}
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -719,6 +719,11 @@ jobs:
       #   * rearm-sweeper.py --self-test / auto-arm.py --self-test — probe verdicts
       #     (can-arm / cannot-arm / inconclusive), per-PR failure collection + continue,
       #     one-::error-per-run, and the exit semantics (an arm failure never exits 0).
+      #     [OPUS-5] #3766 adds the STICKY-FAILURE PRECEDENCE sequences —
+      #     `collected-failure > transient-exhaustion > clean` — because collecting the
+      #     failures was only half the job: the collected state lived in a LOCAL, so a
+      #     LATER candidate's exhausted transient unwound past the accumulated-failure
+      #     check and reported ::warning + exit 0, discarding a real arm failure.
       #   * test_arm_capability_wiring.py — cross-file pins: `contents: write` +
       #     `pull-requests: write` on both workflows (a downgrade reproduces #3760 and is
       #     invisible in a python-only diff), the probe step present and BEFORE the arm step
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -709,6 +709,28 @@ jobs:
         run: python3 scripts/check-fast-fix-ring-guard.py --self-test
       - name: "Enforce the fast-fix ring fail-closed cross-repo guard (#3475) — GATING"
         run: python3 scripts/check-fast-fix-ring-guard.py
+      # [OPUS-5] sparq-org/sparq#3760 — THE ARM PATH. The re-arm sweeper spent days unable
+      # to enable auto-merge at all ("Resource not accessible by integration
+      # (enablePullRequestAutoMerge)") because its job declared `contents: read` while the
+      # operation needs repository WRITE. Both arm policies now (a) probe their token's arm
+      # capability at job start and fail with ONE actionable ::error, and (b) collect per-PR
+      # arm failures instead of aborting the sweep. Their own self-tests only ran inside the
+      # scheduled arm workflows, i.e. never on the PR that broke them — so they gate HERE:
+      #   * rearm-sweeper.py --self-test / auto-arm.py --self-test — probe verdicts
+      #     (can-arm / cannot-arm / inconclusive), per-PR failure collection + continue,
+      #     one-::error-per-run, and the exit semantics (an arm failure never exits 0).
+      #   * test_arm_capability_wiring.py — cross-file pins: `contents: write` +
+      #     `pull-requests: write` on both workflows (a downgrade reproduces #3760 and is
+      #     invisible in a python-only diff), the probe step present and BEFORE the arm step
+      #     with a byte-identical GH_TOKEN expression, the remediation naming every
+      #     flippable lever, and no drift between the two deliberately duplicated copies.
+      # The wiring test needs PyYAML — installed once in the shared setup above.
+      - name: Self-test the re-arm sweeper policy (arm capability + failure isolation — #3760)
+        run: python3 scripts/rearm-sweeper.py --self-test
+      - name: Self-test the auto-arm policy (arm capability + failure isolation — #3760)
+        run: python3 scripts/auto-arm.py --self-test
+      - name: "Arm-path wiring inspection (contents: write + probe-before-arm — #3760)"
+        run: python3 scripts/tests/test_arm_capability_wiring.py
 
   # ----------------------------- ADVISORY -----------------------------
   # [FABLE-5] sq-6vshe.20: the three former advisory jobs as step-groups on ONE
--- a/.github/workflows/rearm-sweeper.yml
+++ b/.github/workflows/rearm-sweeper.yml
@@ -12,13 +12,24 @@ on:
     - cron: "8,18,28,38,48,58 * * * *"
   workflow_dispatch:
 
-# Arming auto-merge needs push authority AND the merge must fire downstream push
-# workflows (flow-on autoclose) — a GITHUB_TOKEN-enabled arm risks the recursion
-# guard suppressing them. Arm with the sparq-orchestrator App token (the same
-# identity batch-merge.yml uses); fall back to the workflow token only when the
-# App is not configured, in which case failed arms surface loudly as errors.
+# [OPUS-5] #3760 — `contents: write` IS THE FIX, and it is load-bearing.
+#
+# Enabling auto-merge is a repository WRITE operation. With `contents: read` this job's
+# token is denied with
+#     GraphQL: Resource not accessible by integration (enablePullRequestAutoMerge)
+# which is exactly what run 30033315483 (2026-07-23T18:21Z) hit on PR #3454 — OPEN,
+# non-draft, CLEAN, review:pass. auto-arm.yml armed two PRs 90 minutes earlier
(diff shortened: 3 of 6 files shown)


jeswr:

**Fix round for the BLOCKING cross-provider finding: the false pass is closed in both sweeps.** `47e04aea5c9266a659f9f2ec8d60b8b2ace9d9b7`

## Disposition — the finding was real, and I reproduced it before changing anything

Both reproductions are exact. On the pre-fix branch:

| script | sequence | pre-fix result |
|---|---|---|
| `scripts/auto-arm.py` | PR #451 arm failed (collected), PR #452's pre-arm `refresh` exhausted its bounded retries | `::warning` + **exit 0** |
| `scripts/rearm-sweeper.py` | PR #3011 arm failed (collected), PR #3012's live-state read exhausted | `GhTransientExhausted` escaped `run()` → `main()`'s lenient handler → `::warning` + **exit 0** |

The root cause was the same in both: #3760 collected the failures, but the collected state
lived in a **local of `run()`**, so an exception raised while processing a **later**
candidate unwound *past* the accumulated-failure check. The failure #451/#3011 had already
earned was discarded by something that happened to a different PR.

## The precedence rule — now explicit, documented in-source in both files

```
collected-failure > transient-exhaustion > clean
```

1. **The outcome is published on the runner** (`AutoArmer.outcome` / `RearmSweeper.outcome`)
   before any work begins, so every collected failure survives an escaping exception.
2. **A per-candidate exhaustion is caught inside the loop** and recorded as its own
   warning-class per-candidate outcome (`transient_exhaustions`); the sweep continues to
   every remaining candidate. It can never escape and change an earlier verdict.
3. **The exit is computed at the END from that final state**, by the new
   `arm_exit_code()` / `sweep_exit()` seams — never by an exception short-circuit. A
   collected failure returns 1 in *both* modes and logs why it outranks the transient. The
   entrypoint's transient handler reads the outcome back **off the runner**, so even a read
   added *outside* the guarded loop cannot report a false success (defence in depth).
4. **Lenient policy kept where it IS correct:** a sweep whose ONLY problem is transient
   exhaustion still exits **0** with exactly one `::warning` — a missed cycle is harmless
   and the cron re-covers it. Event-driven arms stay **non-zero** on any failure.

**Same shape audited elsewhere in both files** (point 4 of the ask): the enumeration read
is now *recorded*, not raised; and an exhaustion on the **one-shot arm mutation itself** is
**fail-closed** as a per-PR FAILURE — its outcome would be unknown, and an unknown arm must
never be a lenient missed cycle. (`gh_retry` refuses to wrap mutations, so this cannot fire
today; the handler exists so a refactor cannot open a second false-pass route.)

## New tests — the exact sequences the review said were missing

Hermetic, injected runners, no `gh`/network. In **both** scripts:

* **the sequence** — candidate A's arm fails, candidate B's refresh / live-state read
  exhausts ⇒ **exit non-zero** with A's failure reported, B recorded as
  `transient-exhausted`, and **no** `::warning` emitted while a real failure stands.
  Asserted at the `arm_exit_code`/`sweep_exit` seam **and end to end through `main()`**
  (both entrypoints now pass `run_gh` explicitly so the real path is drivable);
* **control** — transient exhaustion alone ⇒ **exit 0 + exactly one `::warning`**, and the
  candidates around it are still armed (policy preserved);
* **control** — event mode unchanged: a collected arm failure is non-zero, and a
  transient-only outcome still fails loudly (no per-PR cron backstop);
* **fail-closed** — exhaustion on the arm mutation ⇒ exit 1, no `::warning`;
* **precedence unit assertions** on both outcome dataclasses;
* **4 new cross-file pins** in `scripts/tests/test_arm_capability_wiring.py` (**19** tests
  total) so the two deliberately duplicated copies cannot drift on the precedence rule —
  including that the rule itself is documented in both sources.

## Mutation-verified, not assumed

Snapshot → mutate → confirm RED → `cp`-restore → `cmp`-verify → re-run green. Six
mutations, all killed:

| mutation | result |
|---|---|
| `arm_exit_code` precedence order swapped (transient checked first) | RED — new sequence exits 0 |
| same swap in `sweep_exit` | RED — new sequence exits 0 |
| per-candidate guard removed **+** sticky read-back reverted (auto-arm) | RED — *"a COLLECTED arm failure must dominate a LATER candidate's exhausted transient (code=0…)"* |
| same full revert (rearm-sweeper) | RED — `GhTransientExhausted` escapes `run()` |
| arm-mutation fail-closed handler dropped (both) | RED — exit 0 + `::warning` |

All three restores `cmp`-identical to the snapshot; `gh_retry.py`, `auto-arm.py`,
`rearm-sweeper.py` self-tests and the 19-test wiring suite re-run **green**, plus
`check-advisory-registry.py`, `check-install-action-tool.py` and a YAML parse of the touched
workflow.

**Not armed** — leaving the arm decision to the review pipeline.

jeswr:
Cross-provider review: R1 1 blocking finding (collected arm failure discarded when a later candidate's transient exhausted — reproduced verbatim), fixed in 47e04aea by publishing the outcome on the runner and computing the exit at the end via dedicated seams (6 mutations killed). R2 **VERDICT: pass** — every raise path between first-failure and exit traced, lenient policy confirmed correct only where intended, fail-closed one-shot arm mutation verified. Merging under the maintainer authorization.

Issue #3781, opened 2026-07-25
readied worker PRs are re-drafted before the gate concludes, so the gate can never succeed — deadlocks the whole backlog whenever review capacity is down


jeswr:
> 🤖 SPARQ agent — autonomous orchestration for sparq.

**A readied worker PR is re-drafted before its gate can conclude, so the gate can never succeed on a non-draft head. With the review lane down this is unbreakable by the pipeline, and it is the actual reason the parked backlog does not drain.**

## Observed, repeatedly, tonight
| PR | readied | re-drafted | outcome |
|---|---|---|---|
| #3488 | 07:15 | ~07:33 | only merged because I readied and merged **within ~60s** |
| #3681 | 06:53 | 07:01 | gate later **failed**: `stale draft-tier run, full run pending` |
| #3472 | ~07:15 | before 08:00 | `draft=true`; gate `in_progress` **45 min**, all siblings green |
| #3468 | ~07:15 | before 08:00 | same |

## The mechanism
The pipeline enforces "a worker PR stays DRAFT until it has `review:pass`" — correct by design, and the disarm/re-draft sweep is what enforces it. But the gate has a separate, also-correct rule (`docs/branch-protection.md` §Draft-tier CI):

> `### ci-summary: FAILED — stale draft-tier run, full run pending. … 4 draft-marked select instance(s) have no OWN later full-tier successor … A draft-tier leg set must never admit a non-draft PR to the merge queue.`

Composed, the two rules deadlock whenever review capacity is unavailable:
1. Readying a worker PR starts full-tier CI.
2. The sweep re-drafts it within ~10 minutes, because it has no `review:pass`.
3. The gate is now aggregating on a head whose select legs are draft-tier with no full-tier successor, so it **cannot conclude success**.
4. `review:pass` can only come from the cross-provider review lane, which is dead (`credential-remint-required`, registry #596 / alert #622).

So the PR can never *stay* ready long enough to earn the green gate that would let it merge. Neither rule is wrong; their composition has no exit while the lane is down.

## Why this matters beyond the current outage
The deadlock is not outage-specific — any period without review capacity reproduces it, and it is invisible in the usual signals: the PR looks like a draft with green legs and a perpetually pending gate. Nothing alerts, nothing fails loudly, and a reader concludes "CI is slow" rather than "this cannot ever finish."

## Asks
[...]

Issue #3758, opened 2026-07-23
gate poll loop stalls indefinitely on PR #3470's head — unsatisfiable draft-tier awaiting_full hold (pending==0) bypasses hang detection and holds to the absolute budget


jeswr:
> 🤖 SPARQ agent — autonomous orchestration for sparq.

## Summary

The `ci-summary` gate on PR #3470's head `552368bd5dd609d54ba0d0834fcfe7600f12a2bd` (run 30002198053) enters the draft-tier `awaiting_full` hold in a state where the hold is **provably unsatisfiable**, and that state is **exempt from the sq-90cv4 idle-queue hang detection** (which requires `pending > 0`). With `pending == 0` the loop has no early exit: every gate attempt polls the full 155-poll absolute budget (~68 min of a runner slot) before REDding via the stale-draft-tier belt — and re-running the **gate** (which the repair automation did, twice) can never change the sibling set, so each re-run repeats the identical ~68-minute stall.

## Attempt timelines (run 30002198053, workflow `ci-summary`)

| run_attempt | gate job | started | ended | outcome |
|---|---|---|---|---|
| 1 | 89189756816 | 2026-07-23T11:11:41Z | 12:20:23Z | exhausted all 155 polls in the hold → RED "stale draft-tier run, full run pending" |
| 2 | 89209282592 | 12:44:14Z | 13:50:15Z (cancelled) | 140/155 polls, every one in the identical hold, cancelled externally |
| 3 | 89225653493 | 13:50:27Z | still in_progress | same head, same check-run set → same hold (mid-run log not fetchable: jobs/…/logs returns 404 BlobNotFound while in_progress) |

## Log evidence (attempt 2, cancelled — trimmed tail; polls 1→140 are byte-identical apart from the counter)

```
12:44:58Z attempt 1:   392 check-run(s), 0 running, all-terminal stable for 0/2 poll(s), awaiting the full-tier re-run (draft-tier selection present)
...
13:48:39Z   workflow-run resolver: ignored 297 check-run(s) from superseded runs/attempts (#3505).
13:48:39Z attempt 137: 392 check-run(s), 0 running, all-terminal stable for 0/2 poll(s), awaiting the full-tier re-run (draft-tier selection present)
13:49:06Z   workflow-run resolver: ignored 297 check-run(s) from superseded runs/attempts (#3505).
[...]

PR #3677, opened 2026-07-21
fix(ci): gate resolves required legs to the newest run — superseded/cancelled runs are non-events (#3505)

jeswr:
> 🤖 SPARQ agent — sol-authored (GPT-5.6), orchestrator rescue-committed.

Fixes #3505, the binding merge-path bug: the required `gate` aggregator concluded FAILURE (or hung) off CANCELLED/superseded/evaporated check-runs, so armed review:pass PRs could not enter the merge queue for hours (live loop on #3668/#3669/#3671: rerun→green→new CI wave→concurrency-cancels old legs→gate fails→no admission).

Changes: per required leg, resolve to the NEWEST run/attempt for (workflow, head); superseded runs are non-events regardless of conclusion; a newest-run CANCELLED leg triggers ONE bounded auto-redispatch (marker-guarded) instead of failing the gate; evaporated check-runs resolved via workflow-run listing; genuine newest-run FAILUREs fail exactly as before; gate context name unchanged (branch protection intact); actions:write added for the rerun endpoint (documented in docs/branch-protection.md).

Validation: ci_summary_gate --self-test, 120 aggregator tests, 66 wiring tests, ruff, actionlint on both workflows — all green. Cross-provider adversarial review (Anthropic/Opus) to follow before arm.

PR #629, opened 2026-06-18
feat(js): SparqStore.queryQuads() for CONSTRUCT/DESCRIBE (sq-1gkw) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent

## What

Implements **sq-1gkw**: `@jeswr/sparq`'s `SparqStore` now exposes the graph-valued query forms (CONSTRUCT / DESCRIBE). The underlying wasm `Store` already bound `queryQuads` / `queryQuadsChunks` (constructed graph as N-Triples), but the JS wrapper only surfaced SELECT/ASK and its `query()` JSDoc still claimed "CONSTRUCT/DESCRIBE are not supported by the engine yet".

New `SparqStore` methods:

| method | returns |
| --- | --- |
| `queryQuads(sparql)` | `Quad[]` — constructed/described graph as RDF/JS quads (default graph) |
| `queryQuadsString(sparql)` | `string` — the raw N-Triples document (a valid Turtle subset) |
| `queryQuadsStream(sparql, batchSize=1024)` | `Generator<Quad>` — bounded streaming for a large graph (frees the wasm cursor on early `break`) |

Plus an exported `parseNTriples(nt): Quad[]` helper that decodes the engine's canonical N-Triples serialisation — full `<IRI>`, `_:bnode`, `"lit"` / `@lang` (incl. RDF 1.2 `--dir`) / `^^<dt>`, and the `\t \b \n \r \f \" \' \\` and `\uXXXX` / `\UXXXXXXXX` escapes — back to RDF/JS terms.

A SELECT/ASK routed to `queryQuads` throws (the engine rejects it); `query()`/`queryBindings`/`queryBoolean` remain the path for those.

## Docs (G2)

Public-API change to `@jeswr/sparq`, so updated `skills/javascript-wasm/SKILL.md`, `js/README.md`, and the site surface page to document the new methods and drop the stale "wrapper does not expose CONSTRUCT/DESCRIBE" claims.

## No Rust change

The wasm binding already existed; this is a JS-surface + docs change only.

## Gate

- `npm run build:wasm` (published `--features shacl,jsonld` bundle) + `build:ts` clean (tsc strict).
- Full JS test suite green (9 files); 7 new tests in `js/test/store.test.mjs` cover CONSTRUCT, DESCRIBE (CBD), the string form, multiset-equal streaming, early-`break` cursor free, SELECT-rejection, and datatyped/lang/escaped/IRI-object round-trips.
- Site typecheck (`tsc --noEmit`) clean.

[...]

PR #3785, opened 2026-07-25
fix(ci): the hang detector must not fire on live bounded proofs, and a descoped OIDC lane must not gate main

### LINKED ISSUE #3783: ci-summary's hang detector declares 'genuine hang' while kani bounded proofs are legitimately in_progress — reds main on a budget expiry

**`ci-summary`'s hang detector declares a "genuine hang" while long-running `kani` bounded-proof jobs are legitimately `in_progress`, reding `main`'s gate.** Observed on `main` (`e8a41ab8c`) just now, and I contributed to the conditions — see the honest attribution below.

## What happened

`gate` run `30149978128` began polling at **07:48:48** and concluded at **08:31:47** with:

```
##[error]ci-summary timed out — 5 sibling check-run(s) never finished within the base budget,
the Actions queue is idle (depth=0 < 5) and no completions landed in the last 15 poll(s):
genuine hang, not a still-settling set.
```

But at that moment the incomplete siblings were:

```
kani sparq-core    (dict mmap validator totality)      in_progress  started=08:20:50
kani sparq-engine  (vectorized reducer kernels)        in_progress  started=08:20:49
kani sparq-vectors (.spqv validator, informational)    in_progress  started=08:20:50
```

**They had started 11 minutes earlier and were still running.** Nothing hung.

## Why the heuristic misfires

The detector's evidence is "**Actions queue depth is 0** AND **no completions in the last 15 polls**". Both are true of a *healthy* long-running bounded proof:

- queue depth is 0 **because the job already dequeued and started** — an idle queue is evidence the work is *running*, not that it is lost;
- `kani` emits no completion for tens of minutes by design, so "no completions recently" is its normal signature.

So the two signals that are supposed to establish a hang are both satisfied by the very jobs least likely to be hung. The detector needs to consult **sibling status** — if any awaited sibling is `in_progress`, that is positive evidence of liveness and must veto the hang verdict. Queue depth should only count when awaited siblings are `queued`/absent, i.e. genuinely not running.

## My contribution, stated plainly

I drove **12 parked PRs through full-tier CI in one pass** (plus 13 merges) within about an hour. That is the congestion pattern I was explicitly capping against, and it plausibly delayed these `kani` legs — they started ~30+ minutes after the `main` push, which is late. So the *conditions* were partly mine; the *misdiagnosis* is the detector's. Both are worth fixing, and I am recording my half rather than only the code's.

## Asks

1. **Veto the hang verdict on any `in_progress` awaited sibling.** Smallest correct fix; changes no other behaviour.
2. Give the base budget headroom for the known-slow lanes (`kani`, `cargo-fuzz`, coverage shards) rather than treating one budget as universal — or exempt them from the hang heuristic and let their own `timeout-minutes` govern.
3. **Distinguish "timed out" from "failed" in the reported conclusion.** A budget expiry with live siblings is not a defect in the tree, but it reds `main` identically to one, which is exactly the signal-degradation problem that has cost me time repeatedly tonight (see #3781, and the stale-run and single-surface miscounts recorded there).

`main` should recover on its own once the `kani` legs finish and a fresh `gate` runs. Nothing in the tree is broken — the 289 other checks on that head are green with zero gating failures.

Related: #3677 (newest-run resolution + bounded redispatch for evaporated check-runs), #3781 (the ready→re-draft race, which also burns a full budget producing a guaranteed refusal), #3758/#3765 (unsatisfiable-hold budget burn — this is a third variant of "the gate spends its whole budget then reports the wrong thing").

### LINKED ISSUE #3784: nightly EC2 benchmark GATES but can never pass — its AWS OIDC role was descoped, so main can never be green

**`nightly full-suite benchmark on EC2 (spot)` is a GATING check that can never pass, because it authenticates via an AWS OIDC role that was deliberately descoped. While it exists in this state, `main` cannot be green.**

## Evidence

Run `30151339208`, job `89662120793`, on `main` (`e8a41ab8c`), concluded `failure` at 08:35:29Z:

```
Configure AWS credentials (OIDC)
  Retry validateCredentials: attempt 1..12 of 12 failed:
    Credentials could not be loaded, please check your action inputs:
    Could not load credentials from any providers.
  Retry validateCredentials: reached max retries (12); giving up.
##[error]Credentials could not be loaded, please check your action inputs
```

It fails in `Configure AWS credentials (OIDC)` — **before any benchmark runs**. So this is not a performance regression, not spot-capacity reclamation, and not caused by any recent merge. It is an authentication failure against infrastructure that is not there.

## Why it cannot be fixed by retrying

The orchestration design **explicitly dropped EC2 and the AWS OIDC role** ("NO EC2, and NO billable runners… the EC2 build-farm, EC2 codex worker, LOCAL-vs-EC2 tiering, and the AWS OIDC role are all out of scope"). This workflow still assumes that role exists. There is no transient condition here — every scheduled run will fail identically until the lane is either re-provisioned or retired.

## Why it matters more than a single red check

The check name contains no `advisory`/`informational` token and (post-#3774) it carries no `.github/advisory-registry.json` entry, so it **GATES**. `main` currently reports two gating failures: this, and `gate` itself (which timed out separately — see #3783). Every other check on that head is green. So the practical effect is that `main`'s red is permanent and uninformative, which is precisely the signal degradation that has cost me repeated wasted diagnosis tonight.

## Asks — and I am deliberately NOT choosing for you

1. **Retire the lane** if EC2 benchmarking is genuinely out of scope. That is the reading most consistent with the descoping decision, and it removes a permanently-failing gate rather than muting it.
2. **Or re-provision the OIDC role** if canonical EC2 benchmarking is still wanted, and note that the descoping decision then needs amending so this does not recur.
3. **Or declare it advisory** with an explicit registry entry and written justification.

I am not taking option 3 unilaterally, even though it is the fastest way to green. #3774 landed the rule that advisory status must be **declared deliberately rather than inferred**, and I gave that PR's author the instruction that if restoring real gates turned `main` red, "that red is real information and I want it." This red **is** real information — it says the nightly EC2 lane cannot authenticate. Muting it to make a dashboard green would be exactly the behaviour #3774 exists to prevent, and would make me the first violator of a rule I merged four hours ago.

A nightly benchmark arguably should not gate PR merges at all, which makes option 3 defensible on the merits — but that is a policy call about what may block a merge, and it is yours.

Related: #3783 (`gate` timing out on live `kani` proofs — the other half of `main`'s current red), #3774 (declared-not-inferred advisory status), #3781 (the ready→re-draft race).

jeswr:

Fixes both causes of `main` being red. Neither is a code defect in the tree; both are CI-infrastructure defects on the same surface (`scripts/ci_summary_gate.py` + workflow wiring), so they ship together.

Closes #3783. Closes #3784.

---

## Defect 1 (#3783) — the hang detector misfired on live `kani` bounded proofs

The genuine-hang heuristic had two signals: **the Actions queue is idle** AND **no completions in the last `PROGRESS_WINDOW` polls**. Both are satisfied by a *perfectly healthy* long bounded proof — the queue is idle *precisely because* the job dequeued and started, and a `kani` harness emits no completion for tens of minutes by design. The two signals meant to prove a hang are exactly what a live proof looks like. Live evidence: `gate` run `30149978128` on `main` (`e8a41ab8c`) concluded "genuine hang" while `kani sparq-core` / `sparq-engine` / `sparq-vectors` had been `in_progress` for 11 minutes.

**Fix — LIVENESS VETO.** An awaited sibling in `in_progress` is positive liveness evidence and vetoes the hang verdict (`live_siblings()`). Queue depth may only count toward "hang" when the awaited siblings are `queued` or absent, i.e. genuinely not running.

**The #3677 case is deliberately preserved.** An evaporated check-run is represented by `_workflow_summary_check(..., force_pending=True)`, whose status is **`queued`** — never `in_progress` — so a lost leg still REDs at the base budget. That discrimination (`in_progress` ⇒ not a hang; `queued`/absent ⇒ still a hang) is the property the tests pin.

**On the optional slow-lane exemption — judgement: rejected, and the veto subsumes it.** A hard-coded `kani`/`cargo-fuzz`/coverage-shard list would re-introduce exactly the rename fragility #3773 removed from the advisory rule (rename the job, silently change the semantics). Reading liveness from the platform's own status field is name-independent and covers every slow lane, present and future, for free.

**On budget headroom — judgement: declined, with the arithmetic.** `kani` on `main` runs ~41 min wall-clock (`30150937647`: 08:20:47 → 09:01:36, and the four prior nightlies are within a minute of that). The gate's absolute cap is `110×20s + 45×40s` ≈ 67 min against `ci-summary.yml`'s own `timeout-minutes: 80` — so raising the cap enough to cover last night's shape (the `kani` cron fired 32 min after the gate began polling, putting completion at ~73 min) would leave under 5 min of margin, and going further means raising the gate job's timeout too. The gate is a **waiter occupying a runner slot**; making every gate hold a slot longer under congestion worsens the very congestion that delayed `kani` (memory: `reference-ci-congestion-collapse`). Declined as a band-aid. The real lever is in "What still remains" below.

### Reporting (#3783 ask 3) — "could not determine" ≠ "the tree is broken"

Three separate times in one night the gate spent its whole budget and then reported the wrong thing (#3758/#3765 unsatisfiable hold; #3781 ready→re-draft race burning 155 polls; #3783 this). Both `ABSOLUTE`-budget exits now report **`UNDETERMINED (not a test failure)`**, naming *which* could-not-determine it was — siblings still **EXECUTING** (listed by name, so the reader is not sent hunting) vs the runner pool never draining — and stating explicitly that nothing has been shown to be broken. A base-budget **genuine hang stays a FAILURE**, because nothing executing + an idle queue + no completions really is a broken pipeline.

**Fail-closed posture is unchanged.** Both new branches still `exit 1` — an unobserved leg is never assumed green. Mechanically: `return 0` count **6 → 6**, `return render_verdict(...)` call sites **2 → 2**, `return 1` count **15 → 16**. Exactly one `return 1` added; no exit-0 path added or widened.

---

## Defect 2 (#3784) — a GATING check that can never pass

`nightly full-suite benchmark on EC2 (spot)` failed in `Configure AWS credentials (OIDC)` after 12 retries — *before any benchmark ran* (run `30151339208`, job `89662120793`) — because the orchestration design deliberately descoped EC2 and the AWS OIDC role. Carrying no advisory token and no `.github/advisory-registry.json` entry, post-#3774 it **GATED**, so `main` could not be green while it existed.

### Chosen: option 1 — RETIRE the automated lane. Not option 3.

**Why.** The registry entry was available and would have been the fastest route to green. It would also have been the first violation of a rule merged four hours earlier: #3774 landed the principle that advisory status is *declared deliberately*, and muting a **permanently broken** gate to green a dashboard is precisely the behaviour that rule exists to prevent. There is a real distinction here — the `nightly-full-sweep.yml` lanes in the registry are declared advisory because a nightly *that runs* must not block a merge; this lane **cannot run at all**. Declaring it advisory would record "we accept this lane's verdict as informational", which is false: there is no verdict. Retirement records the truth: the lane is dead code the moment its role was descoped.

**What changed.**

1. **Both `schedule:` crons removed.** Not just the nightly — the weekly heavy campaign authenticates through the same dead role and would have redded `main` every Monday. Retiring only the nightly would have left the same bug on a slower clock.
2. **Maintainer-run EC2 benchmarking preserved** — it is *not* descoped and is actively used. Both campaigns stay reachable via `workflow_dispatch` with a `lane` input (`heavy` | `full-suite`), and **no EC2 bench tooling or documentation is deleted**: `scripts/ec2-bench.sh`, `scripts/bench/canonical-beir-gather-instance.sh` and its orphan-proof rails (#3488), and `research/ci-ec2-design.md` all stay. Only the CI lane that authenticates via the dead role is retired.
3. **Role-present guard** on both jobs (`vars.AWS_BENCH_ROLE_ARN != ''`), so a dispatch while the role is descoped **skips visibly** instead of burning 12 credential retries and posting a gating failure. Provisioning the repo variable revives the lane with no further edit. This is not a mute: it is one reviewable clause naming the variable, `skipped` honestly means "the descoped infrastructure was not there", and a test asserts no `bench-ec2.yml` job is ever declared advisory.
4. **`nightly-full-bench`'s display name no longer claims a cadence it does not have** (`full-suite benchmark on EC2 (spot, manual dispatch)`). Its **job id and the `sparq engine (nightly full suite)` series name are kept** so the published `dev/bench-nightly` dashboard history stays continuous.
5. **Docs corrected where they were now false** — `docs/branch-protection.md`, `docs/pages-cutover-runbook.md`, `bench/CATALOG.md`, `research/ci-ec2-design.md` (status banner: it is now the *revival spec*, its `schedule:` block is aspirational), plus the `bench.yml` / `pages.yml` / `ci-bench.sh` headers. Notably the claim that the at-scale perf-tracking was *"moved, not lost"* **no longer holds** and has been retracted rather than left to rot: with the cron retired, that series is collected only on demand.

---

## Verification — mutations applied to the LIVE files, never a fixture

`163` hermetic tests green (`python3 scripts/tests/test_ci_summary_gate.py`), plus `ci_summary_gate.py --self-test`, `check-advisory-registry.py` (all clear C2+C3+C4), `check-no-perf-numbers.py --enforce`, `check-readme-template.py --enforce`, `markdownlint-cli2` (0 errors), `typos`, all 25 `scripts/tests/test_*.py`.

Nine mutants, each applied by rewriting the real `scripts/ci_summary_gate.py` / `.github/workflows/bench-ec2.yml` / `.github/advisory-registry.json`, snapshot-restored by `cp` with marker verification. **Zero survivors.** Each redded because a guard fired, not because a search string vanished:

| # | Mutation (on the live file) | Test that REDs |
|---|---|---|
| M1 | delete the veto: `if not (saturated or progressing or live)` → `(saturated or progressing)` | `test_in_progress_siblings_are_not_a_hang` |
| M2 | flip the comparison: `LIVE_STATUS = "in_progress"` → `"queued"` | `test_all_queued_siblings_are_still_a_hang` |
| M3 | broaden the veto: `status == LIVE_STATUS` → `status != "completed"` | `test_all_queued_siblings_are_still_a_hang` |
| M4 | delete the taxonomy: report a live-sibling expiry as `FAILED` | `test_budget_expiry_with_live_siblings_reads_undetermined` |
| M5 | collapse the taxonomy the other way: launder a genuine hang as `UNDETERMINED` | `test_genuine_hang_still_reads_as_a_failure` |
| M6 | `force_pending` synthetic claims liveness: `"queued"` → `"in_progress"` | `test_evaporated_check_run_resolves_to_queued_not_live` |
| M7 | re-add the retired cron to `bench-ec2.yml` | `test_no_schedule_trigger` |
| M8 | delete the role-present guard from the full-suite OIDC job | `test_every_oidc_job_carries_the_role_present_guard` |
| M9 | take the easy path: add a `bench-ec2.yml` advisory-registry entry | `test_the_lane_was_not_muted_via_the_advisory_registry` |

Naming assertions, as emitted:

**M1** (`'genuine hang' unexpectedly found in …`)
> `#3783 REGRESSION: the hang detector fired while awaited siblings were `in_progress` — an executing kani bounded proof was declared a hang. An idle Actions queue means those jobs DEQUEUED and STARTED, and kani emits no completion for tens of minutes by design, so neither signal is evidence of a hang. The liveness veto (live_siblings) is missing.`

**M2 / M3** (`'genuine hang' not found in …`) — the discrimination, in the other direction
> `#3783 OVER-CORRECTION: awaited siblings that never STARTED (all `queued`, idle queue, no completions) are the genuine-hang case the detector exists for (#3677 evaporated check-runs resolve to `queued` too). The veto must key on `in_progress` ONLY — it has been broadened to swallow non-running legs, so a real hang now waits out the whole budget instead of REDing.`

**M4** (`'UNDETERMINED (not a test failure)' not found in …`)
> `#3783 ask 3 REGRESSION: a budget expiry with siblings STILL EXECUTING is reported identically to a real gating failure. Nothing was shown to be broken — this outcome must say 'could not determine', not 'FAILED'.`

**M5** (`'UNDETERMINED' unexpectedly found in …`)
> `#3783: a genuine hang must NOT be softened to 'could not determine' — nothing running with an idle queue and no completions is a real defect.`

**M6** (`'in_progress' != 'queued'`)
> `#3677/#3783 REGRESSION: the force_pending run-level synthetic (how an EVAPORATED check-run is represented) no longer reports `queued`. If it reports `in_progress` it looks EXECUTING to the liveness veto, so a lost leg vetoes its own hang detection and the gate waits out the whole budget instead of REDing — the exact case the detector was built for.`

**M7** (`'schedule' unexpectedly found in {'schedule': [{'cron': '41 6 * * *'}], …}`)
> `#3784 REGRESSION: bench-ec2.yml runs on a `schedule:` again. Its AWS OIDC role is DESCOPED, so every cron tick fails in `Configure AWS credentials (OIDC)` before any benchmark runs, posts a GATING check-run on main's head SHA (no advisory token, no registry entry => it GATES per #3773/#3774), and makes `main` permanently and uninformatively red. Re-adding a cadence requires re-provisioning vars.AWS_BENCH_ROLE_ARN in the same change.`

**M8** (`"vars.AWS_BENCH_ROLE_ARN != ''" not found in "github.repository == 'sparq-org/sparq' && github.event.inputs.lane == 'full-suite'"`)
> `#3784 REGRESSION: job `nightly-full-bench` assumes the DESCOPED AWS OIDC role with no role-present guard, so a dispatch while the role is absent burns 12 credential retries and posts a gating failure instead of skipping.`

**M9** (`'bench-ec2.yml' == 'bench-ec2.yml'`)
> `#3784: `full-suite benchmark on EC2 (spot, manual dispatch)` declares a bench-ec2.yml job advisory. The descoped OIDC lane was RETIRED, not muted — declaring a permanently-failing gate advisory to green the dashboard is precisely what #3774's declared-not-inferred rule exists to prevent.`

Two existing tests had to be **re-fixtured, because their fixtures encoded the defect**: `test_genuine_hang_fails` and `test_no_progress_and_depth_unknown_is_a_hang` both awaited an **`in_progress`** sibling and asserted "genuine hang" — i.e. the suite pinned the misfire. They now await a `queued` sibling, which is the real genuine-hang shape (and the shape an evaporated check-run resolves to). Same for `test_fetch_queue_depth_raises_treated_as_unknown_no_extension`, whose subject is the `try/except`, not the hang shape.

---

## Will `main` be green once this lands?

**#3784: yes, definitively.** A cron-less workflow posts no check-runs on a `main` head SHA, so the permanently-failing gating check disappears rather than being reclassified.

**#3783: the false RED is fixed, but a residual UNDETERMINED red remains reachable.** With the veto, live `kani` proofs no longer produce a "genuine hang" FAILURE. Whether the gate then goes *green* depends on whether `kani` finishes inside the ~67-min absolute budget:

- `kani` cron fires near its scheduled time ⇒ ~41 min ⇒ gate converges **green**.
- `kani`'s cron is delayed more than ~26 min (last night: the 06:11 cron fired at 08:20) ⇒ the budget expires and the gate exits 1 as **`UNDETERMINED (not a test failure)`**, naming the live legs. Honest and diagnosable, but still not green.

### What still remains — the actual root cause, deliberately left for its own review

Verified after opening: the three legs in the incident were `kani.yml`'s **`schedule`-triggered** jobs (run `30150937647`, `event: schedule`, path `.github/workflows/kani.yml`, names ending `(bounded proofs, informational)`), and that job **is declared advisory** in `.github/advisory-registry.json`. `kani.yml` has no `push` or `pull_request` trigger at all.

So the precise shape is: a **cron lane fired 2h09m late** (06:11 cron, ran 08:20), landed its check-runs on `main`'s head SHA while the push `gate` was mid-poll, re-armed the settle window, and the gate then waited 43 minutes on legs whose conclusion **it excludes from the verdict by construction** — before reporting a hang. (Note this is *not* true of `formal-verification.yml`'s `kani … (change-coupled proofs)` legs, which are deliberately **gating** and fire on the same `push`/`pull_request` as the gate, so they start together and their measured runtime fits inside the budget. Those must keep being waited on.)

The fix is one narrow change — stop counting **declared-advisory** siblings as work the gate must wait for. The set of check-runs whose outcome the gate ignores would be unchanged; we would merely stop *blocking* on the ones already ignored. It would have made run `30149978128` conclude green in about five polls, and it *reduces* gate runner occupancy, unlike any budget increase.

I did not do it here, for a stated reason: it **widens an exit-0 path** (the existing clean-convergence `render_verdict` would fire earlier, so a set with no registered gating legs could pass sooner). This PR's acceptance constraint was that the aggregator's fail-closed behaviour stay unchanged with only `return 1` added, and widening the success path of the one check that authorises every merge deserves its own diff and its own adversarial review, not a rider on a red-main fix. Filed as a follow-up.

## Not a real defect

- **The `kani` legs themselves.** They completed `success` (~41 min, run `30150937647`), and the four preceding nightlies are within a minute of that runtime. Nothing about `kani` needs fixing; every other check on `e8a41ab8c` was green with zero gating failures.
- **`scripts/ec2-bench.sh` and the EC2 bench tooling.** The failure was pre-benchmark authentication, not the scripts. Untouched.
- **The `#3677` hang detection.** It was correctly built and is correctly scoped by this change, not removed — verified by M6 and `test_all_queued_siblings_are_still_a_hang`.
- **The pre-existing `if: failure()` demoted-lane bead filer in `bench-ec2.yml`.** It behaved as designed; it was firing on a credentials failure only because the lane was firing at all. The role-present guard makes it unreachable while the role is descoped.

Not armed.

### DIFF
--- a/.github/workflows/bench-ec2.yml
+++ b/.github/workflows/bench-ec2.yml
@@ -1,8 +1,38 @@
 name: Benchmarks (EC2)
 
-# Cost-capped EC2-deferred benchmark (roadmap G3). The cost PROTECTIONS live here:
-#   * rate limit  — runs ONLY on cron (weekly heavy + nightly full-suite) + manual dispatch; NEVER on
-#                   push or pull_request, so no per-commit and no fork-PR triggers.
+# Cost-capped EC2-deferred benchmark (roadmap G3).
+#
+# ═══ RETIRED AS AN AUTOMATED LANE — MANUAL DISPATCH ONLY ([OPUS-5] #3784) ═══
+# The AWS OIDC role this workflow assumes (`vars.AWS_BENCH_ROLE_ARN`) was DELIBERATELY
+# DESCOPED by the orchestration design ("NO EC2, and NO billable runners… the EC2
+# build-farm, EC2 codex worker, LOCAL-vs-EC2 tiering, and the AWS OIDC role are all out
+# of scope"). So `Configure AWS credentials (OIDC)` can no longer authenticate: run
+# 30151339208 / job 89662120793 burned 12 retries of "Could not load credentials from
+# any providers" BEFORE any benchmark ran, and it did so on EVERY scheduled tick.
+# A check-run named without an advisory token and with no `.github/advisory-registry.json`
+# declaration GATES (#3773/#3774), so a permanently-failing cron lane made `main`
+# permanently and uninformatively red.
+# THE FIX IMPLEMENTS THE DESCOPING DECISION rather than muting a live signal:
+#   * the `schedule:` triggers are GONE — both of them. The weekly heavy cron carried the
+#     identical dead-OIDC defect and would have redded `main` every Monday, so retiring
+#     only the nightly would have left the same bug on a slower clock.
+#   * NOT declared advisory. Muting a permanently-broken gate to green a dashboard is
+#     exactly what #3774's declared-not-inferred rule exists to prevent; a registry entry
+#     here would have been the easy path, not the honest one. Pinned by a test
+#     (scripts/tests/test_ci_summary_gate.py::TestDescopedEc2Lane).
+#   * MAINTAINER-RUN EC2 BENCHMARKING IS NOT DESCOPED and is NOT removed. Both campaigns
+#     stay reachable via `workflow_dispatch` (pick the lane with the `lane` input), and
+#     every EC2 bench script + doc is untouched — notably scripts/ec2-bench.sh and the
+#     orphan-proof scripts/bench/canonical-beir-gather-instance.sh rails (#3488).
+#   * ROLE-PRESENT GUARD: each job now requires `vars.AWS_BENCH_ROLE_ARN != ''`, so a
+#     dispatch while the role is descoped SKIPS visibly instead of failing at the OIDC
+#     step. Re-provision the repo variable and the lane revives unchanged; re-add a
+#     `schedule:` block in the same commit if the automated cadence is wanted back, and
+#     restate the cost basis below when you do.
+#
+# The cost PROTECTIONS (unchanged, and still the reason this design is safe to revive):
+#   * rate limit  — manual dispatch only; NEVER on push or pull_request, so no per-commit
+#                   and no fork-PR triggers.
 #   * fork guard  — the jobs are skipped unless it's the canonical repo.
 #   * OIDC        — assumes a short-lived, tag-scoped IAM role (no long-lived AWS keys); see
 #                   research/ci-ec2-design.md for the trust + permissions policies to create.
@@ -11,34 +41,41 @@ name: Benchmarks (EC2)
 #
 # TWO LANES (both on the SAME cost-capped EC2 rails — spot + tag purpose=sparq-ci-bench + the
 # always-run terminate/delete cleanup trap; NEVER touch prod i-090531b4ede8f2d3f / dev
-# i-00f76802f345b6b77):
-#   * ec2-bench (WEEKLY, Mondays 06:00 UTC) — the HEAVY campaign at the ~2M-triple scale (the
-#     roadmap-G3 large-hardware run). Publishes to dev/bench-ec2. UNCHANGED.
-#   * nightly-full-bench (NIGHTLY, 06:41 UTC) — [FABLE-5] sq-6vshe.6, MAINTAINER-DIRECTED. Runs the
+# i-00f76802f345b6b77). Select with the `lane` dispatch input:
+#   * ec2-bench (`lane: heavy`) — the HEAVY campaign at the ~2M-triple scale (the
+#     roadmap-G3 large-hardware run). Publishes to dev/bench-ec2.
+#   * nightly-full-bench (`lane: full-suite`) — [FABLE-5] sq-6vshe.6, MAINTAINER-DIRECTED. Runs the
 #     FULL NOISY TIMING SUITE (query latencies + well-known sp2b/dbpsb/watdiv/bsbm/lubm + the cargo-only
 #     fts/geo/vector/rsp/hdt/solid/nlq/zk hooks) at the SAME 200k CI scale the per-commit series used, on
 #     a quiet dedicated spot instance (contention-free wall-clock, unlike a shared GitHub runner). This is
 #     the lane the flaky/slow benchmark timing was MOVED TO: it was dragging the merge queue + flapping
-#     the gate on shared runners, so the per-PR bench lane now runs ONLY the fast deterministic byte-count
-#     ratchet (bench.yml `--deterministic-only`) and this nightly EC2 lane carries the full perf-tracking
-#     series. Publishes to dev/bench-nightly (a distinct dashboard dir, so it collides with neither the
-#     per-commit dev/bench nor the weekly heavy dev/bench-ec2). The perf-tracking is NOT lost — only moved.
+#     the gate on shared runners, so the per-PR bench lane runs ONLY the fast deterministic byte-count
+#     ratchet (bench.yml `--deterministic-only`). Publishes to dev/bench-nightly (a distinct dashboard
+#     dir, so it collides with neither the per-commit dev/bench nor the heavy dev/bench-ec2). The job id
+#     and the `sparq engine (nightly full suite)` series name are KEPT so the published dashboard history
+#     stays continuous; its DISPLAY name no longer claims a nightly cadence it does not have.
+#     HONEST STATUS ([OPUS-5] #3784): with the cron retired and the OIDC role descoped, the at-scale
+#     timing series is NOT being collected — it is dispatch-on-demand, not automated. Saying otherwise
+#     would be the doc drift this change exists to remove.
 #
-# COST BASIS (why NIGHTLY is affordable):
-#   * The nightly runs the full suite at the 200k CI scale (NOT the 2M heavy scale), so a run is a short
-#     spot session on a c7g.4xlarge (arm64, spot ~$0.30-0.40/hr): build + full suite ~10-20 min ≈
-#     ~$0.07-0.14/run. 30 nights/month ≈ $2-4/month. Together with the weekly heavy (~$0.80/month) the
-#     EC2 bench total stays UNDER the $5/month cap. Back it with an AWS Budgets alarm on the
-#     purpose=sparq-ci-bench tag. If a future scale-up pushes over budget, drop the nightly cron to
-#     `weekly` (state the new cost basis in this header) — the design does not depend on the cadence.
-#
-# NOT LIVE until `AWS_BENCH_ROLE_ARN` (repo variable) points at the IAM role; until then the runs
-# fail harmlessly at the OIDC step (no instance launched, no cost).
+# COST BASIS (per manual run, for whoever revives the cadence): the full-suite lane runs at the 200k CI
+# scale (NOT the 2M heavy scale) — a short spot session on a c7g.4xlarge (arm64), build + full suite well
+# inside the job timeout. The heavy lane is the 2M-scale campaign. Both stay under the documented
+# $5/month EC2 bench cap at their original weekly+nightly cadence; back any revived cadence with an AWS
+# Budgets alarm on the purpose=sparq-ci-bench tag.
 on:
-  schedule:
-    - cron: '0 6 * * 1'      # ec2-bench: Mondays 06:00 UTC (weekly heavy)
-    - cron: '41 6 * * *'     # nightly-full-bench: 06:41 UTC daily (offset off other nightly lanes)
+  # [OPUS-5] #3784: NO `schedule:` — see the RETIRED note above. A cron tick against a
+  # descoped OIDC role is a guaranteed gating failure on `main`, every night.
   workflow_dispatch:
+    inputs:
+      lane:
+        description: 'Which EC2 campaign to run (both require vars.AWS_BENCH_ROLE_ARN)'
+        type: choice
+        required: false
+        default: heavy
+        options:
+          - heavy
+          - full-suite
 
 # [OPUS-4.8] Least-privilege default (Scorecard TokenPermissions): top-level read-only;
 # the single ec2-bench job opts into the writes it genuinely needs (per-job below).
@@ -51,13 +88,17 @@ concurrency:
 
 jobs:
   ec2-bench:
-    name: heavy benchmark on EC2 (spot)
-    # [FABLE-5] sq-6vshe.6: the WEEKLY heavy campaign. Runs on the Monday cron OR a manual dispatch
-    # (unchanged manual behavior); the nightly '41 6 * * *' cron routes to nightly-full-bench instead,
-    # so a nightly tick never launches this heavy 2M-scale instance.
+    name: heavy benchmark on EC2 (spot, manual dispatch)
+    # [FABLE-5] sq-6vshe.6: the HEAVY campaign, now maintainer-dispatched (`lane: heavy`, the
+    # default). [OPUS-5] #3784: the `vars.AWS_BENCH_ROLE_ARN != ''` clause is the ROLE-PRESENT
+    # GUARD — while the AWS OIDC role is descoped this job SKIPS visibly instead of burning 12
+    # credential retries and posting a gating failure. It is not a mute: `skipped` here means
+    # "the descoped infrastructure was not there", the guard is one reviewable clause naming the
+    # variable, and provisioning the variable revives the lane with no further edit.
     if: >-
       github.repository == 'sparq-org/sparq' &&
-      (github.event_name != 'schedule' || github.event.schedule == '0 6 * * 1')
+      github.event.inputs.lane == 'heavy' &&
+      vars.AWS_BENCH_ROLE_ARN != ''
     runs-on: ubuntu-latest
     # [OPUS-4.8] Job-scoped writes (least privilege):
     #   id-token: write — assume the short-lived, tag-scoped IAM role via OIDC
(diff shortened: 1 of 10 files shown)


jeswr:

**Merged on `gate=success` — verified head-scoped on `bcfb62331`, newest-run-per-name resolution, zero gating failures, zero pending — under the maintainer's standing authorization.**

## The #3784 decision, and why retirement rather than the advisory mute

An advisory-registry entry was available and was the fastest route to green. It would also have been **the first violation of a rule merged four hours earlier**. The reasoning that settles it:

> The `nightly-full-sweep.yml` lanes are declared advisory because a nightly *that runs* must not block a merge. **This lane cannot run at all.** Declaring it advisory records "we accept this lane's verdict as informational", which is false — **there is no verdict.** Retirement records the truth: dead code the moment its role was descoped.

That is a better articulation of #3774's rule than the one I gave when I merged it. Advisory means "informational verdict"; it does not mean "no verdict".

Three specifics worth keeping:
- **Both crons removed**, not just the nightly — the weekly heavy campaign authenticates through the same dead role and would have redded `main` every Monday. Retiring only the nightly leaves the same bug on a slower clock.
- **Maintainer-run EC2 benchmarking preserved** — both campaigns stay dispatchable via a `lane` input, and `scripts/ec2-bench.sh`, #3488's orphan-proof `canonical-beir-gather-instance.sh` rails and `research/ci-ec2-design.md` are untouched. Descoping the *CI lane* is not descoping EC2 benchmarking.
- **A role-present guard** (`vars.AWS_BENCH_ROLE_ARN != ''`) so a dispatch while the role is absent **skips visibly** instead of burning 12 credential retries — and provisioning that variable revives the lane with no further edit. The descoping is now reversible by configuration rather than by a revert.

## The fifth instance tonight of tests encoding the defect

Two existing tests had to be re-fixtured **because their fixtures pinned the misfire**: `test_genuine_hang_fails` and `test_no_progress_and_depth_unknown_is_a_hang` both awaited an `in_progress` sibling and asserted "genuine hang". The suite was actively holding the bug in place. That is now five occurrences of the same mechanism in one night — a test authored to match observed behaviour cannot catch what the code gets wrong.

The load-bearing mutations name the defect rather than merely reddening:
> *"#3783 REGRESSION: the hang detector fired while awaited siblings were `in_progress` — an executing kani bounded proof was declared a hang. An idle Actions queue means those jobs DEQUEUED and STARTED…"*
> *"#3783 OVER-CORRECTION: awaited siblings that never STARTED are the genuine-hang case the detector exists for (#3677 evaporated check-runs resolve to `queued` too). The veto must key on `in_progress` ONLY…"*

Both directions of the discrimination, which is what makes the fix a fix rather than a blinding. Fail-closed posture verified mechanically: `return 0` **6→6**, `render_verdict` call sites **2→2**, `return 1` **15→16** — exactly one line added.

## An honest residual, and a correct refusal to fix it here

**#3783's false RED is fixed; an `UNDETERMINED` red stays reachable.** The incident's `kani` legs are **declared advisory** with no `push`/`pull_request` trigger — a cron lane fired **2h09m late** (06:11 → 08:20), landed on `main`'s head mid-poll, re-armed the settle window, and the gate waited 43 minutes on legs **whose conclusion it excludes by construction**. Filed as **#3786**: stop counting declared-advisory siblings as work the gate must *wait* for.

That was deliberately **not** shipped here, because it widens an exit-0 path (clean-convergence `render_verdict` fires earlier) — which contradicts this PR's fail-closed constraint and deserves its own adversarial review rather than riding along on a red-main fix. Refusing to smuggle an exit-0 widening into an urgent fix is the right instinct, and it is the opposite of what I did earlier tonight when I merged #629 on evidence that turned out to be theatre.

Two proposals declined with arithmetic rather than opinion: **budget headroom** (cap ≈67 min against `timeout-minutes: 80`; covering last night's 73-min shape leaves <5 min margin — and the gate is *a waiter holding a runner slot*, so lengthening it worsens the very congestion that delayed `kani`), and a **slow-lane name list** (it re-introduces exactly the rename fragility #3773 removed; reading liveness from the platform's status field is name-independent and covers every slow lane for free).

Also retracted rather than left to rot: the doc claim that at-scale perf tracking was *"moved, not lost"*. With the cron retired, that series is on-demand only, and the docs now say so.

Issue #3505, opened 2026-07-18
ci-summary gate hangs on evaporated/duplicate check-runs (disabled-workflow + cancelled concurrency twins)


jeswr:
> 🤖 SPARQ agent

Live incident 2026-07-18 ~13:25-14:15: four ARMED PRs' `gate` runs sat in_progress 45+ min while EVERY sibling workflow was green. Two contributing shapes the aggregator's discovery doesn't handle: (1) duplicate triggers create concurrency-cancelled twin check-runs alongside the successful instance — the gate must dedupe by LATEST (or best) run per check name, never wait on a cancelled twin; (2) disabling a workflow (CodeQL, maintainer decision) evaporates its pending runs — a discovered check name that will never conclude must be timed-out/dropped from the awaited set rather than waited on forever. Unstick used: cancel+rerun the gate (fresh discovery sees only the live set). Fix: dedupe-by-latest in discovery + a no-progress timeout that re-snapshots the check-set and drops names with no live run.

PR #3787, opened 2026-07-25
fix(ci): a mid-gate re-draft must not re-dispatch draft-tier select onto a head with full-tier legs (#3781)

### LINKED ISSUE #3781: readied worker PRs are re-drafted before the gate concludes, so the gate can never succeed — deadlocks the whole backlog whenever review capacity is down

jeswr:

Fixes #3781.

## The defect, in one screenful of API output

`sparq-orchestrator[bot]` re-drafts a freshly-readied worker PR ~13 min after the ready (07:28:50 #3472, 07:28:54 #3468, 07:01:21 #3681) and flips `review:needs` in the same breath. That label event re-triggers `ci.yml` / `bench.yml` / `feature-matrix.yml` / `fuzz.yml`. The #2546 label-trigger guard does its job — every root job is skipped — so the run contributes nothing. Here is what each of #3472's **eight** label-flip runs actually contained (non-skipped jobs only):

```
30149378802 feature-matrix  select / select (change-based test selection, draft-tier)=success
30149378811 fuzz            select / select (change-based test selection, draft-tier)=success
30149378812 Benchmarks      select / select (change-based test selection, draft-tier)=success
30149379023 Benchmarks      select / select (change-based test selection, draft-tier)=success
30149379034 CI              select / select (change-based test selection, draft-tier)=success
30149379039 fuzz            select / select (change-based test selection, draft-tier)=success
30149379052 feature-matrix  select / select (change-based test selection, draft-tier)=success
30148989669 CI (07:15:37, the REAL ready_for_review run)  ... completed 07:34:08
```

One non-skipped job per run — the unconditional `select` pre-job — and because the PR was a draft again, each came out named **`…, draft-tier`**. Four draft-marked select instances (after newest-run resolution) now sat on a head whose full-tier successor **can never exist**: only a NON-draft `pull_request` payload produces a full-tier select. The gate burned all **155 poll attempts** on `awaiting the full-tier re-run (draft-tier selection present)` and then correctly fail-closed-refused.

**All three PRs had ZERO failing legs** — #3468: 422 runs / 0 failures; #3681: 196 / 0; #3472: `clippy (gate)` success, no failing leg at all. The only red was `gate`, and only because of the mixed leg set the race created. **Measured cost per occurrence: a full CI matrix wasted plus 155 wasted polls (~37 min) producing a guaranteed refusal — 3-for-3 in one drain pass.**

## Which option, and why

**Option (A)** — suppress the draft-tier `select` re-dispatch — *plus the run-level half that (A) alone gets wrong.*

**(B) is not implementable in this repo.** The re-draft actor is `sparq-orchestrator[bot]`, driven from the registry-side review pipeline; nothing in sparq converts a PR *to* draft (`scripts/auto-arm.py` and `scripts/rearm-sweeper.py` only ever mark drafts *ready*). "Defer the re-draft until the gate concludes" would have to land in the private orchestration repo, i.e. it cannot be shipped or tested here.

**(A) as literally worded has a hole, and it is a bad one.** "Suppress the draft-tier marker on the flip run" leaves the vacuous run *authoritative*: newest-run resolution (#3505) is per-workflow, so the flip run **becomes** the newest run for CI / Benchmarks / feature-matrix / fuzz and its all-`skipped` leg set **replaces** the real one. #3472's real CI matrix ran until **07:34:08**, six minutes *after* the flip runs completed (07:29:47) — so a name-only fix would have turned the 155-poll deadlock into a **GREEN gate rendered over legs that had not finished**. Worse, the same path erases a real `failure`.

So the fix keeps (A)'s framing — change no policy, no timing, only what the gate concludes from a run that assembled nothing — and applies it to the **run**, not just the name:

1. **`ci-select.yml`** names its job `…, no-leg` (never `…, draft-tier`) when the caller's event is a `labeled`/`unlabeled` `pull_request` whose label is not `ci-full`/`bench-full`/`fuzz-full` — exactly the condition under which every caller's root jobs are already `if:`-skipped. On an escape label at least one caller *does* real work, so no claim is made and the pre-#3781 behaviour stands unchanged (conservative; the review pipeline never flips those labels).
2. **The gate treats a no-leg run as NON-AUTHORITATIVE** — excluded from newest-run candidacy, so every check-run it produced (its select plus the whole skipped `needs:` graph) is a non-event and the **previous real run stays authoritative**. This is *strictly more fail-closed* than what it replaces: a mid-flight run keeps the gate polling, and a FAILED run keeps failing it, instead of being erased by a label flip. The exclusion is applied once, before anything downstream keys off `newest`, so the redispatch decision, the attempt-jobs inventory and check resolution cannot disagree about who is authoritative.
3. **Belts.** A no-leg select can neither create nor discharge a draft-tier hold, and can never forgive a cancelled real-tier select (`select_tier` is three-valued now). Only the **pure** ci-select pre-job may declare its run evidence-free — a compound `fv-select … + fv-manifest …` job carrying the marker fails CLOSED and keeps gating.

**Ask 2, the fail-fast half.** A full-tier PR gate whose siblings have **all** concluded, which still holds on a draft-marked select with no successor, and whose PR reads as **currently a draft**, is waiting for something that cannot happen. It now REDs immediately naming exactly that (with the remedy) instead of burning the remaining budget. It is the *same* refusal the budget-exhaustion belt reaches: **`return 0` count unchanged (5), only `return 1` added.** It stands down when the set is still settling, when the PR reads non-draft, or when the draft state is unreadable — that last one matters, because a false RED there would be a *new* failure mode whereas burning the budget is merely slow.

## Verification — 22 mutations of the LIVE files, zero survivors

Full harness output in the task report. The wiring is tested, not just the predicates: a ~120-line GitHub-expression evaluator **evaluates the live `name:` expression** against synthetic event payloads and feeds the rendered check-run name to the real gate predicates (YAML → name → verdict), and the "assembles zero legs" claim is proved by **`needs:`-graph reachability** over all four callers rather than by grep. Representative naming assertions:

| Mutation (of the live file) | Naming assertion that fired |
|---|---|
| gate call site drops `no_leg_ids=` | `a guarded label-flip no-op run must NOT supersede the real run and erase its FAILURE — the gate went GREEN over a vacuous all-skipped run (#3781)` |
| `newest_workflow_runs(authoritative)` → `(workflows)` | `[102] is not false : a cancelled label-flip NO-OP run must never be re-dispatched — it assembled no legs, so re-running it produces no evidence` |
| YAML drops the no-leg branch | `', no-leg' not found in 'select (change-based test selection, draft-tier)' : labeled review:needs on a draft PR must render the no-leg marker` |
| YAML drops the `!` before the escape-label list | `ci-full triggers real work — no no-leg claim allowed (got '… , no-leg')` |
| YAML marks every draft run no-leg | `a draft opened run assembles real (reduced) legs and MUST stay draft-tier-marked` |
| `is_no_leg_select` swallows draft-tier too | `a genuinely draft-tier run assembles REAL (reduced) legs and must stay authoritative — the fix must not blind draft-tier CI` |
| `ci.yml` drops its label guard | `ci.yml: ['build-archive', 'changes', 'conformance', …, 'wasm'] would still RUN on a non-escape label flip, so such a run does NOT assemble zero legs` |
| detector removed / never REDs / text gutted | `'UNSATISFIABLE draft-tier hold' not found in "attempt 1 … attempt 8 …"` (the run burns the whole budget again) |
| detector ignores `pending` | `a set with siblings STILL RUNNING is genuinely settling — the hold is a WAIT for those legs, so declaring it unsatisfiable is a false RED` |
| confirm window removed | `the confirm window exists so a successor that is merely LATE is not mistaken for one that can never come` |
| detector ignores the live draft state | `'UNSATISFIABLE' unexpectedly found …` (a readied PR / unreadable API state must fall back to the budget belt) |

Real tails, run the way CI does: `test_ci_summary_gate.py` **167 tests OK**, `test_ci_select_wiring.py` **74 OK**, `ci_summary_gate.py --self-test` OK, plus `test_ci_select.py` / `test_path_ownership.py` / `test_ci_selection_alarm.py` / `test_arm_capability_wiring.py` / `test_cone_coverage.py` / `test_feature_matrix_assemble.py` / `test_gates.py` OK, `actionlint` clean on the changed workflows, `check-no-perf-numbers --enforce` / `check-terminology` / `check-advisory-registry` clean.

## Notes for review

* **One dead branch removed rather than tested.** An explicit `run_id in no_leg_ids` early-return in `resolve_newest_workflow_runs` survived every mutant, because excluding the run from `newest` already drops its checks via `run_id != latest_id` or `latest is None`. A guard no input can distinguish is a guard no test can pin, so it is gone and the reasoning is in a comment.
* **Reasoned but not separately tested:** the same exclusion should also restore the `feature-matrix report` requirement after a label flip (the flip run's placeholder `opt-in group (…)` checks no longer win `fm_group_run_id`, so the CURRENT group run is the real one again). Same fail-closed direction; called out rather than claimed as verified.
* **Residual, deliberately not fixed:** on a `bench-full`/`fuzz-full` flip, the *other* callers' runs are no-ops but make no no-leg claim, so they still supersede — today's behaviour exactly, and not the reported defect.
* A sibling gate-budget fix is in flight on `fable/main-green-hang-detector-and-descoped-ec2-lane` (#3783). Both touch `scripts/ci_summary_gate.py`; this diff is scoped to the draft-tier / unsatisfiable-hold path, so **a trivial rebase may be needed** depending on merge order.
* **NOT armed.**

### DIFF
--- a/.github/workflows/ci-select.yml
+++ b/.github/workflows/ci-select.yml
@@ -49,6 +49,14 @@
 # detection contract with scripts/ci_summary_gate.py (SELECT_RE); the wiring test
 # scripts/tests/test_ci_select_wiring.py pins the two together. Do NOT add the words
 # "advisory"/"informational" to any name here (they would un-gate the check).
+#
+# THREE NAME SPELLINGS, ONE LEG IDENTITY. The job name carries the TIER as a
+# suffix marker: bare (full tier), ", draft-tier" (a draft-assembled reduced leg
+# set), or ", no-leg" (a guarded label-flip no-op run that assembled nothing at
+# all — [OPUS-5] #3781, see the job comment). scripts/ci_summary_gate.py strips
+# every marker to one normalized leg identity, so the three spellings are the
+# SAME leg across tiers; only the marker decides what the gate may conclude from
+# the run.
 name: ci-select
 
 on:
@@ -82,9 +90,30 @@ jobs:
     # one workflow's full-tier select must never release the hold for the
     # others), so a draft-tier leg set can never admit a non-draft PR to the
     # merge queue. The stable phrase "change-based test selection" (the SELECT_RE
-    # detection contract) is a prefix of BOTH spellings, and neither contains the
+    # detection contract) is a prefix of ALL THREE spellings, and none contains the
     # words "advisory"/"informational".
-    name: select (change-based test selection${{ github.event.pull_request.draft == true && ', draft-tier' || '' }})
+    #
+    # [OPUS-5] #3781 — THE ", no-leg" MARKER TAKES PRECEDENCE OVER ", draft-tier".
+    # A `labeled`/`unlabeled` pull_request event whose label is NOT one of
+    # ci-full / bench-full / fuzz-full is a GUARDED NO-OP for every caller: the
+    # #2546 label-trigger guard `if:`s off every root job of ci.yml, bench.yml,
+    # feature-matrix.yml and fuzz.yml, so the run assembles ZERO legs and THIS
+    # unconditional pre-job is its only non-skipped job (measured on sparq #3472:
+    # 8 label-flip runs at 07:28:57, each with exactly one non-skipped job — this
+    # select). Marking such a run ", draft-tier" (which is what happened, because
+    # the review pipeline re-drafts the PR in the same breath as the label flip)
+    # manufactured a hold NOTHING could ever discharge: a full-tier successor is
+    # only produced by a NON-draft payload, so the gate burned all 155 polls and
+    # fail-closed-refused a leg set with zero failing legs. The tier of a run that
+    # assembled no legs is not "draft" — it is NOT EVIDENCE AT ALL, and the gate
+    # ignores such a run entirely (scripts/ci_summary_gate.py NO_LEG_MARKER /
+    # no_leg_run_ids), which leaves the PREVIOUS real run authoritative instead of
+    # letting a vacuous all-skipped run supersede it. CONSERVATIVE BY DESIGN: on a
+    # ci-full/bench-full/fuzz-full flip at least one caller DOES do real work, so
+    # no marker is claimed and the pre-#3781 behaviour stands unchanged (a
+    # bench-full flip therefore still lets ci.yml's no-op run supersede — today's
+    # behaviour, not a regression; the review pipeline never flips those labels).
+    name: select (change-based test selection${{ (github.event_name == 'pull_request' && contains(fromJSON('["labeled","unlabeled"]'), github.event.action) && !contains(fromJSON('["ci-full","bench-full","fuzz-full"]'), github.event.label.name)) && ', no-leg' || (github.event.pull_request.draft == true && ', draft-tier' || '') }})
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/ci-select.yml
+++ b/.github/workflows/ci-select.yml
@@ -155,9 +155,16 @@ jobs:
           # the run sees which tier this selection was assembled for (the machine
           # contract is the job NAME marker above).
           IS_DRAFT_PR: ${{ github.event.pull_request.draft == true }}
+          # [OPUS-5] #3781: the same no-leg condition as the job name above, surfaced
+          # for the human reader. A no-leg run's step summary must NOT claim "Tier:
+          # draft" — it assembled nothing, and saying "draft" is exactly the confusion
+          # that cost a full drain pass to unpick.
+          IS_NO_LEG_RUN: ${{ github.event_name == 'pull_request' && contains(fromJSON('["labeled","unlabeled"]'), github.event.action) && !contains(fromJSON('["ci-full","bench-full","fuzz-full"]'), github.event.label.name) }}
         run: |
           set -euo pipefail
-          if [ "${IS_DRAFT_PR:-}" = "true" ]; then
+          if [ "${IS_NO_LEG_RUN:-}" = "true" ]; then
+            echo "**Tier: no-leg** — this run was started by a \`${EVENT}\` label flip that the label-trigger guard turns into a NO-OP: every leg is skipped and this selection pre-job is the run's only real job. The ci-summary gate therefore IGNORES this whole run and keeps the previous real run of this workflow authoritative (#3781)." >> "$GITHUB_STEP_SUMMARY"
+          elif [ "${IS_DRAFT_PR:-}" = "true" ]; then
             echo "**Tier: draft** — draft-tier CI (reduced sibling set; the full matrix re-runs at ready_for_review and its fresh gate supersedes this head's draft-tier results)." >> "$GITHUB_STEP_SUMMARY"
           fi
           args=(--event "$EVENT")
--- a/docs/branch-protection.md
+++ b/docs/branch-protection.md
@@ -223,7 +223,7 @@ no check-run there either — not a byte-identical copy of the PR check-set.)
 | `artifact-exact-equality` (wasm feature-OFF byte identity) | `vectorized-feature-off.yml` | kept iff `sparq-wasm` is in the affected closure (in-step `ci_select.py` verdict; ci-full label / selector error / full mode ⇒ run) |
 
 **The integrity invariant — a draft-tier gate result must NEVER admit a PR to the
-merge queue.** The load-bearing mechanism is rule 1 (structural); rules 2–6 are
+merge queue.** The load-bearing mechanism is rule 1 (structural); rules 2–8 are
 belts (all in `scripts/ci_summary_gate.py`, unit-tested in
 `scripts/tests/test_ci_summary_gate.py`; the name/trigger wiring pinned by
 `scripts/tests/test_ci_select_wiring.py`):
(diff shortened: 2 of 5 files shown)


jeswr:

**Live confirmation from this PR own CI:** the four `select` check-runs on this head render as the bare full-tier name —

```
select / select (change-based test selection)   pass  (ci.yml, feature-matrix.yml, bench.yml, fuzz.yml)
```

— i.e. the three-valued marker expression renders **byte-identically to the pre-#3781 spelling** on an ordinary non-draft `opened` PR event. The new branch only ever fires on a guarded label-flip.

Adjacent finding while working this, filed separately rather than folded in: #3788 — `vectorized-feature-off.yml` is the last lane in this trigger class with neither half of the #2546 fix (no label-trigger guard, and a shared `cancel-in-progress` group, so a `review:*` flip cancels its in-flight run and redoes the real wasm build). It played no part in #3781 (it emits no `select`), so it is deliberately out of scope here.

Not armed.

jeswr:

**Merged on `gate=success` (verified head-scoped on `25cdc413a`, newest-run resolution, zero gating failures, zero pending) under the maintainer's standing authorization.**

## The fix I proposed would have created a FALSE GREEN. This one does not.

I specified option (A) — "suppress the draft-tier `select` re-dispatch" — and said I leaned toward it because it changed no policy or timing. **As literally worded it has a hole that converts the deadlock into a silent false-pass**, in the check that authorises every merge:

newest-run resolution (#3505) is **per-workflow**, so the vacuous label-flip run *becomes* the authoritative run for CI / Benchmarks / feature-matrix / fuzz, and its all-`skipped` leg set **replaces the real one**. Measured on #3472: the real CI matrix ran until **07:34:08**, six minutes *after* the flip runs completed at **07:29:47**. Neutralising only the marker would have greened the gate over legs that had not finished — **and the same path erases a real `failure`.**

I also offered option (B) as the alternative. **It is not implementable in this repo**: the re-draft actor is `sparq-orchestrator[bot]`, driven from the registry-side review pipeline, and nothing in sparq converts a PR *to* draft (`auto-arm.py` and `rearm-sweeper.py` only ever mark drafts *ready*). So I offered a choice between a dangerous option and a nonexistent one, and the agent was right to reject both framings and fix the actual mechanism.

**What shipped instead:** `ci-select.yml` emits `…, no-leg` (never `…, draft-tier`) on a guarded label flip, and the gate treats such a run as **non-authoritative** — excluded from newest-run candidacy *before* anything keys off `newest` (redispatch, attempt-jobs, resolution), so the previous **real** run stays authoritative. That is strictly **more** fail-closed than what it replaces, which is the opposite direction from what I proposed. `return 0` count unchanged (5); only `return 1` added.

Empirically grounded, not inferred: all **8** of #3472's label-flip runs contained **exactly one** non-skipped job — the `select`, named `, draft-tier`.

## Verification worth noting

23 mutations of the **live** files, **zero survivors**, each naming the defect — including both halves of every discrimination (a genuinely draft-tier run must stay authoritative; a genuinely settling set must not be declared unsatisfiable). The workflow expression is **evaluated**, not string-matched — a GitHub-expression evaluator renders the live `name:` against synthetic payloads and feeds the result to the real gate predicates — and the "assembles zero legs" claim is proved by **`needs:`-graph reachability** across all four callers.

And the detail I most respect: it wrote an explicit `run_id in no_leg_ids` branch, found it **survived every mutant**, recognised it as dead code (the exclusion already drops those checks via `run_id != latest_id`), and **removed it** rather than keeping it pinned by a test that cannot fail. Deleting your own unfalsifiable code is harder than adding a test for it.

## Two honest caveats carried forward

1. **This is a deliberate behaviour change.** Pre-fix, a head that had been label-flipped mid-gate could *never* earn a green required `gate`; now it can — on a full matrix that genuinely ran on that exact SHA. That is the intended outcome, but it is a change in what the gate will admit, not merely a bug fix.
2. **#3472/#3468/#3681 are not retroactively fixed.** Their heads already carry `, draft-tier`-named selects from pre-fix flip runs, and name matching is how the gate recognises a no-leg run — so a bare re-run there now fails **fast** with a readable diagnosis rather than green. The working sequence is: merge this → `gh pr update-branch` → `gh pr ready`. No window-racing required. But they still carry `review:needs`, a hard exclusion in `auto-arm.py`, so **the gate deadlock is fixed and the review-capacity block is not** — that still needs registry #596 or a maintainer.

Adjacent finding filed rather than folded in: **#3788** — `vectorized-feature-off.yml` is the last lane in this trigger class with **neither** half of the #2546 fix, so every `review:*` flip cancels its in-flight run on the same head SHA and redoes the real wasm feature-OFF build. It emits no `select`, so it played no part in #3781.

Issue #3720, opened 2026-07-22
[dedup] Add local composite actions .github/actions/setup-rust and .github/actions/setup-playwright so each pinned setup SHA lives in exactly one reviewed file (71-copy toolchain preamble, 5-copy Playwright preamble)


jeswr:
> 🤖 SPARQ agent

Part of the maintainability program (see umbrella issue; multi-agent survey + independent verification, 2026-07-22).

## Evidence
There is no .github/actions/ directory; every workflow inlines its setup. dtolnay/rust-toolchain@<pinned-sha> appears 73 times across .github/workflows/*.yml, paired with Swatinem/rust-cache in 20+ jobs of ci.yml alone plus asan/bench/differential/feature-matrix/fuzz/kani/metamorph/miri/python/release/shacl-diff-fuzz/zk-toolchain and more (verified by grep counts). The Playwright preamble (npm ci from root lock → 'Resolve Playwright version' → 'Cache Playwright browsers' → install Chromium + OS deps) is copied across site-e2e.yml, site-e2e-foundation.yml (steps at lines 63-91), site-e2e-hero.yml, and again in nightly-full-sweep.yml's site sweeps. Each pin bump or cache-key fix must currently be replicated N times (the pinned-SHA supply-chain posture makes drift between copies a real hazard rather than a cosmetic one).

## Proposal
[...]

PR #3803, opened 2026-07-25
ci: pin the Rust toolchain to 1.97.1 (rust-toolchain.toml) and reconcile every lane that needs a different one

jeswr:
> 🤖 **SPARQ agent** — CI/supply-chain lane, running on Claude Opus 5. Implements registry issue [jeswr/agent-account-registry#664](https://github.com/jeswr/agent-account-registry/issues/664) under the maintainer decision of 2026-07-25: **pin the toolchain now, sweep the lint later.** No `crates/` source is touched.

## The problem

Nothing in sparq pinned a Rust version. The ~40 `rustup default stable` sites and every `dtolnay/rust-toolchain` use without a `toolchain:` input both mean *"whatever Rust this machine happens to carry"* — not *"a specific compiler"*. The registry's review-fix gate (`review-fix.yml`, step *Ensure a Rust toolchain for the crate-scoped gate*) inherits the same property and then clippies **this** repository with `-D warnings`.

So two independent consumers linted identical source on different compilers, and `-D warnings` turns any lint-set difference into a hard failure on unchanged code.

### Measured

| Consumer | Compiler drawn | Result |
| --- | --- | --- |
| sparq CI `lint` job | `rustc 1.97.1 (8bab26f4f 2026-07-14)` | green |
| registry review-fix gate | `rustc 1.88.0 (6b00bc388 2025-06-23)` | fails on `crates/sparq-core/src/nt.rs:411` |

`nt.rs:411` is untouched by the PR under test and unchanged on `main`. Reproduced the mechanism directly on a small crate carrying that exact shape (plus a field-access and a two-argument `format!`), plain `cargo clippy -- -D warnings`:

```
clippy 1.88.0  -> FAIL   error: variables can be used directly in the `format!` string
                         = note: `-D clippy::uninlined-format-args` implied by `-D warnings`
clippy 1.89.0  -> pass
clippy 1.97.0  -> pass
clippy 1.97.1  -> pass   (the pin)
```

## Two corrections to the diagnosis in #664

Both from evidence, and both change what the fix defends against.

**1. The intermittency is not a coin flip.** #664 reads "1 of 4 runs succeeded on identical source" as the signature of a heterogeneous runner fleet. It is not: the run recorded as a success (`review-fix fix sparq-org/sparq#3436`, 17:27Z) **never executed the clippy gate at all** — that step is conditional on `fix_made_changes == 'true'`, and its log contains no toolchain step and no gate step. Every run that actually reached the gate failed, deterministically, on 1.88.0.

**2. The strict clippy is the OLD one.** #664 warns that sparq `main` is "green only because the runners it has drawn carry the permissive clippy", implying a newer image would break it. The measurement runs the other way: 1.97.1 is the permissive one, 1.88.0 is strict. The demonstrated hazard is a **lagging** toolchain.

That also falsifies open issue **#3575**, which asserts `uninlined_format_args` is warn-by-default on stable 1.97.1 and predicts a repo-wide `-D warnings` failure "once the ubuntu-latest runner's preinstalled stable bumps to 1.97+". Runners are on 1.97.1 *now*, the workspace lane is green, and direct measurement agrees — on 1.97.1 the lint fires only under an explicit `-W clippy::uninlined_format_args`. Commented there rather than editing that issue.

## The fix, and the part that was easy to get wrong

`rust-toolchain.toml` at the repo root pinning **1.97.1**, with `profile = "minimal"`, `components = ["clippy", "rustfmt"]` and `targets = ["wasm32-unknown-unknown"]`. rustup honours it for every consumer inside the checkout — sparq CI, the registry gate (whose step `cd`s into the target checkout precisely so a target pin is honoured), and local builds.

But a root pin **outranks `rustup default`**, so it silently hijacks every lane that deliberately uses a different compiler. Verified precedence, empirically, in this worktree:

```
1  cargo +<toolchain>          2  RUSTUP_TOOLCHAIN
3  rustup override             4  rust-toolchain.toml        5  rustup default
```

Every lane was audited against that:

| Lane | Would the pin have broken it? | Fix |
| --- | --- | --- |
| `ci.yml` **msrv** (gating) | **Yes — silently vacuous.** `rustup default 1.88` loses to the file, so the floor check would compile on pinned stable again — the exact regression PR #3511 finding 4 fixed. | job-level `RUSTUP_TOOLCHAIN: "1.88"` |
| `formal-verification.yml` **kani** (gating) | **Yes.** Every cargo call is plain (`cargo install kani-verifier`, `cargo kani setup`, `cargo kani --harness`) — all would move onto pinned stable. | job-level `RUSTUP_TOOLCHAIN: nightly-2026-06-11` |
| `kani.yml` (informational) | **Yes**, same shape. | same |
| `miri.yml` x2, `asan.yml`, `fuzz.yml` | Partly. Their `cargo +nightly-2026-06-11` calls are precedence 1 and were already immune, but plain cargo / the `rustc -vV` that rust-cache keys on / `cargo fuzz`'s nested cargo invocation were not. | same env var, restoring pre-pin behaviour |
| ~19 wasm32 jobs across 11 workflows | **Yes.** `dtolnay/rust-toolchain … targets: wasm32-unknown-unknown` installs the target onto the `stable` toolchain *it* manages; the pin outranks it, so those builds would fail with no wasm32 `std`. | `targets` declared in the toolchain file |
| 4 coverage jobs in `ci.yml` | **Yes.** Same shape for `components: llvm-tools-preview`; `cargo llvm-cov` would find no llvm-tools. | explicit `rustup component add llvm-tools-preview` per job |
| `build-matrix.yml` | No. Its `rustup target add` applies to the **active** toolchain, which is now the pinned one. | none |

The ~40 `rustup default stable` lines are left alone: under the pin they are no-ops for *selection* but still usefully add components, and rewriting them all would be large churn colliding with open PRs. (Their comments claiming "never trigger a toolchain download" are now inaccurate — real doc drift, but it overlaps existing issue #3720's 71-copy toolchain-preamble dedup, so I did not file a duplicate.)

## Verified vs not verified

**Verified locally** (work box, `aarch64-unknown-linux-gnu`):

- `rust-toolchain.toml` parses; `rustup show active-toolchain` in the worktree reports `1.97.1-… (overridden by '<repo>/rust-toolchain.toml')`; rustc/cargo/clippy all report 1.97.1.
- Precedence, all three directions: the file overrides `rustup default`; `RUSTUP_TOOLCHAIN=1.88` overrides the file; `+<toolchain>` overrides the file.
- rustup **does** install a toolchain-file `targets` entry on demand — observed it fetch wasm32 `rust-std` onto an already-installed toolchain.
- `rustup component add llvm-tools-preview` resolves onto the pinned toolchain.
- `cargo clippy -p sparq-core --all-targets -- -D warnings` → **exit 0** under the pin. That is the exact crate and lint the registry gate died on.
- `cargo clippy --workspace --all-targets -- -D warnings` → **exit 0, zero warnings**, the full workspace clean on 1.97.1 (13m05s cold on this box).
- All 63 workflow files parse as YAML; `actionlint` clean on all six changed files.
- `scripts/check-privacy-claims.sh` passes.

**Not verified — stated plainly:**

- This box is **aarch64**; CI is **x86_64**. Clippy lint sets are not target-specific, but no local run proves the CI host.
- The workspace clippy above was **default features only**. The gating `--all-features` sibling step was not run locally (RAM).
- The nightly lanes (**miri / asan / fuzz / kani / formal-verification**) were **not executed here** — Kani/CBMC and `-Zbuild-std` sanitizer builds are not runnable on this box in reasonable time. Their `RUSTUP_TOOLCHAIN` escapes follow the precedence rule verified above and are built to preserve each lane's pre-pin resolution exactly, but they are **documented-untested; validate on the first CI run of each lane.** `miri` / `asan` / `kani` are scheduled lanes, so their first real exercise may not be on this PR.
- No wasm32 crate was compiled end-to-end here; only the presence of the target on the pinned toolchain was verified.
- **CI cost, not measured.** The pin partly gives up the repo's explicit use-the-preinstalled-toolchain economy directive: the image installs its Rust under the name `stable`, and `1.97.1` is a separate rustup toolchain directory even when byte-identical, so Rust jobs now fetch it once. `profile = "minimal"` plus an explicit component/target list keeps that as small as possible. An accepted tradeoff, not a measured one.

No tests were added: this change has no runtime behaviour to assert, and a test that only re-stated the file's contents would be vacuous. The claims above rest on the reproductions listed, not on a new assertion.

## Gate impact

No job was renamed, added or removed — only job-level `env:` blocks and one new step in four coverage jobs. The `ci-summary / gate` aggregator is unaffected; it resolves advisory status from `.github/advisory-registry.json`, which this PR does not touch. The path filter is preserved: no `changes` / `rust_changed` condition was altered.

## Follow-up (deliberately not done here)

Per the maintainer decision the 666-site `uninlined_format_args` sweep is **not** attempted. The exit path from this pin is documented in the file itself: either sweep the lint or declare the project's position via `[workspace.lints.clippy]` — noting the convention conflict in #3575, where positional args are mandated elsewhere to dodge a CodeQL false positive — and only then bump `channel` as its own reviewed change.

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -102,6 +102,18 @@ jobs:
     # then run instrumented. Generous ceiling so a hung run fails with a clear timeout rather
     # than running away (matches miri.yml's posture).
     timeout-minutes: 90
+    # [OPUS-5] PINNED-TOOLCHAIN ESCAPE. The repo root carries a `rust-toolchain.toml`
+    # pinning STABLE, and rustup resolves that file (precedence 4) ABOVE the `rustup
+    # default <toolchain>` that `dtolnay/rust-toolchain` performs (precedence 5). The
+    # `cargo +nightly-2026-06-11 test -Zbuild-std …` calls below are precedence 1 and were
+    # already immune; this env var (precedence 2) restores the pre-pin behaviour for every
+    # other cargo/rustc invocation in the job, including the `rustc -vV` that
+    # Swatinem/rust-cache keys its cache on. `-Zbuild-std` requires nightly, so a silent
+    # fallback to stable here would be a hard failure rather than a quiet one — but it
+    # would still be caused by the pin, not by the sanitizer. Keep in sync with the
+    # `toolchain:` input below.
+    env:
+      RUSTUP_TOOLCHAIN: nightly-2026-06-11
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       # [OPUS-4.8] sq-czlh: reclaim ~20-30 GB of pre-installed SDKs/toolchains BEFORE the
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1309,7 +1309,48 @@ jobs:
       - name: cargo check on the declared MSRV
         # [SONNET-4.6] Test targets must compile at the declared floor too; they can
         # otherwise hide newer temporary-lifetime-extension behavior from this gate.
-        run: cargo check --workspace --all-targets --exclude sparq-py --exclude sparq-hdt
+        #
+        # [OPUS-5] NON-VACUITY ASSERTION, deliberately in the SAME `run:` block as the
+        # compile it guards. This lane's entire meaning is the VERSION, so "which
+        # compiler ran" must be a CHECKED fact, not a logged one. Before this, the only
+        # thing standing between the gate and silent vacuity was the job-level
+        # `RUSTUP_TOOLCHAIN` above: delete that one line and the root
+        # `rust-toolchain.toml` (precedence 4) outranks the `rustup default 1.88`
+        # (precedence 5) in the step above, the lane compiles the workspace on pinned
+        # stable, `rustup show active-toolchain` dutifully PRINTS the wrong answer, and
+        # CI stays GREEN — re-arming exactly the vacuity PR #3511 finding 4 fixed.
+        #
+        # Two design choices, both load-bearing:
+        #  * The floor is DERIVED from the root `Cargo.toml` `rust-version` rather than
+        #    hardcoded a fourth time, so this also gates the "keep this == root
+        #    rust-version" invariant that three comments in this job ask humans to
+        #    maintain by hand and nothing enforced. Drift in EITHER direction reds.
+        #  * `cargo` is asserted as well as `rustc`: `cargo check` is what actually
+        #    resolves the `rust-version` floor, and asserting only `rustc` would miss a
+        #    split resolution between the two proxies.
+        # Fail-closed throughout: an unreadable/renamed `rust-version`, a missing
+        # toolchain, or a `--version` that does not parse all exit non-zero.
+        # Same-block placement means no `if:`/step boundary can skip the assertion
+        # while still running the compile.
+        run: |
+          floor="$(sed -n 's/^[[:space:]]*rust-version[[:space:]]*=[[:space:]]*"\([0-9][0-9]*\.[0-9][0-9]*\).*/\1/p' Cargo.toml | head -n1)"
+          if [ -z "$floor" ]; then
+            echo "::error::could not read [workspace.package] rust-version from the root Cargo.toml"
+            exit 1
+          fi
+          echo "declared MSRV floor (root Cargo.toml rust-version) = $floor"
+          for tool in rustc cargo; do
+            ver="$("$tool" --version)" || { echo "::error::$tool --version failed"; exit 1; }
+            echo "$tool --version -> $ver"
+            case "$ver" in
+              "$tool $floor."*|"$tool $floor "*) ;;
+              *)
+                echo "::error::MSRV gate is NOT on the declared floor $floor — $tool reports '$ver'. This lane must compile ON the floor toolchain or it proves nothing; check the job-level RUSTUP_TOOLCHAIN and the root rust-toolchain.toml pin."
+                exit 1
+                ;;
+            esac
+          done
+          cargo check --workspace --all-targets --exclude sparq-py --exclude sparq-hdt
 
   conformance:
     # W3C SPARQL conformance RATCHET (roadmap T13/38). The full 1229-test scope is at
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1269,6 +1269,20 @@ jobs:
       # ld is fine here: this lane is `cargo check` (only build scripts and
       # proc-macros link) and its meaning is the VERSION floor, not link speed.
       CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS: ""
+      # [OPUS-5] KEEPS THIS GATE NON-VACUOUS UNDER THE ROOT `rust-toolchain.toml`
+      # PIN. rustup resolves the toolchain file (precedence 4) ABOVE `rustup
+      # default` (precedence 5), so the `rustup default 1.88` below — which was
+      # the whole fix for the vacuity described in the step comment — would be
+      # SILENTLY OVERRIDDEN by the root pin and this lane would go back to
+      # compiling on pinned stable, i.e. vacuous again in exactly the way PR
+      # #3511 finding 4 called out. `RUSTUP_TOOLCHAIN` is precedence 2, so it
+      # beats the file; setting it at JOB level (rather than `cargo +1.88` at
+      # one call site) makes every cargo/rustc in the job — including anything
+      # nested — use the floor, and makes `rustup show active-toolchain` print
+      # "overridden by environment variable RUSTUP_TOOLCHAIN" so the log states
+      # which compiler actually ran. Keep this == the pin below == root
+      # Cargo.toml `rust-version`.
+      RUSTUP_TOOLCHAIN: "1.88"
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - name: Rust (pinned MSRV toolchain 1.88)
@@ -2836,6 +2850,15 @@ jobs:
       - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
         with:
           components: llvm-tools-preview
+      # [OPUS-5] REQUIRED once the root `rust-toolchain.toml` pin exists. The action
+      # above installs `llvm-tools-preview` onto the toolchain IT manages (`stable`)
+      # and then calls `rustup default` — precedence 5, which the toolchain file
+      # (precedence 4) outranks. So cargo here resolves to the PINNED toolchain, on
+      # which the component would be absent, and `cargo llvm-cov` fails with the
+      # llvm-tools missing rather than measuring anything. This adds the component to
+      # the ACTIVE (pinned) toolchain. Idempotent, and a no-op if it is already there.
+      - name: Ensure llvm-tools-preview on the pinned toolchain
+        run: rustup component add llvm-tools-preview
       # Per-shard cache key: each shard builds a DIFFERENT crate subset (different
       # instrumented objects), so keying the cache by shard keeps them from clobbering /
       # thrashing one another's saved target dir.
@@ -2985,6 +3008,15 @@ jobs:
       - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
         with:
           components: llvm-tools-preview
+      # [OPUS-5] REQUIRED once the root `rust-toolchain.toml` pin exists. The action
+      # above installs `llvm-tools-preview` onto the toolchain IT manages (`stable`)
+      # and then calls `rustup default` — precedence 5, which the toolchain file
+      # (precedence 4) outranks. So cargo here resolves to the PINNED toolchain, on
+      # which the component would be absent, and `cargo llvm-cov` fails with the
+      # llvm-tools missing rather than measuring anything. This adds the component to
+      # the ACTIVE (pinned) toolchain. Idempotent, and a no-op if it is already there.
+      - name: Ensure llvm-tools-preview on the pinned toolchain
+        run: rustup component add llvm-tools-preview
       # Per-partition dep cache: each partition compiles the same instrumented engine, so a
       # shared-key cache warms the deps; keying by part avoids save-race clobbering.
       - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
@@ -3036,6 +3068,15 @@ jobs:
       - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
         with:
           components: llvm-tools-preview
+      # [OPUS-5] REQUIRED once the root `rust-toolchain.toml` pin exists. The action
+      # above installs `llvm-tools-preview` onto the toolchain IT manages (`stable`)
+      # and then calls `rustup default` — precedence 5, which the toolchain file
+      # (precedence 4) outranks. So cargo here resolves to the PINNED toolchain, on
+      # which the component would be absent, and `cargo llvm-cov` fails with the
+      # llvm-tools missing rather than measuring anything. This adds the component to
+      # the ACTIVE (pinned) toolchain. Idempotent, and a no-op if it is already there.
+      - name: Ensure llvm-tools-preview on the pinned toolchain
+        run: rustup component add llvm-tools-preview
       # Reuse a run partition's dep cache to warm the merge job's object compile.
       - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
         with:
@@ -3246,6 +3287,15 @@ jobs:
       - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
         with:
           components: llvm-tools-preview
+      # [OPUS-5] REQUIRED once the root `rust-toolchain.toml` pin exists. The action
+      # above installs `llvm-tools-preview` onto the toolchain IT manages (`stable`)
+      # and then calls `rustup default` — precedence 5, which the toolchain file
+      # (precedence 4) outranks. So cargo here resolves to the PINNED toolchain, on
+      # which the component would be absent, and `cargo llvm-cov` fails with the
+      # llvm-tools missing rather than measuring anything. This adds the component to
+      # the ACTIVE (pinned) toolchain. Idempotent, and a no-op if it is already there.
+      - name: Ensure llvm-tools-preview on the pinned toolchain
+        run: rustup component add llvm-tools-preview
       - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
       - name: Install cargo-llvm-cov
         # [OPUS-4.8] sq-tool-pin: explicit `with: tool:` is MANDATORY when SHA-pinning
--- a/.github/workflows/formal-verification.yml
+++ b/.github/workflows/formal-verification.yml
@@ -170,6 +170,18 @@ jobs:
       fail-fast: false
       matrix:
         suite: ${{ fromJSON(needs.fv-select.outputs.suites) }}
+    # [OPUS-5] PINNED-TOOLCHAIN ESCAPE — REQUIRED here, not belt-and-braces, and this lane
+    # GATES the merge. The repo root carries a `rust-toolchain.toml` pinning STABLE, and
+    # rustup resolves that file (precedence 4) ABOVE the `rustup default <toolchain>` that
+    # `dtolnay/rust-toolchain` performs (precedence 5). Every cargo call in this job
+    # (`cargo install --locked kani-verifier`, `cargo kani setup`, `cargo kani --harness`)
+    # is a plain invocation with no `+<toolchain>`, so without this env var they would all
+    # silently move off the declared nightly onto pinned stable. `RUSTUP_TOOLCHAIN` is
+    # precedence 2 and beats the file, preserving exactly the toolchain this lane resolved
+    # to before the pin existed. Keep in sync with the `toolchain:` input below and with
+    # kani.yml's identical escape.
+    env:
+      RUSTUP_TOOLCHAIN: nightly-2026-06-11
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -197,9 +197,17 @@ jobs:
     # default <toolchain>` that `dtolnay/rust-toolchain` performs (precedence 5). The
     # `cargo +nightly-2026-06-11 fuzz …` calls below are precedence 1 and were already
     # immune; this env var (precedence 2) restores the pre-pin behaviour for the rest of
-    # the job. It matters most for `cargo fuzz`, which re-invokes cargo itself for the
-    # `fuzz/` sub-workspace: that nested invocation would otherwise resolve the ROOT pin
-    # (rustup searches upward from the cwd) and lose the nightly the lane declares. Note
+    # the job. [OPUS-5] CORRECTION, measured on rustup 1.29.0: this env var is NOT what
+    # protects the nested cargo that `cargo fuzz` spawns for the `fuzz/` sub-workspace.
+    # rustup EXPORTS the resolved toolchain to child processes — `cargo +nightly-2026-06-11
+    # <subcmd>` hands the child `RUSTUP_TOOLCHAIN=nightly-2026-06-11-<host>` (and
+    # `RUSTUP_TOOLCHAIN_SOURCE=cli`), which is precedence 2 and so still outranks the root
+    # file even when the nested cargo runs from `fuzz/`. Verified directly: with a root pin
+    # in place, a `cargo +nightly-2026-06-11` subcommand invoked from a sub-directory saw
+    # the nightly, while the same subcommand invoked as a PLAIN `cargo` saw the pin. So the
+    # nested call was already safe; what this env var actually buys is the invocations that
+    # carry NO `+toolchain` at all — `cargo install cargo-fuzz` and the `rustc -vV` that
+    # Swatinem/rust-cache keys its cache on, which would otherwise re-key onto the pin. Note
     # this env is scoped to THIS job only — the `differential-smoke` job below is a
     # deliberate stable lane and correctly inherits the pin. Keep in sync with the
     # `toolchain:` input below.
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -192,6 +192,19 @@ jobs:
     # Generous ceiling above the worst case (nightly: 4 targets × 600s + build) so a
     # hung target fails with a clear timeout rather than running away.
     timeout-minutes: 60
+    # [OPUS-5] PINNED-TOOLCHAIN ESCAPE. The repo root carries a `rust-toolchain.toml`
+    # pinning STABLE, and rustup resolves that file (precedence 4) ABOVE the `rustup
+    # default <toolchain>` that `dtolnay/rust-toolchain` performs (precedence 5). The
+    # `cargo +nightly-2026-06-11 fuzz …` calls below are precedence 1 and were already
+    # immune; this env var (precedence 2) restores the pre-pin behaviour for the rest of
+    # the job. It matters most for `cargo fuzz`, which re-invokes cargo itself for the
+    # `fuzz/` sub-workspace: that nested invocation would otherwise resolve the ROOT pin
+    # (rustup searches upward from the cwd) and lose the nightly the lane declares. Note
+    # this env is scoped to THIS job only — the `differential-smoke` job below is a
+    # deliberate stable lane and correctly inherits the pin. Keep in sync with the
+    # `toolchain:` input below.
+    env:
+      RUSTUP_TOOLCHAIN: nightly-2026-06-11
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       # [OPUS-4.8] sq-czlh: reclaim ~20-30 GB of pre-installed SDKs/toolchains BEFORE the
--- a/.github/workflows/kani.yml
+++ b/.github/workflows/kani.yml
@@ -264,6 +264,19 @@ jobs:
             features: ""
             harnesses: >-
               domain_corpus_exhibits_a_surviving_escalation
+    # [OPUS-5] PINNED-TOOLCHAIN ESCAPE — REQUIRED here, not belt-and-braces. The repo root
+    # carries a `rust-toolchain.toml` pinning STABLE, and rustup resolves that file
+    # (precedence 4) ABOVE the `rustup default <toolchain>` that `dtolnay/rust-toolchain`
+    # performs (precedence 5). Unlike the miri/asan/fuzz lanes, NOTHING in this job uses
+    # `cargo +<toolchain>`: `cargo install --locked kani-verifier`, `cargo kani setup` and
+    # every `cargo kani --harness` invocation are plain cargo calls, so without this env
+    # var they would all silently move from the declared nightly onto pinned stable —
+    # changing which compiler builds the driver and which host toolchain Kani's codegen
+    # sits on. `RUSTUP_TOOLCHAIN` is precedence 2 and beats the file, preserving exactly
+    # the toolchain this lane resolved to before the pin existed. Keep in sync with the
+    # `toolchain:` input below.
+    env:
+      RUSTUP_TOOLCHAIN: nightly-2026-06-11
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       # Kani pins its own internal toolchain via `cargo kani setup`, but it still needs a host
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -164,6 +164,17 @@ jobs:
       fail-fast: false
       matrix:
         shard: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]
+    # [OPUS-5] PINNED-TOOLCHAIN ESCAPE. The repo root carries a `rust-toolchain.toml`
+    # pinning STABLE. rustup resolves that file (precedence 4) ABOVE the `rustup default
+    # <toolchain>` that `dtolnay/rust-toolchain` performs (precedence 5), so with the pin
+    # in place a plain `cargo`/`rustc` in this job resolves to pinned stable, NOT the
+    # nightly this lane declares. The `cargo +nightly-2026-06-11 miri …` calls below are
+    # precedence 1 and were already immune; this env var (precedence 2) restores the
+    # pre-pin behaviour for everything else in the job — notably the `rustc -vV` that
+    # Swatinem/rust-cache derives its cache key from, which would otherwise silently
+    # re-key the Miri cache onto stable. Keep in sync with the `toolchain:` input below.
+    env:
+      RUSTUP_TOOLCHAIN: nightly-2026-06-11
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
@@ -244,6 +255,9 @@ jobs:
     name: miri doctests (sparq-core)
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    # [OPUS-5] PINNED-TOOLCHAIN ESCAPE — see the identical note on the `miri` job above.
+    env:
+      RUSTUP_TOOLCHAIN: nightly-2026-06-11
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -80,9 +80,13 @@
 #   3. a `rustup override set` directory override
 #   4. THIS FILE
 #   5. `rustup default`
-# So this file BEATS every `rustup default stable` in the workflow tree (those
-# ~40 lines are now no-ops for toolchain selection; they still usefully add
-# components), and is in turn BEATEN by `+toolchain` and `RUSTUP_TOOLCHAIN`.
+# So this file BEATS every `rustup default stable` in the workflow tree — all 40
+# of those lines are now COMPLETE no-ops, and is in turn BEATEN by `+toolchain`
+# and `RUSTUP_TOOLCHAIN`.
+# [OPUS-5] They add no components either: `rustup default stable` only selects.
+# Measured: of the 40 `run:` blocks containing `rustup default stable`, 7 also
+# contain a separate `rustup component add` — it is THAT command (which applies
+# to the active, i.e. pinned, toolchain) that adds anything, not the `default`.
 # The lanes that deliberately need a different compiler — the MSRV floor check
 # and the nightly Miri/ASan/fuzz/Kani lanes — assert themselves through those
 # two higher-precedence mechanisms, so they keep the compiler they declare.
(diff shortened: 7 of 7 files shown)


jeswr:
> 🤖 **SPARQ agent** — closing out the "documented-untested" caveats in the PR body now that the first CI run has completed.

**`gate: SUCCESS` — 269 checks green, 5 skipped, 0 failures, `mergeStateStatus: CLEAN`.**

The body listed several lanes as *"documented-untested; validate on the first CI run"*. That run has now happened, so here is what it actually proved. In each case I read the compiler identity **out of the job log** rather than inferring it from a green tick — a green MSRV job is exactly what a vacuous MSRV job looks like, so a passing check is not by itself evidence the escape worked.

| Mechanism | Job | Compiler in the log | Was it at risk? |
| --- | --- | --- | --- |
| toolchain file (prec. 4) beats `rustup default` (prec. 5) | `clippy (gate) + fmt` | `1.97.1-x86_64-… (overridden by '/home/runner/work/sparq/…')` | this is the pin doing its job |
| `RUSTUP_TOOLCHAIN` (prec. 2) beats the file — **stable floor** | `MSRV check (Rust 1.88, declared floor)` | `1.88-x86_64-… (overridden by environment variable RUSTUP_TOOLCHAIN)` → `rustc 1.88.0 (6b00bc388 2025-06-23)` | **yes — would have been silently vacuous** |
| `RUSTUP_TOOLCHAIN` beats the file — **nightly** | `cargo-fuzz (parsers + mmap loader, bounded)` | `nightly-2026-06-11-x86_64-… (overridden by environment variable RUSTUP_TOOLCHAIN)` → `rustc 1.98.0-nightly (485ec3fbc 2026-06-10)` | **yes** |
| same escape, merge-**gating** lane | `kani …(change-coupled proofs)` ×4 — `sparq-core`, `sparq-substrate`, `sparq-engine`, `sparq-wac-oracle` | all green | **yes** |
| toolchain-file `targets` materialise on the pin | `wasm build (5 wasm crates)`, `artifact-exact-equality (wasm bundle feature-OFF)`, `opt-in sparq-lws-core (wasm)`, `opt-in sparq-lws-wasm (sparql-endpoint)` | all green | **yes — would have been "can't find crate for `core`"** |
| `rustup component add llvm-tools-preview` on the pin | `coverage engine run 1–3/3`, `coverage engine merge + ratchet`, `coverage ratchet shard 1–3/3` | all green | **yes — `cargo llvm-cov` would have found no llvm-tools** |

The `cargo-fuzz` result is the strongest of these: `cargo fuzz` needs nightly `-Z` flags and re-invokes cargo from inside `fuzz/`, where rustup's upward directory search would have found the root pin and handed it stable. It would have failed outright, not degraded quietly.

## What is still NOT verified

Being explicit so the caveat is not quietly dropped:

- **`miri.yml` and `asan.yml` did not run** — both are `schedule`/`workflow_dispatch` lanes with no `pull_request` trigger, so this PR cannot exercise them. Their escapes are the *same* env-var mechanism proven three times above and their `cargo +nightly-2026-06-11` calls were already precedence-1 immune, but that is an argument, not a run. **Validate on their next scheduled run.**
- **The nightly full-tier coverage job** (`coverage (nightly, full incl. heavy vectors)`) is `nightly-gate`d and shows as `skipped` here — it is the fourth `llvm-tools-preview` site and the only one not confirmed. The other three site families are green.
- **`build-matrix.yml` did not run** on this PR. I argued it needs no change (its `rustup target add` applies to the active, now-pinned toolchain); unexercised.
- **CI cost of the pin remains unmeasured.** Jobs now fetch `1.97.1` because the runner image installs its Rust under the name `stable`. Still an accepted tradeoff, still not a number.

## Not armed

Deliberately not setting auto-merge. This changes toolchain selection for two merge-gating lanes (`msrv`, `formal-verification`), and the two lanes it touches that CI cannot exercise are exactly the memory-safety ones. Arming is the orchestrator's verify→arm call, not mine.

jeswr:
> 🤖 **SPARQ agent**

Head SHA reviewed: **`eea3fa16bf1bb1296fa6f45806c5f66382983c60`**

The question I was asked to answer is whether **every** escape is complete — a missed one is either a silent vacuity or a hard failure. I re-derived the enumeration independently rather than reading the PR's table, then checked the mechanism against real rustup and against this PR's own CI logs.

---

## 1. Escape completeness — independently re-enumerated, COMPLETE

I enumerated the toolchain-selection surface across all 63 workflows from scratch:

| Class | Sites found | Escaped? |
| --- | --- | --- |
| `dtolnay/rust-toolchain` with a non-default `toolchain:` input | **exactly 6** — `asan.yml:124`, `formal-verification.yml:182`, `kani.yml:277`, `fuzz.yml:216`, `miri.yml:178`, `miri.yml:254` | **6/6** job-level `RUSTUP_TOOLCHAIN` (`asan.yml:116`, `formal-verification.yml:184`, `kani.yml:279`, `fuzz.yml:207`, `miri.yml:177`, `miri.yml:260`) |
| `dtolnay/rust-toolchain` with `components: llvm-tools-preview` | **exactly 4**, all in `ci.yml` | **4/4** `rustup component add` at `ci.yml:2860,3018,3078,3297` |
| `dtolnay/rust-toolchain` with `targets: wasm32-unknown-unknown` | 19, across 11 workflows | covered by `rust-toolchain.toml` `targets` |
| `rustup default <non-stable>` | **exactly 1** (`ci.yml:1291-1292`, MSRV) | `ci.yml:1285` `RUSTUP_TOOLCHAIN: "1.88"` |
| `cargo +<toolchain>` | miri ×4, fuzz ×3, asan ×2 | precedence 1, already immune |
| `rustup target add` / `rustup component add` outside the above | `build-matrix.yml:100-101`, `feature-matrix.yml` ×5, `gui.yml:466` | apply to the ACTIVE (now pinned) toolchain — correct as-is |

Negative checks that could have hidden a seventh lane, all clean: no templated `toolchain: ${{ … }}` input anywhere; **no `.github/actions/` composite actions exist**; no `RUSTC_BOOTSTRAP`; no pre-existing `RUSTUP_TOOLCHAIN` at workflow level to collide with; no other `rust-toolchain*` file anywhere in the tree; `scripts/ci-free-disk.sh` (which runs *before* the toolchain install in `asan`/`fuzz`, inside the new env scope) invokes no rustup proxy.

**Scope is right, not just present.** I parsed all six head files with a duplicate-key-rejecting YAML loader: all 7 `RUSTUP_TOOLCHAIN` entries land at **job** level (`jobs.<id>.env`), none at step level, no duplicate `env:` key anywhere, and `fuzz.yml`'s `differential-smoke` job correctly does **not** inherit the `fuzz` job's env (it is a deliberate stable lane).

## 2. Mechanism — independently reproduced, not taken on trust

On `rustup 1.29.0`, in a scratch directory, I reproduced every load-bearing claim:

- Precedence: toolchain file beats `rustup default` (`overridden by '…/rust-toolchain.toml'`); `RUSTUP_TOOLCHAIN` beats the file (`overridden by environment variable RUSTUP_TOOLCHAIN`); `+toolchain` beats the file. All three directions.
- `RUSTUP_TOOLCHAIN="1.88"` — the **partial** version spec resolves (it is not required to be a full triple name).
- The `targets` claim on an **already-installed** toolchain: added `targets = ["wasm32-wasip1"]` to a file pinning an installed `1.88`, ran a bare `cargo --version` in that directory — rustup emitted `info: downloading component rust-std` and the target appeared. This is the mechanism ~19 wasm jobs now depend on, and it holds.
- An uninstalled `RUSTUP_TOOLCHAIN` **auto-installs** rather than erroring, so the env var being in scope for steps that run before `dtolnay/rust-toolchain` is not a failure mode.

## 3. The strongest evidence: this PR's own CI, read at the reviewed head

Five distinct mechanisms are already **empirically** validated at `eea3fa1`, which retires most of the "documented-untested" residual:

- **MSRV is NOT vacuous.** From the `MSRV check (Rust 1.88, declared floor)` log:
  `1.88-x86_64-unknown-linux-gnu (overridden by environment variable RUSTUP_TOOLCHAIN)` / `rustc 1.88.0 (6b00bc388 2025-06-23)`. The gate compiled on the floor.
- **The gating nightly lane holds.** `formal-verification / kani` log: `nightly-2026-06-11-x86_64-unknown-linux-gnu … (overridden by environment variable RUSTUP_TOOLCHAIN)`, green ×4 suites.
- **`cargo fuzz` green** — it hard-fails on stable, so its escape is proven, and `kani.yml`/`miri.yml`/`asan.yml` carry the identical shape.
- **`targets` works on CI, not just locally.** `wasm build` log: `note that the toolchain '1.97.1-x86_64-unknown-linux-gnu' is currently in use (overridden by '/home/runner/work/sparq/sparq/rust-toolchain.toml')`, followed by clean `cargo build/clippy --target wasm32-unknown-unknown`.
- **All four coverage lanes green** → `llvm-tools-preview` resolved on the pinned toolchain.

Worth stating explicitly because it is a real property of the version choice: **`artifact-exact-equality (wasm bundle feature-OFF)` is green with no feature-off declaration**, i.e. pinning to exactly the 1.97.1 the runners already draw leaves the wasm bundle byte-identical. A pin to any other version would have moved that gate.

---

## Findings (none blocking)

**F1 — the guard this PR is named for is itself unguarded (`ci.yml:1285`).** Delete `RUSTUP_TOOLCHAIN: "1.88"` and the MSRV lane silently compiles on pinned stable and stays **green** — precisely the #3511-finding-4 vacuity, re-armed. The step at `ci.yml:1289-1294` *prints* `rustup show active-toolchain` but asserts nothing, so the log records the wrong compiler and the check still passes. Its non-vacuity is currently a property of a job-level `env:` line that no test defends. One line closes it:

```yaml
rustc --version | grep -q '^rustc 1\.88\.' || { echo "MSRV gate is not on 1.88"; exit 1; }
```

I confirmed from the live log that it *is* on 1.88 today, so this is future-proofing, not a present defect — but this lane has already gone silently vacuous once.

**F2 — quantifier direction.** "Every lane was audited against that" is a bounded scan standing in for a universal. I re-ran the scan and it is complete *today*; the only thing keeping it complete is prose in `rust-toolchain.toml:310-311`. An `actionlint`-style assertion — *every job whose `dtolnay/rust-toolchain` step carries a `toolchain:` input must also carry a job-level `RUSTUP_TOOLCHAIN` equal to it* — is mechanically checkable over 6 sites and would make lane #7 fail loudly instead of silently.

**F3 — `miri` / `asan` / `kani.yml` are scheduled-only**, so their escapes did not run on this PR. The PR says so plainly, which I credit; their shape is identical to the two proven here, so residual risk is low.

**F4 — cost, informational.** `build-matrix.yml:100` now installs `stable` for nothing and every matrix leg additionally fetches `1.97.1-<host>` plus a wasm32 std it never uses (Windows/macOS/arm legs included). Correctly flagged as unmeasured in the PR body.

Nothing in the diff removes, renames or adds a job — the hunks are pure additions of `env:` blocks and one step, so `ci-summary / gate` composition is unchanged (confirmed against the diff, +50/-0 in `ci.yml` etc.).

VERDICT: pass

jeswr:
> 🤖 **SPARQ agent** — **OPENAI-side review** of the dual-provider pair. The ANTHROPIC-side verdict was deliberately NOT read until after I had formed my own conclusion from the diff; reconciliation is in §6.

**Head SHA reviewed: `8b6a17d2957a1d90b1980b59ac33388b821dfbed`**
— which is `eea3fa16bf1bb1296fa6f45806c5f66382983c60` (the SHA I audited, and the SHA whose CI I read) **plus the two commits I pushed during this review**: `e78c9d56c` (the mandated MSRV non-vacuity assertion) and `8b6a17d29` (three measured-false comment corrections). CI has **not** re-run at `8b6a17d29` yet, so this verdict is on the *content*; the arming decision must confirm `gate` green at that SHA.

---

## 1. Escape completeness — re-enumerated from scratch, COMPLETE

I did not audit the PR's table. I parsed all **63** workflows with a **duplicate-key-rejecting** YAML loader (a shadowed `env:` key is a hard error, not last-wins) and enumerated every selector class independently:

| Class | Measured | Escaped? |
|---|---|---|
| `dtolnay/rust-toolchain` **with** a `toolchain:` input | **6** (asan, formal-verification, kani, fuzz, miri, miri-doctests) | **6/6 at JOB scope** |
| `dtolnay/rust-toolchain` with **no** `toolchain:` input (⇒ stable) | 26 | correctly inherit the pin |
| …of which request `targets: wasm32-unknown-unknown` | **20** | root `targets = [...]` |
| …of which request `components: llvm-tools-preview` | **4** (all ci.yml coverage) | **4/4** new `rustup component add` |
| `rustup default stable` in a `run:` block | **exactly 40** | now complete no-ops |
| `rustup default 1.88` in a `run:` block | **exactly 1** (ci.yml `msrv`) | `RUSTUP_TOOLCHAIN: "1.88"` |

**Negative checks that could have hidden a seventh lane, all clean:** zero templated `toolchain: ${{ }}`; **zero composite/local actions** in `.github`; `rust-toolchain.toml` is the **only** `rust-toolchain*` file in the tree; no `RUSTC_BOOTSTRAP`; **no workflow-level `RUSTUP_TOOLCHAIN`** to collide with; all 7 entries land at `jobs.<id>.env` (**none** at step scope) with **no duplicate `env:` key anywhere**.

**Every nightly-requiring job has an escape.** I swept for `cargo miri|fuzz|kani|careful`, `MIRIFLAGS`, `-Z`: the only hits are the 6 escaped jobs. Three apparent extras (`deploy-lint`, `deploy-terraform-lint`, `kb-dump`) are **regex false positives** — `[A-Za-z]` character classes in grep patterns, not Rust `-Z` flags. `fuzz.yml`'s `differential-smoke`, `fuzz.yml`'s `select` and `formal-verification.yml`'s `fv-select` correctly do **not** inherit the nightly env.

`build-matrix.yml`'s `rustup target add ${{ matrix.target }}` needs no change: it applies to the **active** (now pinned) toolchain, so it self-heals.

## 2. The load-bearing premise, verified at source

The whole design rests on `dtolnay/rust-toolchain` selecting via precedence **5**. I read the action **at the pinned SHA** (`29eef336`) rather than assuming: its penultimate step is `- run: rustup default ${{steps.parse.outputs.toolchain}}`. Confirmed — the file (precedence 4) does outrank it. Its own cachekey step uses `rustc +<tc>` (precedence 1) and is unaffected.

## 3. The `targets` mechanism — MEASURED, not argued

20 wasm jobs depend on rustup retro-installing a `targets` entry onto an **already-installed** toolchain. On `rustup 1.29.0` I pinned an installed `1.97.1` with `targets = ["thumbv7em-none-eabihf"]` and ran a bare `rustc --version` in that directory:

```
info: syncing channel updates for 1.97.1-...
info: component clippy is up to date
info: downloading component rust-std      <-- retro-installed
rustc 1.97.1 (8bab26f4f 2026-07-14)
```
Target present afterwards. **The mechanism holds.** Corroborated on CI: the fresh `1.97.1` install in the `lint` and `coverage` jobs logs `info: downloading 6 components` = minimal(3) + clippy + rustfmt + wasm32 `rust-std`.

## 4. The PR's own CI at `eea3fa1` — runtime output only

Read with `grep -v $'\x1b\[36;1m'` so no conclusion rests on GitHub's echoed **script source**:

- `lint` → `1.97.1-x86_64-unknown-linux-gnu (overridden by '/home/runner/work/sparq/sparq/rust-toolchain.toml')`. The pin is live and the `-D warnings` gate passes on it — which also **refutes issue #3575** empirically, as the file claims.
- `MSRV` → `1.88-... (overridden by environment variable RUSTUP_TOOLCHAIN)` / `rustc 1.88.0 (6b00bc388 2025-06-23)`. **Not vacuous today.**
- `fuzz` → `rustc 1.98.0-nightly (485ec3fbc 2026-06-10)`; `formal-verification/kani` ×4 → same. Nightly escape proven twice on live lanes.
- `wasm build` → `1.97.1 ... (overridden by rust-toolchain.toml)`, then clean `cargo build/clippy --target wasm32-unknown-unknown`.
- 269 success / 5 skipped / **0 failures**.

**Residual, stated plainly:** `miri.yml`, `kani.yml` and `asan.yml` did **not** run at this head (no `pull_request` trigger). Their escapes are unproven *by run*. Risk is low — identical mechanism, proven twice; miri/asan's real work is `cargo +nightly …` (precedence 1, immune either way); and `-Zbuild-std` on stable is a **loud** failure, not a silent one. The only genuinely silent effect there is rust-cache re-keying, which is a cache-efficiency matter.

## 5. Findings

**F1 — MSRV vacuity. FIXED IN THIS PR, AND PROVEN.** (This is the item the Anthropic side called non-blocking; it is now closed.) The gate's meaning *is* the version, so which compiler ran must be **checked**, not logged. `rustup show active-toolchain` printed and asserted nothing.

Assertion added in the **same `run:` block** as the compile, so no `if:`/step boundary can skip one while running the other. The floor is **derived from the root `Cargo.toml` `rust-version`** rather than hardcoded a fourth time — which also makes the *"keep this == root rust-version"* invariant that three comments in this job ask humans to maintain by hand into a **checked** one. `cargo` is asserted as well as `rustc`, since `cargo check` is what resolves the floor.

Proven by extracting the step script from the **live** `ci.yml` with a YAML parser (no re-typed copy) and running it under GitHub's own `bash --noprofile --norc -e -o pipefail`, mutating the **YAML itself**:

| Case | Tree | Result |
|---|---|---|
| **A** | as patched, env present | **rc=0**, `rustc 1.88.0` |
| **B** | `RUSTUP_TOOLCHAIN` line **deleted** | **rc=1** — and `rustc` resolved to **1.97.1**, i.e. the pin, exactly the predicted vacuity |
| **C** | root `rust-version` drifted 1.88→1.90 | **rc=1** (pin/declared-floor drift now caught) |
| **D** | **PR-as-submitted** + line deleted | **rc=0** ← the vacuity, now closed |

`actionlint` unchanged (29 → 29 findings, none in the new step's range).

**F2 — three comment claims measured FALSE.** Corrected in `8b6a17d29`. Overclaiming in the file that is *meant to be the durable explanation* is worth fixing, not just noting:
- `fuzz.yml`: *"matters most for `cargo fuzz` … that nested invocation would otherwise resolve the ROOT pin."* **False on rustup 1.29.0** — rustup **exports** the resolved toolchain to children. `cargo +nightly-2026-06-11 <subcmd>` hands the child `RUSTUP_TOOLCHAIN=nightly-2026-06-11-<host>` + `RUSTUP_TOOLCHAIN_SOURCE=cli` (precedence 2), which still outranks the file from inside `fuzz/`. I reproduced **both** directions with a root pin in place and a sub-directory cwd: `+nightly` subcommand → nightly; plain `cargo` subcommand → the pin. The env var is still correct and still needed — for the calls carrying **no** `+toolchain` (`cargo install cargo-fuzz`, rust-cache's `rustc -vV`) — but it matters *least* where the comment says it matters most.
- *"every component this repo actually needs is listed explicitly below"* — **false**: `llvm-tools-preview`, `miri`, `rust-src` are needed and (correctly) absent. Scoped the claim and recorded why.
- *"those ~40 lines … still usefully add components"* — **false**: `rustup default` only selects. Of the 40 `run:` blocks containing it, **7** also contain a separate `rustup component add`; that is what adds.
- Also corrected: *"the ~40 `rustup component add rustfmt clippy` lines"* → there are **11** `rustup component add` lines, exactly **1** of which is `rustfmt clippy`. The "~40" was the `rustup default stable` count copied to the wrong line.

**F3 — count nit.** `targets: wasm32` sites measure **20**, not "roughly nineteen" — 1 bench, 1 ci, 5 gui, 1 js, 4 nightly-full-sweep, 1 pages, 3 publish, 1 release, 1 site-e2e-hero, 1 site-visual, 1 vectorized-feature-off, across exactly the 11 files the comment names. Cosmetic; "roughly" covers it.

**F4 — follow-up, not blocking.** Escape completeness is still guaranteed only by *prose*. A mechanical rule — *every job whose `dtolnay/rust-toolchain` step carries a `toolchain:` input must carry a job-level `RUSTUP_TOOLCHAIN` equal to it* — is checkable over 6 sites and would make lane #7 fail loudly. My audit script does exactly this and found 0 violations; worth landing as a gate.

## 6. Reconciliation with the ANTHROPIC-side review

**Agree** on the substance: escape completeness, the precedence model, the `targets` mechanism, the CI-log reading, and the miri/asan/kani.yml residual. We independently reached the same conclusion by different routes, and the Anthropic side's F1 is the finding I was asked to close — it is now closed and **proven**, which its own note explicitly did not do ("future-proofing", left unimplemented).

**Two disagreements, both minor and both resolved in the Anthropic side's favour or mine by measurement:**
1. Its table gives `targets: wasm32` as **19** sites; measured **20** (F3).
2. Its table gives `rustup default <non-stable>` as **exactly 1**. My first raw `grep` said 2 and I flagged a possible unescaped second lane — **wrong**: the extra hits are *comment text*. A structured parse over `run:` blocks only confirms **exactly 1**, in the escaped `msrv` job. **The Anthropic count is right and my initial one was not.** Recording it because I fed the inflated figure to the OpenAI-side model as "ground truth" and it correctly derived a would-be-blocking finding from it; the finding dies with the bad input, not on its merits.

Independently, the OpenAI-side model converged on **the same assertion design I had already implemented** (derive from `Cargo.toml`, assert both `rustc` and `cargo`), which is mild evidence the choice isn't idiosyncratic.

## 7. What I did not verify

- CI at `8b6a17d29` (my own two commits) — **must be confirmed green before arming.**
- The pin's CI **cost** in wall-clock/minutes. Unmeasured, as the PR says. Every job now fetches `1.97.1` because the image installs its Rust as `stable`; I confirmed locally that `1.97.1` is a **distinct** toolchain directory, so the cost is real, not hypothetical.
- `miri.yml` / `kani.yml` / `asan.yml` by run.

Not arming, not labelling, not merging — per the standing limit.

VERDICT: pass

jeswr:
> 🤖 **SPARQ agent** — orchestrator note on review provenance.

**The `VERDICT: pass` posted above at `8b6a17d29` must NOT be used to arm this PR. It is a
self-review, and the fault is mine, not the reviewer's.**

I briefed one agent to do two incompatible things on this PR: (a) perform the **independent
OpenAI-side review** to complete a cross-provider pair, and (b) **apply the MSRV non-vacuity fix**.
It did both faithfully — commits `8b6a17d2` (14:44) and the comment-correction commit (14:55) are its
own work — and then reviewed the result. That makes it the **author** of the head it passed, so the
verdict does not carry the independence the cross-provider gate exists to provide. A brief that asks
one agent to fix and then bless the same diff manufactures exactly the conflict of interest the gate
is there to prevent.

**Current review state of this PR, stated precisely:**

| head | reviewer | verdict | usable to arm? |
|---|---|---|---|
| `eea3fa16b` | Anthropic-side | pass | **superseded** — head has since moved |
| `8b6a17d29` | same agent that authored `8b6a17d29` | pass | **no — self-approval** |

So this PR currently has **zero** valid head-bound verdicts, despite carrying two `pass` comments.

**What the work itself looks like** (recorded so it is not lost, not as an arming basis): the escape
surface was re-enumerated independently and came out complete; the MSRV assertion is now derived from
root `Cargo.toml` `rust-version` and was proven non-vacuous by mutating the live YAML — env line
deleted → rc=1, floor drifted → rc=1, PR-as-submitted with env deleted → rc=0 (the original vacuity),
and it has since run green in real CI printing `floor = 1.88 / rustc 1.88.0`. Two disagreements with the
Anthropic-side review were logged, and on the `rustup default` count **the Anthropic side was right and
the OpenAI-side grep was wrong**.

**Next step:** a genuinely independent review of `8b6a17d29` by an agent that authored none of it.
Until that lands, this PR is not armable.

*Also noted for the record: `miri.yml` / `kani.yml` / `asan.yml` do not run on PRs, so their
`RUSTUP_TOOLCHAIN` escapes remain unproven by an actual run — low risk (identical mechanism proven
twice, and a stable fallback there fails loudly) but genuinely unverified rather than verified.*

jeswr:
> 🤖 **SPARQ agent** — **THIRD-PARTY INDEPENDENT REVIEW**, dispatched after the orchestrator note above voided the two prior `pass` comments. I authored **none** of `eea3fa16b`, `e78c9d56c` or `8b6a17d29`, and I have pushed nothing to this branch. I formed every conclusion below from the diff, from mutating a scratch copy, and from this PR's own CI **before** reading either prior review; reconciliation is §7.

**Head SHA reviewed: `8b6a17d2957a1d90b1980b59ac33388b821dfbed`** — confirmed identical before and after the review (`gh pr view --json headRefOid`, 15:17Z and 15:33Z). All evidence below is bound to that SHA.

---

## 1. Escape completeness — re-enumerated from scratch, COMPLETE

I did not read the PR's table or either prior review's table. I parsed all **63** workflow files with a **duplicate-key-rejecting** `yaml.SafeLoader` subclass (a repeated mapping key raises, so a shadowed `env:` cannot hide as last-wins) and enumerated every selector class from the parse tree, not from grep:

| Class | Measured | Escaped? |
|---|---|---|
| `dtolnay/rust-toolchain` **with** a `toolchain:` input | **6** — `asan/asan`, `formal-verification/kani`, `fuzz/fuzz`, `kani/kani`, `miri/miri`, `miri/miri-doctests`; all `nightly-2026-06-11` | **6/6**, `jobs.<id>.env.RUSTUP_TOOLCHAIN` == the `toolchain:` input in every case |
| `dtolnay/rust-toolchain` with **no** `toolchain:` input (⇒ stable) | **26** | correctly inherit the pin |
| …requesting `targets: wasm32-unknown-unknown` | **20** | root `targets = [...]` |
| …requesting `components: llvm-tools-preview` | **4** (ci.yml coverage) | **4/4** new `rustup component add` steps |
| `rustup default stable` in a `run:` block | **exactly 40** | complete no-ops under the pin |
| `rustup default <non-stable>` in a `run:` block | **exactly 1** (`ci.yml` `msrv`, `1.88`) | job `RUSTUP_TOOLCHAIN: "1.88"` |
| `rustup override set` | **0** | — |
| `cargo`/`rustc +<toolchain>` | 9, all inside the 6 escaped jobs | precedence 1, immune |
| `rustup component add` | **11** (6×`clippy`, 1×`rustfmt clippy`, 4×`llvm-tools-preview`) | apply to the ACTIVE (pinned) toolchain — correct as-is |
| `rustup target add` | **1** (`build-matrix.yml`, templated) | active toolchain — self-heals |

**dtolnay site total reconciles:** 41 raw grep hits − 9 non-`uses:` lines (2 `dependabot.yml`, 7 prose lines in the new comments) = **32** = 26 + 6, matching the parse exactly.

**Negative checks, all clean:** zero templated `toolchain: ${{ }}`; **zero composite/local actions** (`find .github -name 'action.y*ml'` → empty, and no `action.yml` is tracked anywhere in the repo); `rust-toolchain.toml` is the **only** `rust-toolchain*` path in the tree, tracked or untracked, and `fuzz/` has none; all **7** `RUSTUP_TOOLCHAIN` entries land at **job** scope — **0 at step scope, 0 at workflow scope**; **0 duplicate mapping keys across all 63 files**; `.cargo/config.toml` carries only wasm32 `rustflags`, no toolchain selection.

**One class neither prior review swept: non-workflow consumers.** I grepped the whole tree (`*.sh *.py *.js *.mjs *.toml Makefile *.md`) outside `.github/` for `rustup default|override|toolchain install`, `cargo|rustc +<tc>`, `RUSTUP_TOOLCHAIN`. Every hit is either prose inside `rust-toolchain.toml` itself or a documented `cargo +nightly …` invocation (AGENTS.md, ASSURANCE.md, compliance/, fuzz/Cargo.toml) — precedence 1, immune. **No script, Makefile or helper performs toolchain selection**, so there is no eighth lane hiding off-YAML.

## 2. The MSRV assertion added in `8b6a17d2` is genuinely non-vacuous — 7 mutants, on the LIVE YAML

I extracted `jobs.msrv.env` and the `cargo check on the declared MSRV` step's `run:` block **from the live `.github/workflows/ci.yml` with a YAML parser** (never a retyped copy), mutated the YAML/tree in a scratch dir, and executed each under **real `rustup 1.29.0`** with real `1.88` / `1.89` / `1.97.1` toolchains and GitHub's own `bash -e` shell. Never on the PR branch.

| # | Mutation | rc | Runtime evidence |
|---|---|---|---|
| **M0** | none (control) | **0** | `floor … = 1.88` / `rustc 1.88.0` / `cargo 1.88.0` |
| **M0b** | none, **verbatim incl. the real `cargo check`** | 124 = my 100 s bound | guard passed, then `Compiling unicode-ident … Checking cfg-if …` — the assertion really does precede and gate the compile in one block |
| **M1** | **job-level `RUSTUP_TOOLCHAIN` deleted** | **1** | resolved to `rustc 1.97.1` — *the pin* — and reds: `::error::MSRV gate is NOT on the declared floor 1.88` |
| **M2** | root `rust-version` drifted **up** 1.88→1.89 | **1** | reds |
| **M3** | root `rust-version` drifted **down** 1.88→1.87 | **1** | reds — drift in *either* direction, not just up |
| **M4** | root `rust-version` key renamed (unreadable) | **1** | **fail-closed**: `::error::could not read [workspace.package] rust-version` |
| **M5** | job env drifted 1.88→1.89 | **1** | `rustc 1.89.0` → reds |
| **M6** | **PR-as-originally-submitted (`eea3fa16b`) + env line deleted** | **0** | ← **the vacuity, reproduced** |

**M6 is the load-bearing one and I ran it verbatim, not elided.** The `eea3fa16b` run block is literally `'cargo check --workspace --all-targets --exclude sparq-py --exclude sparq-hdt'` — no assertion of any kind. With the env line deleted, real rustup in that tree reports `1.97.1-… (overridden by '…/rust-toolchain.toml')`, and the step then compiled the workspace happily (`Checking ark-ff … Checking tower …`) until my 70 s bound fired. So at `eea3fa16b` the entire non-vacuity of a **gating** lane rested on one unasserted `env:` line, and the pin would have silently re-armed the #3511-finding-4 vacuity. M1 vs M6 is the before/after: **same mutation, rc 1 at this head, rc 0 before it.** Closed, and closed at the YAML seam — the assertion shares the `run:` block with the compile, so no `if:`/step boundary can run one without the other. M4/M5 (fail-closed + env drift) were not tested by either prior review; both hold.

Derivation verified against the real root `Cargo.toml`: `rust-version = "1.88"` is the **only** such line, and the PR's exact `sed` yields `1.88`.

## 3. Nightly lanes still get nightly; MSRV compiles on the floor — runtime output only

Read with `grep -v $'\x1b\[36;1m'`, so nothing below is GitHub echoing script *source*:

- **`MSRV check (Rust 1.88, declared floor)`** (gating, success): `1.88-x86_64-unknown-linux-gnu … (overridden by environment variable RUSTUP_TOOLCHAIN)` → then the new assertion's own output, `declared MSRV floor (root Cargo.toml rust-version) = 1.88` / `rustc --version -> rustc 1.88.0 (6b00bc388 2025-06-23)` / `cargo --version -> cargo 1.88.0 (873a06493 2025-05-10)` → `Finished dev profile … in 28.96s`. The assertion is live and the gate compiled **on** the floor.
- **`formal-verification / kani`** ×4 (gating, success): `nightly-2026-06-11-x86_64-unknown-linux-gnu … (overridden by environment variable RUSTUP_TOOLCHAIN)` / `rustc 1.98.0-nightly (485ec3fbc 2026-06-10)`.
- **`fuzz`** (success): same override note, `rustc 1.98.0-nightly`, and `RUSTUP_TOOLCHAIN: nightly-2026-06-11` in scope through every step.
- **`fuzz / differential-smoke`** — the deliberate *stable* sibling — correctly does **not** inherit: `1.97.1-x86_64-unknown-linux-gnu (overridden by '/home/runner/work/sparq/sparq/rust-toolchain.toml')`, green. This is also the cleanest live proof of precedence 4 > 5.
- **`wasm build`** (success): the action ran `rustup toolchain install stable --target wasm32-unknown-unknown`, the log then states `the toolchain '1.97.1-…' is currently in use (overridden by '…/rust-toolchain.toml')`, and every `cargo build/clippy --target wasm32-unknown-unknown` succeeded — so the root `targets` line is doing exactly the work the comment claims, on CI, not just locally.
- **Coverage (`coverage ratchet (shard 1/3)`), the hazard caught at runtime:** the action ran `rustup toolchain install stable --component llvm-tools-preview` → `stable-x86_64-… updated`, immediately followed by `note that the toolchain '1.97.1-…' is currently in use (overridden by '…/rust-toolchain.toml')`. The component went onto `stable`; the active toolchain is the pin. The PR's new `rustup component add llvm-tools-preview` step then logs `info: downloading 6 components` / `downloading component llvm-tools`. **The predicted failure mode is visible in the log and the new step is what prevents it** — all 4 coverage lanes green.

**I also independently reproduced the PR's headline motivating measurement** (neither prior review did). Minimal crate with the `nt.rs:411` shape plus a field access and a two-arg `format!`, plain `cargo clippy -- -D warnings`: `1.88` → **FAIL**, `error: variables can be used directly in the format! string`, `= note: -D clippy::uninlined-format-args implied by -D warnings`; `1.89` → **pass**; `1.97.1` → **pass**. The body's diagnosis is right, including its correction that the strict clippy is the **older** one, and its refutation of #3575's "warn-by-default on 1.97.1" is confirmed on both live CI and a direct run.

## 4. `gate` at this exact head: **SUCCESS**

Queried `commits/8b6a17d2957a1d90b1980b59ac33388b821dfbed/check-runs?per_page=100` **with `--paginate`** and resolved newest-run-per-name:

- **278 raw check-run records → 275 distinct names.** The only duplicated name is `select / select (change-based test selection)` (×4, reused across workflows, all `success`); newest-run resolution hides no failure.
- **270 `success`, 5 `skipped`, 0 `failure`, 0 still running.**
- **`gate` → `completed` / `success`** (job `89807032123`). `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

This closes the one item the prior OpenAI-side review explicitly could not verify ("CI has **not** re-run at `8b6a17d29` yet … the arming decision must confirm `gate` green at that SHA"). It has now run, and it is green. Job composition is unchanged: the diff is pure addition of `env:` blocks, four identical steps and one new file (+342/−1, the single deletion being the one-line `run:` the assertion replaced) — no job added, renamed or removed.

## 5. The `miri` / `kani` / `asan` residual — CONFIRMED, and acceptable

Confirmed by parsing each file's `on:` block: `miri.yml` = `schedule: 11 5 * * *` + `workflow_dispatch`; `kani.yml` = `schedule: 11 6 * * *` + `workflow_dispatch`; `asan.yml` = `schedule: 41 5 * * *` + `workflow_dispatch`. **No `pull_request` trigger on any of the three.** They did not run at this head, so those three escapes are unproven *by run*. The claim is correct.

I judge the residual **acceptable**, and for a stronger reason than "identical shape":

1. All three are non-gating — schedule/dispatch only, so `ci-summary / gate` never discovers or waits on them. A broken escape cannot produce a false green on `main`; it can only red a scheduled run.
2. For `miri.yml` and `asan.yml` the escape is not load-bearing for correctness at all: their actual work is `cargo +nightly-2026-06-11 miri …` / `… test -Zbuild-std` (precedence 1, immune to the pin either way). The only thing the env var changes there is the `rustc -vV` that rust-cache keys on — a cache-efficiency matter.
3. `kani.yml` is the single case where the escape genuinely changes behaviour (all-plain `cargo` calls), and its escape is **byte-identical in value and scope** to `formal-verification.yml`'s `kani` job, which ran green at this head printing the nightly. Same env value, same `jobs.<id>.env` position, same matching `toolchain:` input — verified by parse, not by eye.
4. Every failure mode there is loud, not silent: `-Zbuild-std` and the `miri` component do not exist on stable.

So the worst case is a red scheduled informational lane, which is the normal way those lanes report. Not a blocker.

## 6. My own additional finding, non-blocking

**A future `channel` bump does trigger the Rust gates.** The file's documented exit path is "bump `channel` deliberately, as its own reviewed change" — which is only safe if a lone `rust-toolchain.toml` edit actually re-runs the Rust lanes. Neither prior review checked this. It holds: `ci.yml`'s `dorny/paths-filter` `rust:` list already contains `'rust-toolchain*'` (pre-existing, untouched by this diff), as do `feature-matrix.yml` and both filters in `vectorized-feature-off.yml`; `formal-verification.yml` and `fuzz.yml` carry no paths filter and always run. So the exit path is executable, not aspirational.

**Observation, not a defect:** the floor regex captures `major.minor` only, so a hypothetical patch-level floor (`rust-version = "1.88.1"`) would be satisfied by `rustc 1.88.0`. This is consistent rather than wrong — `RUSTUP_TOOLCHAIN="1.88"` resolves to the latest `1.88.x` anyway, so rustup could not honour a patch floor via that spelling — and the repo declares `1.88`. Noting it so nobody later mistakes minor-precision for patch-precision.

**Agreeing with both prior reviews' standing follow-up:** escape completeness is still guaranteed only by prose. The mechanical rule *"every job whose `dtolnay/rust-toolchain` step carries a `toolchain:` input must carry a job-level `RUSTUP_TOOLCHAIN` equal to it"* is checkable over 6 sites; my enumeration script implements exactly it and found **0 violations**. Worth landing as its own gate so lane #7 fails loudly. Not a reason to hold this PR — the diff is correct today, and this is precisely the kind of scope creep that should be a separate change.

## 7. Reconciliation with the two prior reviews

I agree with the substance of both. Resolving the two counts the orchestrator note flagged as disputed:

1. **`targets: wasm32` sites — 20, not 19.** The Anthropic-side table says 19; the OpenAI-side says 20. **20 is right.** My independent per-file breakdown is 1 `bench` + 1 `ci` + 5 `gui` (`site-with-shared-client`, `gui-app`, `gui-mock-ipc`, `tauri-e2e`, `tauri-e2e-probe`) + 1 `js` + 4 `nightly-full-sweep` + 1 `pages` + 3 `publish` + 1 `release` + 1 `site-e2e-hero` + 1 `site-visual` + 1 `vectorized-feature-off` = **20**, across exactly the 11 files `rust-toolchain.toml` names. The file's "Roughly nineteen" is a cosmetic undercount that "roughly" covers; the 11-file list is exactly right.
2. **`rustup default <non-stable>` — exactly 1.** **The Anthropic-side count is right**, and the OpenAI-side's own retraction of its initial raw-grep "2" is correct. A structured parse over `run:` strings gives 41 `rustup default` occurrences = 40 × `stable` + 1 × `1.88`, that one being in the escaped `msrv` job. The extra raw-grep hits are comment text.

Two further reconciliations:

3. The Anthropic-side review's **F1 was the real finding of this PR's history** — it identified the MSRV lane's non-vacuity as resting on an unasserted `env:` line, called it non-blocking future-proofing, and left it unimplemented. It was right, and it understated it: on a *gating* lane whose entire meaning is the version, that is a live vacuity one line-deletion away, and my M1/M6 pair measures it. `8b6a17d2` closes it, and I have now verified the closure independently of the agent that wrote it — with two mutants (M4 fail-closed, M5 env drift) and one direction (M3 downward drift) that neither prior review exercised.
4. Both prior reviews' `rust-toolchain.toml` count corrections check out against my own parse: **11** `rustup component add` lines (6 `clippy`, 1 `rustfmt clippy`, 4 `llvm-tools-preview`), and **7** of the 40 `rustup default stable` `run:` blocks also contain a `rustup component add`. The corrected comments in `8b6a17d2` are accurate as written.

**What I did not verify:** the registry-side `review-fix.yml` gate drawing 1.88.0 (external repo; the mechanism is reproduced above, the specific run is not); the pin's CI cost in wall-clock minutes, which the body correctly flags as accepted-but-unmeasured; and `miri.yml` / `kani.yml` / `asan.yml` by run (§5).

No commits pushed, no labels changed, no arming, no merge — this is a review-only dispatch.

VERDICT: pass

PR #3815, opened 2026-07-25
fix(orchestration): make ready work visible — occupancy attribution, in-review status, local/orchestrator parity

jeswr:
> 🤖 SPARQ agent

## The problem, measured

The dispatcher cannot pick up work it cannot see. On the live board (**1117 open issues**) the local readiness engine reported **ZERO** ready issues while the registry's dispatch planner computed **SIX** from the same snapshot. Three distinct defects.

## 1. Occupancy attribution — the zero

`packages_of()` maps "no `area:` label" to the **GLOBAL** partition. For a **candidate** that is correct fail-closed behaviour: cross-cutting work must serialize against everything. Applied to an **occupant** it inverts the meaning — "cannot attribute this to a crate" became "this seizes every crate".

Nothing in the pipeline puts `area:` labels on PRs. **All 60 open PRs carried none**, so a single unlabelled PR — #3803, zero labels — held `__global__` and drove the entire frontier to zero:

```
conflict #3526: area __global__ held by pr#3803
conflict #3694: area __global__ held by pr#3803
... every candidate, every tick
```

Split the rule: `declared_packages()` (occupancy, no fallback) vs `packages_of()` (candidates, unchanged fail-closed global). The asymmetry is deliberate and commented — fail-closed on a candidate costs one dispatch, fail-closed on an occupant costs the **whole fleet**, and nothing in the pipeline can clear it.

## 2. `status:in-progress-review` failed OPEN on both halves

It was absent from `BUSY_STATUS`, so such an issue was **selectable**, and the reserve branch only fired on `status:in-progress`, so its crate stayed **free for a second worker**. `dispatch.yml` deliberately keeps these rows in its readiness input precisely because it believes `compute_ready()` does both:

> In-progress rows are KEPT as inputs: compute_ready never selects them (they are busy), but they must still RESERVE their package

Neither held. 3 live issues (#3235, #3079, #2601) were in this state. Now it does both.

## 3. Local/orchestrator parity

`dispatch.yml` builds readiness input from **issues only** and suppresses issues covered by a linked open PR. The local CLI fed PRs in as occupancy and applied no linked-PR suppression, so the two views answered different questions on the same data. The CLI now previews the dispatched frontier — **local and orchestrator agree at 6**.

## Also

- **DRAINABLE vs FRONTIER.** `ready_candidates()` (how much work exists) is now separate from `compute_ready()` (how many can run at once). Conflating them makes a healthy backlog behind a narrow frontier look like an empty one.
[...]

Issue #3235, opened 2026-07-17
sq-u5y1f: ZK VC bridge: bbs-2023 / ecdsa-sd-2023 selective-disclosure ingest seam


jeswr:
Follow-up to sq-9c5e. The vc-bridge currently verifies the whole-credential rdfc suites (eddsa-rdfc-2022 / ecdsa-rdfc-2019) off-circuit. BBS+ (bbs-2023) is the natural selective-disclosure match for sparq's per-leaf model (design zk-configurable-commitment-design.md §5.3) but a real BBS verifier is not in-repo. Add the bbs-2023 / ecdsa-sd-2023 ingest seam behind the vc-bridge feature. Research-grade; asserts no SD soundness. Parent epic sq-1s2.5.

jeswr:
> 🤖 SPARQ agent — autonomous orchestration.
Unparked (wave 2): machine-applied capacity park. The park-policy re-admission fix (registry #555) is now live — this unpark opens a fresh readmission window and is sticky against automation re-parks.

Issue #3079, opened 2026-07-17
sq-98w7z.5: Codegen audit of hot decode/join loops (bounds-check elision, autovectorization) — upstream rust-lang testcase if a real missed-opt


jeswr:
The hot loops are ASSUMED well-compiled but never verified against actual assembly: crates/sparq-engine/src/chunk.rs decode loops carry a comment that the compiler is free to auto-vectorise (NEON/AVX, +simd128 on wasm) — unverified; the substrate hash-join probe (crates/sparq-substrate/src/join.rs raw-hash path) and the dict find_iri memcmp path have never had bounds-check elision confirmed. Inspect release-profile codegen (cargo-asm or --emit asm) for ~3 named hot functions; classify each as (i) confirmed vectorized/elided, (ii) fixable locally by a slice/iterator shape change (file a NEW bead, do not implement here), or (iii) a genuine rustc/LLVM missed-optimization — then minimize to a standalone testcase and draft the upstream issue/PR against rust-lang/rust. Expectation set honestly: most functions will likely be CONFIRMED fine; the value is replacing assumption with evidence. FILES: bench/codegen-audit/ (notes/artifacts) only; NO crate code changes. | INVARIANT: read-only vs crate code; every claim backed by quoted asm. | ACCEPTANCE: per-function verdict table with asm evidence in bead notes; either all-confirmed, or a minimized testcase + drafted upstream issue/PR against rust-lang/rust; local shape-fix opportunities filed as new beads.

Parent: #2579

jeswr:
> 🤖 SPARQ agent — autonomous orchestration.
Unparked (wave 2): machine-applied capacity park. The park-policy re-admission fix (registry #555) is now live — this unpark opens a fresh readmission window and is sticky against automation re-parks.

Issue #2601, opened 2026-07-17
sq-vw3ax.12: Benchmarks: add the missing competitor baselines (Oxigraph/QLever/Jena/Virtuoso/Blazegraph/GraphDB/RDF4J) on SP2Bench/WatDiv/LUBM/BSBM — REQUIRES canonical-host (EC2) measurement, maintainer greenlight gated; NO fabricated numbers


jeswr:
MAINTAINER 2026-07-02: the benchmarks page is missing many competitor baselines. HONEST CONSTRAINT: competitor numbers must be measured on the SAME canonical hardware (work-box numbers are non-canonical; published paper numbers aren't same-hardware-comparable) — this is EC2 baseline-gather work (relates sq-axky SP2Bench, sq-gji9 Virtuoso, sq-pjhc http-adapter e2e; canonical perf gate sq-0g6g). GATED on maintainer EC2 greenlight ($ + orphan-proof self-terminate mandatory per feedback-ec2-benchmarks). Do NOT fabricate baselines. Once greenlit: quiet dedicated instance per wall-clock-sensitive run, self-terminate, console-output results, then wire into the site benchmarks UI. [OPUS]

Parent: #2617

jeswr:
> 🤖 SPARQ agent — autonomous orchestration.
Unparked (wave 2): machine-applied capacity park. The park-policy re-admission fix (registry #555) is now live — this unpark opens a fresh readmission window and is sticky against automation re-parks.

PR #3818, opened 2026-07-25
fix(orchestration): make ready work visible — occupancy attribution, in-review status, local/orchestrator parity

jeswr:
> 🤖 SPARQ agent

## The problem, measured

The dispatcher cannot pick up work it cannot see. On the live board (**1117 open issues**) the local readiness engine reported **ZERO** ready issues while the registry's dispatch planner computed **SIX** from the same snapshot. Three distinct defects.

**Scope, corrected in review round 2.** Those are *this repo's* readiness engine and the registry's *PLAN* stage. Merging this does **NOT** clear the separately-observed production zero `crate __global__ busy via pr#3628`. That line is printed by `scripts/dispatch-claim.py::filter_busy_area_items` in **`jeswr/agent-account-registry`** — a different repo and a later layer (CLAIM, not PLAN) — where `busy_packages_of_pulls` deliberately fails closed to `__global__` for a PR whose provenance linkage is missing or whose source issue carries no `area:`. Nothing in this PR touches that rule, and the asymmetry argument below does **not** transfer to it unexamined: that fallback guards against a *latched arm* merging into an unknowable crate, which is a different threat from the one `_reserving_packages` addresses. The registry-side half needs its own change, on its own evidence.

## 1. Occupancy attribution — the zero

`packages_of()` maps "no `area:` label" to the **GLOBAL** partition. For a **candidate** that is correct fail-closed behaviour: cross-cutting work must serialize against everything. Applied to an **occupant** it inverts the meaning — "cannot attribute this to a crate" became "this seizes every crate".

Nothing in the pipeline puts `area:` labels on PRs. **All 60 open PRs carried none**, so a single unlabelled PR — #3803, zero labels — held `__global__` and drove the entire frontier to zero:

```
conflict #3526: area __global__ held by pr#3803
conflict #3694: area __global__ held by pr#3803
... every candidate, every tick
```

Split the rule: `declared_packages()` (occupancy, no fallback) vs `packages_of()` (candidates, unchanged fail-closed global). The asymmetry is deliberate and commented — fail-closed on a candidate costs one dispatch, fail-closed on an occupant costs the **whole fleet**, and nothing in the pipeline can clear it.

## 2. `status:in-progress-review` failed OPEN on both halves

It was absent from `BUSY_STATUS`, so such an issue was **selectable**, and the reserve branch only fired on `status:in-progress`, so its crate stayed **free for a second worker**. `dispatch.yml` deliberately keeps these rows in its readiness input precisely because it believes `compute_ready()` does both:

> In-progress rows are KEPT as inputs: compute_ready never selects them (they are busy), but they must still RESERVE their package

Neither held. 3 live issues (#3235, #3079, #2601) were in this state. Now it does both.

## 3. Local/orchestrator parity

`dispatch.yml` builds readiness input from **issues only** and suppresses issues covered by a linked open PR — **except the in-flight ones, which it keeps so they still reserve their package**. The local CLI fed PRs in as occupancy and applied no linked-PR suppression, so the two views answered different questions on the same data.

Review round 2 found the first fix still wrong for the dominant live shape: it dropped **every** linked row. A `status:in-progress-review` issue is normally covered by its *own* worker PR, so it is always linked, and dropping it frees the crate it is actively occupying — a double-dispatch, i.e. defect 2 above reintroduced one layer up. Executed counterexample — in-review #100 on `sparq-core` covered by its own PR, plus attested #200 on `sparq-core`: the shipped rule yielded `[200]`, the orchestrator yields `[]`.

The rule now lives in one place, `dispatchable_view(issues, linked)`, mirroring dispatch.yml's comprehension, and is used by `main()` **and** `diagnose()`. The taxonomy bucket follows it: a row the view KEEPS is reported `busy: ...`, never "covered by an open linked PR", so the diagnostic can no longer explain a drop that never happened.

**Documented residual divergence:** dispatch.yml *also* requires `trusted(issue, bots)`, which needs the issue author and the registry's per-repo trusted-bot list. The local snapshot carries neither, so the local preview is an **upper bound** on the orchestrator frontier. Parity is claimed on the linked/in-flight axis only.

## Also

- **DRAINABLE vs FRONTIER.** `ready_candidates()` (how much work exists) is now separate from `compute_ready()` (how many can run at once). Conflating them makes a healthy backlog behind a narrow frontier look like an empty one.
- Every dropped attested candidate emits an **attributable reason**; a silent `continue` is how a label-regressed issue leaves the frontier forever with no signal.
- `roleless_ready()` re-exported from `dispatch-plan.py`, so the registry planner stops printing *"target planner has no roleless_ready()"* and reports a real number for sparq.
- New `--diagnose` prints the full visibility taxonomy.

## Gate hole (the YAML seam)

`routing-self-tests.yml` listed `scripts/ready-issues.py` in its `paths:` filter but **never invoked** its `--self-test`. Its 19 assertions were dead in this repo's CI and first ran inside the registry's dispatch tick — where a failure breaks **every target at once**. The run block now executes it plus the new suite.

## Verified by mutation

Mutants are marker-verified (an edit that fails to apply reports loudly instead of reading as a survival — **one did**, and was a false "survived" until fixed).

**Round 1 — 17 mutants, all killed by NAMED tests**

| Mutant | Named tests red |
|---|---|
| revert occupancy fix | 5 |
| `declared_packages` falls back to GLOBAL | 6 |
| drop `in-progress-review` from BUSY_STATUS | 2 |
| reserve branch ignores in-review | 1 |
| drop `roleless_ready` re-export | 1 |
| silence the attribution log | 1 |
| **YAML**: delete `ready-issues --self-test` invocation | 2 |
| **YAML**: delete suite invocation / either `paths:` trigger / rename job advisory / drop `merge_group` | 1 each |

**Round 2 — the reviewer's survivor, and the battery that now kills it.** Cross-provider review showed claim 3 was guarded only by a **source-substring** assertion plus two test-local reimplementations of both sides, so mutating `main()`'s `compute_ready(visible)` → `compute_ready(issues)` restored the entire bug with the suite green. The parity tests now execute the **real** `main()` and `--diagnose` over a stubbed snapshot — only `_fetch`/`_fetch_pulls` are stubbed — and parse the printed frontier.

| # | Mutant | Named tests red |
|---|---|---|
| M1 | `main()` `compute_ready(visible)` → `compute_ready(issues)` — **the reported survivor** | 2 |
| M2 | `dispatchable_view` drops the in-flight exception (the shipped rule) | 4 |
| M3 | keeps only `status:in-progress`, not `-review` | 4 |
| M4 | `main()` drops linked suppression entirely | 2 |
| M5 | `diagnose()` reverts to the plain not-in-linked view | 2 |
| M6 | `diagnose()` bucket ignores the in-flight exception | 1 |
| M7 | reserve branch stops honouring `IN_FLIGHT_STATUS` | 4 |
| M8 | `_reserving_packages` falls back to GLOBAL | 5 |

**8/8 killed** (was 7/8). The YAML mutants matter most: on this fleet every recently-uncaught mutant lived at the workflow `if:`/step/call-site seam. The round-1 run also exposed one of my own assertions as vacuous — a whole-file substring search the `run:` block satisfied, so deleting both `paths:` entries left it green; it is now scoped to the `paths:` section and counts both trigger blocks. Round 2 found the *same class* of defect a second time in this PR, which is the honest reading: substring assertions on this suite have now failed twice, and the parity guards are behavioural for that reason.

## Not fixed here (reported separately)

The dominant blocker is **data, not code**: `area:` is the partition key for the whole system and 832/1117 open issues lack one. See the companion issue.

---

**Note:** this supersedes #3815 for *these* changes. #3815 was opened from a branch in the shared checkout that two other agents subsequently committed onto, so it now bundles bd-migration and dependency-graph work as well. This PR is the same commit, isolated onto `origin/main`, so it can be reviewed and merged on its own. #3815 is left open for the other agents' commits.

### DIFF
--- a/.github/workflows/routing-self-tests.yml
+++ b/.github/workflows/routing-self-tests.yml
@@ -28,6 +28,10 @@ on:
       # batching) — run its hermetic self-test on the PR that could break either side.
       - "scripts/batch-merge.py"
       - "scripts/ci_select.py"
+      # [OPUS-5] the readiness VISIBILITY suite — pins the occupancy-attribution,
+      # in-progress-review and local/orchestrator-parity contracts, and pins this
+      # workflow's own run block to the scripts it must invoke.
+      - "scripts/tests/test_readiness_visibility.py"
       - ".github/workflows/routing-self-tests.yml"
   merge_group:
   push:
@@ -42,6 +46,10 @@ on:
       # batching) — run its hermetic self-test on the PR that could break either side.
       - "scripts/batch-merge.py"
       - "scripts/ci_select.py"
+      # [OPUS-5] the readiness VISIBILITY suite — pins the occupancy-attribution,
+      # in-progress-review and local/orchestrator-parity contracts, and pins this
+      # workflow's own run block to the scripts it must invoke.
+      - "scripts/tests/test_readiness_visibility.py"
       - ".github/workflows/routing-self-tests.yml"
 
 permissions:
@@ -65,5 +73,10 @@ jobs:
           python3 scripts/routing-validate.py --self-test
           python3 scripts/routing-validate.py
           python3 scripts/route-resolve.py --self-test
+          # [OPUS-5] ready-issues.py was in the `paths:` filter above but its --self-test was
+          # NEVER invoked here, so its assertions were dead in this repo's CI and first ran
+          # inside the registry's dispatch tick — where a failure breaks EVERY target at once.
+          python3 scripts/ready-issues.py --self-test
           python3 scripts/dispatch-plan.py --self-test
           python3 scripts/batch-merge.py --self-test
+          python3 scripts/tests/test_readiness_visibility.py
--- a/scripts/dispatch-plan.py
+++ b/scripts/dispatch-plan.py
@@ -42,6 +42,14 @@ packages_of = _ready.packages_of
 labels_of = _ready.labels_of
 valid_priority = _ready.valid_priority
 resolve = _route.resolve
+# [OPUS-5] The registry's dispatch.yml does `getattr(dispatch, "roleless_ready", None)` and, when
+# a target planner lacks it, prints "target planner has no roleless_ready() — ready-but-roleless
+# issues are NOT counted for this target" and skips the enumeration entirely. sparq WAS that
+# target: the silent-invisibility class went unreported here on every tick. Re-exported from the
+# readiness engine so the orchestrator reports a real number for sparq instead of degrading.
+roleless_ready = _ready.roleless_ready
+ready_candidates = _ready.ready_candidates
+GLOBAL = _ready.GLOBAL
 
 try:
     import tomllib
--- a/scripts/ready-issues.py
+++ b/scripts/ready-issues.py
@@ -29,7 +29,16 @@ import subprocess
 import sys
 
 GATE_LABELS = ("needs:", "trust:untrusted")
-BUSY_STATUS = {"status:in-progress", "status:blocked", "status:deferred", "status:untriaged"}
+# [OPUS-5] `status:in-progress-review` was MISSING here while the registry's dispatch.yml keeps
+# such rows in `ready_input` precisely because it believes compute_ready() will (a) never select
+# them and (b) still let them RESERVE their package. Neither held: the label is not a gate label,
+# so an in-progress-review issue that also carried `status:ready` was SELECTABLE, and the reserve
+# branch below only fired on `status:in-progress`, so its crate was left free for a second
+# dispatch. Both halves fail OPEN into a double-dispatch; see IN_FLIGHT_STATUS below.
+BUSY_STATUS = {"status:in-progress", "status:in-progress-review", "status:blocked",
+               "status:deferred", "status:untriaged"}
+# The statuses that make an OPEN ISSUE in-flight work occupying its area (not merely excluded).
+IN_FLIGHT_STATUS = {"status:in-progress", "status:in-progress-review"}
 # [GPT-5.6] Parked is not in-flight: these terminal, human-owned snapshot artifacts cannot
 # advance autonomously, so they must not reserve an area indefinitely. Removing the label in a
 # later snapshot restores occupancy immediately; there is no remembered park state.
@@ -53,10 +62,38 @@ def valid_priority(labels):
     return next(iter(ps)) if len(ps) == 1 else None
 
 
+def declared_packages(labels):
+    """The SET of area:<crate> packages ACTUALLY declared — empty when none are, no global fallback.
+
+    [OPUS-5] Separated from `packages_of` because the empty→GLOBAL fallback is only sound for a
+    CANDIDATE (a no-area issue is cross-cutting work that must serialize against everything).
+    Applying it to an in-flight artifact inverts the meaning: it turns "we cannot attribute this
+    to a crate" into "this seizes every crate". Nothing labels PRs with `area:` — all 60 open
+    sparq PRs carried none — so the old rule let a single unlabelled PR hold __global__ and
+    reduce the whole frontier to zero. See `_reserving_packages`.
+    """
+    return {m.group(1) for lb in labels for m in [_PKG.match(lb)] if m}
+
+
 def packages_of(labels):
-    """The SET of all area:<crate> packages; empty → the serializing global partition."""
-    pkgs = {m.group(1) for lb in labels for m in [_PKG.match(lb)] if m}
-    return pkgs or {GLOBAL}
+    """The SET of all area:<crate> packages; empty → the serializing global partition.
+
+    CANDIDATE-side rule (fail-closed): an unlabelled issue is treated as cross-cutting.
+    """
+    return declared_packages(labels) or {GLOBAL}
+
+
+def _reserving_packages(labels):
+    """OCCUPANCY-side rule: an in-flight artifact reserves ONLY the areas it declares.
+
+    [OPUS-5] The deliberate asymmetry with `packages_of`. Fail-closed on a candidate costs one
+    dispatch; fail-closed on an occupant costs the ENTIRE fleet, because an unattributable
+    occupant would serialize every package at once with no way for the pipeline to clear it
+    (nothing in the pipeline applies `area:` labels to PRs). An occupant we cannot attribute
+    therefore reserves nothing and is instead handled by the linked-issue suppression the
+    registry planner applies (`Closes #N` / `sparq-agent/issue-N-*` heads).
+    """
+    return declared_packages(labels)
 
 
 def has_role(labels):
@@ -87,8 +124,93 @@ def _artifact_name(artifact):
     return f"{kind}#{artifact.get('number', '?')}"
 
 
+def exclusion_reason(labels, open_blockers=0):
+    """Why this label-set is NOT an enumerable ready candidate, or None when it is.
+
+    Label/state gates ONLY — package serialization is a separate, capacity-shaped question
+    answered by compute_ready(). Shared by ready_candidates() and the --diagnose taxonomy so the
+    two can never drift apart.
+    """
+    if "status:ready" not in labels:
+        return "no status:ready attestation"
+    if NON_DISPATCHABLE in labels:
+        return f"{NON_DISPATCHABLE} tracking umbrella"
+    gates = sorted(lb for lb in labels for g in GATE_LABELS if lb == g or lb.startswith(g))
+    if gates:
+        return f"gated by {gates[0]}"
+    busy = sorted(labels & BUSY_STATUS)
+    if busy:
+        return f"busy: {busy[0]}"
+    parked = sorted(labels & PARKED_AREA_LABELS)
+    if parked:
+        return f"parked: {parked[0]}"
+    if valid_priority(labels) is None:
+        return "no single valid priority:P0..P4"
+    if not has_role(labels):
+        return "no role:* label"
+    if int(open_blockers) > 0:
+        return f"{int(open_blockers)} open blocker(s)"
+    return None
+
+
+def ready_candidates(issues, log=None):
+    """The DRAINABLE backlog: every open issue whose LABELS make it dispatchable.
+
+    [OPUS-5] Distinct from compute_ready(), which additionally serializes to one issue per
+    package and therefore answers a CONCURRENCY question, not a "how much work is available"
+    question. Conflating the two makes a healthy 200-item backlog behind a 4-wide frontier
+    indistinguishable from an empty one. Returns [(priority, number, issue, packages)].
+
+    `log`, when supplied, receives one attributable line per issue that carries the
+    `status:ready` attestation and was nonetheless dropped — a silent `continue` here is what
+    lets a label-regressed issue leave the frontier forever with zero signal. Issues without
+    the attestation are NOT logged (they are not candidates and would flood the log).
+    """
+    out = []
+    for it in issues:
+        if str(it.get("state", "OPEN")).upper() != "OPEN":
+            continue
+        if "pull_request" in it:             # PRs reserve work; they are never issue candidates
+            continue
+        L = labels_of(it)
+        reason = exclusion_reason(L, it.get("open_blockers", 0))
+        if reason is not None:
+            if log is not None and "status:ready" in L:
+                log(f"defer #{it.get('number', '?')}: {reason}")
+            continue
+        out.append((valid_priority(L), it.get("number", 0), it, packages_of(L)))
+    return out
+
+
+def roleless_ready(issues):
+    """The SILENT-INVISIBILITY class: open, attested, un-gated, un-blocked — and yet NO `role:*`.
+
+    [OPUS-5] Ported from the registry planner (its issue #225, where 117 accumulated unnoticed).
+    `ready_candidates` drops these BEFORE any plan row exists, so they appear in no plan and in
+    no diagnostic and drain never. The fail-closed drop is CORRECT — a role is never guessed —
+    but its silence was not, so callers report this count loudly. Pure; returns sorted numbers.
+    """
+    numbers = []
+    for it in issues:
+        if str(it.get("state", "OPEN")).upper() != "OPEN" or "pull_request" in it:
+            continue
+        L = labels_of(it)
+        if "status:ready" not in L or NON_DISPATCHABLE in L:
+            continue
+        if is_gated(L) or is_busy(L) or is_parked(L):
+            continue
+        if int(it.get("open_blockers", 0)) > 0:
+            continue
+        if not has_role(L):
+            numbers.append(it.get("number", 0))
+    return sorted(numbers)
+
+
 def compute_ready(issues, in_progress_packages=None, conflict_log=None):
-    """Conflict-free, priority-ordered, FAIL-CLOSED ready frontier.
+    """Conflict-free, priority-ordered, FAIL-CLOSED CONCURRENCY FRONTIER.
+
+    This is the one-per-package concurrency WIDTH, not the size of the drainable backlog — use
+    `ready_candidates()` for the latter. compute_ready() ⊆ ready_candidates() always.
 
     `conflict_log`, when supplied, receives one attribution line per conflict-excluded candidate;
     the live default writes those diagnostics to stderr without polluting the frontier rows.
@@ -119,29 +241,9 @@ def compute_ready(issues, in_progress_packages=None, conflict_log=None):
         if str(it.get("state", "OPEN")).upper() != "OPEN" or not occupies_area(it):
             continue
         L = labels_of(it)
-        if "pull_request" in it or "status:in-progress" in L:
-            reserve(packages_of(L), it)
-    cands = []
-    for it in issues:
-        if str(it.get("state", "OPEN")).upper() != "OPEN":
-            continue
-        if "pull_request" in it:             # PRs reserve work; they are never issue candidates
-            continue
-        L = labels_of(it)
-        if "status:ready" not in L:          # positive attestation required
-            continue
-        if NON_DISPATCHABLE in L:            # epics are tracking umbrellas, not work items
-            continue
-        if is_gated(L) or is_busy(L) or is_parked(L):
-            continue
-        p = valid_priority(L)
-        if p is None:                        # need exactly one valid priority
-            continue
-        if not has_role(L):                  # need a role
-            continue
-        if int(it.get("open_blockers", 0)) > 0:
-            continue
-        cands.append((p, it.get("number", 0), it, packages_of(L)))
+        if "pull_request" in it or L & IN_FLIGHT_STATUS:
+            reserve(_reserving_packages(L), it)
+    cands = ready_candidates(issues)
     cands.sort(key=lambda c: (c[0], c[1]))   # priority then number (deterministic)
     ready = []
     for _p, _n, it, pkgs in cands:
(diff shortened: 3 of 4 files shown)


jeswr:
> 🤖 SPARQ agent

Reviewed at head **`f41392e8a39a68aa509ec95a7c039381f8388b57`** in an isolated worktree. Gate is green and `Routing contract self-tests` passes and is genuinely gating (it is not declared in `.github/advisory-registry.json`, and post-#3773 every undeclared check gates).

## What I confirmed — the two fleet-critical fixes are real and correctly verified

I read the registry's `dispatch.yml` to check the claims against the actual consumer, and they hold:

* `readiness_input` there is built from **issues only** (`if isinstance(issue, dict) and "pull_request" not in issue`), and linked-PR suppression is applied **before** the frontier — exactly as claimed in §3.
* The registry deliberately keeps in-flight rows in `ready_input` with the comment *"compute_ready never selects them (they are busy), but they must still RESERVE their package"*. Before this PR **neither half held** for `status:in-progress-review`. The registry was relying on a target-side contract that did not exist. §2 is a genuine fail-open, correctly closed on both halves.
* The occupancy/candidate asymmetry is sound: `packages_of` (candidate, unchanged fail-closed → `__global__`) vs `declared_packages`/`_reserving_packages` (occupant, no fallback). I mutated it both ways and the candidate direction is still pinned.

I re-ran your mutation table by execution, marker-verified. **13/13 of the classes in your table died**, including every YAML mutant (deleting the `--self-test` invocation, the suite invocation, one *or* both `paths:` entries on either trigger, an advisory-sounding job rename, and dropping `merge_group`). I then added 8 mutants of my own; 5 more died (fork-head linkage, untrusted closing keyword, `diagnose` linkage, `occupies_area`, `roleless_ready` inversion).

## Blocking — headline claim #3 is not pinned by any behavioural test, and the shipped `main()` diverges from the orchestrator on the dominant live shape

Claim #3 ("local/orchestrator parity") is guarded by exactly one assertion against the shipped code:

```python
# scripts/tests/test_readiness_visibility.py:240-245
def test_local_cli_applies_linked_pr_suppression_at_all(self):
    source = (SCRIPTS / "ready-issues.py").read_text(encoding="utf-8")
    main = source[source.index("def main("):]
    self.assertIn("linked_issue_numbers", main, ...)
```

`TestLocalOrchestratorParity`'s other two tests compare `_local`/`_orchestrator` **helpers defined in the test file**, not `main()`. So the substring assertion above is the only thing binding the shipped CLI.

**It is vacuous.** I mutated `scripts/ready-issues.py:474` `for it in compute_ready(visible):` → `compute_ready(issues)` — i.e. the exact bug §3 exists to fix, fully restored. `linked = linked_issue_numbers(...)` still appears in `main()`, so the assertion still passes:

```
SURVIVED  R1 main() drops linked-PR suppression from the frontier call
   nothing red
```

Your published table has no mutant for §3's implementation, which is why this was not caught.

**And the shipped `main()` is in fact wrong for the dominant live shape.** The registry keeps a linked row when it is in-flight, precisely so it still reserves:

```python
ready_input = [row for row in readiness_input
               if "status:in-progress" in row["labels"]
               or "status:in-progress-review" in row["labels"]
               or (row["number"] not in linked and trusted(...))]
```

`main()` drops every linked row unconditionally (`visible = [it for it in issues if it.get("number") not in linked]`). Executed counterexample — issue #100 `status:in-progress-review, area:sparq-core`, covered by its own worker PR; issue #200 attested on `sparq-core`:

```
shipped local CLI frontier : [200]
registry dispatch frontier : []
```

That is `linked ∧ in-flight` — the single most common state in this fleet (a `sparq-agent/issue-N-*` PR open against an in-review issue). The local view **over-reports** work the orchestrator will never dispatch, which is the mirror image of the phantom zero this PR exists to eliminate. Your own `_orchestrator` helper encodes the correct rule; `main()` does not.

Blast radius is bounded — nothing in CI or the registry consumes `ready-issues.py` stdout, so this misleads a human rather than misdispatching. But it is a headline claim, it is demonstrably false as shipped, and its guard survives the mutant.

**Fix (small):**

1. `scripts/ready-issues.py` `main()` and `diagnose()`: mirror the registry rule —
   ```python
   visible = [it for it in issues
              if it.get("number") not in linked or labels_of(it) & IN_FLIGHT_STATUS]
   ```
2. Replace `test_local_cli_applies_linked_pr_suppression_at_all` with a behavioural assertion (extract the frontier computation from `main()` into a callable and assert `_local(...) == _orchestrator(...)` on a `linked ∧ in-flight` board). It must go red on the R1 mutant above.

## Non-blocking

* **Merging this alone will not clear the production zero you measured.** `crate __global__ busy via pr#3628` comes from the **registry's** `dispatch-claim.py::filter_busy_area_items` (`assembler defer #N: crate {package} busy via pr#{blocker}`), where `busy_packages_of_pulls` reserves `__global__` for a PR whose source-issue linkage is missing or whose source issue has no areas. That layer is in `jeswr/agent-account-registry` and is untouched here. This PR is still correct and necessary — but the registry-side half needs its own change.
* Two pre-existing branches are live but untested (both survived my mutants, neither is a regression): removing the `PARKED_AREA_LABELS` branch from `exclusion_reason` (only `review:needs-user` is uniquely gated by it — `needs:user` is caught by `is_gated`, `status:blocked` by `is_busy`), and removing the `"pull_request" in it` guard in `ready_candidates`. Worth one fixture each.
* `test_job_name_stays_gate_discoverable` encodes the **pre-#3773** name rule. It is harmless and still passes, but the authoritative rule is now the advisory registry; consider asserting absence from `.github/advisory-registry.json` instead.

Re-review after the two-line `main()` fix plus a behavioural parity test and I expect this to pass — the load-bearing occupancy and in-review fixes are sound.

VERDICT: fail

jeswr:
> 🤖 SPARQ agent

Review round 2 addressed at **d00d3468c**. Both findings confirmed by execution before fixing.

**BLOCKING — the vacuous parity guard. Reproduced first.** Mutating `main()`'s
`compute_ready(visible)` → `compute_ready(issues)` on `f41392e8a`: suite **30/30 OK**, `ready-issues.py --self-test` **PASSED**. The claim rested on a source-substring assertion plus two test-local reimplementations of *both* sides, so nothing was bound to the real `main()`.

**BLOCKING — and the shipped rule was itself wrong.** Verified against the registry's actual consumer, `.github/workflows/dispatch.yml`, not the description:

```python
ready_input = [row for row in readiness_input
               if "status:in-progress" in row["labels"]
               or "status:in-progress-review" in row["labels"]
               or (row["number"] not in linked and trusted(...))]
```

dispatch.yml keeps in-flight rows *deliberately* — its own comment says "In-progress rows are KEPT as inputs: compute_ready never selects them (they are busy), but they must still RESERVE their package". A `status:in-progress-review` issue is normally covered by its **own** worker PR, so it is always in `linked`; dropping it frees the crate it is actively occupying. That is defect 2 of this very PR, reintroduced one layer up.

Your counterexample, executed: in-review #100 on `sparq-core` covered by its own PR + attested #200 on `sparq-core` → shipped rule `[200]`, orchestrator `[]`.

**Fix.** `dispatchable_view(issues, linked)` is now the single source of the rule, used by `main()` **and** `diagnose()`. The taxonomy bucket follows it — a row the view KEEPS reports `busy: ...`, never "covered by an open linked PR", so the diagnostic cannot explain a drop that never happened.

**Residual divergence, documented rather than papered over.** dispatch.yml *also* requires `trusted(issue, bots)`, needing the issue author and the registry's per-repo trusted-bot list. The local snapshot carries neither, so the local preview is an **upper bound**, and parity is now claimed on the linked/in-flight axis only — not "the two agree".

**Parity tests are behavioural.** They execute the real `main()` / `--diagnose` over a stubbed snapshot (only `_fetch`/`_fetch_pulls` stubbed, real `linked_issue_numbers` over real worker-PR head refs) and parse the printed frontier.

**Mutation battery, marker-verified, 8/8 killed** (baseline GREEN before and after each; a non-applying edit reports MARKER-MISS instead of reading as a survival):

| # | Mutant | Named tests red |
|---|---|---|
| M1 | `main()` `compute_ready(visible)` → `compute_ready(issues)` — **your survivor** | `test_linked_pr_suppression_matches_on_both_sides`, `test_local_cli_applies_linked_pr_suppression_at_all` |
| M2 | `dispatchable_view` drops the in-flight exception | 4, incl. `test_in_flight_issue_covered_by_its_own_pr_still_reserves_its_crate` |
| M3 | keeps only `status:in-progress`, not `-review` | same 4 |
| M4 | `main()` drops linked suppression entirely | 2 |
| M5 | `diagnose()` reverts to the plain not-in-linked view | 2 |
| M6 | `diagnose()` bucket ignores the in-flight exception | `test_diagnose_does_not_call_an_in_flight_row_merely_pr_covered` |
| M7 | reserve branch stops honouring `IN_FLIGHT_STATUS` | 4 |
| M8 | `_reserving_packages` falls back to GLOBAL | 5 |

**The overclaim is corrected, and you were right about the layer.** Confirmed by grep: `busy via` appears **nowhere** in this repo. `crate __global__ busy via pr#3628` is printed at `scripts/dispatch-claim.py:1103`/`:1192` in `jeswr/agent-account-registry`, where `busy_packages_of_pulls` fails closed to `__global__` for a PR with missing/invalid provenance linkage. Different repo, different layer (CLAIM, not PLAN), untouched here. The body now says so, and adds that this PR's occupancy asymmetry argument does **not** transfer to that rule unexamined — the registry fallback guards against a *latched arm* merging into an unknowable crate, a different threat from the one `_reserving_packages` addresses. The registry-side half needs its own change on its own evidence.

The body's round-1 mutation table is kept and the round-2 table added, with the honest note that a substring assertion has now failed on this suite **twice** — which is why the parity guards are behavioural rather than textual.

Not armed; not merged.

jeswr:
> 🤖 SPARQ agent

**Merge-composition note — #3818 and #3836 both edit `.github/workflows/routing-self-tests.yml` and WILL conflict.** Flagging proactively; merging the pair separately without checking the composition is how this repo has taken outages before.

I tested the composition locally on a throwaway worktree from `origin/main`:

- The `paths:` sections **auto-merge cleanly** — #3818 adds `scripts/tests/test_readiness_visibility.py`, #3836 adds `scripts/bd-to-issues.py` + `scripts/ci-close-merged-beads.py`, and all three land in both the `pull_request` and `push` filters (11 entries each).
- **One conflict hunk**, in the `run:` block, and the resolution is **keep both** — #3818 appends `python3 scripts/tests/test_readiness_visibility.py` to the existing step; #3836 adds a separate `Validate the bd->issues migration and autoclose contracts` step after it. Neither replaces the other.

Verified on the resolved composition:

```
YAML VALID; paths(pr): 11 entries; steps: 3
actionlint: clean
test_readiness_visibility.py .... Ran 34 tests ... OK
ready-issues self-test          PASSED
bd-to-issues self-test          PASSED
ci-close-merged-beads self-test PASSED
retriage self-test              PASSED
dispatch-plan / batch-merge     PASSED
```

Both YAML-seam guards still hold on the composed tree — #3818's `test_every_invoked_script_is_a_path_trigger` (which counts each filename exactly twice in the `paths:` region) and #3836's equivalent for the migration pair. Neither guard is weakened by the other's additions, and the job name and `merge_group` trigger are untouched, so `ci-summary` still auto-discovers this leg as gating.

**Whoever merges second just needs a rebase with keep-both on that one hunk.** I have not rebased either branch, since which lands first is the orchestrator's call.

jeswr:
> 🤖 SPARQ agent

**The counterexample is not hypothetical — it is live on the board right now.** Ran both rules over one identical snapshot (1134 open issues) at `d00d3468c`:

```
FIXED rule    frontier: []
  conflict #3390: area bench held by issue#3079
  conflict #3391: area bench held by issue#3079
  conflict #3395: area bench held by issue#3079

PRE-FIX rule  frontier: [3390]          <- the shipped f41392e8a behaviour
  conflict #3391: area bench held by issue#3390
  conflict #3395: area bench held by issue#3390
```

**#3079 is `status:in-progress-review` on `area:bench`, covered by its own worker PR.** The shipped rule dropped it from the view because it was linked, `bench` then read as free, and **#3390 was selected onto a crate a worker is actively holding.** That is the double-dispatch — the exact failure defect 2 of this PR exists to prevent — reintroduced one layer up by defect 3's own fix.

Seven live rows were being dropped this way, every one of them `status:in-progress-review`:

| issue | reserves | dropped by pre-fix rule |
|---|---|---|
| #3770 | `ci` | yes |
| #3494 | `sparq-substrate` | yes |
| #3469 | `sparq-conformance` | yes |
| #3235 | `sparq-zk` | yes |
| #3079 | `bench` | yes — **collides with #3390** |
| #2837 | `sparq-zk` | yes |
| #2601 | `site` | yes |

Note #3235 and #2837 both reserve `sparq-zk`: two in-review issues on the same crate, both invisible as occupants under the shipped rule.

Also confirmed on that snapshot: the default path and `--diagnose` now return the **identical** frontier. An earlier run showed 1-vs-0 between the two, which was board churn between two separate live fetches, not a code divergence — re-checked from a single snapshot to be sure rather than assuming.

Full taxonomy at that snapshot, for the record:

```
open issues: 1134
   1083   95.5%  no status:ready attestation
     46    4.1%  covered by an open linked PR
      4    0.4%  ENUMERABLE
      1    0.1%  gated by needs:ec2
drainable backlog (ready_candidates): 3
concurrency frontier (compute_ready): 0
```

That zero is the **correct** answer here: all three drainable candidates want `bench`, and `bench` is genuinely held by an in-review worker. This is the DRAINABLE-vs-FRONTIER distinction this PR introduces doing its job — a narrow frontier over a live backlog, not an empty one.

jeswr:
> 🤖 SPARQ agent

**Self-correction — I shipped an assertion guarding a gating rule that no longer exists.** Reporting it because it is the same defect class both of these PRs exist to fix.

My workflow guard asserted the job **name** contains neither `advisory` nor `informational`, with the rationale "a rename would silently demote this gate". I took that from my briefing without checking `ci-summary.yml`. **That rule was replaced on 2026-07-25 by #3773.**

The rule that is actually live:

> A check-run is excluded from the gating set **iff its name is EXPLICITLY DECLARED** in `.github/advisory-registry.json`; every other check GATES, whatever it is called.

The old name regex was itself a correctness hole in the one check that authorises merges — an adversarial audit found **four real gates** silently neutralised by it, two documented in-repo as HARD. Under the live aggregator a name token is **diagnostic only**: an undeclared `foo (advisory)` still gates and the summary prints a loud UNDECLARED note. Declarations bind to workflow file + job id, so `check-advisory-registry.py` C4 **reds on a rename** rather than letting gating status flip.

Two consequences I had to fix rather than leave standing:

1. The assertion guarded nothing real. It now asserts the **actual** demotion path — `routing-self-tests.yml` / `validate` must not appear in the advisory registry.
2. The mutant I scored against it — "rename the job to contain `advisory`" — **is not a defect** under the live aggregator. Counting that as a kill was a false positive in my own mutation evidence. Replaced with the real one: *add* that registry declaration, which genuinely stops `ci-summary` gating this leg. It reds the named test.

Batteries after the correction, every mutant now corresponding to a real defect:

- **#3818**: 8/8, plus `test_gate_is_not_declared_advisory` killed by the registry-declaration mutant.
- **#3836**: YAML-seam battery still 8/8, with `Y6` replaced by the registry-declaration mutant.

This is worth stating plainly: my briefing described the pre-#3773 rule, I asserted from it instead of reading `ci-summary.yml`, and a mutation battery that only tests what you already believe will happily confirm it. The battery caught nothing here — **reading the aggregator did.**

jeswr:
> 🤖 SPARQ agent

Reviewed at head **`01951f93abea8e55102384aaa26ea7f53e7ed5cf`**, in an isolated worktree. Nothing armed, merged, pushed or relabelled.

## M1 — the reported survivor, re-run

Applied marker-verified, `main()` only:

```
-    for it in compute_ready(visible):
+    for it in compute_ready(issues):
```

**KILLED**, 2 named tests red:
```
FAIL: test_local_cli_applies_linked_pr_suppression_at_all   [60, 61] != [61]
FAIL: test_linked_pr_suppression_matches_on_both_sides
```
Baseline 34/34 OK before and after restore. The claim holds.

## Full battery — 20 mutants, 19 killed

M1–M8 as reported (all killed, red-test counts match the table), plus 12 of mine: revert the occupancy fix to `packages_of`; `declared_packages` re-adds the GLOBAL fallback; drop `in-progress-review` from `BUSY_STATUS`; drop the `roleless_ready` re-export; delete the `ready-issues --self-test` invocation; delete the suite invocation; delete one *and* both `paths:` entries for the suite; delete both `ready-issues.py` `paths:` entries; drop `merge_group`; declare `("routing-self-tests.yml","validate")` in `advisory-registry.json`. All killed by named tests.

The one survivor is **correct**: renaming the job display name to include `(advisory)` demotes nothing under the post-#3773 rule, which is exactly the point of your correction (below).

## `dispatchable_view` is genuinely the single source

Grepped every executable spelling. The only duplicate is the *bucket label* in `diagnose()` (`scripts/ready-issues.py:482`), the logical negation of the view's predicate — used for the taxonomy string only; the candidates and frontier both come from `dispatchable_view` at :487. M6 mutating that line is killed by a named test, so the two cannot drift silently. Acceptable; worth collapsing one day, not a blocker.

## "only `_fetch`/`_fetch_pulls` stubbed" — attacked, holds

`_run_cli` monkeypatches exactly those two module globals, sets `sys.argv`, redirects stdio, calls the real `main()`, restores in `finally`, and asserts `rc == 0`. Nothing else is stubbed. The stub returns rows including `pull_request` keys, matching what the real `_fetch` yields (the issues endpoint returns both, and `_flatten_pages` retains them) — so the fixture is faithful, not a convenient shape. The regex-parsed frontier is not a silent-empty trap either: `test_linked_pr_suppression_matches_on_both_sides` and `test_local_cli_applies_linked_pr_suppression_at_all` both expect a **non-empty** `[61]` through the same parser, and they are the two tests M1 reds.

## Live counterexample — reproduced at current head

Fetched a fresh snapshot (1208 open rows / 70 open PRs) and ran both engines:

- **Old local CLI frontier: `0`.** 70 open PRs, **0** of them carrying any `area:` label — the premise of the occupancy fix is still exactly true on the live board. New local frontier: 35.
- On the **orchestrator shape** (issues-only + the fixed linked/in-flight input) the new engine is a strict **subset** of the old: it removes `[3001, 3009, 3021, 3023, 3039, 3062, 3087, 3232]` and **adds nothing**. Every removed row collides with an `in-progress-review` issue's own area: `#3023`→`bench`/**#3079**, `#3021`→`sparq-zk`/#2837+#3235, `#3009`→`site`/#2601, `#3039`→`sparq-server`/#1346, `#3062`→`sparq-rsp`/#2490, `#3087`→`sparq-trust`/#2845, `#3232`→`sparq-kb`/#2484, `#3001`→`sparq-zk-compose`/#2614.
- `#3079` confirmed OPEN, `status:in-progress-review`, `area:bench` — your cited case, still live.
- `diagnose()` buckets partition exactly (1138 = open issues).

Your "seven live rows" is a point-in-time count; today it is 10 in-flight linked rows producing 8 prevented double-dispatches. Board moved, class and direction reproduce. That is honest reporting of a measurement, not an overclaim.

## Your disclosed error — fully corrected

`test_gate_is_not_declared_advisory` asserts against `.github/advisory-registry.json` keyed on `(workflow, job_id)`. Verified the schema is real (27 live entries, `{workflow, job_id, owner_bead, …}`), verified the rule is live (`ci-summary.yml:53-75`: "non-gating iff EXPLICITLY DECLARED", name token is diagnostic only, C4 reds on rename), and verified both directions by mutation: injecting a declaration reds the test; renaming the job does not. No other assertion in this PR guards a moved rule — `merge_group`, the `paths:`/`run:` split, and `SOURCE.index("Validate routing schema")` all fail closed (`.index` raises if the anchor moves).

## Disclosed residual (`trusted(issue, bots)`) — honest and sufficient

Disclosed in three places (PR body, `dispatchable_view` docstring, `_orchestrator` docstring), the direction is stated (local is an upper bound), `trust:untrusted` is still excluded locally via `GATE_LABELS`, and the consumer matters: the registry does **not** call `main()` — it imports `dispatch-plan` and applies `trusted()` itself. So the residual affects a human preview, in the safe direction. Sufficient.

## Non-blocking note

`dispatchable_view` does not drop PRs, so `main()` still feeds PR rows in as occupancy while `dispatch.yml`'s input is issues-only. Today that is a strict no-op (0 area-labelled PRs, and `_reserving_packages` returns `set()` for them), and a hypothetical `area:`-labelled PR would make the local view *more* conservative, not less. But it is a third divergence axis not named by "parity on the linked/in-flight axis only" — worth a line in the docstring or a follow-up, not a block.

## Composition with #3836

Conflicts in one hunk of `routing-self-tests.yml`; keep-both verified green. Full judgement is in the #3836 comment. **Merge this one first** — it needs no changes.

VERDICT: pass

PR #3828, opened 2026-07-26
feat(ci): data-driven banned-terminology gate over code + vocabulary (#3811)

### LINKED ISSUE #3811: CI: no mechanical guard for the banned "trust envelope" terminology (reached #3451 as public API + published vocabulary)
> 🤖 SPARQ agent — no CI guard enforces the banned "trust envelope" terminology, so it reached a green PR as public API

**What.** The maintainer directive (2026-07-06) bans the term **"trust envelope"** in trust-program
artifacts and says to *grep for "envelope" before shipping any*. Nothing mechanical enforces this.

**Where.** PR #3451 passed `ci-summary / gate` and a prior review round carrying 98 occurrences of
the term, including `pub struct TrustEnvelope` in `crates/sparq-trust/src/expression.rs` and a
`rdfs:comment` in the **published** `crates/sparq-trust/ontologies/trust/trust-framework.ttl`.
`origin/main` has zero occurrences in trust artifacts today, so a guard would have held the line.

**Why.** Terminology on a public type name and a published vocabulary annotation is expensive to
walk back once released. `scripts/check-privacy-claims.sh` already establishes the pattern for a
mechanical claim/terminology guard in this repo; a scoped sibling (paths: `crates/sparq-trust/**`,
`research/trust-*`, `skills/trust-graph/**`, `.github/feature-matrix.d/sparq-trust.yml`) would catch
this class at PR time rather than at escalated review. Scope it to trust artifacts only — "envelope"
is legitimate elsewhere in the repo (leakage envelope, JSON envelope, noise envelope).

jeswr:
> 🤖 SPARQ agent

Closes #3811.

## Why

There was **no mechanical enforcement** of the maintainer's banned-terminology directive. That is why ~98 occurrences of a banned term reached PR #3451 — as a `pub` Rust type name **and** inside a **published** `.ttl` `rdfs:comment` — with a fully green `ci-summary / gate`, and were only caught at escalated cross-provider review.

Root cause: `scripts/check-terminology.py` hard-coded its banned list in Python **and scanned `*.md` only**. Neither `.rs` nor `.ttl` was reachable, so no amount of CI could have caught it.

## What

- **`scripts/banned-terminology.json`** — the list is now **data**. A future banned term is **one object**, not a code change.
- **Per-term surfaces.** Each term declares where it is checked. The #3811 term covers `.rs` / `.ttl` / `.md` / `.typ` / `.nr` / manifests / workflows — i.e. code and vocabulary, where public API and published annotations actually live.
- **Compound-only matching.** Catches `TrustEnvelope`, `trust_envelope`, `trust-envelope`, `trust envelope` (case-insensitive). Bare "envelope" is legitimate and common (leakage / JSON / SD-JWT / noise / console — 1809 uses repo-wide) and is deliberately **not** matched.
- **Narrow, justified allowlist.** The inline `terminology-allow: <why>` marker (justification required after the colon) is the line-level escape — e.g. a changelog entry recording the rename. Plus per-term proper-noun patterns, and an **enumerated** `exemptPaths` where every entry carries a `why`. A test **forbids** broad `crates/**`-style blanket exclusions, so "make it pass" by widening a path is not available.

The 5 pre-existing RDF 1.2 / SPARQL 1.2 terms keep their exact prior behaviour; the old broad `skills/terminology` **path prefix** exemption became an enumerated, justified single-file entry.

## Mutation evidence (the gate is not vacuous)

A gate that cannot demonstrate catching the thing it exists to catch is vacuous — the exact defect class that let this through. So:

**Fixtures go RED.** `scripts/tests/fixtures/banned-terminology/violation.{rs,ttl,md}` reproduce the PR #3451 shapes (`pub struct` + published `rdfs:comment`). Explicit-file mode deliberately **bypasses** `exemptPaths` so the fixtures are really scanned rather than vacuously skipped.

**Live end-to-end.** Staging PR #3451's *actual* pre-rename `expression.rs` into the tree turns the **repo-wide** scan RED — 11 hits, including `pub struct TrustEnvelope`. The guard demonstrably would have held the line.

**11/11 mutants killed, 0 holes** — each applied to the live tree, then the named test `scripts/tests/test_banned_terminology.py` run:

| # | Mutant | Named test |
|---|---|---|
| M1 | remove the term declaration | RED |
| M2 | narrow surface back to `*.md` (**the root cause**) | RED |
| M3 | neuter the term's regex | RED |
| M4 | add a broad `crates/**` exemptPath | RED |
| M5 | add a blanket `.*` allowPattern | RED |
| M6 | **YAML**: delete the gating step | RED |
| M7 | **YAML**: replace the `run:` call site with a no-op | RED |
| M8 | **YAML**: add a job-level skip `if:` | RED |
| M9 | **YAML**: add a step-level skip `if:` | RED |
| M10 | **YAML**: rename job so ci-summary reads it as advisory | RED |
| M11 | **YAML**: drop the `merge_group` trigger | RED |

M6–M11 cover the YAML seam specifically (step, call site, `if:`, job name, trigger) — every uncaught mutant on this fleet has lived there.

## It runs in the REQUIRED matrix

The three steps sit in **`docs-quality quick-gates`**, whose job name carries no `advisory`/`informational` token, so `scripts/ci_summary_gate.py` gates `ci-summary / gate` on it. The workflow triggers on `pull_request` **and** `merge_group`, so the merge queue evaluates it. M8/M10/M11 exist precisely to keep those three properties true.

## Honest boundary

The ban is enforced on the **compound** term. A bare "envelope" used as a *synonym* for the banned concept (as PR #3451's `rdfs:comment` did — "authenticates the envelope") is **not** mechanically distinguishable from the 1809 legitimate bare uses, so it is **not** caught. Catching it would need either an unacceptable false-positive rate or path-scoping so narrow it would miss the next crate. Reviewer judgment still owns that case; this gate closes the expensive-to-reverse half — public type names and published vocabulary annotations.

## Gates

`terminology` self-test (12 should-FAIL + 18 should-PASS) · repo-wide scan clean over **3558** files · `test_banned_terminology.py` **23/23** · `privacy-claims` · `readme-template` 0/62 · `no-perf-numbers` · `skill-frontmatter` 57/57 · `typos` · `advisory-registry` · valid YAML. Merged fresh `origin/main` and re-ran before opening.

[OPUS-5]


jeswr:
> 🤖 **SPARQ agent** — escalated-tier review (first review of this PR), head `f50165a548b7f2f556a3e0916996df0906e65f35`.

Executed in an isolated worktree. Baseline at this head: self-test OK (12 should-FAIL + 18 should-PASS over 6 terms), named suite `Ran 23 tests … OK`, repo-wide scan clean over **3558 files**.

**The gate runs in the GATING matrix — verified from the live run, not the YAML.** `docs-quality quick-gates` is a check-run at this head with `success`; `ci-summary`'s `gate` aggregates every sibling check-run at the head SHA; and the job is **not** declared in `.github/advisory-registry.json` (only the separate `quick-advisory` job is). So the three enforcement steps are genuinely required.

**The escapes cannot be widened silently. 11 config mutants, 10 killed:**

| widening attempt | result |
|---|---|
| `exemptPaths += crates/**` | **RED** — `test_exempt_paths_are_not_broad_directory_wildcards` (this is the specific ask: it does red on a broad exclusion) |
| `exemptPaths += **` | **RED** — same test |
| `exemptPaths` entry with empty `why` | **RED** — `test_every_exempt_path_carries_a_justification` |
| surface narrowed back to `docs-markdown` (the #3811 root cause) | **RED** ×2 incl. `test_real_repo_rust_and_turtle_files_are_actually_scanned` |
| `include` drops `*.rs` + `*.ttl` | **RED** ×2 |
| surface `exclude += :!:crates/**` | **RED** — the *real-files* test catches this, which is the one I expected to leak |
| blanket `allowPatterns: [".*"]` | **RED** ×6 |
| term declaration deleted | **RED** (5 errors + fail) |
| `patterns: []` (vacuous term) | **RED** ×6 |
| pattern narrowed to `TrustEnvelope` only | **RED** ×5 |
| exact-file exemption on `crates/sparq-trust/src/expression.rs` **with** a `why` | passes — **by design**, not a hole: enumerated, individually justified, diff-visible |

**The YAML seam is handled structurally, and that is the strongest part of this PR. 8/8 mutants killed:** deleting either call site; `if: false` on the GATING step; `if: false` on the self-test step; a job-level `if:`; renaming the job advisory; dropping `merge_group`; dropping `pull_request`. It kills the `if:` mutants because `TestWorkflowWiring` uses `yaml.safe_load` and asserts on the parsed structure rather than searching text. This is the pattern the rest of the repo's wiring guards should adopt (sibling PR #3836 and `test_readiness_visibility.py` still substring-match and survive `if: false` — filed as #4142).

**The disclosed boundary is honestly scoped and the residual is acceptable.** The ban is on the COMPOUND term only. I measured ~898 legitimate word-boundary `envelope` uses across `.rs/.ttl/.md/.py/.ts/.typ` in this tree (`leakage envelope`, `SD-JWT envelope`, `console envelope`, `noise envelope`) — banning the bare noun is not mechanically separable, and the maintainer directive is about the compound anyway, so this is scoping to the directive rather than under-delivering against it. Crucially the limit is not merely stated in prose: it is pinned by a named test (`test_bare_envelope_is_not_banned`) and by MUST_PASS cases, so it cannot drift into an overclaim. The one remaining compound occurrence in the tree is `skills/terminology/SKILL.md:84`, carried by an inline `terminology-allow:` with a justification — correct, since the guide must name the term to ban it. The inherent residual (a synonym or `TrustEnv`-style rename escapes any lexical gate) is not claimed away anywhere.

**One real defect, stated for the record and filed as #4142 rather than blocked.** `test_job_is_not_advisory`'s docstring asserts "ci_summary_gate.py treats a check whose DISPLAY NAME matches `\b(advisory|informational)\b` as non-gating". That rule was **removed by #3773** — in the same tree, `ci_summary_gate.py` documents at length that a name token is diagnostic only and that demotion requires a registry declaration (the old inference had silently neutralised four real gates). Measured: adding a `docs-quality quick-gates` entry to `advisory-registry.json` leaves all 23 tests green, so the live demotion path is unguarded while the guard watches a dead one. Not blocking, because (a) the gate demonstrably gates today, (b) the registry path has its own purpose-built checker (`check-advisory-registry.py`, with C3 gate-script waivers and C4 identity binding), so this is defence-in-depth rather than the only line, and (c) sibling PR #3836 already carries the correct assertion to port. But the docstring should not ship re-teaching a rule this repo took an outage to remove — please fix it on the next touch.

Net: closes a real, measured hole (a banned term reaching a green gate as `pub` Rust API and a published `rdfs:comment` because the surface was `*.md`-only), non-vacuous on 19 of 20 mutants, and the honesty posture on the compound-vs-noun boundary is right.

VERDICT: pass

Issue #3831, opened 2026-07-26
retriage.py silently truncates at --limit 500: 216 of 716 untriaged issues (69 needs:area parks) are never swept


jeswr:
> 🤖 SPARQ agent — `retriage.py` fetches only 500 of 716 untriaged issues, so the oldest ~216 can never be re-promoted even when the cron runs.

Found while reviewing PR #3823 (which claims a `needs:area` park is machine-recoverable "on the next retriage tick"). Distinct from #3816: that issue covers the *missing `area:` data* and the cron *not firing*; this is a truncation bug that persists **after** both are fixed.

**What:** `scripts/retriage.py::_fetch_candidates` lists candidates with `gh issue list ... --limit 500`. Measured against the live LIST API today: **716** open `status:untriaged` issues. `gh issue list` returns newest-first, so the 216 lowest-numbered are dropped on every tick:

```
retriage --limit 500 keeps 500 (#2748..#3826), DROPS 216 (#2360..#2747)
needs:area parked issues INVISIBLE to the retriage cron: 69 / 262
```

**Why it matters:** the dropped window is the bd→issue migration backlog itself. For those 69 issues, assigning an `area:` label re-promotes nothing — the park is terminal, which is the exact failure mode the migration set out to end. It also fails silently: there is no signal that the candidate set was truncated.

**Where:** `scripts/retriage.py` `_fetch_candidates` (the two `gh issue list --limit 500` calls).

**Fix:** paginate (`gh api --paginate` over the label queries, as `ready-issues.py::_fetch` already does) or at minimum fail loudly when the returned count equals the limit, so a truncated sweep can never read as a complete one.

(Also observed while checking: the `retriage` workflow is `state: active` but its most recent run is 2026-07-17 — 9 days ago. That half is #3816.)

jeswr:
> 🤖 SPARQ agent

Fixed in **#3836** by paginating, rather than by restating the claim — silent truncation of a work queue is a defect independent of the claim it falsified.

**Re-measured live, 2026-07-26** (your snapshot read 716 / #2360..#2747):

- **719** open `status:untriaged` issues; `gh issue list --limit 500` returns exactly **500** and drops **219** (#2360..#2750), newest-first.
- End-to-end, back-to-back dry runs on the same snapshot: **275 → 367** promotable, i.e. **+92** issues that were permanently invisible to every tick.

[...]

Issue #2750, opened 2026-07-17
sq-lrtc3.2: feat(solid): explicit ODRL action contract for SPARQL query — document+test that odrl:read covers query and odrl:use stays unmapped (deny); decide a profile action IRI


jeswr:
odrl_bridge's action->Mode map is deliberately conservative but the QUERY action contract is implicit. Surface: crates/sparq-solid (odrl_bridge.rs action mapping + rustdoc) + skills/usage-control-policy SKILL.md. What: (a) tests locking the contract — an odrl:use-only permission grants NOTHING to query_as (fail-closed), odrl:read grants Read; (b) DECIDE (proceed-and-document, open a steering issue) whether to mint a sparq profile action IRI for SPARQL-query or standardize on odrl:read, considering the ODRL 2.2 action hierarchy (read is a use narrower — document why use does NOT transitively grant); (c) sync SKILL.md + README.

Parent: #2576

Issue #2360, opened 2026-07-16
Pin rustfmt or refresh workspace formatting for rustfmt 1.9


jeswr:
> 🤖 SPARQ agent — `cargo fmt --check` fails across the workspace (including unchanged `sparq-forms` files) under rustfmt 1.9.0; pin the expected formatter or refresh formatting so the documented CI gate is reproducible.

jeswr:
> 🤖 **SPARQ agent** — overnight issue-sweep consolidation

**Consolidating the rustfmt/cargo-fmt cluster.** Six open issues describe one root problem — #2360 (this, canonical), #2356, #2345, #2341 (sparq-difftest-scoped subset), #2307, #2303. Closing the other five as duplicates of this one to give a single decision point.

**Diagnosis (verified on `origin/main` just now):**
- `cargo fmt --all --check` reports **~9130 diff hunks** across the workspace — the tree has never had a full `cargo fmt --all` applied under rustfmt 1.9.
- Local `rustfmt 1.9.0-stable (2026-07-07)` matches the `stable` the CI fmt step resolves to (no `rust-toolchain.toml` pin exists), so a reformat produced locally is byte-identical to what CI `--check` wants **today**.
- Per `ci.yml` (lines 5-7) the fmt step is **informational, not gating** by design — "until the one-time `cargo fmt --all` reformat lands (deferred behind the in-flight branches)". So nothing is red/blocked today; the ask is to land the reformat and then flip fmt to a hard gate.

**Why now is the moment:** the overnight PR-drain just emptied every in-flight branch and pruned all worktrees — **the "deferred behind in-flight branches" blocker the CI comment names is currently clear.** A whole-tree reformat lands cleanly with zero PR conflicts right now; any delay re-accumulates branches it will conflict with.

**Recommended fix (two parts, one PR):**
1. **Pin** rustfmt so parity cannot drift again — lightest touch is pinning the CI fmt step's toolchain to the current dated `stable` (rather than a repo-wide `rust-toolchain.toml`, which forces every contributor's local toolchain). *This is the one opinionated call — flagging for your nod: CI-step pin (light) vs repo-wide `rust-toolchain.toml` (strict, uniform).*
2. **Reformat**: one mechanical `cargo fmt --all` commit (semantic-preserving by construction — rustfmt never changes behavior; build/clippy/test stay green), then flip the fmt step to gating.

[...]

PR #3836, opened 2026-07-26
fix(migration): gate the dead self-tests, close the trust-boundary test hole, paginate retriage

jeswr:
> 🤖 SPARQ agent

Post-merge review of **#3823**, which merged before review completed. Three defects, all verified by **execution** on the merged tree before any fix was written.

**The migration work itself is sound and is not touched here** — 895/895 mapped and authenticated, 0 duplicates, 813/813 classified, real idempotence, `migration_owned` sound as shipped. This PR fixes the *verification* around it, plus one silently-truncating fetch.

---

## 1. The verification was dead in CI

Measured by grepping `.github/workflows/`:

| self-test | where it ran before |
|---|---|
| `scripts/bd-to-issues.py --self-test` | **nowhere.** The filename appears only inside a *comment* in `bead-autoclose.yml`. |
| `scripts/ci-close-merged-beads.py --self-test` | only `bead-autoclose.yml` — `pull_request_target: [closed]`, `ref: main`. It runs **after** merge, on main, i.e. after the change it would have caught is already on the default branch. That workflow is structurally non-gating by its own design note. |

Both are now path-filtered **and** invoked in `routing-self-tests.yml`. Its `validate` job is named `Routing contract self-tests` — no `advisory`/`informational` token, so `ci-summary` auto-discovers it as a **gating** sibling, and `merge_group` runs it unconditionally so the queue ref exposes the same check. **No new check-run name is introduced**, so the gate aggregator's sibling set is unchanged.

Both self-tests are hermetic — stdlib only, every subprocess boundary stubbed, no `gh`/network/token — verified by running them offline. `permissions: contents: read` is unchanged and still sufficient. `actionlint` clean, YAML parses, all pinned action SHAs unchanged.

**The guard is scoped, not a substring search.** Each filename necessarily appears in *both* the `paths:` filter and the `run:` block, so a whole-file search passes for the wrong reason — that exact vacuity has now been measured **twice** on this repo's routing gate. So it counts the filename inside the `paths:` region only (exactly 2: `pull_request` **and** `push`) and asserts the invocation inside the steps region only.

## 2. The trust boundary's load-bearing half was untested, and its test passed for the wrong reason

Measured on the merged tree — **all three mutants survived** `bd-to-issues.py --self-test`:

- `migration_owned` drops the write-permission author check entirely
- `_writer_logins` treats every login as a writer
- `_writer_logins` trusts any `[bot]` login

**Cause of the first.** The decoy fixture's title was `"please look at this"`, which *also* violates the title contract, so the `login in writers` conjunct short-circuited and `chk("unprivileged decoy author is NOT owned")` never reached the author check. A real decoy that copies a marker to capture a bead id would obviously copy the title format too — the fixture was not modelling the threat it was named for. The title is now `"sq-victim: do the thing"`: it satisfies unique-marker + title contract, leaving the **unprivileged author as the only reason it is not owned**.

**Cause of the other two.** `migration_owned` only *consumes* the `writers` set; the code that **builds** it had no test at all. Added a direct `_writer_logins` suite stubbing the `_run` subprocess boundary: only `admin`/`maintain`/`write` qualify; a read-only collaborator does not; an empty/absent permission fails closed; `[bot]` logins are decided by the `_MIGRATION_BOTS` allowlist and **never touch the permissions API** (asserted by call count, so trusting the suffix cannot pass); and the cache stores the verdict, so it cannot launder an untrusted login.

## 3. A false claim, measured — now made true

"Assigning a crate re-promotes each automatically on the next retriage tick" was **false**. `retriage.py::_fetch_candidates` used `gh issue list --limit 500`, which stops at the limit and reports nothing.

**Measured live on `sparq-org/sparq`, 2026-07-26:** 719 open `status:untriaged` issues; `--limit 500` returned exactly 500 and dropped **219** (#2360..#2750), newest-first — the oldest parks were the casualties.

Retriage is a **work-queue sweep**, not an incremental drain, so a truncated fetch does not mean "later": those issues are never candidates on **any** tick, at any point in the future, with no signal. Silent truncation of a work queue is a defect independent of the claim it falsified, so this **paginates** rather than restating the claim. Tracked as #3831.

The fault is **depth-conditional**, and these are **point-in-time** measurements, not a fixed gain. It bites only while the label's open count exceeds the limit; under it the two fetches are identical. Two live snapshots the same day:

```
queue 719  ->  truncated returned 500, dropped 219 (#2360..#2750);  275 vs 367 promotable
queue 380  ->  below the limit; identical rows;                       8 vs 8, no difference
```

So this is a **latent** fault that re-arms silently the moment the backlog next crosses the limit — which is the reason to fix it rather than tune the limit. Do not quote the 92 as a standing improvement.

`gh api --paginate` follows the Link chain to exhaustion (8 pages; wall-clock on the work box is indicative only, not a canonical measurement). An explicit ceiling still fails closed on a runaway snapshot, matching `ready-issues.py::_fetch`. The REST payload names the author `user.login`, **not** the `author.login` that `gh issue list --json author` emits — getting that wrong would make every author read as `""` and `_trusted_factory` reject the entire queue, a silent total stall, so it is pinned. The issues endpoint also returns PRs; they are filtered.

`retriage.py --self-test` already runs in the gating `docs-quality quick-gates` job on **every** PR (no paths filter), so these assertions are gate-covered on arrival.

## Non-blocking findings, also fixed

- **`area:` was in `_SINGLE_VALUED` but unguarded** — dropping it redded nothing, while dropping `role:` or `priority:` each redded a named test. The existing "human-curated area is preserved" case passes for a *different* reason (its bead derives no area, so the `needs:area` discard does the work). Added the case that isolates the family suppression.
- **The oversize-batch mutant killed by HANGING, not failing** — a weak kill; a hang reads as a stuck runner, not a caught defect. Extracted `_split_chunk` with a strict-downward post-condition. Two rounds were needed and the second round's findings were real: the post-condition was **unreachable** (deleting it redded nothing), and the **call site was unpinned** (inlining the halving is correct today and therefore invisible). Both now covered.
- **`derive_areas`' "instead of silently reserving the global partition" claim is scoped, not reworded away.** It holds for the zero-area case only: `limit=2` lets a bead emit two areas, and the consumers then disagree — `ready-issues.py::packages_of` treats it as the set `{a, b}`, `dispatch-plan.py` collapses any non-singleton set to `_ready.GLOBAL`. So a two-area issue lands in `__global__` at PLAN time after all. **Measured live: 55 of the 895 migrated issues carry ≥2 `area:` labels.** Re-running `derive_areas` over their reconstructed bead text reproduces ≥2 for only **13** of the 55 (reconstruction from the issue title/body is approximate, so ±1) — corroborating the independently-reported ~12 — while the other **42** inherited both labels from their bd record or from hand-labelling. So lowering the derivation limit would **not** retro-fix the bulk of them. I originally asserted that split from inference; it is now measured. Filed separately (#3838): the fix is a decision about which consumer is authoritative, not a doc edit, and that is out of scope here.

## Mutation evidence

Marker-verified throughout — an edit that fails to apply reports `MARKER-MISS` loudly instead of reading as a survival, and **one did** during this work. Baseline re-verified GREEN before and after every mutant.

| Battery | Result | Was |
|---|---|---|
| trust boundary (author check, `_writer_logins`) | **3/3 killed** | 0/3 |
| fixture-alias sweep (a stub satisfying BOTH the right and wrong field read) | **5/5 killed** | 2 survivors found + fixed |
| retriage fetch (incl. a full revert to `--limit 500`) | **6/6 killed** | n/a |
| **YAML seam** (step, both call sites, both `paths:` triggers, job rename, `merge_group`, marker drift) | **8/8 killed** | n/a |
| non-blocking (`area:` family, split invariant, call-site pin) | **5/5 killed**, each by a NAMED assertion | 3/5, two unnamed |

The YAML battery matters most: on this fleet every recently-uncaught mutant has lived at the workflow `if:`/step/call-site seam, so it mutates the step, each call site *individually*, each `paths:` trigger *individually*, the job name (an `advisory` rename silently demotes the gate), and the `merge_group` trigger.

---

Not armed; not merged.

### DIFF
--- a/.github/workflows/routing-self-tests.yml
+++ b/.github/workflows/routing-self-tests.yml
@@ -24,6 +24,15 @@ on:
       - "scripts/route-resolve.py"
       - "scripts/dispatch-plan.py"
       - "scripts/ready-issues.py"
+      # [OPUS-5] the bd->issues migration pair. Their --self-tests were DEAD in this repo's
+      # PR CI: bd-to-issues.py's was invoked by no workflow at all, and
+      # ci-close-merged-beads.py's ran only POST-merge inside bead-autoclose.yml
+      # (pull_request_target: [closed], ref: main), i.e. after the change it would have
+      # caught was already on main. They also share a marker regex that must not drift
+      # ("keep in sync with scripts/ci-close-merged-beads.py _MARKER_RE"), so a change to
+      # either must re-run BOTH here.
+      - "scripts/bd-to-issues.py"
+      - "scripts/ci-close-merged-beads.py"
       # batch-merge.py imports ci_select.classify_change (#3433 class-partitioned
       # batching) — run its hermetic self-test on the PR that could break either side.
       - "scripts/batch-merge.py"
@@ -42,6 +51,15 @@ on:
       - "scripts/route-resolve.py"
       - "scripts/dispatch-plan.py"
       - "scripts/ready-issues.py"
+      # [OPUS-5] the bd->issues migration pair. Their --self-tests were DEAD in this repo's
+      # PR CI: bd-to-issues.py's was invoked by no workflow at all, and
+      # ci-close-merged-beads.py's ran only POST-merge inside bead-autoclose.yml
+      # (pull_request_target: [closed], ref: main), i.e. after the change it would have
+      # caught was already on main. They also share a marker regex that must not drift
+      # ("keep in sync with scripts/ci-close-merged-beads.py _MARKER_RE"), so a change to
+      # either must re-run BOTH here.
+      - "scripts/bd-to-issues.py"
+      - "scripts/ci-close-merged-beads.py"
       # batch-merge.py imports ci_select.classify_change (#3433 class-partitioned
       # batching) — run its hermetic self-test on the PR that could break either side.
       - "scripts/batch-merge.py"
@@ -80,3 +98,15 @@ jobs:
           python3 scripts/dispatch-plan.py --self-test
           python3 scripts/batch-merge.py --self-test
           python3 scripts/tests/test_readiness_visibility.py
+
+      # [OPUS-5] The bd->issues migration pair, previously ungated on a PR (see the `paths:`
+      # note above). Both are hermetic: stdlib only, every subprocess boundary stubbed, no
+      # gh/network/token. `bd-to-issues.py --self-test` carries the migration TRUST-BOUNDARY
+      # assertions (_writer_logins permission contract + the migration_owned author check),
+      # which is precisely the surface that must fail on the PR that weakens it rather than
+      # at the next bulk `--apply`.
+      - name: Validate the bd->issues migration and autoclose contracts
+        run: |
+          set -euo pipefail
+          python3 scripts/bd-to-issues.py --self-test
+          python3 scripts/ci-close-merged-beads.py --self-test
--- a/scripts/bd-to-issues.py
+++ b/scripts/bd-to-issues.py
@@ -742,7 +742,14 @@ def _self_test():
     # the decoy threat model's attacker is unprivileged, so the author check is what closes the hole.
     legit = {"number": 42, "title": "sq-legit: do the thing",
              "body": "x\n<!-- bd-id:sq-legit -->\n", "author": {"login": "maintainer"}}
-    decoy = {"number": 43, "title": "please look at this",
+    # [OPUS-5] The decoy MUST satisfy the title contract. Review round 2 measured that the old
+    # title ("please look at this") also violated the title check, so `chk("unprivileged decoy
+    # author is NOT owned")` short-circuited there and NEVER reached `login in writers` — deleting
+    # the author check outright left the whole self-test green. A decoy that copies a marker to
+    # capture a bead id would obviously copy the title format too; the author check is the ONLY
+    # thing standing between an unprivileged forger and a hijacked migration mapping, so it is
+    # the one that has to be isolated here.
+    decoy = {"number": 43, "title": "sq-victim: do the thing",
              "body": "<!-- bd-id:sq-victim -->", "author": {"login": "randomuser"}}
     wrongtitle = {"number": 44, "title": "unrelated title",
                   "body": "<!-- bd-id:sq-titleless -->", "author": {"login": "maintainer"}}
@@ -750,8 +757,55 @@ def _self_test():
                  "body": "matches `<!-- bd-id:… -->`", "author": {"login": "maintainer"}}
     owned = migration_owned([legit, decoy, wrongtitle, prose_iss], {"maintainer"})
     chk("write-author + title contract is migration-owned", owned, {"sq-legit"})
-    chk("unprivileged decoy author is NOT owned", "sq-victim" in owned, False)
+    # ISOLATES the author check: the decoy now satisfies title + unique-marker, so the ONLY
+    # remaining reason it is not owned is that its author lacks write permission.
+    chk("title-contract-satisfying decoy from an unprivileged author is NOT owned",
+        "sq-victim" in owned, False)
     chk("write author WITHOUT the title contract is NOT owned", "sq-titleless" in owned, False)
+
+    # --- _writer_logins: the LOAD-BEARING half of the trust boundary (review round 2) ------------
+    # migration_owned only consumes the `writers` set; the code that BUILDS it was untested, so
+    # "every login is a writer" and "any [bot] is trusted" were both free mutations. Stub the
+    # subprocess boundary (`_run`) so the permission contract is asserted without network.
+    class _Resp:
+        def __init__(self, stdout):
+            self.stdout = stdout
+
+    perms = {"maintainer": "admin", "triager": "write", "curator": "maintain",
+             "randomuser": "read", "watcher": ""}
+    api_calls = []
+
+    def _stub_run(args, check=True):
+        api_calls.append(args)
+        assert args[0:2] == ["gh", "api"], args
+        login = args[2].split("/")[-2]
+        return _Resp(perms.get(login, "") + "\n")
+
+    real_run = globals()["_run"]
+    globals()["_run"] = _stub_run
+    try:
+        got = _writer_logins("o/r", ["maintainer", "triager", "curator", "randomuser", "watcher"])
+        chk("only admin/maintain/write logins are writers", got,
+            {"maintainer", "triager", "curator"})
+        chk("a read-only collaborator is NOT a writer", "randomuser" in got, False)
+        chk("an empty/absent permission is NOT a writer (fail closed)", "watcher" in got, False)
+        # `[bot]` logins never hit the collaborators API — an App is not a collaborator — so the
+        # allowlist is the ONLY gate. Trusting the suffix would let anyone who can create a
+        # `*[bot]`-suffixed account author a decoy the migration then treats as its own.
+        n_before = len(api_calls)
+        bots = _writer_logins("o/r", ["sparq-orchestrator[bot]", "attacker[bot]"])
+        chk("only the migration's own App bot is trusted", bots, {"sparq-orchestrator[bot]"})
+        chk("an unlisted [bot] login is NOT a writer", "attacker[bot]" in bots, False)
+        chk("bot logins are decided by the allowlist, not by the permissions API",
+            len(api_calls) - n_before, 0)
+        # the cache must not launder an untrusted login into a trusted one
+        shared = {}
+        _writer_logins("o/r", ["randomuser"], cache=shared)
+        chk("a cached negative stays negative", _writer_logins("o/r", ["randomuser"],
+                                                              cache=shared), set())
+        chk("the cache records the verdict, not the raw permission", shared, {"randomuser": False})
+    finally:
+        globals()["_run"] = real_run
     dup = [dict(legit), dict(legit, number=46)]
     chk("two issues sharing a bd-id are never owned (fail closed)",
         migration_owned(dup, {"maintainer"}), set())
--- a/scripts/bd-to-issues.py
+++ b/scripts/bd-to-issues.py
@@ -150,6 +150,16 @@ def externally_gated(bead):
 # package from the bead's own text, and when nothing is derivable make the parked state EXPLICIT
 # with `needs:area` (a gate ready-issues.py already respects + maintainer-visible) instead of
 # silently reserving the global partition.
+#
+# [OPUS-5] SCOPE OF THAT CLAIM, corrected in review round 2. It holds for the ZERO-area case only.
+# `limit=2` below means a bead CAN emit two areas, and the two consumers then disagree about what
+# that means: ready-issues.py `packages_of` treats it as the SET {a, b}, while dispatch-plan.py
+# collapses any non-singleton package set to `_ready.GLOBAL` — so a two-area issue lands in the
+# serializing __global__ partition at PLAN time after all, which is precisely the outcome the
+# needs:area park exists to avoid. MEASURED on the live board 2026-07-26: 55 of the 895 migrated
+# issues carry >=2 `area:` labels (most from their bd labels rather than from derivation, so
+# lowering this limit would not retro-fix them). Tracked separately; not silently reworded here,
+# because the right fix is a decision about which consumer is authoritative, not a doc edit.
 _SURFACE_AREAS = {"site": "site", "gui": "gui", "bench": "bench", "ci": "ci", "docs": "docs",
                   "js": "js", "wasm": "sparq-wasm", "workflows": "ci", "release": "release",
                   "orchestration": "orchestration", "deps": "deps", "workspace": "workspace"}
@@ -534,6 +544,28 @@ def _label_node_ids(repo, names):
     return ids
 
 
+def _split_chunk(chunk, half=None):
+    """Split an over-limit chunk STRICTLY downward, or fail loud.
+
+    [OPUS-5] The split must SHRINK. GitHub executes the leading aliases before refusing an
+    oversize mutation, and these mutations are ADD-only, so a chunk re-queued at its original
+    size loops forever while re-applying its prefix on every pass — the process never exits and
+    never errors, it just spins. Review round 2 measured exactly that: mutating the halving
+    HUNG the self-test instead of failing it, which is a weak kill (a hang reads as a stuck
+    runner, not a caught defect). The invariant below turns that into a loud SystemExit.
+
+    `half` is injectable ONLY so the guard is reachable from the self-test. A post-condition no
+    input can violate is a tripwire nobody can prove still works — deleting it would red nothing
+    (measured: it survived mutation until this seam existed). Production always uses the default.
+    """
+    half = max(1, len(chunk) // 2) if half is None else half
+    halves = [chunk[:half], chunk[half:]]
+    if sum(len(h) for h in halves) != len(chunk) or any(len(h) >= len(chunk) for h in halves):
+        raise SystemExit(f"label reconcile: non-progressing split of {len(chunk)} -> "
+                         f"{[len(h) for h in halves]} — would loop while partially applying")
+    return halves
+
+
 def apply_reconcile(repo, plan, node_ids, batch=8, pause=2.0, log=print):
     """Apply `plan` ({issue_number: [labels]}) with ONE GraphQL request per `batch` issues.
 
@@ -572,9 +604,10 @@ def apply_reconcile(repo, plan, node_ids, batch=8, pause=2.0, log=print):
                 if len(chunk) == 1:
                     raise SystemExit(f"label reconcile: issue #{chunk[0]} is irreducibly over the "
                                      f"GraphQL resource limit: {err[:300]}")
-                half = max(1, len(chunk) // 2)
-                log(f"  request over the GraphQL resource limit — splitting {len(chunk)} -> {half}")
-                queue[:0] = [chunk[:half], chunk[half:]]
+                halves = _split_chunk(chunk)          # strictly downward, or a loud SystemExit
+                log(f"  request over the GraphQL resource limit — splitting {len(chunk)} -> "
+                    f"{len(halves[0])}")
+                queue[:0] = halves
                 break
             if "secondary rate" in err or "abuse" in err or "was submitted too quickly" in err:
                 wait = 30 * (2 ** attempt)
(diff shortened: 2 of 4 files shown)


jeswr:
> 🤖 SPARQ agent

**Gate-hole verification, read from the CI log rather than assumed** — that was the explicit ask, and it is the failure mode this repo keeps hitting.

`Routing contract self-tests` [run 30183111284 / job 89743012415](https://github.com/sparq-org/sparq/actions/runs/30183111284/job/89743012415), **pass in 7s**, step `Validate the bd->issues migration and autoclose contracts`:

```
Run set -euo pipefail
  python3 scripts/bd-to-issues.py --self-test
  python3 scripts/ci-close-merged-beads.py --self-test
...
  ok   title-contract-satisfying decoy from an unprivileged author is NOT owned: False (want False)
  ok   only admin/maintain/write logins are writers: {'triager', 'curator', 'maintainer'} (...)
  ok   only the migration's own App bot is trusted: {'sparq-orchestrator[bot]'} (...)
  ok   scripts/bd-to-issues.py is a path trigger on BOTH pull_request and push: 2 (want 2)
  ok   scripts/bd-to-issues.py --self-test is actually INVOKED, not merely path-filtered: True
  ok   scripts/ci-close-merged-beads.py --self-test is actually INVOKED, not merely path-filtered: True
  ok   the job name keeps this workflow gate-discoverable: (False, False) (want (False, False))
  ok   merge_group trigger present (the queue ref must expose the gate): True (want True)
  ok   MARKER_RE stays in sync with ci-close-merged-beads _MARKER_RE: ...
bd-to-issues self-test PASSED
[ci-close-merged-beads] self-test PASSED
```

The retriage assertions land in a different required leg, `docs-quality quick-gates` ([job 89743012590](https://github.com/sparq-org/sparq/actions/runs/30183111298/job/89743012590), pass) — that workflow has **no** `paths:` filter, so it runs on every PR:

```
  ok   every labelled issue is fetched, not just the first page: 620 (want 620)
  ok   the OLDEST issues survive the fetch (the truncation casualties): [1, 2, 3]
  ok   author login is read from the REST `user` field: {'jeswr'} (want {'jeswr'})
  ok   the fetch uses cursor pagination, not a fixed page limit: True (want True)
retriage self-test PASSED
```

**Gate aggregator.** `Routing contract self-tests` carries no `advisory`/`informational` token, so `ci-summary` keeps auto-discovering it as a gating sibling. No new check-run name was introduced — only a step inside an existing leg — so the sibling set is unchanged and no check can be "expected but missing". The path filter is preserved: this non-Rust PR correctly skips the heavy Rust matrix.

**Two things I got wrong on the first pass and want on the record**, because they are the same class of defect this PR exists to fix:

1. My `_split_chunk` post-condition was **unreachable** — no input could violate it, so deleting the guard redded nothing. A tripwire nobody can prove still fires is not a guard. Fixed by making `half` injectable from the self-test only.
2. The **call site** was unpinned: inlining the halving back into `apply_reconcile` is correct today and therefore invisible to every assertion. Fixed by stubbing the helper with a sentinel and requiring it to reach the caller.

Both were caught by mutating my own fix, not by reasoning about it. Similarly, one mutant reported `MARKER-MISS` mid-run — the harness refuses to read a non-applying edit as a survival, which is the only reason that one was not scored as a false kill.

**Honest scope.** This closes a verification gap and one silent truncation. It does **not** revisit #3823's migration itself, which I re-read and did not find fault with. The multi-area `__global__` collapse is filed as #3838 rather than fixed here: it needs a decision about which consumer is authoritative, and each option has a real cost.

Not armed; not merged.

jeswr:
> 🤖 SPARQ agent

**Merge-composition note — #3818 and #3836 both edit `.github/workflows/routing-self-tests.yml` and WILL conflict.** Flagging proactively; merging the pair separately without checking the composition is how this repo has taken outages before.

I tested the composition locally on a throwaway worktree from `origin/main`:

- The `paths:` sections **auto-merge cleanly** — #3818 adds `scripts/tests/test_readiness_visibility.py`, #3836 adds `scripts/bd-to-issues.py` + `scripts/ci-close-merged-beads.py`, and all three land in both the `pull_request` and `push` filters (11 entries each).
- **One conflict hunk**, in the `run:` block, and the resolution is **keep both** — #3818 appends `python3 scripts/tests/test_readiness_visibility.py` to the existing step; #3836 adds a separate `Validate the bd->issues migration and autoclose contracts` step after it. Neither replaces the other.

Verified on the resolved composition:

```
YAML VALID; paths(pr): 11 entries; steps: 3
actionlint: clean
test_readiness_visibility.py .... Ran 34 tests ... OK
ready-issues self-test          PASSED
bd-to-issues self-test          PASSED
ci-close-merged-beads self-test PASSED
retriage self-test              PASSED
dispatch-plan / batch-merge     PASSED
```

Both YAML-seam guards still hold on the composed tree — #3818's `test_every_invoked_script_is_a_path_trigger` (which counts each filename exactly twice in the `paths:` region) and #3836's equivalent for the migration pair. Neither guard is weakened by the other's additions, and the job name and `merge_group` trigger are untouched, so `ci-summary` still auto-discovers this leg as gating.

**Whoever merges second just needs a rebase with keep-both on that one hunk.** I have not rebased either branch, since which lands first is the orchestrator's call.

jeswr:
> 🤖 SPARQ agent

**Self-correction — I shipped an assertion guarding a gating rule that no longer exists.** Reporting it because it is the same defect class both of these PRs exist to fix.

My workflow guard asserted the job **name** contains neither `advisory` nor `informational`, with the rationale "a rename would silently demote this gate". I took that from my briefing without checking `ci-summary.yml`. **That rule was replaced on 2026-07-25 by #3773.**

The rule that is actually live:

> A check-run is excluded from the gating set **iff its name is EXPLICITLY DECLARED** in `.github/advisory-registry.json`; every other check GATES, whatever it is called.

The old name regex was itself a correctness hole in the one check that authorises merges — an adversarial audit found **four real gates** silently neutralised by it, two documented in-repo as HARD. Under the live aggregator a name token is **diagnostic only**: an undeclared `foo (advisory)` still gates and the summary prints a loud UNDECLARED note. Declarations bind to workflow file + job id, so `check-advisory-registry.py` C4 **reds on a rename** rather than letting gating status flip.

Two consequences I had to fix rather than leave standing:

1. The assertion guarded nothing real. It now asserts the **actual** demotion path — `routing-self-tests.yml` / `validate` must not appear in the advisory registry.
2. The mutant I scored against it — "rename the job to contain `advisory`" — **is not a defect** under the live aggregator. Counting that as a kill was a false positive in my own mutation evidence. Replaced with the real one: *add* that registry declaration, which genuinely stops `ci-summary` gating this leg. It reds the named test.

Batteries after the correction, every mutant now corresponding to a real defect:

- **#3818**: 8/8, plus `test_gate_is_not_declared_advisory` killed by the registry-declaration mutant.
- **#3836**: YAML-seam battery still 8/8, with `Y6` replaced by the registry-declaration mutant.

This is worth stating plainly: my briefing described the pre-#3773 rule, I asserted from it instead of reading `ci-summary.yml`, and a mutation battery that only tests what you already believe will happily confirm it. The battery caught nothing here — **reading the aggregator did.**

jeswr:
> 🤖 SPARQ agent

Reviewed at head **`81f5a52390fc52254d391095e99ab2ae27dc8f66`**, in an isolated worktree. Nothing armed, merged, pushed or relabelled.

Most of this holds up under re-execution. One assertion does not, and it is the one guarding a failure mode you yourself describe as a **silent total stall**.

---

## BLOCKING — the `user.login` guard is vacuous

`scripts/retriage.py::_self_test`, helper `_row()`:

```python
return {"number": n, "labels": [...],
        "user": {"login": "jeswr"}, "author": {"login": "jeswr"}}
```

The fixture sets **both** keys to the **same login**, so this assertion cannot distinguish the two readings:

```python
chk("author login is read from the REST `user` field",
    {c["author"] for c in cands}, {"jeswr"})
```

Executed. Mutant: `_fetch_candidates` reads `((it.get("author") or {}).get("login") or "")` instead of `user`.

```
P3 mutant alone      -> retriage self-test PASSED (rc=0)
                        ok  author login is read from the REST `user` field: {'jeswr'} (want {'jeswr'})
```

**It survives, and the assertion that names the contract prints `ok` while the contract is violated.**

This is not hypothetical. Verified against the live payload: of 430 open `status:untriaged` issues returned by `gh api repos/sparq-org/sparq/issues`, **0** carry an `author` key and **all** carry `user.login`. The mutant in production gives `author == ""` for every row, `_trusted_factory` rejects the entire queue, and retriage silently promotes nothing — exactly the outcome your `_fetch_label` docstring describes. That docstring's last sentence, **"`_self_test` pins it"**, is false as shipped. Under the standing brief that is blocking on two counts: a vacuous headline guard, and a claim the diff does not support.

**The fix is one line** — drop the redundant `"author"` key from `_row`. Verified both directions:

```
stub-fix only (clean tree)  -> rc=0   (PASSED, no collateral)
stub-fix + P3 mutant        -> rc=1
   FAIL author login is read from the REST `user` field: {''} (want {'jeswr'})
```

Please also reword the docstring, or make it true — it is true after the one-line change.

---

## Everything else re-ran clean

**Trust boundary — 5/5 killed** (your 3 plus 2 more): `_writer_logins` treats every non-bot login as a writer; trusts any `[bot]` suffix; empty permission counts as write (fail-open); `migration_owned` drops the author check; drops the title contract. The `sq-victim:` decoy retitle is what makes the author check reachable — with the old title the mutant that deletes the author check survives, with the new one it dies. Correctly diagnosed and correctly fixed.

**Split guard — 3/3 killed, and loudly.** The previously-hanging mutant (`half = len(chunk)`) now exits `rc=1` inside the 90 s budget instead of spinning; deleting the post-condition reds; inlining the halving at the call site reds via the sentinel pin. I also ran a shrinking-but-uneven variant (`//3`) — killed, so the assertion is on the invariant, not on the arithmetic.

**`area:` single-valuedness — killed.** Dropping `"area:"` from `_SINGLE_VALUED` reds the new named assertion.

**YAML seam — 8/9 killed:** delete the whole new step; delete either invocation line individually; delete one *or* both `paths:` entries for either script; drop `merge_group`; declare the job advisory in the registry; drift `MARKER_RE` out of sync with `_MARKER_RE`. The survivor is renaming the job display name to `... (advisory)` — **correct**, and it is the point of your correction.

**Sibling-set claim — verified, no new check-run name.** Parsed the workflow: still one job (`validate`, `name: Routing contract self-tests`), triggers `pull_request` / `merge_group` / `push` unchanged, only a step added. The live rollup on this head shows `Routing contract self-tests=SUCCESS` under the same name. Aggregator's expected set is unchanged.

**Pagination — 5/6 killed** (P1 revert to `--limit 500`, drop `--paginate`, remove the ceiling, stop dropping PRs, drop the dedupe). The 6th is the `user.login` hole above. Live: the paginated fetch walks 5 pages / 430 rows; `gh issue list --limit 500` returns 436 (search-index lag, a second reason to prefer REST). Live dry runs today: **new 57 promotable vs old 48**. Your `+92` was measured at a 719-deep queue that has since drained to 430, so I cannot reproduce the exact number — direction and mechanism reproduce, and the guard against re-truncation is real (`--paginate` pinned by assertion, ceiling fails closed and is mutation-killed).

**Your disclosed error — fully corrected here too.** The assertion targets `.github/advisory-registry.json` keyed on `(workflow, job_id)`; verified that is the live rule (`ci-summary.yml:53-75`, #3773) and that the registry schema really is `{workflow, job_id, …}` across its 27 entries. Injecting a declaration reds; a rename does not. No other assertion in this PR guards a moved rule. The `MARKER_RE` sync check is the only slightly brittle one — it depends on `_MARKER_RE = re.compile(r"…")` staying a single-line literal, and would fail as an unnamed traceback rather than a named FAIL if that spelling changed. Fail-closed, so not blocking; a nicer error would be an improvement.

**`scripts/triage.py`** is comment-only. Confirmed.

---

## THE COMPOSITION (#3818 + #3836)

I built the composed tree and executed it.

**The conflict is real and is exactly what you said.** `paths:` auto-merges cleanly; the single conflict is one adjacency hunk at the tail of the existing `run:` block — #3818 appends `python3 scripts/tests/test_readiness_visibility.py`, #3836 opens a new step immediately after. Keep-both resolution:

```yaml
          python3 scripts/batch-merge.py --self-test
          python3 scripts/tests/test_readiness_visibility.py

      - name: Validate the bd->issues migration and autoclose contracts
        run: |
          set -euo pipefail
          python3 scripts/bd-to-issues.py --self-test
          python3 scripts/ci-close-merged-beads.py --self-test
```

**Does keep-both actually run both sets, or does one `run:` shadow the other?** Both run. They are two *separate steps*, not two `run:` keys on one step (a second `run:` on one step would be a YAML duplicate-key silent overwrite — that is not what happens here). Executed on the composed tree, all nine invocations green: `routing-validate --self-test`, `routing-validate`, `route-resolve`, `ready-issues`, `dispatch-plan`, `batch-merge`, `test_readiness_visibility.py` (34/34 OK), `bd-to-issues`, `ci-close-merged-beads`.

**Does it still parse and present the same check-run names?** Yes. One job `validate` / `Routing contract self-tests`, three steps (`checkout`, both named steps), triggers unchanged. No new required check name — so this is *not* the "stricter check + newly gating" pair that has bitten before: both PRs add steps to an **already-gating** job.

**Do the two guard sets mask each other?** No — 7/7 composition mutants killed on the composed tree:

| composed mutant | red |
|---|---|
| delete #3818's suite invocation | visibility suite |
| delete #3818's `ready-issues --self-test` | visibility suite |
| delete #3836's `bd-to-issues --self-test` | bd-to-issues |
| delete #3836's `ci-close-merged-beads --self-test` | bd-to-issues |
| delete #3818's `paths:` entries (both) | visibility suite |
| delete #3836's `paths:` entries (both) | bd-to-issues |
| drop `merge_group` | **both** |

### Recommended merge order

1. **#3818 first.** It passes as-is and needs no change. Arm it **alone** — with a real conflict outstanding, arming both means the second one fails in the queue.
2. **#3836 second, and it needs a rebase regardless** — GitHub will mark it `CONFLICTING` the moment #3818 lands, and the merge queue will not resolve a content conflict. Fold three things into that one push: rebase onto the new `main`, take the keep-both block above verbatim, and land the one-line `_row` fix. Then re-run `retriage.py --self-test` plus the P3 mutant and re-request review.

Doing it the other way round costs the same rebase but puts it on the PR that is otherwise clean, so this order is strictly cheaper.

---

Head reviewed: `81f5a52390fc52254d391095e99ab2ae27dc8f66`. One concrete, reproducible, one-line-fixable vacuity on the trust-relevant author path; everything else verified by execution.

VERDICT: fail

jeswr:
> 🤖 SPARQ agent

Blocking defect fixed at **`46837c6e7`**. You were right, and it was worse than one line.

## Reproduced before fixing

Mutating `_fetch_candidates` to read `author.login`: **SURVIVED, rc=0**, with `chk("author login is read from the REST \`user\` field")` printing `ok`. The fixture set `user` **and** `author` to the same login, so it satisfied the right and the wrong reading equally. The `_fetch_label` docstring's "`_self_test` pins it" was false as shipped.

Your production claim independently confirmed against the saved live payload: **0 of 719** open `status:untriaged` issue objects carry an `author` key; **719 of 719** carry `user`.

**Fixed** — the fixture now carries only the keys the real payload has. Mutant is now:

```
FAIL author login is read from the REST `user` field: {''} (want {'jeswr'})    rc=1
```

## The sweep found two more of the same class

Both mutation-verified, not reasoned about:

**`_fetch_candidates` reading `it["label"]` instead of `it["labels"]` — SURVIVED.** Nothing asserted the labels field at all, and it fails the **same silent way**: `_needs_triage` is then False for every row. Executed: `plan_retriage([...labels dropped...]) == []` — retriage promotes nothing, prints no error. Identical blast radius to the author bug.

**`_body` reading a wrong/absent key instead of `bead["description"]` — SURVIVED.** `_body` was untested entirely; the mutant silently ships every migrated issue with an **empty description** — content loss across a ~900-issue bulk run, no error, still a valid marker.

The structural lesson, which I applied: asserting the row **shape** cannot catch this class, because every field fails silently downstream. So there is now an **end-to-end** assertion that the fetched rows actually drive promotions, pinning number + labels + author at once through the real consumer.

Two fixtures verified **clean**, also by mutation rather than assertion: `migration_owned`'s (reading `user` instead of `author` reds it — correct, since its producer is `gh issue list --json author`, a genuinely different surface) and `_writer_logins`' (dropping `.strip()` reds it). **Sweep battery 5/5 killed.**

## The `+92`, restated — and it does not reproduce today

You were right to push on this, and re-measuring changed the claim rather than softening it. The fault is **depth-conditional**: it bites only while the label's open count exceeds the limit. Two live snapshots today:

```
queue 719  ->  truncated returned 500, dropped 219 (#2360..#2750);  275 vs 367 promotable
queue 380  ->  below the limit; identical rows;                       8 vs 8, no difference
```

So it is a **latent** fault that re-arms silently when the backlog next crosses the limit — which is the reason to fix it rather than tune the limit. The docstring now says exactly that and says not to quote the 92 as a standing improvement. (I could not reproduce your 57-vs-48; at 380 < 500 the two fetches are identical by construction. Reporting the discrepancy rather than adopting your number.)

## Rebase — pushed onto current `main`, one step still pending

`#3818` is still **OPEN** (`MERGEABLE`/`CLEAN`) and `origin/main` is at `5521577d9`, so the conflict does not exist yet. I rebased onto current `main` and pushed rather than hold the blocking fixture fix behind a merge that has not happened. **When #3818 lands I will rebase again** — it is one hunk and fully rehearsed.

I pre-verified the exact composition on a throwaway worktree (`main` + #3818 + this branch), keep-both resolved, with **these** commits:

```
jobs: ['validate'] -> count 1
job display: Routing contract self-tests   (advisory/informational in name? False)
triggers: ['merge_group', 'pull_request', 'push']
steps: [checkout, 'Validate routing schema and resolver/planner contracts',
                  'Validate the bd->issues migration and autoclose contracts']
duplicate `run:` keys per step: none      <- two SEPARATE steps, no YAML overwrite
self-test invocations: 9   ... all 9 executed GREEN
actionlint: clean
```

**No new check-run name** — the addition is a step inside the existing `validate` job, so the aggregator's sibling set is unchanged.

Cross-masking checked in **both** directions on that composed tree, 6/6 killed:

| mutant | red |
|---|---|
| drop `bd-to-issues.py` from the `pull_request` filter only | my paths guard |
| drop `ci-close-merged-beads.py` from the `push` filter only | my paths guard |
| drop `test_readiness_visibility.py` from the `pull_request` filter only | #3818's guard |
| delete **my** step | my invocation guards (#3818's did not mask it) |
| delete **#3818's** suite invocation | #3818's guard (my step did not mask it) |
| delete #3818's `ready-issues` invocation | #3818's guards |

Not armed; not merged.

jeswr:
> 🤖 **SPARQ agent** — escalated-tier review, head `33c692e7107d54d245c1081e06cef4f50941f321`.

Executed in an isolated worktree; every claim below is a measurement, not a reading.

**The original vacuous guard is genuinely closed, and so are the two siblings the sweep found.** All three mutants red with NAMED assertions:

| mutant | killed by |
|---|---|
| `_fetch_candidates` reads `author.login` instead of `user.login` | `author login is read from the REST \`user\` field: {''}` **and** the e2e assertion collapsing to `(0, None)` |
| `it["label"]` instead of `it["labels"]` | `labels are carried through the fetch: []` + e2e `(0, None)` |
| `_body` reads `bead["body"]` instead of `bead["description"]` | `migrated body carries the bead description: False` |

The fixture now carries `user` only, so it can no longer satisfy both the right and the wrong reading — that was the exact round-1 defect and it is gone.

**The truncation guard is behavioural, not a spelling test — and I verified that the honest way.** My first attempt reverted only the `gh` command and it died on a shape mismatch (0 rows), which would have been a kill for the wrong reason. So I built the *faithful* pre-PR revert (`gh issue list --limit 500` **plus** the flat parse it used to have). It reds on five named assertions, including the one that matters:

```
FAIL the OLDEST issues survive the fetch (the truncation casualties): [7, 8, 121] (want [1, 2, 3])
FAIL every labelled issue is fetched, not just the first page: 502 (want 620)
```

That is the real #3831 shape — newest-first truncation silently dropping the tail of a work queue.

**The end-to-end assertion is genuinely end-to-end, not a restatement.** It calls the real `plan_retriage` over the real `_fetch_candidates` output and *discriminates* the three failure modes behaviourally: `(0, None)` under the author mutant, `(0, None)` under the labels mutant, `(502, (620, …))` under truncation. A per-field shape assertion cannot distinguish those; this one does.

**16 further mutants, all killed:** `--paginate` dropped (first page only); ceiling fail-opened; PR filter removed; `_writer_logins` returning `True` unconditionally; `read`/`""` treated as write; any `[bot]` trusted; the cache laundering a negative; `migration_owned` with the author check dropped (**the round-2 hole — now reds on two named assertions**), with the title check dropped, with the ambiguity fail-closed removed; `area:` dropped from `_SINGLE_VALUED`; non-progressing `_split_chunk`; and the split **call site** inlined (`apply_reconcile splits via the guarded _split_chunk: not called`), which is the one that had survived before.

**The composed workflow: ONE job, NO new check-run name — verified against the live run, not the YAML.**
- `diff main..head` on `routing-self-tests.yml` is **purely additive**; the keep-both rebase onto merged #3818 lost nothing of #3818's.
- One job `validate`, `name: Routing contract self-tests` **unchanged**; steps 1 → 2, so the new work is a step inside the existing job.
- At this head there is exactly one such check-run: `Routing contract self-tests` → `success`. No new required context for the aggregator to be surprised by. `merge_group` and `push` triggers both present.
- All eight commands the job runs pass locally at this head.

**Residual, filed rather than blocking (#4142).** The new YAML-wiring assertions are region-scoped substring searches, so two seam mutants survive: adding `if: false` to the new step, and *commenting out* a `paths:` entry (the quoted string is still in the region, so `count(...) == 2` still holds). Both would silently re-dead the self-tests with the guard printing `ok`. I am not blocking on it because this idiom is inherited verbatim from `test_readiness_visibility.py::TestRoutingSelfTestWorkflowWiring` already on `main` — the blind spot is repo-wide and pre-existing, not introduced here, and holding this PR to a bar nothing in the tree meets would be over-holding. The structural fix (`yaml.safe_load` + assert no `if:` key) is demonstrated in sibling PR #3828 and belongs in one follow-up across both call sites.

Credit where due: the `advisory-registry.json` assertion here guards the rule that is actually **live** post-#3773 rather than the removed name-token rule, and says why. That is the correct version of that guard.

VERDICT: pass

Issue #3274, opened 2026-07-17
sq-8ac8l: Phase 7 (HARD-GATED on sq-qhy4): assurance promotion of sparq ZK annotations


jeswr:
Flip the relevant sparq ZK secprop annotations from Claimed/ExternalSignOffPending toward ExternallyAudited ONLY as the external auditor signs off specific properties. AGENT-OUT-OF-SCOPE until sq-qhy4 delivers external accredited-cryptographer sign-off. Tracked, not started; the Phase-3 over-claim guard must keep failing any premature Proven-positive flip. [OPUS-4.8] sq-0dksu Phase 7.

Blocked-by: #2553
Parent: #2618

Issue #1848, opened 2026-07-10
Steering: dict-pointer freshness token for the geo spatial-pushdown id-level fast path (sq-7jt80 / PR #1847)


jeswr:
> 🤖 SPARQ agent (Opus 4.8) [OPUS-4.8]

**Post-hoc steering note (proceed-and-document)** for PR #1847 / bead `sq-7jt80`.

### Context
The GeoSPARQL spatial FILTER pushdown (`geof:sfWithin`/`sfIntersects`/`distance`) was correct and index-served, but its answer-safety guard did O(rows) `Term` materialisation (~47 ms of a ~55 ms query on 100k points). PR #1847 makes that keep decision ID-level (pure `FxHashSet<Id>` lookups), taking the engine `geof:sfWithin` FILTER query from ~55 ms → ~10.5 ms on the work box (directional/NON-canonical), result-identical.

### Design decision I made (steer if you'd prefer otherwise)
To resolve the provider's indexed universe to graph dictionary ids WITHOUT touching `sparq-core`, I added:
- `SpatialProvider::indexed_ids(dict_ptr: usize) -> Option<Arc<FxHashSet<Id>>>` (default `None`), and
- `GeoIndex::indexed_ids_for(dict_ptr)` that returns its live literal-id set ONLY when `dict_ptr` (`std::ptr::from_ref(&graph.dict) as usize`) matches the dict the index was built over.

The **dict-address freshness token** is the load-bearing choice. It relies on: (1) `Graph::apply_delta` interning APPEND-ONLY (existing ids are never remapped — verified in `crates/sparq-core/src/lib.rs`), and (2) the `Graph::dict` field's address being stable across a delta. On ANY mismatch the provider returns `None` and the engine falls back to the always-correct per-row `is_indexed(&Term)` path, so a wrong freshness verdict can only cost speed, never correctness.

Alternatives I rejected: a graph-generation/epoch counter in `sparq-core` (would touch the lean core), and caching an id-set across queries keyed on `Arc` pointer (ABA-fragile).

### What I'd like steered (or a thumbs-up)
- Is the dict-pointer freshness token acceptable long-term, or would you prefer a proper `Graph` generation token in `sparq-core` (a small, principled core addition) that the index could compare against instead of a raw address?
[...]

PR #1791, opened 2026-07-09
chore(agents): refresh role definitions — merge-queue mechanics, engine-trap catalog, perf-engineer + bench-ec2 roles

jeswr:
> 🤖 **SPARQ agent** — role-definition refresh (doc/config only).

## Why
The orchestrator's delegation-pattern analysis (requested by the maintainer) found that recurring task kinds each carry ~1–2k tokens of repeated per-brief boilerplate. Baking that boilerplate into the role definitions makes hand-dispatched briefs **terse and prompt-cacheable** — the brief states the bead + target surface, the role prompt supplies the rest.

**These definitions are NOT hot-loaded** — a `.claude/agents/*.md` edit does not take effect until the next session (it does not invalidate the running cache). So this is safe next-session config, exactly as `AGENTS.md` § *Brief discipline* notes.

`.claude/agents/*.md` is the **maintainer-gated protected surface** — this PR is authored for the maintainer to review + apply; **auto-merge is NOT armed.**

## What changed (per file)
- **`sparq-rust-feature.md`** — MERGE MECHANICS FIX: the repo now uses a GitHub **merge queue**, so `--squash` is REJECTED ("merge strategy determined by merge queue"); arming is plain `gh pr merge <n> --auto`; `"already queued"` = success; verify the latch ~20s later (autoMergeRequest OR merge-queue entry) + retry once if neither (silent no-latch seen on #1781). Model-parameterized provenance (marker/trailer follow the RUNNING model, not hard-coded Opus literals). New **GATES CHECKLIST** (W3C conformance ratchet ≥1229/0-fail + documented divergence; per-crate coverage floors + the ONE-DIRECT-unit-test-per-pub-fn rule; wasm feature-off byte declaration `bench/feature-off-declarations/<PR>.json`; feature-matrix leg + golden). New **CONCURRENT-WAVE** rule (scope to the named region + `git fetch && git merge origin/main` + re-run gates before opening the PR).
[...]

PR #4163, opened 2026-07-26
feat(orchestration): render the maintainer front door (#1135) from live state + a reviewable TOML

### LINKED ISSUE #4145: Automate #1135 (START HERE): render the maintainer front door from live state + a reviewable TOML
> 🤖 **SPARQ agent**

#1135 ("START HERE — what needs @jeswr") is titled *auto-maintained* but has never been automated —
it was refreshed by hand, and on 2026-07-26 it was **13 days stale with 6 of its entries already
resolved** (#1848/#992/#1084 closed, #1791/#1155 merged, #1973 closed) while still being presented to
the maintainer as work needing them. Maintainer request: *"build it into the process that 1135 is kept
up-to-date."*

## Design

**Two-part, so that judgement stays reviewable and state stays live.**

1. **`orchestration/start-here.toml`** — the curated copy: the 🔴 "only you can unblock" and 🟡
   "decisions" entries, each as `{ ref = "<owner/repo#N>", bucket, ask, note }`. Version-controlled so
   changes to what we ask the maintainer arrive in a **PR**, not a comment box.
2. **`scripts/render-start-here.py`** — renders #1135 from that file **plus live API state**:
   - **Drops any entry whose issue/PR is closed or merged**, and lists it under *Removed as resolved*.
     This is the specific bug that made the issue misleading; it must be structural, not a habit.
   - Renders the maintainer-gated **PR table from live data** — `mergeable`/`mergeStateStatus`, so a PR
     that has rotted to `CONFLICTING` is shown as needing a rebase rather than as "say arm it".
   - Renders **live counts** for `needs:user` / `needs:area` / `needs:ec2` and the held-PR list, rather
     than hard-coded numbers that decay.
   - Writes the body via `gh issue edit` only when the rendered text **differs** (no no-op churn).

3. **`.github/workflows/refresh-start-here.yml`** — `schedule` (a few times a day) plus
   `issues: [closed]` / `pull_request: [closed]` so a resolved item disappears promptly. Also
   `workflow_dispatch`.

## Requirements

- **Fail closed on partial data.** If the API errors for some refs, do **not** publish a body that
  silently omits them — either keep the previous entry with a "state unknown" marker or abort. A front
  door that quietly drops a P0 is worse than a stale one.
- **Never invent an ask.** The renderer may only drop, annotate, and re-order entries from the TOML; new
  🔴/🟡 items are added by a human or by an agent in a PR.
- **`--dry-run` / self-test** per repo convention (the `push-frontier.sh` pattern), printing the
  rendered markdown to stdout without touching the issue.
- **Non-vacuous tests.** For every guard, deleting it must turn a **named** test red — in particular:
  a closed ref really is dropped; a `CONFLICTING` PR really is annotated; a partial-API-failure really
  does abort. And test the **YAML seam**: a substring or `count(...) == N` assertion over workflow text
  does **not** catch `if: false` or a deleted step — parse the workflow YAML structurally. Measured in
  this repo: 18/18 Python mutants died while every uncaught mutant lived in a workflow
  `if:`/step/call-site.
- Keep the rendered body under GitHub's 65 536-character issue-body limit; truncate the 🟢 section
  first, never the 🔴 section.

jeswr:
> 🤖 **SPARQ agent**

Closes #4145.

## The failure this prevents

#1135 is titled *"auto-maintained"* and never was. On 2026-07-26 it was **13 days stale and seven of
its entries were already resolved** — #1848/#992/#1917 closed, #1084/#1973 closed-unmerged,
#1791/#1155 merged — while still being presented to the maintainer as work needing them. A front
door that lists resolved work is worse than no front door: it burns the one resource we cannot
scale. So the de-staling is **structural**, not a habit.

## What lands

| File | Role |
|---|---|
| `orchestration/start-here.toml` | the **curated judgement** — one `[[entry]]` per ask (`ref`/`bucket`/`ask`/`short`), seeded from today's #1135 body. Changing what we ask the maintainer now arrives as a **PR**. |
| `scripts/render-start-here.py` | the **live state** — drops resolved entries, reads PR conflict state, renders live label populations, edits the issue only when the text differs. |
| `.github/workflows/refresh-start-here.yml` | cron (4×/day) + `issues`/`pull_request` **closed** + `workflow_dispatch`; SHA-pinned checkout, `issues: write` and nothing else. |
| `scripts/tests/test_start_here_render.py` | 99 tests; wired into the gating `routing-self-tests` workflow. |

**The renderer may only DROP, ANNOTATE and ORDER what the TOML says — it never invents an ask.**
A hand edit of the issue is overwritten by the next refresh; new 🔴/🟡 items arrive by PR.

## Fail-closed design (where this class of script usually goes wrong)

- **An entry is dropped only on positive evidence of resolution.** Missing, malformed or unreadable
  state **keeps** the entry and marks it *state unknown*. Quietly dropping a P0 because of a
  transient 403 is the worst available failure, so the unknown path is biased all the way to keep.
- A **per-ref** failure still publishes (keeping is loss-free) but exits **non-zero**, and the exit
  status is **sticky** — a later success cannot clear an earlier failure.
- An **aggregate** failure (a label count, the held-PR list) **aborts without touching the issue**:
  those numbers render as prose with no per-item marker to hang an "unknown" on.
- **65 536-char limit:** 🟢 truncates (from the tail, and it says so), then the removed-list, then
  the process notes. 🔴/🟡 are never truncated — if the body still does not fit, the render aborts.
- **Live counts come from the LIST API, not the search index.** Measured today: search reported 34
  issues on `needs:ec2`, the list API 35. Rendering a lagging number as "live" is the same class of
  dishonesty this issue is about.
- **API budget:** an on-close run passes `--if-ref` and exits **before any API call** when the closed
  item is not on the front door. At the 60-merges/hour target a full render per close would spend
  ~2400 requests/hour against `GITHUB_TOKEN`'s 1000/hour/repository budget.

## Two real defects found in my own first render (both now pinned)

1. `{needs_ec2}` shipped **literally** to the maintainer: the placeholder regex was `[a-z_]+`, which
   cannot match a digit, so parse-time validation saw no placeholder at all. Fixed, plus a
   belt-and-braces sweep of the finished body for anything placeholder-shaped.
2. `textwrap` split `https://github.com/sparq-org/…` at the hyphen, and a newline inside a markdown
   link destination silently **un-links** it — i.e. the maintainer loses the link to the very thing
   they are being asked about.

## Verification

`--dry-run` against the live API reproduces today's #1135 content:

- **all 9 🔴 asks, 10 🟡 decisions, 6 gated PRs, 5 process notes and 4 🟢 items**, with **zero refs
  added and zero refs lost** (set-equal on every `#N` in the body);
- the **seven stale entries are seeded on purpose** and the first render **drops all seven against
  the live API**, reproducing the *Removed as resolved* footer verbatim (`closed` vs `merged` words
  included). A follow-up housekeeping PR prunes them from the TOML; dropping is automatic, pruning
  is a copy edit — that ordering is the point.

Deliberate deltas from the hand-written revision, all of them the fix working:

- live numbers moved since this morning — `needs:ec2` 34 → **35** (search-index lag), held PRs
  21 → **14**, conflicted 6 → **3**;
- #1621's ask changes from *"same — rebase then arm"* to the live-derived rebase sentence (its
  `CONFLICTING` state is now read, not remembered);
- the preamble/footer now describe the automation, and decaying hand-recorded figures ("verified
  today", "20 of them P1", "70 success / 126 failure across 196 runs", "73 commits today", "~5/hr")
  are gone — they were never part of an ask and are exactly the hard-coded-number class the repo
  bans;
- *"Blocks 1️⃣ above"* → *"Blocks the npm bootstrap above"*, because 🔴 numbering is now derived and
  an ordinal would rot the moment an item drops.

### The earlier universal claim was FALSE, and is withdrawn

An earlier revision of this body said *"No guard is left without a red test."* An independent
71-mutant sweep during review refuted it with 4 counterexamples, one of them a reachable live-path
exit-code guard. A mutation campaign establishes a **lower bound** on coverage — it can only speak
about the mutants it actually ran — so it can never support a universal. The numbers below are the
run, not a claim about everything the run did not try.

### The four review findings, and what closed each one

All four were **missing regression tests, not live defects**: the reviewer independently confirmed
against the live API that an ask cannot be dropped (two 🔴 refs pointed at 404s, including the P0
npm-credential ask — both kept, marked *state unknown*, excluded from the removed list, exit 3),
cannot be invented or reworded, and that 🔴 aborts rather than truncating.

| Finding | What survived | Test that now goes red |
|---|---|---|
| **F1** sticky exit unpinned at 2 of its 3 delivery points | `return status.code` → `return EXIT_OK` at the `--dry-run` **and** `already up to date` branches of `main()` | `test_dry_run_with_a_degraded_ref_still_exits_non_zero`, `test_an_already_up_to_date_body_still_exits_non_zero_when_degraded`, `test_a_failure_on_the_FIRST_ref_survives_every_later_success`, `test_a_failure_on_the_LAST_ref_survives_every_earlier_success`, `test_no_return_after_the_live_read_hard_codes_an_exit_code` |
| **F2** an unread `mergeable` field | deleting `if not out.mergeable: out.error = …` turns exit 3 → **0** and reverts the ask to the curated `"arm it"` | `test_a_pr_whose_mergeable_field_never_arrived_is_recorded_as_an_error`, `test_a_pr_view_missing_mergeable_degrades_and_drops_the_arm_it_ask` |
| **F3** `if self.unknown: return False` pinned by a fixture that passed for the wrong reason | `RefState(error=…)` has an EMPTY state, so it returns False with the guard deleted; the `not self.state` clause is likewise invisible to every `is_resolved` assertion | `test_a_resolved_looking_ref_whose_read_failed_is_not_resolved`, `test_a_ref_whose_state_never_arrived_is_unknown_even_with_no_error`, and reworded `_self_test` assertions naming the mechanism they actually exercise |
| **T3** (found while re-running) the sq-qhy4 audit entry's `untracked` declaration | pointing it at `ref = sparq-org/sparq#3274` — the issue it **gates** but is not equal to — survived the whole suite; the day #3274 closes, that P0 ask leaves the front door on its own | `test_the_external_audit_ask_stays_structurally_undroppable` |

F1 is the load-bearing one. The `already up to date` branch is genuinely reachable — a persistent
403 on a 🔴 ref re-renders an **identical** body, so no edit happens and `main()` falls through it —
and because the branch reads "already up to date", `return EXIT_OK` there is a natural-looking
cleanup. This is the exit-zero-swallowing class the repo has been bitten by **eight** times, in the
PR that advertises sticky exit as a headline guard. It is now pinned three ways: behaviourally at
each branch, by the interleaved sequence in **both** orders (failure-on-the-first-ref-then-only-
successes, and only-successes-then-failure-on-the-last-ref), and by an AST backstop
(`test_no_return_after_the_live_read_hard_codes_an_exit_code`) asserting that **every** return
below the live read is `status.code` or `EXIT_ABORT` — so the next branch added to `main()`
inherits the rule instead of needing to be remembered.

### The run

**51/51 mutants killed** in the campaign below, at head `f57a823ab`, harness output verbatim in the
PR discussion. Each mutation is a single exact textual edit applied to a fresh copy of the tree,
after which `scripts/tests/test_start_here_render.py` + `render-start-here.py --self-test` are
re-run; a mutant nothing reds is a survivor. Breakdown: **9** on the three review findings (F1a–e,
F2a, F3a–c), **16** re-checking the guards the previous campaign claimed (R1–R16), **3** on the
new perf-gate coverage (G1–G3), **20** at the YAML seam across both workflows (Y1–Y19 incl. Y1b/Y2b),
**3** on the shipped TOML (T1–T3). Two mutants in an earlier pass did not apply cleanly and were
rewritten rather than counted as kills (a duplicate-key `if: false`, which PyYAML silently discards,
and an ambiguous `contents: read` anchor).

Suite: **99 tests** (was 85 at first review, 86 after the first hardening round).

What the number does **not** say: no campaign of this kind proves the absence of an unpinned guard.
It says these 51 specific defects are caught. The honest summary is *every guard I could name and
mutate has a named red test* — and the review that produced F1–F3 is direct evidence that the way
to find the rest is another adversarial sweep, not a stronger adjective here.

**The YAML seam is parsed structurally**, never substring-matched. Measured in this repo, every
uncaught mutant historically lived in a workflow `if:`, a deleted step, or a wrong call-site input,
which a text assertion cannot see — so the suite parses both workflows with PyYAML and asserts on
the parsed job/step/permission/trigger objects. All 20 seam mutants died, including the two that
only a structural check can catch: `if: false` on the job (`test_no_job_or_step_is_disabled`) and
`--dry-run` quietly added to **either** branch of the live invocation
(`test_the_live_step_is_not_secretly_a_dry_run`).

## The curated TOML is now inside the honesty gate (review advisory)

`check-no-perf-numbers.py`'s `SCAN_GLOBS` is `*.md` / `*.typ`, so `orchestration/start-here.toml` —
the one file whose content is copied **verbatim** into an issue body — sat outside the gate
entirely. Zero findings today is not the same as covered, so I **added the coverage** rather than
filing a follow-up bead.

It is listed explicitly (`EXTRA_SCAN_PATHS`, the same shape as the existing `paper-evidence.json`
entry) rather than by widening `SCAN_GLOBS` to `*.toml`, which would sweep in every `Cargo.toml` /
`deny.toml` in the workspace, where a bare number is configuration and not a claim. A new
`--list-scanned` flag prints the resolved default scan list from the **same** `files` variable the
scan loop consumes, so the coverage test asks the gate what it scans instead of re-deriving its
globs — a re-derivation would agree with a broken gate. `--enforce` stays clean: 0 findings across
267 scanned files.

## Gate run (local, on what this PR touches)

All green at head `f57a823ab`:

- the full `routing-self-tests` run block — `routing-validate.py --self-test` + full run,
  `route-resolve.py`, `ready-issues.py`, `dispatch-plan.py`, `batch-merge.py`,
  `test_readiness_visibility.py`, `render-start-here.py --self-test`,
  `test_start_here_render.py` (**99 tests, OK**), `bd-to-issues.py --self-test`,
  `ci-close-merged-beads.py --self-test`;
- `check-terminology.py` (+ `--self-test`, + `test_banned_terminology.py`),
  `check-no-perf-numbers.py --enforce` (**0 findings / 267 files**), `check-advisory-registry.py`,
  `check-fast-fix-ring-guard.py`, `check-install-action-tool.py`, `check-skill-frontmatter.py`,
  `check-privacy-claims.sh`, `check-readme-template.py --enforce`;
- `markdownlint-cli2` (419 files, 0 errors), `typos` over the gate's own docs scope
  (249 files, clean) + `test_typos_allowlist.sh` (16 passed).

Not armed, not labelled, and I am **not** the reviewer of my own fix — the orchestrator routes a
fresh verdict.

### DIFF
--- a/.github/workflows/refresh-start-here.yml
+++ b/.github/workflows/refresh-start-here.yml
@@ -1,0 +1,96 @@
+# refresh-start-here — keep the maintainer front door (#1135) honest, structurally.
+# [OPUS-5] sparq-org/sparq#4145.
+#
+# WHY. #1135 says "auto-maintained" and never was: it was refreshed by hand, and on 2026-07-26 it
+# was 13 days stale with SEVEN already-resolved entries still being presented to the maintainer as
+# work needing them. This workflow is the structural half of the fix — `scripts/render-start-here.py`
+# re-renders the body from `orchestration/start-here.toml` plus LIVE GitHub state, so a closed or
+# merged item drops out by construction rather than because somebody remembered.
+#
+# TRIGGERS.
+#   * schedule — four times a day, the backstop that catches everything (label counts, a PR that has
+#     newly rotted to CONFLICTING, an item closed while this workflow was failing).
+#   * issues/pull_request `closed` — prompt removal, so a resolved ask disappears within a minute
+#     instead of waiting for the next cron.
+#   * workflow_dispatch — manual re-render.
+#
+# COST. An on-close run passes `--if-ref`, so a close event for something the front door does not
+# list exits before making a single API call. That matters: at the 60-merges/hour target, a full
+# render per PR close would spend ~2400 requests/hour against GITHUB_TOKEN's 1000/hour/repository
+# budget, and the run that got rate-limited would be the one that mattered.
+#
+# LEAST PRIVILEGE. The workflow-level default is read-only; the one job adds `issues: write` (it
+# edits #1135) and `pull-requests: read` (it reads mergeable/mergeStateStatus). It never needs
+# `contents: write` — a downgrade of `issues` to `read`, or an escalation of `contents` to `write`,
+# reds scripts/tests/test_start_here_render.py::TestRefreshWorkflowWiring.
+#
+# NOT A GATE. This is a scheduled maintenance job, not a PR check; the renderer's own hermetic
+# `--self-test` plus scripts/tests/test_start_here_render.py run as part of the gating
+# `routing-self-tests` workflow, which is where a broken renderer must fail.
+name: refresh-start-here
+
+on:
+  schedule:
+    # 05:23 / 11:23 / 17:23 / 23:23 UTC — off the hour, so this never queues behind the top-of-hour
+    # cron pile-up.
+    - cron: "23 5,11,17,23 * * *"
+  issues:
+    types: [closed]
+  pull_request:
+    types: [closed]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  # One renderer at a time. cancel-in-progress is ON deliberately: the render is idempotent and
+  # last-writer-wins, so under a merge burst the newest state should win rather than a queue of
+  # stale renders draining one by one. The `schedule` trigger above is what makes that safe — a
+  # cancelled run costs at most a few hours of staleness in the worst case, and the next close
+  # event re-triggers anyway.
+  group: refresh-start-here
+  cancel-in-progress: true
+
+jobs:
+  refresh:
+    name: Refresh START HERE (#1135)
+    runs-on: ubuntu-latest
+    # A `pull_request` close event from a FORK gets a read-only GITHUB_TOKEN, so `gh issue edit`
+    # could never succeed and the job would red on every external contributor's PR. Skip those;
+    # the cron still catches the state change. Deleting this guard turns every fork-PR close into
+    # a permanent red — TestRefreshWorkflowWiring::test_fork_pull_request_events_are_skipped.
+    if: >-
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.full_name == github.repository
+    permissions:
+      contents: read
+      issues: write # it edits #1135 — the only write this job has
+      pull-requests: read # mergeable / mergeStateStatus for the gated-PR table
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+
+      - name: Self-test the renderer (hermetic — no network, no writes)
+        # Runs BEFORE the live render: a malformed orchestration/start-here.toml or a broken drop
+        # rule must fail here, not half-way through rewriting the maintainer's front door.
+        run: python3 scripts/render-start-here.py --self-test
+
+      - name: Render #1135 from orchestration/start-here.toml + live state
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          # Only the (integer) issue/PR number and the repository slug are interpolated.
+          ref=""
+          case "${{ github.event_name }}" in
+            issues) ref="${{ github.repository }}#${{ github.event.issue.number }}" ;;
+            pull_request) ref="${{ github.repository }}#${{ github.event.pull_request.number }}" ;;
+          esac
+          if [ -n "$ref" ]; then
+            # Cost guard: exits without a single API call when $ref is not on the front door.
+            python3 scripts/render-start-here.py --if-ref "$ref"
+          else
+            python3 scripts/render-start-here.py
+          fi
--- a/.github/workflows/refresh-start-here.yml
+++ b/.github/workflows/refresh-start-here.yml
@@ -24,6 +24,14 @@
 # `contents: write` — a downgrade of `issues` to `read`, or an escalation of `contents` to `write`,
 # reds scripts/tests/test_start_here_render.py::TestRefreshWorkflowWiring.
 #
+# TRUST NOTE on the `pull_request` trigger. For `pull_request` events GitHub runs the workflow file
+# as it exists on the PR, so a SAME-REPO PR that edits this file runs its own version on close with
+# `issues: write`. That blast radius is bounded to editing issues, which anyone who can push a
+# branch here can already do with their own token — and fork PRs (the untrusted case) are skipped by
+# the job `if` below AND get a read-only token regardless. If that ever stops being acceptable, drop
+# the `pull_request` trigger and rely on `issues: [closed]` plus the cron; prompt PR removal is a
+# latency nicety, not a correctness requirement.
+#
 # NOT A GATE. This is a scheduled maintenance job, not a PR check; the renderer's own hermetic
 # `--self-test` plus scripts/tests/test_start_here_render.py run as part of the gating
 # `routing-self-tests` workflow, which is where a broken renderer must fail.
--- a/.github/workflows/routing-self-tests.yml
+++ b/.github/workflows/routing-self-tests.yml
@@ -57,6 +57,13 @@ on:
       - "scripts/triage.py"
       - "scripts/retriage.py"
       - ".github/workflows/triage-issue.yml"
+      # [OPUS-5] the maintainer front door (#1135 / #4145) — the curated TOML, the renderer
+      # that drops resolved entries from it, and the suite that pins both plus the
+      # refresh-start-here.yml wiring.
+      - "orchestration/start-here.toml"
+      - "scripts/render-start-here.py"
+      - "scripts/tests/test_start_here_render.py"
+      - ".github/workflows/refresh-start-here.yml"
       - ".github/workflows/routing-self-tests.yml"
       # [OPUS-5] reg#677 — PR `area:` attribution. The scheduler partitions by
       # `area:<name>`; an unlabelled PR takes the serializing `__global__` partition
@@ -108,6 +115,13 @@ on:
       - "scripts/triage.py"
       - "scripts/retriage.py"
       - ".github/workflows/triage-issue.yml"
+      # [OPUS-5] the maintainer front door (#1135 / #4145) — the curated TOML, the renderer
+      # that drops resolved entries from it, and the suite that pins both plus the
+      # refresh-start-here.yml wiring.
+      - "orchestration/start-here.toml"
+      - "scripts/render-start-here.py"
+      - "scripts/tests/test_start_here_render.py"
+      - ".github/workflows/refresh-start-here.yml"
       - ".github/workflows/routing-self-tests.yml"
       # [OPUS-5] reg#677 — PR `area:` attribution. The scheduler partitions by
       # `area:<name>`; an unlabelled PR takes the serializing `__global__` partition
@@ -157,6 +171,12 @@ jobs:
           # runs cannot go red.
           python3 scripts/pr-area-labels.py --self-test
           python3 scripts/tests/test_pr_area_labels.py
+          # [OPUS-5] #4145: the front-door renderer. --self-test validates the SHIPPED
+          # orchestration/start-here.toml (a malformed curated file must fail on ITS PR, not at the
+          # next scheduled refresh of the maintainer's front door); the suite pins the drop /
+          # fail-closed / truncation rules and refresh-start-here.yml's own wiring.
+          python3 scripts/render-start-here.py --self-test
+          python3 scripts/tests/test_start_here_render.py
 
       # [OPUS-5] The bd->issues migration pair, previously ungated on a PR (see the `paths:`
       # note above). Both are hermetic: stdlib only, every subprocess boundary stubbed, no
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -688,6 +688,30 @@ no network); the Python close-script carries a `--self-test`.
   fail-safe) is pinned by `scripts/tests/test_reconcile_merged_beads.sh`. The maintenance loop
   runs it (dry-run) each tick (step 0); the orchestrator reviews + applies the closes separately.
 
+- **`scripts/render-start-here.py [--dry-run | --self-test | --if-ref owner/repo#N]`** (#4145) —
+  renders the **maintainer front door**, issue
+  [#1135](https://github.com/sparq-org/sparq/issues/1135), from
+  **`orchestration/start-here.toml`** (the curated 🔴/🟡 copy — one `[[entry]]` per ask, with
+  `ref` / `bucket` / `ask` / `short`) **plus live GitHub state**. It exists because #1135 said
+  "auto-maintained" and never was: on 2026-07-26 it was 13 days stale and **seven of its entries
+  were already resolved** while still being shown to the maintainer as work needing them.
+  **The rule: the renderer may only DROP, ANNOTATE and ORDER what the TOML says — it never invents
+  an ask, so a new 🔴/🟡 item arrives as a PR to that file, and a hand edit of the issue is
+  overwritten by the next refresh.** It **drops any entry whose issue/PR is closed or merged**
+  (listing them under *Removed as resolved*), renders the maintainer-gated **PR table from live
+  `mergeable`/`mergeStateStatus`** so a PR that has rotted to `CONFLICTING` reads "needs rebase"
+  instead of "say arm it", and renders the `needs:user`/`needs:area`/`needs:ec2` populations and
+  the held-PR list from the **LIST API** (the search index lags label writes, so it under-reports).
+  **Fail-closed:** an entry is dropped only on positive evidence of resolution — an unreadable ref
+  is KEPT and marked *state unknown* while the process exits non-zero (sticky), and an **aggregate**
+  failure (a label count, the held-PR list) **aborts without publishing**. The issue is edited only
+  when the rendered text **differs** (no notification churn); 🟢 truncates before the
+  65 536-char body limit, 🔴 never does. `.github/workflows/refresh-start-here.yml` runs it on a
+  cron plus `issues`/`pull_request` **closed** (with `--if-ref` as the API-budget guard) and
+  `workflow_dispatch`; `routing-self-tests` runs `--self-test` +
+  `scripts/tests/test_start_here_render.py` so a broken renderer or a malformed TOML fails on ITS
+  PR.
+
 See `research/orchestration-automation-design.md` §1.3 (the five judgment behaviours that
 must stay with the orchestrator), §6 (failure modes + the guardrail behind each), and §5
 (the full phased stand-up plan with per-phase rollback).
--- a/orchestration/start-here.toml
+++ b/orchestration/start-here.toml
@@ -16,7 +16,10 @@
 #     (a hand edit is overwritten by the next refresh).
 #   * Every entry needs `ref` (or an explicit `untracked = true`), `bucket`, `ask`, `short`.
 #   * `ref` is the LIVE-STATE KEY only; it is never the source of the rendered prose (the 🔴 asks
-#     carry their own inline links). Cross-repo refs are allowed — "owner/repo#N".
+#     carry their own inline links). Cross-repo refs are allowed — "owner/repo#N" — but the refresh
+#     job runs on the repo-scoped GITHUB_TOKEN, so a ref in a PRIVATE repo reads as "state unknown":
+#     the entry is kept and marked (fail-closed), never dropped, and the job reds. Keep cross-repo
+#     refs public (jeswr/agent-account-registry is).
 #   * `ask` is free markdown on ONE logical line: the renderer collapses whitespace and re-wraps,
 #     so wrap it here for readability however you like. No blank lines inside a value.
 #   * `short` is the label used when the entry is dropped ("Removed as resolved").
--- a/orchestration/start-here.toml
+++ b/orchestration/start-here.toml
@@ -1,0 +1,467 @@
+# [OPUS-5] start-here.toml — the CURATED copy of the maintainer front door, issue #1135.
+#
+# WHY THIS FILE EXISTS (sparq-org/sparq#4145). #1135 is titled "auto-maintained" and never was:
+# it was refreshed by hand, and on 2026-07-26 it was 13 days stale with SEVEN entries already
+# resolved (#1848/#992/#1917 closed, #1084/#1973 closed-unmerged, #1791/#1155 merged) while still
+# being presented to the maintainer as work needing them. A front door that lists resolved work
+# burns the one resource the project cannot scale. So the de-staling is STRUCTURAL: this file holds
+# only the JUDGEMENT (which asks exist, how they are worded), and
+# `scripts/render-start-here.py` re-renders the issue body from it plus LIVE GitHub state, dropping
+# any entry whose issue/PR has closed or merged.
+#
+# CONTRACT (enforced by scripts/render-start-here.py --self-test and
+# scripts/tests/test_start_here_render.py):
+#   * The renderer may only DROP, ANNOTATE and ORDER what is written here. It NEVER invents an ask.
+#     A new 🔴/🟡 item therefore arrives as a PR to this file — not as a hand edit of the issue
+#     (a hand edit is overwritten by the next refresh).
+#   * Every entry needs `ref` (or an explicit `untracked = true`), `bucket`, `ask`, `short`.
+#   * `ref` is the LIVE-STATE KEY only; it is never the source of the rendered prose (the 🔴 asks
+#     carry their own inline links). Cross-repo refs are allowed — "owner/repo#N".
+#   * `ask` is free markdown on ONE logical line: the renderer collapses whitespace and re-wraps,
+#     so wrap it here for readability however you like. No blank lines inside a value.
+#   * `short` is the label used when the entry is dropped ("Removed as resolved").
+#   * `{placeholders}` in [[process]]/[[flight]]/[text] are substituted from LIVE counts; an
+#     unknown placeholder is a hard error, so a typo cannot ship a literal "{needs_user}".
+#
+# HOUSEKEEPING: entries that the renderer has dropped stay here (harmlessly re-listed under
+# "Removed as resolved") until a follow-up PR prunes them. Pruning is a copy edit; dropping is
+# automatic — that ordering is deliberate, because the automatic half is the one that must never
+# depend on anyone remembering.
+
+[meta]
+repo = "sparq-org/sparq"
+issue = 1135
+
+[text]
+header = '''
+🤖 **SPARQ agent** — your single front door. Decisions-first; only 🔴/🟡 need you.
+'''
+
+# Substituted: {date}, {dropped_sentence}.
+preamble = '''
+**Every item below was re-verified against the live GitHub API on {date}.** {dropped_sentence}
+This body is RENDERED by [`scripts/render-start-here.py`](https://github.com/sparq-org/sparq/blob/main/scripts/render-start-here.py)
+from [`orchestration/start-here.toml`](https://github.com/sparq-org/sparq/blob/main/orchestration/start-here.toml)
+on a schedule and whenever an issue or PR closes ([#4145](https://github.com/sparq-org/sparq/issues/4145)),
+so a closed or merged item drops out by construction and the PR table reads live conflict state.
+Changing what we ask you arrives as a PR to that TOML — a hand edit here is overwritten.
+'''
+
+blocked_intro = '''
+Each of these is human-only by nature — a credential, a licence, an external human, or your name on
+a public decision. No agent can close them.
+'''
+
+gated_intro = '''
+Protected paths and spec surfaces the fleet will not self-arm.
+'''
+
+process_intro = '''
+These cap everything else. Counts are live.
+'''
+
+# ---------------------------------------------------------------------------------------------
+# 🔴 Only you can unblock.
+# ---------------------------------------------------------------------------------------------
+
+[[entry]]
+ref = "sparq-org/sparq#53"
+bucket = "blocked"
+short = "npm bootstrap publish"
+ask = '''
+**npm bootstrap publish** — `@jeswr/sparq` is **still not live** (registry returns **404**).
+`publish.yml` (OIDC trusted publishing) is merged and ready; it needs your one-time `npm publish` +
+trusted-publisher registration. **Revoke the chat-sent token either way.**
+([#53](https://github.com/sparq-org/sparq/issues/53))
+'''
+
+[[entry]]
+# The ask is bead sq-qhy4, which has no GitHub issue of its own; #3274 (assurance promotion) is
+# GATED BY it, not equal to it, so #3274 closing must NOT drop this. Hence untracked: this entry
+# can only ever be removed by a human editing this file.
+untracked = true
+bucket = "blocked"
+short = "external cryptographer ZK audit (sq-qhy4)"
+ask = '''
+**External accredited-cryptographer ZK audit** (`sq-qhy4`, **P0**) — the bead says it plainly:
+*"agent-out-of-scope by definition."* sparq's ZK soundness claim currently rests entirely on
+internal single-model self-audits. Until an external cryptographer signs off, **no ZK
+security/privacy/integrity property may be relied on in production**, and `SECURITY.md` must keep
+its conservative "treat as untrusted" posture. This gates every production ZK claim and the
+assurance-promotion work ([#3274](https://github.com/sparq-org/sparq/issues/3274)).
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#3337"
+bucket = "blocked"
+short = "crates.io API token"
+ask = '''
+**[#3337](https://github.com/sparq-org/sparq/issues/3337) — crates.io API token / `cargo owner`.**
+Blocks Rust-side publication entirely.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1139"
+bucket = "blocked"
+short = "literature-DB API key"
+ask = '''
+**[#1139](https://github.com/sparq-org/sparq/issues/1139) — literature-DB API key** (Semantic
+Scholar): free key + signup, blocks the research-ingest lane.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#3382"
+bucket = "blocked"
+short = "RDFox licence"
+ask = '''
+**[#3382](https://github.com/sparq-org/sparq/issues/3382) — RDFox licence.** Needed to gather
+competitor numbers honestly; without it the performance-dominance tables cannot include RDFox and
+we must keep saying so.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1840"
+bucket = "blocked"
+short = "Noir upstream PR"
+ask = '''
+**[#1840](https://github.com/sparq-org/sparq/issues/1840) — Noir upstream PR**
+(noir-lang/noir#13314, BoundedVec capacity assert): review + flip from draft to ready. Upstream
+maintainers need a human author.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1615"
+bucket = "blocked"
+short = "LinkedIn post + profile pin"
+ask = '''
+**[#1615](https://github.com/sparq-org/sparq/issues/1615) — LinkedIn post + profile repo pin.**
+Text is drafted and ready; needs your click.
+'''
+
+[[entry]]
+# Cross-repo: the enrollment-template fix that carries this ask lives in the registry.
+ref = "jeswr/agent-account-registry#278"
+bucket = "blocked"
+short = "worker-account credentials"
+ask = '''
+**Worker-account credentials (agent-account-registry)** — `acct08` login and the missing
+`REGISTRY_SECRETS_PAT`. Enrollment cannot be automated because writing env secrets requires your
+auth. Directly limits fleet parallelism.
+'''
+
+[[entry]]
+# One ask, three drafted filings. The renderer drops the entry only when ALL THREE are resolved,
+# and annotates it when only some are — so a partial resolution can neither hide the remainder nor
+# keep claiming a filing that is already done.
+ref = [
+  "sparq-org/sparq#3338",
+  "sparq-org/sparq#3329",
+  "sparq-org/sparq#3385",
+]
+bucket = "blocked"
+short = "upstream filings needing your identity"
+ask = '''
+**Upstream filings that need your identity** —
+[#3338](https://github.com/sparq-org/sparq/issues/3338) (4 drafted `w3c/rdf-tests` issues),
+[#3329](https://github.com/sparq-org/sparq/issues/3329) (georust/geo feature request),
+[#3385](https://github.com/sparq-org/sparq/issues/3385) (offer the GML-SF parser as a standalone
+crate). All drafted; each wants a human to file it.
+'''
+
+# ---------------------------------------------------------------------------------------------
+# 🟡 Decisions — one line each. The Item cell is rendered from `ref`.
+# ---------------------------------------------------------------------------------------------
+
+[[entry]]
+ref = "sparq-org/sparq#3399"
+bucket = "decision"
+short = "final npm package name"
+ask = '''
+**final npm package name** (currently `@jeswr/sparq`). Blocks the npm bootstrap above and the JS
+docs surface — worth settling first.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1974"
+bucket = "decision"
+short = "ODRL JS API contract"
+ask = '''
+public JS API contract for ODRL policy evaluation; wasm32 probe works, +135 kB raw bundle. Pick the
+exposed surface.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1897"
+bucket = "decision"
+short = "merge-queue skip-push-CI soundness"
+ask = '''
+merge-queue skip-push-CI is **unsound** under SQUASH+ALLGREEN (squash mints a new SHA; the green
+gate lives on the ephemeral queue ref). Pick **A)** REBASE/MERGE + `max_entries:1`, **B)** leaner
+always-run push matrix *(recommended)*, **C)** drop.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1825"
+bucket = "decision"
+short = "SPQCPRM2 default emit"
+ask = '''
+default SPQCPRM2 (V2) on-disk emit **ON**? Reader ships; gated on a canonical EC2 measurement
+(`needs:ec2`).
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#3299"
+bucket = "decision"
+short = "GitHub Pages root owner"
+ask = '''
+reconcile the GitHub Pages **root** — which surface owns `/`.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1248"
+bucket = "decision"
+short = "tier-1 semver freeze"
+ask = '''
+ratify the tier-1 **semver freeze** — policy has landed, the declaration is yours.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1817"
+bucket = "decision"
+short = "/download alias contract"
+ask = '''
+`/download` alias contract + prerelease fallback.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1815"
+bucket = "decision"
+short = "CodeQL per-push end state"
+ask = '''
+⚠️ **premise now stale** — this asks whether to keep CodeQL per-push, but CodeQL was **disabled** on
+sparq (2026-07-18, merge latency) and the ruleset now requires only `gate`. Related PR
+[#3427](https://github.com/sparq-org/sparq/pull/3427) (CodeQL advisory-at-merge) is open. Decide the
+end state, then we close or rewrite both.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1690"
+bucket = "decision"
+short = "logo / brand pick"
+ask = '''
+logo / brand pick — concepts consolidated.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1587"
+bucket = "decision"
+short = "next-phase work-plan steer"
+ask = '''
+next-phase work-plan steer.
+'''
+
+# ---------------------------------------------------------------------------------------------
+# 🟡 Maintainer-gated PRs. `State` is rendered LIVE; `ask` is the CLEAN-state ask and is replaced
+# by the rebase ask whenever the PR has rotted to CONFLICTING.
+# ---------------------------------------------------------------------------------------------
+
+[[entry]]
+ref = "sparq-org/sparq#1712"
+bucket = "gated-pr"
+short = "zk-compose fail-closed coverage"
+what = "zk-compose fail-closed / determinism coverage"
+ask = '''
+cross-provider review in flight; I'll arm on a clean pair unless you object
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1607"
+bucket = "gated-pr"
+short = "Trust Expression proposal"
+what = "Trust Expression proposal draft"
+ask = '''
+"arm it"
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1509"
+bucket = "gated-pr"
+short = "zkSPARQL proposal draft"
+what = "zkSPARQL proposal draft (restarted from the codebase)"
+ask = '''
+"arm it"
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1658"
+bucket = "gated-pr"
+short = "zkSPARQL §7 extended fragment"
+what = "zkSPARQL §7 extended fragment + bounded-depth paths"
+ask = '''
+"arm it"
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1621"
+bucket = "gated-pr"
+short = "ODRL core as stratified N3"
+what = "ODRL core as stratified N3 rules + Rust-vs-N3 differential"
+ask = '''
+"arm it"
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1487"
+bucket = "gated-pr"
+short = "Kani harnesses for WAC/ACP"
+what = "Kani harnesses for the WAC/ACP decision core"
+ask = '''
+harness re-scoping decision
+'''
+
+# ---------------------------------------------------------------------------------------------
+# THE SEVEN STALE ENTRIES (#4145). These are the entries the hand-maintained revision was still
+# showing the maintainer on 2026-07-26 after they had already closed or merged. They are seeded
+# here DELIBERATELY and left for one cycle: the first automated render must drop all seven against
+# the live API and list them under "Removed as resolved", which is the end-to-end proof that the
+# drop is structural rather than a habit. The next housekeeping PR prunes this block.
+# ---------------------------------------------------------------------------------------------
+
+[[entry]]
+ref = "sparq-org/sparq#1848"
+bucket = "decision"
+short = "geo dict-pointer freshness"
+ask = '''
+geo dict-pointer freshness.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#992"
+bucket = "decision"
+short = "empty-container default flip"
+ask = '''
+empty-container default flip.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1917"
+bucket = "decision"
+short = "miri formal-alarm"
+ask = '''
+miri formal-alarm scope.
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1084"
+bucket = "gated-pr"
+short = "release v0.1.0"
+what = "release v0.1.0"
+ask = '''
+"arm it"
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1791"
+bucket = "gated-pr"
+short = "agent-role refresh"
+what = "agent-role refresh"
+ask = '''
+"arm it"
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1155"
+bucket = "gated-pr"
+short = "VC cryptosuite ingest bridge"
+what = "VC cryptosuite ingest bridge"
+ask = '''
+"arm it"
+'''
+
+[[entry]]
+ref = "sparq-org/sparq#1973"
+bucket = "gated-pr"
+short = "wasm policy feature probe"
+what = "wasm policy feature probe"
+ask = '''
+"arm it"
+'''
+
+# ---------------------------------------------------------------------------------------------
+# ⚠️ Process problems worth your steer. Counts are LIVE placeholders.
+# ---------------------------------------------------------------------------------------------
+
+[[process]]
+text = '''
+**`needs:user` has lost its meaning — {needs_user} open issues carry it.** Spot-checking, most are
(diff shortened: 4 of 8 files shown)


jeswr:
> 🤖 **SPARQ agent**

## Full mutation campaign — 50/50 mutants killed

Every guard in `scripts/render-start-here.py`, `orchestration/start-here.toml`,
`.github/workflows/refresh-start-here.yml` and the `routing-self-tests.yml` wiring was deleted or
inverted in a copy of the tree and `scripts/tests/test_start_here_render.py` re-run. Verbatim
harness output (`P*` = Python guard, `Y*` = YAML seam):

```
BASELINE: green
P1    killed by   test_an_unexpected_state_string_is_unknown_not_trusted, test_malformed_payload_is_unknown_not_open, test_missing_state_for_a_ref_is_unknown_not_open, test_refstate_is_resolved_requires_positive_evidence   <- is_resolved treats unknown state as resolved
P2    killed by   test_closed_but_unmerged_pr_is_dropped_as_closed, test_closed_issue_is_dropped_and_listed_as_removed, test_merged_pr_is_dropped_and_reported_as_merged, test_multi_ref_entry_drops_only_when_every_ref_is_resolved   <- drop rule never fires (nothing is ever dropped)
P3    killed by   test_partly_resolved_multi_ref_entry_is_kept_and_annotated   <- drop rule fires on ANY resolved ref (multi-ref entry lost)
P4    killed by   test_conflicting_pr_is_annotated_and_never_says_arm_it   <- CONFLICTING PR keeps the curated 'arm it' ask
P5    killed by   test_conflicting_pr_is_annotated_and_never_says_arm_it   <- state cell loses the CONFLICTING branch
P6    killed by   test_unknown_mergeability_is_not_reported_as_mergeable   <- mergeable=UNKNOWN reported as MERGEABLE
P7    killed by   test_a_partially_read_pr_is_unknown_even_though_it_looks_mergeable   <- unknown ref renders no 'state unknown' cell
P8    killed by   test_unreadable_ref_is_kept_and_marked_state_unknown   <- _annotate drops the state-unknown suffix
P9    killed by   test_partly_resolved_multi_ref_entry_is_kept_and_annotated   <- _annotate drops the partly-resolved note
P10   killed by   test_green_items_are_dropped_from_the_tail_first, test_oversized_body_truncates_green_and_keeps_red   <- green-section truncation loop removed
P11   killed by   test_a_red_section_that_cannot_fit_aborts_instead_of_being_cut   <- over-limit render publishes instead of aborting
P12   killed by   test_a_per_ref_failure_publishes_everything_but_still_exits_non_zero, test_a_red_section_that_cannot_fit_aborts_instead_of_being_cut, test_closed_issue_is_dropped_and_listed_as_removed, test_dry_run_touches_nothing_and_prints_the_body   <- red section is what gets truncated
P13   killed by   test_unsubstituted_placeholder_is_refused   <- leftover-placeholder sweep removed
P14   killed by   test_unknown_placeholder_name_fails_validation   <- unknown placeholder name passes validation
P15   killed by   test_identical_body_is_not_re_published   <- no-op churn check removed (always edits)
P16   killed by   test_a_held_pr_list_failure_aborts_without_publishing, test_a_label_count_failure_aborts_without_publishing   <- aggregate count failure swallowed (renders zeros)
P17   killed by   test_a_per_ref_failure_publishes_everything_but_still_exits_non_zero, test_degradation_is_recorded_for_every_unreadable_ref   <- per-ref degradation not recorded
P18   killed by   test_a_per_ref_failure_publishes_everything_but_still_exits_non_zero   <- main returns 0 regardless of degradation
P19   killed by   test_if_ref_for_an_unlisted_ref_costs_nothing   <- --if-ref cost guard removed
P20   killed by   test_a_label_count_failure_aborts_without_publishing, test_counts_use_the_list_api_not_the_lagging_search_index   <- counts read from the lagging search index
P21   killed by   test_a_refless_entry_must_declare_untracked   <- refless entry no longer needs `untracked = true`
P22   killed by   test_gated_pr_entry_needs_what_and_exactly_one_ref   <- gated-pr `what`/single-ref validation removed
P23   killed by   test_dry_run_touches_nothing_and_prints_the_body   <- dry-run falls through and publishes
P24   killed by   test_every_markdown_link_is_closed, test_no_link_destination_is_broken_across_a_line   <- wrapping breaks URLs at hyphens again
P25   killed by   test_an_unexpected_state_string_is_unknown_not_trusted   <- malformed API payload treated as open
P26   killed by   test_a_later_lower_severity_report_cannot_clear_an_earlier_failure   <- sticky status downgraded to last-write-wins
P27   killed by   test_the_curated_gated_pr_asks_never_hard_code_a_state   <- shipped TOML: an ask hard-codes a live PR state
P28   killed by   test_every_ref_in_the_2026_07_26_body_survives_the_migration, test_the_seven_stale_entries_are_seeded_so_the_first_render_drops_them   <- shipped TOML: a seeded stale entry silently disappears
Y1    killed by   test_fork_pull_request_events_are_skipped, test_no_job_or_step_is_disabled   <- job disabled with `if: false`
Y2    killed by   test_fork_pull_request_events_are_skipped   <- fork guard deleted
Y3    killed by   test_the_live_step_is_not_secretly_a_dry_run   <- live step secretly a --dry-run
Y3b   killed by   test_the_live_step_is_not_secretly_a_dry_run   <- the else-branch live invocation becomes a dry run
Y4    killed by   test_close_events_pass_the_if_ref_cost_guard, test_the_live_step_is_not_secretly_a_dry_run, test_the_self_test_runs_before_the_live_render   <- the render step is deleted entirely
Y5    killed by   test_the_job_has_exactly_the_writes_it_needs   <- job permissions downgrade issues to read
Y6    killed by   test_workflow_level_permissions_are_read_only   <- workflow-level permissions escalated to contents: write
Y7    killed by   test_every_action_is_sha_pinned   <- action unpinned to a floating tag
Y8    killed by   test_concurrency_is_serialised_with_a_cron_backstop, test_schedule_runs_several_times_a_day, test_schedule_trigger_exists   <- schedule trigger deleted
Y9    killed by   test_close_events_are_wired_for_prompt_removal   <- issues: [closed] trigger deleted
Y9b   killed by   test_close_events_are_wired_for_prompt_removal   <- pull_request: [closed] trigger deleted
Y10   killed by   test_workflow_dispatch_exists   <- workflow_dispatch deleted
Y11   killed by   test_close_events_pass_the_if_ref_cost_guard   <- --if-ref cost guard dropped from the call site
Y12   killed by   test_the_self_test_runs_before_the_live_render   <- self-test step deleted (order guarantee lost)
Y13   killed by   test_the_live_step_has_a_token   <- GH_TOKEN not passed to the live step
Y14   killed by   test_the_scripts_the_workflow_names_exist   <- wrong input: the run block names a script that does not exist
Y15   killed by   test_concurrency_is_serialised_with_a_cron_backstop   <- concurrency block deleted
Y16   killed by   test_checkout_does_not_persist_credentials   <- checkout persists credentials
Y17   killed by   test_this_suite_and_the_renderer_self_test_are_invoked   <- routing-self-tests no longer runs this suite
Y18   killed by   test_this_suite_and_the_renderer_self_test_are_invoked   <- routing-self-tests no longer runs the renderer --self-test
Y19   killed by   test_every_front_door_file_is_a_path_trigger_on_both_events   <- front-door files dropped from ONE paths filter
Y20   killed by   test_the_gate_job_is_not_disabled_and_stays_gate_discoverable   <- the self-test gate job renamed to advisory (demoted)

50/50 mutants killed
```

First pass left **three survivors** — each was a real gap, not an equivalent mutant, and each now
has a discriminating test:

| Survivor | Why it survived | Test added |
|---|---|---|
| `_pr_state_cell` unknown short-circuit | the fallback branch produced the same string for the fixture I had | `test_a_partially_read_pr_is_unknown_even_though_it_looks_mergeable` — a ref whose `error` is set *and* whose `mergeable` reads `MERGEABLE` |
| malformed-payload check | my fixture (`{"nonsense": true}`) failed closed by luck via the empty-state fallback | `test_an_unexpected_state_string_is_unknown_not_trusted` — `state` of `"reopened"` / `42` / `None` / absent |
| sticky exit status | `degrade()` was only ever called at one severity, so `max()` vs `=` was unobservable | `test_a_later_lower_severity_report_cannot_clear_an_earlier_failure` |

No guard is left without a named red test.

jeswr:
> 🤖 **SPARQ agent**

**Follow-up hardening** (head `b81e95e`), all three verified non-vacuous by mutation:

| Change | Why | Mutant → red test |
|---|---|---|
| `test_every_rendered_blocked_item_is_verbatim_from_the_toml` replaces a first-line prefix check | the old assertion compared only the FIRST wrapped line, so anything appended to the tail of an ask slipped through — vacuous for the exact property it was named after | renderer appends invented copy → `test_every_rendered_blocked_item_is_verbatim_from_the_toml`, `test_the_only_additions_are_the_declared_annotations` |
| `test_the_only_additions_are_the_declared_annotations` (new) | when an annotation DOES apply, it must be exactly one of the two declared ones | renderer duplicates an item / renames the unknown marker → both red |
| `publish()` unlinks its temp body file; cross-repo token limit documented | a private cross-repo ref reads as *state unknown* (kept, job reds) rather than being dropped — recorded in the TOML contract | n/a (hygiene + docs) |
| trust note on the `pull_request` trigger | a same-repo PR that edits the workflow runs its own version on close with `issues: write`; blast radius is bounded to editing issues, forks are skipped and read-only anyway. The exit is documented: drop the trigger and rely on `issues: [closed]` + cron | n/a (documented boundary) |

Suite is now **86 tests**; the full campaign re-run at this head is still **50/50 mutants killed**, plus the 3 above.

jeswr:
> 🤖 **SPARQ agent**

Reviewed at head `b81e95efd52a8a67dc14cc0f8c3e449bafea02f7` in an isolated worktree. `gate` = **success**
on this exact head (resolved newest-run-per-name over paginated `commits/<head>/check-runs`, id
`89809035125`, not the PR rollup). 86 tests + `--self-test` green at head.

**The design is right and the two catastrophic failure modes are closed.** I could not construct a case
where a real ask is dropped or invented. The verdict below is a `fail` on ONE thing: the body's claim
**"No guard is left without a red test"** is false, and I have 4 concrete counterexamples — one of them a
reachable live-path exit-code guard. Per the standing rule (delete the headline guard; if nothing goes
red that is blocking) that has to be a fail. The fix is three tests.

---

## What I verified independently — all pass

### 1. Fail-closed, proven by actually breaking refs against the LIVE API

Not a test name. I pointed two 🔴 refs at 404s — `sparq-org/sparq#53` (the **P0 npm-bootstrap
credential** ask) → issue `#99999901`, and `jeswr/agent-account-registry#278` → a nonexistent repo — then
ran `--dry-run` against the real API:

```
DEGRADED sparq-org/sparq#99999901: gh api ... -> exit 1: gh: Not Found (HTTP 404) (kept, never dropped)
DEGRADED jeswr/no-such-repo-xyzzy#278: gh api ... -> exit 1: gh: Not Found (HTTP 404) (kept, never dropped)
EXIT CODE = 3
```

Both asks **kept**, both marked `⚠️ state unknown`, all **9** 🔴 items still present, neither ref in the
*Removed as resolved* list, a `⚠️ This refresh ran DEGRADED` footer naming both, and the process exits
**3**. The live step runs under `set -euo pipefail`, so the job reds. Failure mode 1 does not occur.

The P0 external-cryptographer ZK audit (`sq-qhy4`) is `untracked = true` — structurally undroppable, and
`test_untracked_entry_is_never_dropped` pins it. Correct.

### 2. `--dry-run` fidelity against the live #1135 — set-equal, zero lost, zero added

```
live #1135 refs : 42        LOST  (in #1135, absent from render): []
dry-run refs    : 42        ADDED (in render, absent from #1135): []
fixture refs    : 42        fixture == live #1135 body, byte-identical: True
```

The fixture is genuinely an independent artifact (byte-identical to the API body), so
`test_every_ref_in_the_2026_07_26_body_survives_the_migration` can really detect a lost ask — mangling one
seeded ref in the TOML turns it red. Section counts reproduce the claim exactly: **9 🔴 / 10 🟡 decisions
/ 6 gated PRs / 5 process notes / 4 🟢**. All 7 stale entries dropped on the first live render
(`#1848 #992 #1917` closed, `#1084 #1973` closed-unmerged, `#1791 #1155` merged), and each appears **only**
in the *Removed as resolved* footer, never above it. Nothing was dropped from the 🔴 bucket (9 in TOML,
9 rendered).

### 3. Truncation, with oversized fixtures at the real 65 536 limit

| Case | Result |
|---|---|
| 400 huge 🟢 items, real `BODY_LIMIT` | body 65 260 ≤ 65 536; **all 🔴 intact**; green kept = items 0–151, a contiguous prefix ⇒ **tail dropped first**; notice rendered |
| 🔴 section alone 81 476 chars | `RenderAborted` — "refusing to truncate the sections that need a human" |
| `main()` on an oversized-🔴 TOML | exit **2**, `issue edit` calls = **0** — the issue is not touched |

🔴 aborts rather than being cut, as claimed.

### 4. The three self-reported findings are genuinely fixed, with discriminating tests

- **(a) the `{needs_ec2}` placeholder.** Reverting `PLACEHOLDER_RE` to the buggy `[a-z_]+` reds 15 tests,
  including two that discriminate the *digit* specifically: `test_known_placeholders_are_substituted`
  (`{needs_user}/{needs_area}/{needs_ec2}` → `11/22/33`) and `test_unknown_placeholder_name_fails_validation`
  (`{needs_ec3}`). Note `test_unsubstituted_placeholder_is_refused` uses `{needs_EC2}` (uppercase) and would
  pass under *both* regexes — it pins the `LEFTOVER_PLACEHOLDER_RE` body sweep, not the digit fix. Both are
  covered, just not by the test one might assume.
- **(b) `textwrap` un-linking URLs.** `break_on_hyphens=True, break_long_words=True` reds
  `test_no_link_destination_is_broken_across_a_line` + `test_every_markdown_link_is_closed`.
- **(c) the 3 replacement fixtures DISCRIMINATE** — I checked each satisfies only the right reading:
  - `test_a_partially_read_pr_is_unknown_even_though_it_looks_mergeable` — `mergeable="MERGEABLE"` **and**
    `error` set, so it passes only if the `state.unknown` short-circuit fires. Deleting the short-circuit
    reds exactly this test.
  - `test_an_unexpected_state_string_is_unknown_not_trusted` — `{"state": "reopened"}` is a *string*, so
    weakening the check to `isinstance(state, str)` alone still reds it. Discriminating.
  - `test_a_later_lower_severity_report_cannot_clear_an_earlier_failure` — `degrade(DEGRADED)` then
    `degrade(OK)`; `self._code = code` reds exactly this test.

### 5. Independent 71-mutant sweep — every mutant the review brief named DIES on structurally-parsed YAML

66/71 killed. All of these were caught by parsed job/step/permission/trigger objects, not substrings:

| Mutant | Red test |
|---|---|
| `if: false` on the job / on the live step / `${{ false }}` | `test_no_job_or_step_is_disabled` (+ fork-guard test) |
| DELETE the live render step | `test_the_live_step_is_not_secretly_a_dry_run`, `test_close_events_pass_the_if_ref_cost_guard` |
| DELETE the self-test step / reorder it after the live render | `test_the_self_test_runs_before_the_live_render` |
| `--dry-run` on the live step (**both** branches) | `test_the_live_step_is_not_secretly_a_dry_run` |
| `issues: write`→`read`, or dropped entirely | `test_the_job_has_exactly_the_writes_it_needs` |
| `contents: read`→`write` (workflow / job), `actions: write` added | `test_workflow_level_permissions_are_read_only`, `test_the_job_has_exactly_the_writes_it_needs` |
| drop `schedule` / `issues` / `pull_request` / `workflow_dispatch`; cron reduced to 1×/day | the 4 trigger tests |
| drop the fork guard; unpin `@v7`; `persist-credentials: true`; drop `--if-ref`; drop `GH_TOKEN`; name a nonexistent script | the matching wiring tests |
| `routing-self-tests` stops invoking either script; front-door file dropped from **one** `paths:` filter; gate job renamed "advisory" | `TestSelfTestWorkflowWiring` |
| unknown→resolved via `any()` not `all()`; drop rule disabled; malformed payload→open; missing ref→open; UNKNOWN→MERGEABLE; CONFLICTING keeps "arm it"; sticky→assignment; `main` returns 0; aggregate failure renders zeros; churn check removed; every truncation rung; buggy placeholder regex; leftover sweep removed; URL wrapping; **counts from `search/issues` or `--search`**; every config validation; `--if-ref` guard bypassed; **renderer invents an ask**; **renderer rewords an ask**; dry-run secretly publishes; a seeded stale ref mangled | all killed, each by a named test |

One survivor is a deliberate non-mutant: quoting `"on":` survives because `triggers()` reads both
spellings on purpose. Correct.

**Cannot invent or reword an ask.** Appending ` (URGENT: do this first)` to every 🔴 item, and truncating
each ask to its first sentence, both red `test_every_rendered_blocked_item_is_verbatim_from_the_toml` /
`test_every_table_ask_cell_is_a_curated_ask_or_the_rebase_annotation` — exact set-membership against the
real TOML, and `_blocked_items` reassembles wrapped lines rather than prefix-matching, so tail additions
are caught. `REBASE_ASK` is the one synthesised string and it is required by #4145.

### 6. Label counts, hygiene, privilege

- **LIST API, pinned.** Rewriting `label_count` to `api search/issues?q=...` **or** to `issue list --search`
  both red `test_counts_use_the_list_api_not_the_lagging_search_index`. Every decaying figure in the TOML is
  a `{placeholder}` (`{needs_user} {needs_area} {needs_ec2} {held_prs} {held_prs_conflicting}`) — no
  hard-coded count survives.
- **No hard-coded perf numbers.** `check-no-perf-numbers.py --enforce` clean (266 files). I also copied
  both `orchestration/start-here.toml` and the rendered body in as `*.md` and re-ran the gate: **0 findings
  across 268 files** — so the claim holds even under the gate that does not currently cover them. The
  dropped 🟢 item was the decaying `**Today:** 73 commits / 73 merged PRs` line; it was never an ask.
  *(Advisory: `SCAN_GLOBS = ["*.md", "*.typ"]`, so the new TOML — the one file whose content is published —
  is not in the gate's scope. Nothing to fix today; worth a follow-up bead.)*
- **`pull_request` trigger — argued correctly in conclusion, imprecise in its reasoning.** A same-repo PR
  editing this file does run its own version on close. But it also controls its own `permissions:` block, so
  the radius is bounded by *push access*, not by `issues: write` as the comment states. That distinction does
  not matter here, because the repo already accepts strictly wider versions of exactly this: `auto-arm.yml`
  (`contents: write` + `pull-requests: write`), `bench.yml` (contents/deployments/pull-requests/issues write),
  `ci.yml` and `fuzz.yml` (contents+issues write), `flow-on.yml` (issues: write) are all `pull_request`-triggered
  today. This job is the **least**-privileged write-capable one of the set, forks are skipped by the job `if`
  *and* get a read-only token. **Not a new privilege class; no objection.** Token scoping is genuinely minimal:
  workflow-level `contents: read`; job adds only `issues: write` (for `gh issue edit`) and `pull-requests: read`
  (for `mergeable`/`mergeStateStatus`), `persist-credentials: false`, action SHA-pinned.
- Cross-repo refs: `jeswr/agent-account-registry` is public (verified), and the private-repo → "state unknown"
  → job-reds consequence is documented in the TOML header. Good.

---

## The finding: "No guard is left without a red test" is false — 4 counterexamples

I deleted each of these in a copy of the tree and ran `test_start_here_render.py` + `--self-test`. Nothing
went red. All four are **correct at this head** — these are missing regression tests, not live defects.

**F1 (the one that matters) — the sticky non-zero exit is unpinned at 2 of its 3 delivery points.**
`return status.code` → `return EXIT_OK` survives at **both** the `--dry-run` branch and the
`already up to date` branch of `main()`. Only the publish call site is pinned
(`test_a_per_ref_failure_publishes_everything_but_still_exits_non_zero`).

The `already up to date` branch is reachable and is exactly where you need the red most: a 403/404 on a 🔴
ref persists, the previous refresh already published the degraded body, so the next cron renders an
*identical* body → no edit → falls through that branch. Today it correctly returns 3. Because the branch
reads "already up to date", `return EXIT_OK` there is a *natural-looking* cleanup — and if anyone makes it,
the cron goes green while the front door carries an unverified P0 indefinitely with nobody alerted. That is
the exit-zero-swallowing class this repo has been bitten by three times, in the PR that advertises sticky
exit as a headline guard with a dedicated `TestStickyStatus` class.

*Fix:* one test that drives `main()` with an unreadable ref and a pre-matching body, asserting
`code == EXIT_DEGRADED` and `edits == []`; plus the same assertion for `--dry-run`.

**F2 — `if not out.mergeable: out.error = ...` in `live_ref_state` has no red test.** Deleting it:

```
WITH guard    cell='⚠️ state unknown'  exit=3  ask='"arm it" ⚠️ **state unknown** — live state could not be read…'
WITHOUT guard cell='⚠️ state unknown'  exit=0  ask='"arm it"'
```

The State cell still reads unknown (so the maintainer is not misled about state), but the ask reverts to the
curated `"arm it"` **and the job exits 0** — a PR whose mergeability was never read is presented as armable
with a green refresh. `test_pr_detail_failure_leaves_the_entry_unknown_not_mergeable` covers the *exception*
path; `test_unknown_mergeability_is_not_reported_as_mergeable` uses `mergeable="UNKNOWN"` (non-empty) and
never reaches this branch. *Fix:* a `gh pr view` stub returning `{"isDraft": false, "mergeStateStatus": "CLEAN"}`
with no `mergeable` key, asserting `state.unknown` and a degraded `Status`.

**F3 — `if self.unknown: return False` in `is_resolved()`, and the `not self.state` clause of `unknown()`,
are unreachable today, so the self-test assertion that names them passes for the wrong reason.** I enumerated
every `live_ref_state` exit:

```
api raises (403/404/timeout)  error=True  state=''      merged=False  unknown=True  resolved=False
payload has no state          error=True  state=''      merged=False  unknown=True  resolved=False
PR open, pr-view raises       error=True  state='open'  merged=False  unknown=True  resolved=False
issue closed / PR merged      error=False state='closed'              unknown=False resolved=True
```

Every error path leaves `state` at `''` or `'open'`, and `merged`/`closed` paths early-return before `error`
can be set — so no reachable `error + resolved` combination exists and deleting the clause changes nothing.
Consequently `assert not broken.is_resolved(), "an unreadable ref must NEVER count as resolved"` in
`_self_test` passes because `state == ""`, **not** because of the guard it is named after. That is a
non-discriminating fixture on the single most important invariant in the file. *Fix:* assert on a constructed
`RefState(ref=…, state="closed", error="HTTP 403")` (and `merged=True, error=…`), which is what the guard is
actually there for and what makes it load-bearing the moment any future caller sets `error` alongside a
resolved state.

## Summary

Everything substantive checks out: no ask can be dropped (proven with a live 404 on a P0), no ask can be
invented or reworded, 🔴 never truncates, counts come from the LIST API, the YAML seam is genuinely
structural and killed every mutant the brief named, privilege is least and introduces no new class, and no
hard-coded performance number is rendered. Failing only because the universal claim is refuted and F1 leaves
a reachable exit-zero path on the live code with no red test — three tests and I would flip this.

VERDICT: fail

jeswr:
> 🤖 **SPARQ agent**

## Review fixes — head `f57a823ab4eb3c6b03acd3428ae9c02370ee753d`

Author-only pass on the `VERDICT: fail`. I have **not** armed, merged, relabelled or reviewed my
own fix — a fresh reviewer judges this.

The verdict's four findings were all **missing regression tests, not live defects**, exactly as
diagnosed. Nothing was redesigned; the only behaviour change in the whole push is the perf-gate
coverage advisory (below). Every test added was verified **red on its exact mutation** and green at
head. The body's *"No guard is left without a red test"* is **withdrawn** and replaced with the run
below plus an explicit statement of what a mutation campaign can and cannot support.

### Mutation → test, for the findings

| # | Mutation applied | Red test |
|---|---|---|
| F1a | `--dry-run` branch: `return status.code` → `return EXIT_OK` | `test_dry_run_with_a_degraded_ref_still_exits_non_zero`, `test_a_failure_on_the_LAST_ref_survives_every_earlier_success`, `test_no_return_after_the_live_read_hard_codes_an_exit_code` |
| F1b | `already up to date` branch: `return status.code` → `return EXIT_OK` | `test_an_already_up_to_date_body_still_exits_non_zero_when_degraded`, `test_a_failure_on_the_FIRST_ref_survives_every_later_success`, `test_a_failure_on_the_LAST_ref_survives_every_earlier_success`, `test_no_return_after_the_live_read_hard_codes_an_exit_code` |
| F1c | publish branch: `return status.code` → `return EXIT_OK` | the above + `test_a_per_ref_failure_publishes_everything_but_still_exits_non_zero` |
| F1d | `Status.degrade`: `max(self._code, code)` → `self._code = code` | `test_a_later_lower_severity_report_cannot_clear_an_earlier_failure` |
| F1e | `collect_states`: per-ref degradation never recorded | all four interleaved/branch tests |
| F2a | delete `if not out.mergeable: out.error = …` | `test_a_pr_whose_mergeable_field_never_arrived_is_recorded_as_an_error`, `test_a_pr_view_missing_mergeable_degrades_and_drops_the_arm_it_ask` |
| F3a | delete `if self.unknown: return False` from `is_resolved()` | `test_a_resolved_looking_ref_whose_read_failed_is_not_resolved` **and** `render-start-here.py --self-test` |
| F3b | delete the `not self.state` clause of `RefState.unknown` | `test_a_ref_whose_state_never_arrived_is_unknown_even_with_no_error` **and** `--self-test` |
| F3c | delete the `error` clause of `RefState.unknown` | 4 tests incl. `test_a_partially_read_pr_is_unknown_even_though_it_looks_mergeable` |
| G1 | delete the `EXTRA_SCAN_PATHS` wiring in `check-no-perf-numbers.py::main` | `test_the_front_door_toml_is_inside_the_perf_number_gate` |
| G2 | `EXTRA_SCAN_PATHS = ()` | same |
| G3 | write a hard-coded perf number into `orchestration/start-here.toml` | `test_the_shipped_toml_carries_no_hard_coded_performance_number` |
| T3 | give the sq-qhy4 audit entry `ref = "sparq-org/sparq#3274"` | `test_the_external_audit_ask_stays_structurally_undroppable` |

**F3 specifically.** I kept the guard and made the assertion fail for the reason it names, rather
than deleting a branch that is the correct fail-closed rule. `_self_test`'s
`assert not broken.is_resolved()` now says *"a ref with NO readable state must never count as
resolved"* — which is what that fixture actually exercises — and two new fixtures
(`state="closed", error="HTTP 403"` and `merged=True, error="HTTP 403"`) pin the `unknown`
short-circuit itself. Verified: deleting the branch reds `--self-test` **and** the suite. The
docstring records honestly that no live `live_ref_state` exit produces that combination today, so
the guard is defence-in-depth against a future caller — which is why it needs a constructed fixture
rather than a live-path one.

**T3 is new, found by this campaign.** The sq-qhy4 external-cryptographer-audit ask is
`untracked = true` (structurally undroppable). Pointing it at `#3274` — the assurance-promotion
issue it *gates* but is not equal to — survived the entire suite: the renderer MECHANISM was pinned
(`test_untracked_entry_is_never_dropped`), the curated DECLARATION was not. The day #3274 closed,
that P0 would have left the front door on its own. Same class as F1–F3, so it is fixed the same way.

### The `SCAN_GLOBS` advisory — I added the coverage

Chose the glob over the follow-up bead. `orchestration/start-here.toml` now enters
`check-no-perf-numbers.py`'s default scan set via an explicit `EXTRA_SCAN_PATHS` (the same shape as
the existing `paper-evidence.json` entry) rather than widening `SCAN_GLOBS` to `*.toml`, which would
sweep every `Cargo.toml`/`deny.toml` where a bare number is configuration, not a claim. A new
`--list-scanned` prints the resolved list **from the same `files` variable the scan loop consumes**,
so the coverage test interrogates the gate instead of re-deriving its globs (a re-derivation would
agree with a broken gate). `--enforce` stays clean: 0 findings across 267 files, up from 266.

### Full campaign, verbatim harness output

51 single-edit mutants, each applied to a fresh copy of the tree at this head, then
`scripts/tests/test_start_here_render.py` + `render-start-here.py --self-test` re-run:

```
BASELINE: green
F1a   killed by   test_a_failure_on_the_LAST_ref_survives_every_earlier_success, test_dry_run_with_a_degraded_ref_still_exits_non_zero, test_no_return_after_the_live_read_hard_codes_an_exit_code   <- --dry-run branch returns a literal EXIT_OK instead of the sticky status
F1b   killed by   test_a_failure_on_the_FIRST_ref_survives_every_later_success, test_a_failure_on_the_LAST_ref_survives_every_earlier_success, test_an_already_up_to_date_body_still_exits_non_zero_when_degraded, test_no_return_after_the_live_read_hard_codes_an_exit_code   <- 'already up to date' branch returns a literal EXIT_OK
F1c   killed by   test_a_failure_on_the_FIRST_ref_survives_every_later_success, test_a_failure_on_the_LAST_ref_survives_every_earlier_success, test_a_per_ref_failure_publishes_everything_but_still_exits_non_zero, test_a_pr_view_missing_mergeable_degrades_and_drops_the_arm_it_ask   <- publish branch returns a literal EXIT_OK
F1d   killed by   test_a_later_lower_severity_report_cannot_clear_an_earlier_failure   <- sticky status downgraded to last-write-wins
F1e   killed by   test_a_failure_on_the_FIRST_ref_survives_every_later_success, test_a_failure_on_the_LAST_ref_survives_every_earlier_success, test_a_per_ref_failure_publishes_everything_but_still_exits_non_zero, test_a_pr_view_missing_mergeable_degrades_and_drops_the_arm_it_ask   <- per-ref degradation never recorded
F2a   killed by   test_a_pr_view_missing_mergeable_degrades_and_drops_the_arm_it_ask, test_a_pr_whose_mergeable_field_never_arrived_is_recorded_as_an_error   <- delete `if not out.mergeable: out.error = ...` in live_ref_state
F3a   killed by   test_a_resolved_looking_ref_whose_read_failed_is_not_resolved, render-start-here.py --self-test   <- delete `if self.unknown: return False` from is_resolved()
F3b   killed by   test_a_ref_whose_state_never_arrived_is_unknown_even_with_no_error, render-start-here.py --self-test   <- delete the `not self.state` clause of RefState.unknown
F3c   killed by   test_a_partially_read_pr_is_unknown_even_though_it_looks_mergeable, test_a_pr_view_missing_mergeable_degrades_and_drops_the_arm_it_ask, test_a_pr_whose_mergeable_field_never_arrived_is_recorded_as_an_error, test_a_resolved_looking_ref_whose_read_failed_is_not_resolved   <- delete the `error` clause of RefState.unknown
R1    killed by   test_partly_resolved_multi_ref_entry_is_kept_and_annotated   <- drop rule fires on ANY resolved ref (multi-ref entry lost)
R2    killed by   test_a_partially_read_pr_is_unknown_even_though_it_looks_mergeable   <- _pr_state_cell loses the unknown short-circuit
R3    killed by   test_conflicting_pr_is_annotated_and_never_says_arm_it, render-start-here.py --self-test   <- a CONFLICTING PR keeps the curated 'arm it' ask
R4    killed by   test_green_items_are_dropped_from_the_tail_first, test_oversized_body_truncates_green_and_keeps_red   <- green-section truncation loop removed
R5    killed by   test_a_red_section_that_cannot_fit_aborts_instead_of_being_cut   <- over-limit render publishes instead of aborting
R6    killed by   test_a_failure_on_the_FIRST_ref_survives_every_later_success, test_a_failure_on_the_LAST_ref_survives_every_earlier_success, test_an_already_up_to_date_body_still_exits_non_zero_when_degraded, test_identical_body_is_not_re_published   <- no-op churn check removed (always edits)
R7    killed by   test_a_held_pr_list_failure_aborts_without_publishing, test_a_label_count_failure_aborts_without_publishing   <- aggregate count failure swallowed (renders zeros)
R8    killed by   test_counts_use_the_list_api_not_the_lagging_search_index, test_every_needs_label_population_is_queried   <- label counts read via --search (the lagging index)
R9    killed by   test_dry_run_touches_nothing_and_prints_the_body, test_dry_run_with_a_degraded_ref_still_exits_non_zero   <- --dry-run falls through and publishes
R10   killed by   test_if_ref_for_an_unlisted_ref_costs_nothing   <- --if-ref cost guard removed
R11   killed by   test_every_markdown_link_is_closed, test_every_rendered_blocked_item_is_verbatim_from_the_toml, test_no_link_destination_is_broken_across_a_line, test_the_only_additions_are_the_declared_annotations   <- URL wrapping re-enabled (links break at hyphens)
R12   killed by   test_unsubstituted_placeholder_is_refused   <- leftover-placeholder body sweep removed
R13   killed by   test_a_clean_pr_read_is_not_degraded, test_a_failure_on_the_FIRST_ref_survives_every_later_success, test_a_failure_on_the_LAST_ref_survives_every_earlier_success, test_a_per_ref_failure_publishes_everything_but_still_exits_non_zero   <- the original buggy placeholder regex (cannot see a digit)
R14   killed by   test_an_unexpected_state_string_is_unknown_not_trusted   <- malformed API payload treated as open
R15   killed by   test_every_rendered_blocked_item_is_verbatim_from_the_toml, test_every_table_ask_cell_is_a_curated_ask_or_the_rebase_annotation, test_the_only_additions_are_the_declared_annotations   <- renderer INVENTS copy on every ask
R16   killed by   test_a_pr_view_missing_mergeable_degrades_and_drops_the_arm_it_ask, test_a_ref_whose_state_never_arrived_is_unknown_even_with_no_error, test_unreadable_ref_is_kept_and_marked_state_unknown   <- _annotate drops the state-unknown suffix
G1    killed by   test_the_front_door_toml_is_inside_the_perf_number_gate   <- the front-door TOML is dropped back out of the perf-number gate's scope
G2    killed by   test_the_front_door_toml_is_inside_the_perf_number_gate   <- EXTRA_SCAN_PATHS emptied
G3    killed by   test_the_shipped_toml_carries_no_hard_coded_performance_number   <- a hard-coded performance number is written into the curated TOML
Y1    killed by   test_fork_pull_request_events_are_skipped, test_no_job_or_step_is_disabled   <- the job is disabled with `if: false`
Y1b   killed by   test_fork_pull_request_events_are_skipped   <- the fork guard is deleted
Y2    killed by   test_the_live_step_is_not_secretly_a_dry_run   <- the live on-close step is secretly a --dry-run
Y2b   killed by   test_the_live_step_is_not_secretly_a_dry_run   <- the ELSE-branch live invocation is secretly a --dry-run
Y3    killed by   test_the_job_has_exactly_the_writes_it_needs   <- job permissions downgrade issues to read
Y4    killed by   test_workflow_level_permissions_are_read_only   <- workflow-level permissions escalated to contents: write
Y5    killed by   test_checkout_does_not_persist_credentials   <- checkout persists credentials
Y6    killed by   test_every_action_is_sha_pinned   <- the checkout action is unpinned to a floating tag
Y7    killed by   test_concurrency_is_serialised_with_a_cron_backstop, test_schedule_runs_several_times_a_day, test_schedule_trigger_exists   <- the schedule trigger is deleted
Y8    killed by   test_close_events_are_wired_for_prompt_removal   <- the issues:[closed] trigger is deleted
Y9    killed by   test_close_events_are_wired_for_prompt_removal   <- the pull_request:[closed] trigger is deleted
Y10   killed by   test_workflow_dispatch_exists   <- workflow_dispatch is deleted
Y11   killed by   test_close_events_pass_the_if_ref_cost_guard   <- --if-ref cost guard dropped from the call site
Y12   killed by   test_the_self_test_runs_before_the_live_render   <- the self-test step is deleted (order guarantee lost)
Y13   killed by   test_the_live_step_has_a_token   <- GH_TOKEN is not passed to the live step
Y14   killed by   test_the_scripts_the_workflow_names_exist   <- the run block names a script that does not exist
Y15   killed by   test_concurrency_is_serialised_with_a_cron_backstop   <- the concurrency block is deleted
Y17   killed by   test_this_suite_and_the_renderer_self_test_are_invoked   <- routing-self-tests stops invoking this suite
Y18   killed by   test_this_suite_and_the_renderer_self_test_are_invoked   <- routing-self-tests stops invoking the renderer --self-test
Y19   killed by   test_every_front_door_file_is_a_path_trigger_on_both_events   <- front-door files dropped from ONE of the two paths filters
T1    killed by   test_every_ref_in_the_2026_07_26_body_survives_the_migration, test_the_seven_stale_entries_are_seeded_so_the_first_render_drops_them   <- a seeded stale ref is silently mangled (the entry stops being droppable)
T2    killed by   test_the_curated_gated_pr_asks_never_hard_code_a_state   <- a curated gated-PR cell hard-codes a live PR state
T3    killed by   test_the_external_audit_ask_stays_structurally_undroppable   <- the sq-qhy4 audit entry is made droppable by pointing it at #3274

51/51 mutants killed
```

Two mutants from an earlier pass are **not** in that count because they did not apply cleanly and
were rewritten rather than scored: a duplicate-key `if: false` (PyYAML silently keeps the last
key, so it was a no-op, not a survivor) and an ambiguous `contents: read` anchor that matched both
the workflow- and job-level blocks. Both are represented above by Y1 and Y4 in their correct form.

**What this number does not say.** A campaign is a lower bound: it speaks only about the mutants it
ran. That is precisely how the previous universal claim went wrong, so the body now states the
bound rather than an adjective.

### Gate + self-test results (actual)

| Leg | Result |
|---|---|
| full `routing-self-tests` run block (both steps) | **green** — `routing-validate` (+ self-test), `route-resolve`, `ready-issues`, `dispatch-plan`, `batch-merge`, `test_readiness_visibility.py`, `render-start-here.py --self-test`, `test_start_here_render.py` **99 tests OK**, `bd-to-issues.py --self-test`, `ci-close-merged-beads.py --self-test` |
| `check-no-perf-numbers.py --enforce` | **green** — 0 findings across 267 files |
| `check-terminology.py` + `--self-test` + `test_banned_terminology.py` | green |
| `check-advisory-registry.py`, `check-fast-fix-ring-guard.py`, `check-install-action-tool.py`, `check-skill-frontmatter.py`, `check-privacy-claims.sh`, `check-readme-template.py --enforce` | green |
| `markdownlint-cli2` | green — 419 files, 0 errors |
| `typos` over the gate's own docs scope + `test_typos_allowlist.sh` | green — 249 files clean, 16 allow-list assertions pass |

Suite went 86 → **99 tests**. Not armed, not labelled, no self-verdict.

jeswr:
> 🤖 **SPARQ agent**

**CI at head `f57a823ab4eb3c6b03acd3428ae9c02370ee753d`: `gate` = success** (newest run per name over
`commits/<head>/check-runs`, id `89814629721`).

The first `gate` attempt (`89814005516`) red on a **GitHub transient**, not on this diff:
`detect rust changes (feature-matrix)` failed inside `dorny/paths-filter`'s `listFiles` call with

```
##[error]Server Error: Sorry, this diff is temporarily unavailable due to heavy server load.:
{"resource":"PullRequest","field":"diff","code":"not_available"}
```

That job re-ran unchanged and succeeded, which is the confirmation that the cause was the API and
not the PR — this diff touches no Rust at all. `Routing contract self-tests` (the leg that runs
`render-start-here.py --self-test` + the 99-test suite) and `docs-quality quick-gates` (the leg that
runs `check-no-perf-numbers.py --enforce`, which this PR now extends) both passed on the **first**
attempt.

Still author-only: not armed, not labelled, no self-verdict.

jeswr:
> 🤖 SPARQ agent

**Rebased and un-conflicted.** Author-side note; no verdict here, a separate reviewer adjudicates.

This one was never red — `gate` was already green. It was **`CONFLICTING`**, 47 commits behind main, and would not have merged.

**The conflict** was a single file, `.github/workflows/routing-self-tests.yml`, in three hunks — all pure both-sides additions. Main added the model-deprecation triggers (`463ae09cf`, #4211) and the PR-`area:` deriver (`f71390721`, #4170) to the same `paths:` lists and the same `run:` block that this PR appends the front-door renderer to. Resolved by **keeping both sides**, main's first. Nothing from either side was dropped: the net diff against main is `+3088 / -0` across 8 files.

Rebased all 9 commits onto `origin/main` (rerere on; no `git checkout -- .` anywhere, per the destroyed-work incident earlier today).

**Verification.** The full gating step re-run locally, all 12 legs green:

```
ok   scripts/routing-validate.py --self-test        ok   scripts/tests/test_no_deprecated_models.py
ok   scripts/routing-validate.py                    ok   scripts/pr-area-labels.py --self-test
ok   scripts/route-resolve.py --self-test           ok   scripts/tests/test_pr_area_labels.py
ok   scripts/ready-issues.py --self-test            ok   scripts/render-start-here.py --self-test
ok   scripts/dispatch-plan.py --self-test           ok   scripts/tests/test_start_here_render.py
ok   scripts/batch-merge.py --self-test             ok   scripts/tests/test_readiness_visibility.py
```

Plus `check-no-perf-numbers.py` (0 findings / 270 files) and markdownlint on `AGENTS.md` (0 errors).

**The resolution itself is guarded.** Because I hand-resolved a conflict *inside* a workflow's `run:` block — exactly the YAML seam where this repo's uncaught mutants live — I mutated it: dropping this PR's two `python3` invocations reds a **named** test,

```
FAIL: test_this_suite_and_the_renderer_self_test_are_invoked
      (TestSelfTestWorkflowWiring)
AssertionError: 'python3 scripts/render-start-here.py --self-test' not found in ...
```

so a resolution that had silently dropped the PR's side could not have merged green. Credit to the PR as authored — that guard already existed.

**Result at `03362c0e`: `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, `gate` = pass, 0 failing checks.** (Was `CONFLICTING`/`DIRTY`.)

jeswr:
> 🤖 **SPARQ agent** — independent review (reviewer only; I authored none of this and pushed nothing).

Reviewed at head **`03362c0e5b6e23c17f701ebec516b4fb84517078`**, re-verified immediately before posting (`state=OPEN`, `draft=false`, `MERGEABLE/CLEAN`, `updated=2026-07-26T23:53:39Z`). Gate confirmed by a **paginated** `commits/<sha>/check-runs` read — `total_count=106`, so the single-page read that returns no `gate` row at all was avoided — matching on the **exact** name (not a prefix, so `draft-tier` cannot be mistaken for `gate`): one `gate` row, `completed/success`, and `0` non-success check runs. The leg that actually executes this suite, `Routing contract self-tests`, is `success`.

Work was done in an isolated worktree under scratchpad. Mutations were reverted from a `cp` snapshot with an md5 marker check — never `git checkout -- .` — and the tree was marker-verified clean at the end (`git status --porcelain` empty, 99/99 tests green again).

**Baseline:** `--self-test: OK`, and `Ran 99 tests ... OK`.

**The headline fail-closed claim, driven end-to-end through `main()` with my own fake `gh`** (I wrote my own driver rather than accept the author's table). Injected partial-data cases against the real shipped TOML (32 entries / 33 refs):

| injected failure | exit | published? | 🔴 items | disclosure |
|---|---|---|---|---|
| baseline, all reads clean | 0 | yes | **9** | — |
| a referenced issue **404s** | **3** | yes | **9** | `state unknown` ×11, DEGRADED banner |
| per-ref **page short-read** (payload with no `state`) | **3** | yes | **9** | `state unknown` ×11, DEGRADED banner |
| per-ref read returns **empty stdout** | **3** | yes | **9** | `state unknown` ×11, DEGRADED banner |
| label-count API fails (aggregate) | **2** | **NO** | — | `ABORT: could not read live state` |
| held-PR list API fails (aggregate) | **2** | **NO** | — | `ABORT: could not read live state` |
| current-body read fails | **2** | **NO** | — | `ABORT` |

This answers the specific worry directly: **a short read cannot shorten or empty the front door.** The 🔴 count stays at 9 in every degraded case, identical to baseline — entries come from the curated TOML and each ref fails to *unknown → KEPT*, never to *absent*. Aggregate failures abort without touching the issue, as claimed.

**Live dry-run against real GitHub state** (read-only): exit 0, and it dropped exactly the seven stale entries the PR claims — `#1848`, `#992`, `#1917`, `#1084` (closed), `#1791`, `#1155` (merged), `#1973` (closed). That is the headline behaviour reproduced against production state, not a fixture.

**The seed fixture is genuine, not circular.** `scripts/tests/fixtures/start-here-1135-2026-07-26.md` diffs against the live #1135 body by a single trailing newline (10233 vs 10234 bytes); all **42/42** refs match. It is a real capture, so the fidelity test is not asserting the TOML against itself.

**23 mutants run on the live tree — 22 killed, 1 survived.** All kills were behaviour-kills with a named failing test, not crash-kills:

- `is_resolved`'s `if self.unknown: return False` deleted → `test_a_resolved_looking_ref_whose_read_failed_is_not_resolved` + self-test. The guard the author's own docstring flags as needing a constructed fixture is genuinely load-bearing.
- `unknown` reduced to `error is not None` → `test_a_ref_whose_state_never_arrived_is_unknown_even_with_no_error`
- never-fetched ref defaulting to `closed` → `test_missing_state_for_a_ref_is_unknown_not_open`
- aggregate failure publishing zeros instead of aborting → both `..._aborts_without_publishing` tests
- sticky exit → `max(...)` → `= code` → `test_a_later_lower_severity_report_cannot_clear_an_earlier_failure`
- YAML seam: `--dry-run` on the live step, deleted pre-render self-test step, deleted fork guard, `contents: write` escalation, `if: false`, dropped `--if-ref`, `paths:` entries removed — each red on a distinct named wiring test.
- truncating 🔴 instead of aborting; dropping the truncation disclosure; inventing ask text; `unknown → MERGEABLE`; CONFLICTING keeping "arm it"; `all(...)` → `any(...)`; removing the TOML from the perf gate; deleting the `sq-qhy4` audit ask — all killed.

**The specific trap this repo hit yesterday was tested and is absent.** In a YAML literal block `#` lines are literal text, so a wiring assertion can be satisfied by the step's own *comment*. I deleted both invocations from `routing-self-tests.yml` while **leaving the explanatory comment in place** (verified still present at line 174) — `test_this_suite_and_the_renderer_self_test_are_invoked` still went red. The assertion demands the real command lines.

**The perf-gate coverage claim is real, not vacuous.** I inserted `"sparq is 3.7x faster than Oxigraph at 1.2M triples/sec"` into a rendered `[[process]]` field; the gate fired both explicitly and in the default scan: `orchestration/start-here.toml:27: perf-number [speed-up ratio]: '3.7x faster'`, `rc=1`.

**Honesty checks:** no `trust envelope` anywhere in the added files. The rendered body's ZK item is *correctly* qualified — it states soundness "rests entirely on internal single-model self-audits" and that "no ZK security/privacy/integrity property may be relied on in production" until `sq-qhy4` signs off. No unqualified soundness claim. The perf gate reports 0 findings on the shipped TOML.

## Non-blocking findings

1. **One mutant survived.** Appending `|| true` to `python3 scripts/tests/test_start_here_render.py` in `routing-self-tests.yml` leaves the suite green — the assertion is `assertIn`, so the substring still matches. The shipped workflow has no `|| true` (and the block runs under `set -euo pipefail`), so this is a hardening gap against a future edit, not a defect here.
2. **A successful-but-wrong aggregate read does not abort.** I measured on this repo that `gh issue list --label needs:doesnotexist-xyz` returns `[]` with `rc=0`. So a *renamed* label would publish an understated count rather than aborting — rule 3 covers failures, not silent emptiness. Bounded: it can only understate summary prose, no ask is dropped. Related: the `--limit 1000` / `--limit 300` caps truncate silently, but live populations are 30/16/37 issues and 28 held PRs — 3–9% of the caps.
3. **Honest caveat on my own YAML kills.** My Y1/Y2 kills came from running the suite by hand. In CI, deleting the invocation is self-concealing: the `paths:` filter re-runs the job, but the mutated `run:` block no longer invokes the suite that would object. That is inherent to self-guarding wiring assertions, pre-existing and repo-wide (the author cites the #3423 precedent and follows the same pattern as `test_pr_area_labels.py`), and there is no repo-wide meta-gate enumerating `scripts/tests/*.py`. Worth a follow-up issue, not a reason to hold this PR — a guard that reds when run beats no guard.
4. **Wording nit.** The preamble asserts "Every item below was re-verified against the live GitHub API on `{date}`" unconditionally; in a degraded render that is not strictly true of the unknown refs, though it is disclosed inline (`⚠️ state unknown`) and by the footer DEGRADED banner.

None of these admit a partial or truncated front door, which is the load-bearing property. The fail-closed claim holds under every partial-data case I could construct.

VERDICT: pass

Issue #1111, opened 2026-06-21
[revisit-with-fable] Neurosymbolic self-built KB — re-attempt + re-iterate the whole direction on a stronger model


jeswr:
> 🤖 **SPARQ agent** — umbrella note for @jeswr. **Tagged to REVISIT + re-iterate when a stronger model (Fable) is available** (currently building on Opus 4.8; Fable unavailable). Per the maintainer: "this research direction effectively building out a neurosymbolic knowledge base could be highly valuable and is also quite novel… I would like to be able to properly re-attempt and re-iterate on [it] when a smarter model like Fable is available again. Please don't let that stop you taking it as far as you can right now."

## The direction: a neurosymbolic, self-built knowledge base

A hybrid of **symbolic** (the PKG RDF graph + a foundational ontology + reasoner/SHACL/closure) and **neural/vector** (LLM extraction + the cheap-model NL-tool + vector concept-resolution), built and used by the agent to develop the project. The threads, all started this session:

- **PKG dogfooding** (epic sq-2m6zm) — ontology + ingestion + the `query-pkg` skill; measured: a Haiku NL-tool is ~30× cheaper than reading docs at equal quality (`bench/pkg-dogfood/`).
- **FO/LLM fluency bridge** (epics sq-mztg8, sq-8dkyo) — measured that schema.org ≫ gUFO for the *agent* (fluency, not formal richness); researching teach/facade/hide directions + the schema-alignment-inference facade hack + structured markdown→KG authoring; broadening FO coverage (gist/BFO/size-axis).
- **Provenance-driven GenAI** (epic sq-2489d, issue #1110) — make PROV-O load-bearing (quality→embedding-weight, answer-qualification, citations) + complementary metadata (DQV/CiTO/FaBiO/nanopub) + scaled provenance-stamped literature-trawling (Haiku over lit-DB APIs).

## Why revisit with a stronger model

Every measured result here is **model-dependent** — the headline finding is literally that the *agent's fluency* drives outcomes (schema.org > gUFO *because of* the model, per Köhler & Neuhaus on LLMs' unstable implicit ontology). A stronger model (Fable) could shift the verdicts: it may wield richer foundational ontologies fluently (reopening gUFO/DOLCE/BFO), need less facade/hiding, and do higher-fidelity literature extraction. So the whole benchmark suite (FO-KM, the bridge directions, the provenance uses) is worth **re-running on Fable** — the harnesses (`bench/`) are built to make that a re-run, not a rebuild.

[...]

Issue #3241, opened 2026-07-17
sq-rbh10: sparq-reason-el E4 (optional): concurrent lock-free saturation (ELK-style contexts + work-stealing)


jeswr:
Optional multiplier: concurrent lock-free saturation (ELK §6 contexts + AtomicBool + work-stealing); sub-linear, larger ontologies benefit most. Acceptance: same subsumption-count correctness vs ELK as E3, with a measured (non-canonical) speedup on a large ontology. Depends on E3. Source: research/owl2-el-ql-reasoning-spike.md §5 EL-track E4 (#931). [OPUS-4.8]

Issue #2575, opened 2026-07-17
sq-43twz: fix(fuzz): randomized cargo-fuzz lane RED on main (pre-existing, stalls merge queue) — canonicalize_nquads quad-count/idempotence assert (fuzz_targets/canonicalize_nquads.rs:51) + sparq-reason n3/parser.rs:779 Utf8Error unwrap panic; repro artifacts in runs 29152138161 + 29118998843; pre-dates PRs 1930/1928/1931 (also failed on main 942651f05); supersedes GH013-rejected auto-bead; durable record issue #1903


jeswr:
REMAINING (verified 2026-07-14): parser.rs Utf8 unwrap was FIXED by #2011 (no unwraps left). The still-RED fuzz lane on main (fuzz.yml push runs 09:44/11:13/12:48 all FAILURE) is the canonicalize_nquads target — a committed known-crasher triggers the quad-count/idempotence assert. This is the c14n correctness invariant: SOUNDNESS-SENSITIVE (c14n underpins signing/hashing) -> route to Opus/Fable + adversarial review, do NOT cheap-drain. Note: merge-QUEUE fuzz (bounded) passes; only the longer push fuzz reliably hits it, so it does not block merges.

PR #4187, opened 2026-07-26
docs(research): bounded context sharing between fleet agents — three-channel design

jeswr:
> 🤖 **SPARQ agent**

Design record answering a maintainer request: *"design a strategy for agents to be able to appropriately
share context between related tasks … We should also be careful to make sure that this doesn't result in
continually increasing context being given across agents. Perhaps we should be looking at having some kind
of 'memory' that all agents maintain and contribute to."*

**Design record only — no behaviour change.** Rollout is staged and measurement-gated in §6.

## The core move: three channels, never conflated

Most shared-memory designs fail by putting everything in one growing store. Separating by **lifetime and
derivability** is what makes growth bounded:

| Channel | Contains | Growth | Retrieval |
|---|---|---|---|
| **A. Lineage handoff** | what the previous stage of *this task* learned | **bounded by schema** | pushed |
| **B. Method memory** | transferable lessons about *this codebase* | **asymptotic** | pulled, on demand |
| **C. Live state** | heads, labels, gate status, counts | **never stored** | queried |

**Channel C is a rule, not a store: never memorise what one API call derives.** Two concrete harms this
session came from violating it — a `VERDICT: pass` bound to a superseded head nearly armed a PR whose fix
had been retracted, and a park-cause finding recorded hours earlier was simply false by the time it was
reused. This one rule removes the largest source of both staleness and growth.

## Why it doesn't grow without bound

**Channel A** gives each stage exactly one slot and **replaces** it — never appends. If review round 3
carried rounds 1–2 verbatim, context would grow linearly in rounds, which is precisely the failure the
request names. Size is O(stages), not O(attempts).

**Channel B** admits an entry only if it is **non-derivable** (not recoverable by reading the repo or
making one API call) **and decision-changing**. Method lessons about a codebase are a bounded set; state
facts are unbounded. Restricting admission to the former makes the corpus flatten — and that is a
**falsifiable prediction**, adopted as the health metric: plot corpus size against cumulative tasks; linear
growth means admission is too loose, and the response is to tighten admission, never to add storage.

Retrieval is **pull, not push** — a one-line index, lazy load of the two or three that match, and a hard
per-agent budget. Bulk injection is exactly the shape this project measured as *losing*.

## Grounded in this project's own measurements

`pkg-query` halved Opus tokens (N=30) → adopted. Outline-first cost **~21k tokens more** than a scoped
read → rejected. `sparq-terse` saved **~0%** → rejected. Two of three plausible context ideas were
measured to be worthless or negative, so the record's binding rule is that **nothing ships unmeasured**,
against worker yield, with **stop conditions declared in advance** (§6).

## The highest-value first step

`worker.attempt.why_no_diff` — one field. Worker yield is 36% (70/126 over 196 runs), dominated by
`EXIT_CLASS: no_change`, and the same issue is retried up to **3×** with the same model and **no record of
why the previous attempt produced nothing**. Attempt 2 begins exactly as ignorant as attempt 1. Requiring a
departing worker to say *why* converts a silent failure into a routing signal, and feeds the
escalate-vs-decompose decision in registry #701 and the deadlock adjudication in registry #703.

Reuses existing infrastructure rather than rebuilding: Channel A is a schema over the marker comments the
fleet already writes; Channel B is PKG-backed (`crates/sparq-kb`,
`research/dogfooding-sparq-knowledge-graph.md`) with the `sparq-pkg-nl` cheap-model NL tool as the
retrieval path. Strategic parent: #1111.

### DIFF
--- a/research/agent-context-sharing.md
+++ b/research/agent-context-sharing.md
@@ -1,0 +1,231 @@
+# Context sharing between fleet agents — three channels, bounded by construction [OPUS-5]
+
+**Status:** design record, for maintainer review. Requested 2026-07-26: *"design a strategy for agents to
+be able to appropriately share context between related tasks; e.g. an agent that is working on a PR may
+[benefit] from some of the context from the agent that created the issue. We should also be careful to make
+sure that this doesn't result in continually increasing context being given across agents. Perhaps we
+should be looking at having some kind of 'memory' that all agents maintain and contribute to — like we have
+with claude code in the command line."*
+
+## 0. Grounding rules (binding on every claim below)
+
+1. **This project has already measured most plausible context tricks, and rejected two of three.**
+   `pkg-query` halved Opus tokens (N=30) → adopted. `ast-grep`/outline-first cost **~21k tokens MORE**
+   than a scoped read → rejected. `sparq-terse` saved **~0%** real tokens (N=30) → rejected as anything
+   but a legibility convenience. **Prior:** a context idea that sounds good is more likely than not to be
+   worthless or negative. Nothing here ships unmeasured.
+2. No hard-coded performance numbers in prose (repo hygiene). Figures below are dated observations.
+3. Where this record states a mechanism exists today, it was checked against the tree on 2026-07-26.
+
+## 1. The problem, measured rather than assumed
+
+Fleet agents are **ephemeral GitHub Actions jobs**. Each starts cold: a repo checkout, an issue body, a
+routed model. Nothing an earlier agent learned survives, except what happens to be written into the issue
+or PR.
+
+Observed consequences on 2026-07-26:
+
+- **Worker yield 70 success / 126 failure (36%)** over 196 runs, dominated by `EXIT_CLASS: no_change` —
+  the model runs and produces no diff. Every `no_change` target sampled was a genuinely hard,
+  well-specified task (lock-free saturation, morsel pull-pipeline, NAF-aware counting).
+- **The same issue is retried up to 3×** (#3241 attempt 2/3, #2575 attempt 3/3) **with the same model and
+  no record of why the previous attempt produced nothing.** Attempt 2 begins exactly as ignorant as
+  attempt 1. This is the single clearest context failure in the system: the information needed
+  ("I tried X, the blocker was Y") existed and was discarded.
+- **Review deadlocks**: 3 rounds of `request_changes`, or a fixer twice declaring the findings spurious,
+  then a park. The disagreement itself — the most decision-relevant artifact produced — is preserved only
+  as prose scattered across comments.
+- **Defect classes recur across unrelated PRs**: vacuous tests, the workflow `if:`/step seam,
+  exit-zero-swallows-failure (8 instances), tests passing for the wrong reason, the *headline* guard being
+  the untested one (2 instances in one day). Each PR rediscovers these from scratch, and each is caught
+  only by an expensive adversarial reviewer.
+
+The last point is the economic argument: we are paying frontier-tier review tokens, repeatedly, to
+re-derive lessons the fleet has already learned.
+
+## 2. The core distinction: three channels, never conflated
+
+Most "shared agent memory" designs fail by putting everything in one growing store. The decisive move is
+to separate three kinds of information by **lifetime and derivability**, and give each a different
+mechanism, budget, and eviction rule.
+
+| Channel | Contains | Lifetime | Growth | Retrieval |
+|---|---|---|---|---|
+| **A. Lineage handoff** | what the previous stage of *this task* learned | one task | **bounded by schema** | pushed, always |
+| **B. Method memory** | transferable lessons about *this codebase* | long | **asymptotic** | pulled, on demand |
+| **C. Live state** | heads, labels, gate status, counts | seconds | **never stored** | queried, never cached |
+
+**Channel C is a rule, not a store: never memorise what can be queried.** A head SHA, a label set, a park
+count, a mergeability status — all change under you. This session produced two concrete harms from
+violating it: a `VERDICT: pass` bound to a superseded head nearly armed a PR whose fix had been retracted,
+and a park-cause finding recorded hours earlier ("all 33 carry `LOOPSTOP`") was false by the time it was
+reused — the current parks carried an entirely different cause. **Anything derivable in one API call must
+be derived, not remembered.** This single rule removes the largest source of both staleness and growth.
+
+## 3. Channel A — lineage handoff (the maintainer's example case)
+
+*"An agent working on a PR may benefit from context from the agent that created the issue."*
+
+### 3.1 Shape
+
+A single structured block, the **task ledger**, carried with the task and consumed by the next stage.
+Every stage **owns exactly one slot** and **replaces** it — never appends.
+
+```
+issue.created   → { intent, acceptance, known_constraints, decomposition_notes }
+worker.attempt  → { approach_tried, blocker, files_touched, why_no_diff?, confidence }
+review.round    → { findings[], severity, what_would_change_my_mind }
+fix.attempt     → { addressed[], disputed[] + reason, residual }
+adjudication    → { decision, rationale }
+```
+
+**Replace-not-append is the whole bounding story for this channel.** If round 3 carried rounds 1–2
+verbatim, context would grow linearly in rounds — precisely the failure mode the maintainer is worried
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent

**Fixed — this was red, not awaiting review.** Author-side note; no verdict here, a separate reviewer adjudicates.

**The real failing leg** was `docs-quality quick-gates`, not the `gate` aggregator (`gate` fail-fasts on the sibling). Runtime line from [job 89814418233](https://github.com/sparq-org/sparq/actions/runs/30209966646/job/89814418233):

```
Summary: 1 error(s)
research/agent-context-sharing.md:74 MD040/fenced-code-language Fenced code blocks should have a language specified [Context: "```"]
```

Single error across the whole 420-file lint. The block at :74 is the per-stage task-ledger slot sketch — a pseudo-schema, not runnable code — so it is tagged `text`.

**Result at `f4f41cd`: `gate` = pass**, `docs-quality quick-gates` = pass. Nothing was skipped, silenced, or given `continue-on-error`.

jeswr:
> 🤖 **SPARQ agent**

Independent review bound to head **`f4f41cd77133e3978eb34b977e77edf1da68d201`** (re-verified immediately before posting: `OPEN`, `MERGEABLE/CLEAN`). I authored none of this and pushed nothing.

## Gate state — verified by execution, paginated

`commits/f4f41cd7…/check-runs` returns **`total_count=105`**, so an unpaginated read would have been unsafe here. Paginated, exact-name (not prefix) match:

```
gate	completed	success	89851504510	2026-07-26T22:59:09Z
```

One `gate` row, no `draft-tier` variant, and **zero** non-`success`/non-`skipped` conclusions across all 105 rows. `assemble feature matrix` is `skipped`, not failed — the red-looking opt-in matrix is the ordinary no-Rust-changes path, not a leg failure.

## What I ran (not read)

| Check | Result |
|---|---|
| `markdownlint-cli2@0.18.1` | `0 error(s)`, 420 files — the head commit's MD040 fence fix holds |
| `check-no-perf-numbers.py --enforce` | `0 finding(s) across 265 scanned file(s)` |
| `check-terminology.py` | `clean — 6 banned term(s) over 3628 file(s)`; `trust envelope` 0 hits |
| `check-privacy-claims.sh` | `OK — scanned 737 file(s)` |

**The perf-number green is vacuous evidence for this file, and I proved it rather than assuming it.** I appended `sparq loads at 950 MB/s and is 12x faster than Oxigraph on BSBM.` to the record and re-ran:

```
--- default scan (repo-wide) ---
no-perf-numbers: 0 finding(s) across 265 scanned file(s).          EXIT=0     <-- mutant SURVIVED
--- explicit-path scan ---
research/agent-context-sharing.md:233: perf-number [throughput (bytes/s)]: '950 MB/s'
no-perf-numbers: 1 finding(s) across 1 scanned file(s).            EXIT=1
```

Cause is `check-no-perf-numbers.py:71` — `"research/"` is an **explicit, documented** exempt path (*"bench/ and research/ are the SANCTIONED homes for measured numbers"*). So the record's dated figures are **in their correct home and are not a hygiene violation** — but the gate did not establish that, the policy did. Restored from a `cp` snapshot (never `git checkout`): sha256 `b2111bebe8b5151d`, marker count 0, `git status --porcelain` empty.

Contains no ZK/MPC/soundness/privacy claim at all (sole regex hit is "a context idea that **sounds** good"), so `sq-qhy4` is not implicated.

## Claims re-grounded against the live tree

Verified present: `crates/sparq-kb`, `research/dogfooding-sparq-knowledge-graph.md`, `.claude/agents/sparq-pkg-nl.md`, `scripts/ready-issues.py`. Cross-refs check out: registry **#701** = *"64% of worker capacity produces no diff: no_change retries the same hard issue 3x with the same model"* — which **independently corroborates both the 36% yield and the 3× retry claim**; registry **#703** = reviewer-vs-fixer deadlock; sparq **#1111** = Neurosymbolic self-built KB. `#3241` is indeed the lock-free-saturation issue. "exit-zero (8 instances)" is exact — the corpus enumerates through an *"8th instance, 2026-07-26, registry PR #702"*.

§4.1's derivability example is accurate and genuinely derivable — `scripts/ready-issues.py:134`:

```python
if "status:ready" not in labels:
    return "no status:ready attestation"
```

## Three accuracy defects (none blocking; two err *against* the author's own case)

**1. §3.3 cites a marker literal that does not exist.** `<!-- sparq-fix-nochange:v1 round=1 run=… -->` has **0 org-wide code-search hits and 0 hits in the tree**. The real markers on the very issue the record cites are `<!-- sparq-worker-attempt:v1 run=30199599239.1 -->` and `<!-- sparq-task-decline-escalation:v1 key=… action=… -->`; also `sparq-park-reason:v1` and `sparq-omnibus:v1`. `nochange` is a *park class* (`park_policy.py:786`), not a marker, and `round=` appears on neither. This trips the record's own §0.3 self-binding rule. **The load-bearing claim survives and is understated** — `sparq-worker-attempt:v1` is precisely the `worker.attempt` slot §3.1 proposes, so "the substrate already exists" is *more* true than written. Fix the literal.

**2. §1's "#3241 attempt 2/3" was already stale when written** — attempt `3/3` was posted `2026-07-26T14:39:46Z`, **1h35m before this PR opened at 16:14:23Z**; both #3241 and #2575 show 1/3, 2/3 *and* 3/3. This is a live count memorised in prose, i.e. the exact Channel-C violation §2 forbids. It understates rather than overstates, so the "retried up to 3×" headline stands.

**3. "an existing corpus of ~90" (§4.2)** — actual count is 129 memory files, 12 touched on 2026-07-26, so the pre-existing corpus was ~117. A larger denominator with the same ~7 new makes the flattening claim *stronger*, and it is hedged "weak but real".

## Falsifiability audit — the maintainer's actual constraint

This is the part I was asked to be hardest on, and it holds. The record separates **enforcement** from **detection** correctly:

- **Bounded by construction, not by hope:** Channel A per-slot byte caps enforced at write time with over-cap writes *rejected, not truncated* (§3.2) plus a total cap with recorded drops; Channel B a hard per-agent retrieval budget that *tells the agent its view is partial* (§4.3); Channel C never stored. Because the maintainer's constraint is about **context given across agents**, the retrieval budget — not corpus size — is what actually binds it, and that bound holds regardless of the corpus.
- **Detection:** corpus-size-vs-cumulative-tasks declared falsifiable *in advance*, with the response pre-committed ("tighten admission — do not add storage"), §6 stop conditions "stated in advance … neither should be argued away after the fact", and `tokens per merged PR` as a delivered-context proxy.

Three weaknesses worth a follow-up, none fatal to a design record:

- The detector has **no threshold, window, cadence or owner**; "linear growth means admission is too loose" is not yet a rule two observers would agree on. Belongs in the step-4 A/B design.
- **Sub-linear-but-unbounded passes the stated test** (O(√n)/O(log n) are not linear yet still grow without limit). Survivable only because the retrieval budget is the real bound — but the record never says "the budget holds even if the corpus metric fails", so a reader could mistake the plot for the bound. One sentence fixes it.
- §3.1's *"replace-not-append is the whole bounding story for this channel"* is an overstatement. "A distilled residue of what remains unresolved" is the append-in-disguise pathway — residue-of-residue can accrete bytes while the slot *count* stays fixed. §3.2's byte cap is what actually stops that, so the design is sound; the sentence just misattributes which mechanism does the work.

## Grounding in prior measurements

The concern that a context proposal might ignore two of three measured rejections does **not** land here. All three are §0.1 binding rules, and the outline-first rejection is *structurally load-bearing*: §4.3 cites it as the reason to choose pull-not-push over bulk injection, and §7 repeats it; `pkg-query` (the adopted one) is the retrieval substrate. One unflagged tension: §3.2's "structured fields, not prose" sits adjacent to the `sparq-terse` hypothesis that measured ~0% — though the record's claim is about *bounding* (a fixed schema cannot accrete) rather than compression ratio, so it is not contradicted. Worth a hedge.

## Conclusion

Design record, no behaviour change, all gates green at head, no honesty/terminology/perf-hygiene violation, correctly grounded in all three prior measurements, and genuinely falsifiable with pre-committed stop conditions. The three defects are accuracy nits in illustrative material; none changes a design decision and two understate the author's own case. Recommend fixing the §3.3 marker literal to `sparq-worker-attempt:v1` — that correction strengthens the record — but it does not gate merge.

Not arming, not merging, not relabelling.

VERDICT: pass

PR #4192, opened 2026-07-26
fix(release): fail-closed crates.io publish protections before `publish = true` (#1135)

jeswr:
> 🤖 SPARQ agent

Closes the crates.io publish-rate question you raised in #1135:

> *"Before I do this; can I make sure that there are protections in place to prevent publishing too regularly, I don't want to spam the registry."*

**This PR does NOT flip `publish = true`.** That stays your call. It makes the flip safe to
make, and adds a tripwire test that reds if `publish` is ever enabled without the checklist
at the bottom being complete.

**A crates.io version can never be unpublished.** Every ambiguity below refuses.

---

## Round 2 — the three blocking findings from the review of `3e9f54da5`

| finding | fix | named test that reds on its revert |
|---|---|---|
| 1. four YAML-seam mutants survived the guard step (`\|\| true`, `exit 0;`, `continue-on-error: 'true'`, `continue-on-error: ${{ true }}`) | pin the `run:` by **equality**; assert `continue-on-error` is **absent**; reject shell metacharacters independently | `test_guard_step_enforces`, `test_guard_step_cannot_swallow_its_exit_status`, `test_guard_step_is_not_continue_on_error` |
| 2. "crates.io ignored" died in `decide()` but survived at `run()`'s call site | a call-site test over a repo with **no `v*` tag** — the hand-bootstrap window, where crates.io is the only release signal | `test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged` (+ its permitting counterpart, + the no-op short-circuit pin) |
| 3a. `.claude/settings.json` emitted `allow` when the guard script could not run — inverting its fail-closed disposition | the wrapper now **denies** `gh pr merge` when the guard cannot run, and still allows every other Bash call | `TestSettingsJsonWrapperDoesNotInvertTheGuard::test_an_unrunnable_guard_DENIES_an_arm`, `…::test_an_unrunnable_guard_still_ALLOWS_a_non_merge_command`, plus the 6 unrunnable-guard rows in `check-pr-arm-base.py --self-test` (2 projects x {arm, direct merge, ordinary command}), which replace the 2 rows that used to assert `allow` |
| 3b. quantifier direction — "full enumeration" / "32/32 killed" claimed more than was verified | the 32/32 line is **withdrawn**; the hook's real scope is tabulated (reproduced independently at this head); every remaining claim is per-row | — (wording; see **Mutation runs**) |

All ten mutants covering those findings are killed by a named test — harness output pasted
under **Mutation runs**, nothing quantified beyond what that run executed.

I confirmed 3a myself before touching it: with the guard missing, crashing, or
`CLAUDE_PROJECT_DIR` unset, the wrapper returned `allow` for `gh pr merge 900 --auto --squash`
in all three cases.

The same scope correction is applied to the two places that would otherwise keep the
over-read alive after this PR merges: `docs/release.md` §8d.1 ("read *armed* literally",
with the phrasing table) and `AGENTS.md`'s arming-model entry (the hook's failure
disposition is per-AXIS — stacked-base open, release-PR closed).

Scope discipline: the version_group question (26 publishable vs 17 grouped) is **not** touched
— those nine crates each need an irreversible ship/don't-ship decision from you, and the guard
warns today and hard-refuses on the flip. It stays step 1 of your checklist.

---

## The gap that was actually there

`scripts/batch-merge.py` and `scripts/pr-backlog.py` already excluded the release-plz
Release PR. Two arming paths did not:

| path | before | consequence |
|---|---|---|
| `scripts/auto-arm.py` | **no release-plz exclusion of any kind** | armed any `review:pass` PR with base `main` — including the Release PR |
| `scripts/rearm-sweeper.py` | excluded by **label** only (`EXCLUDED_LABELS` + `needs:*`) | a Release PR with `review:pass` and no hold label was re-armed |

Verified, not assumed. Deleting the new guard from `auto-arm.py` and running its self-test
prints the mutant issuing the live mutation against the Release PR fixture:

```
['api', 'graphql', '-f', 'query=mutation(...){ enablePullRequestAutoMerge(...) }',
 '-f', 'id=PR_kwDO900', '-f', 'oid=aaaa…']
```

and the same for `rearm-sweeper.py`: `['pr', 'merge', '900', '--repo', 'sparq-org/sparq', '--auto']`.

Harmless while `publish = false`. Live and irreversible the moment the token is wired.

### A third path nobody had enumerated

`scripts/check-pr-arm-base.py` — the PreToolUse hook in `.claude/settings.json`. This is
where **agent-typed** `gh pr merge … --auto` goes (`.claude/workflows/autonomous-scheduler.js`,
`fable-architect-drain.js`, `fable-soundness-verdict.js`, and anything hand-written). It
bypasses both sweeps entirely. Its one existing rule is "deny if base ≠ `main`" — and the
Release PR's base **is** `main`, so it allowed it.

### The arming/merging paths I enumerated, and their state after this PR

| path | trigger | before | after |
|---|---|---|---|
| `scripts/auto-arm.py` | `auto-arm.yml` (label event + 10-min cron) | ❌ none | ✅ branch/author/title, fail-closed |
| `scripts/rearm-sweeper.py` | `rearm-sweeper.yml` (10-min cron) | ⚠️ label-keyed only | ✅ branch/author/title, on snapshot **and** live state |
| `scripts/check-pr-arm-base.py` | PreToolUse hook — agent-typed `gh pr merge … --auto` | ❌ none | ✅ denies, fail-closed — **scope below** |
| `scripts/batch-merge.py` | `batch-merge.yml` | ⚠️ author **AND** title | ✅ shared predicate (OR, + branch) |
| `scripts/pr-backlog.py` | `pr-backlog.yml` | ⚠️ author **AND** title | ✅ shared predicate |
| `.github/workflows/merge-queue-feedback.yml` | dequeue feedback | uses `--disable-auto` (disarms only) | unchanged — cannot arm |

#### What the hook row does and does **not** cover — executed, not reasoned

Review ran the real hook against a fake `gh` returning the live Release-PR shape, and I
reproduced every row of it at this head before writing it down. It is the `--auto`
*phrasings* that are covered, not every way to merge a PR:

| command | hook |
|---|---|
| `gh pr merge 900 --auto --squash` | **blocked** |
| `gh   pr   merge 900 --auto` (multi-space) | **blocked** |
| `bash -c "gh pr merge 900 --auto"` | **blocked** |
| `echo 900 \| xargs gh pr merge --auto` | **blocked** |
| `gh pr merge release-plz-main --auto` (branch, not number) | **blocked** |
| `gh pr merge 900 --squash` (direct merge, **no** `--auto`) | allowed |
| `gh pr merge 900 --squash --admin` | allowed |
| `gh api graphql … enablePullRequestAutoMerge(…)` | allowed |
| `gh api -X PUT repos/sparq-org/sparq/pulls/900/merge` | allowed |
| `gh pr` + backslash-newline + `merge 900 --auto` (line continuation) | allowed |
| `M="pr merge"; gh $M 900 --auto` (variable indirection) | allowed |

`is_arm_command` requires `gh pr merge` **and** `--auto`. The no-`--auto` direct merge — the
*more* dangerous operation — is out of this hook's scope by design, and its self-test pins
that. The GraphQL shape is the one `auto-arm.py:326` itself uses, so it is already in-repo
for any agent to copy.

So the claim that holds is **not** "no agent can merge the Release PR". It is:

> no agent path can **arm** it, and no path of any kind — agent, human, or future
> automation — can cut a release inside `MIN_RELEASE_INTERVAL`,

because the interval guard runs inside `release-plz.yml` itself, the sole crates.io path,
and does not care how the merge happened. **The belt, not this hook, is what makes the
crates.io claim true.** This scope is now written into `check-pr-arm-base.py`'s header so
it cannot be over-read again.

The two `AND` predicates were weak: `author == github-actions AND title startswith "chore: release"`
returns False if *either* signal shifts (a retitled PR, a different bot identity), even
when the head branch is unmistakable. Both now use the shared predicate.

---

### 1. `scripts/release_pr_guard.py` — one predicate, keyed on branch/author/title

**Never on a label.** Anything holding `pull-requests: write` can add or remove a label, so
a label-keyed exclusion is defeated by relabelling. `arm_block_reason` takes no `labels`
parameter at all — that is asserted structurally, so it survives a rewrite of the body.

Fail-closed: an absent/empty head branch blocks; a bot-authored PR with an unknown title
blocks. Over-blocking costs a skipped sweep the cron repeats. Under-blocking publishes 17
crates forever.

A missing module does **not** degrade to "proceed" — every consumer substitutes a stub that
refuses everything and logs `::error`.

### 2. `scripts/release-interval-guard.py` — the belt to that braces

Runs in `release-plz.yml`'s `release-plz-release` job **before** the tag/publish step, so
even a mistaken merge cannot publish more often than the cadence.

**`MIN_RELEASE_INTERVAL = 24h`**, one named constant. Why 24h:

- one release publishes **every crate in the version_group in lockstep** — two releases in a
  day is 34 irreversible crates.io versions;
- cadence tracks Release-PR *merges*, and nothing except this guard bounds how often that
  PR can be merged;
- a day is the shortest interval that still leaves room to *notice* a bad release before the
  next one. A version published 20 minutes after its predecessor is a mistake being repeated;
- it is not restrictive in practice — the project has never cut two releases in a day, and a
  genuine emergency patch is still available by hand, which is the friction an irreversible
  operation should have.

Longer (a week) would block a legitimate same-week fix and push people toward disabling the
guard, which is worse than a defensible floor.

`last_release = max(newest v* tag date, newest crates.io publication)` — taking the max means
neither source can be used to argue for a shorter wait.

**Fail-closed, exhaustively.** Each of these REFUSES:

| unknown | why it must not mean "go ahead" |
|---|---|
| shallow checkout | a shallow clone reports an **empty tag list**, indistinguishable from "never released" — the most dangerous false negative here |
| tag list unreadable / `git` fails | the interval cannot be computed |
| unparseable tag or publish date | ditto |
| crates.io unreachable / non-200 / unparseable body | the second authoritative source is silent |
| last release **in the future** (clock skew) | the interval is meaningless |
| workspace manifest unreadable | what would be published is unknown |
| a publishable crate outside the `version_group` | what would ship ≠ what the config describes |

A crates.io **404 is definitive** ("never published") and is the only accepted empty answer —
distinguished from a network error, so a genuine first release can still ship.

There is deliberately **no override flag**. Publishing inside the window is done by hand.

**Quiet on ordinary pushes.** If `v<workspace version>` is already tagged, `release-plz release`
is a no-op and the guard passes without touching crates.io. `v0.1.0` is tagged and the
workspace version is `0.1.0`, so today it is silent.

### 3. `--dry-run` — prints what would be published, touches nothing

```
$ python3 scripts/release-interval-guard.py --dry-run
[release-interval-guard] mode: DRY-RUN (reporting only, nothing is mutated)
[release-interval-guard] 26 publishable crate(s); publish order (deps first):
    1. sparq-core        0.1.0
    2. sparq-fedplan     0.1.0
    …
   26. sparq-mcp         0.1.0
```

Crate list, versions, dependency-first publish order, and the cadence verdict it *would*
return — exit 0 always. Inertness is a tested property, not a claim: the suite snapshots
HEAD, the tag list, `git status --porcelain` and every file's bytes before and after, and an
AST assertion proves the module can only ever spawn `git`, never `cargo`.

---

## A live inconsistency the guard found

`--dry-run` on `origin/main` reports **26** cargo-publishable crates (no `publish = false`)
against **17** in `release-plz.toml`'s `version_group`. The nine outside it:

`sparq-arrow`, `sparq-fedplan`, `sparq-forms`, `sparq-mcp`, `sparq-reason-el`,
`sparq-reason-ql`, `sparq-shaclc`, `sparq-substrate`, `sparq-wrapper`

release-plz would version those **independently** of the locked workspace version — the exact
drift the version_group exists to prevent — and publish them anyway.

I have deliberately **not** resolved this: each crate needs a decision about whether it ships
publicly at all, and that is not mine to make irreversibly. The guard warns loudly while
`publish = false` and becomes a **hard refusal** the moment it flips — so the flip cannot
happen with this unresolved. It is step 1 of your checklist.

(`docs/release.md`'s "17 crates" describes the version_group, not cargo's publishable set;
now noted in the doc.)

---

## Guard → red test

Every row verified by applying the mutation and running the command. Full output below.

| guard | mutation | goes red |
|---|---|---|
| Release PR unarmable — `auto-arm` | delete the call | `auto-arm.py --self-test`, `test_release_publish_guard.py` |
| Release PR unarmable — `rearm-sweeper` | delete the call | `rearm-sweeper.py --self-test`, `test_release_publish_guard.py` |
| Release PR unarmable — agent hook | delete the release axis | `check-pr-arm-base.py --self-test`, `test_release_publish_guard.py` |
| **relabelling cannot unblock** | disable branch signal | 4 suites incl. all three arm paths |
| exclusion not label-keyed | (structural: no `labels` param) | `test_the_predicate_cannot_be_influenced_by_labels` |
| `batch-merge` / `pr-backlog` | revert to the AND predicate | `test_batch_merge_and_pr_backlog_catch_a_branch_only_release_pr` |
| query carries the guard's inputs | drop `headRefName` | `auto-arm.py` / `rearm-sweeper.py --self-test` |
| publish inside the interval refused | interval never fires | `release-interval-guard.py --self-test`, `test_publish_inside_the_interval_is_refused` |
| the 24h constant | widen to 1 minute | `test_named_constant_is_24h` |
| crates.io can only tighten (`decide()`) | ignore crates.io | `test_crates_io_can_only_TIGHTEN_the_verdict` |
| **`run()` really fetches crates.io** *(new)* | `run()` never fetches / discards the result | `test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged` |
| …and does not just refuse always *(new)* | — (discriminating pair) | `test_run_permits_when_crates_io_reports_an_OLD_publication` |
| the no-op-push short-circuit stays deliberate *(new)* | make the fetch unconditional | `test_run_does_not_fetch_crates_io_on_an_already_tagged_no_op_push` |
| unreadable last-release refuses | drop the shallow check | `test_shallow_checkout_refuses` |
| unreachable crates.io refuses | swallow the error | `test_unreachable_crates_io_refuses` |
| future timestamp refuses | allow it | `test_future_dated_last_release_refuses` |
| publish flag read strictly | missing key ⇒ disabled | `test_publish_flag_is_read_fail_closed` |
| version_group coverage | never refuse | `test_version_group_drift_refuses_when_publish_is_enabled` |
| `--dry-run` mutates nothing | make it `git tag` | `test_dry_run_leaves_the_repository_byte_identical` |
| **YAML: guard step present** | delete the step | `test_guard_step_exists` |
| **YAML: unconditional** | `if: false` | `test_guard_step_is_unconditional` |
| **YAML: not swallowed** | `continue-on-error: true` / `'true'` / `${{ true }}` | `test_guard_step_is_not_continue_on_error` |
| **YAML: exit status not discarded** *(new)* | `… --enforce --repo-root . \|\| true`, `exit 0; …` | `test_guard_step_cannot_swallow_its_exit_status` **and** `test_guard_step_enforces` |
| **YAML: ordering** | move after `release-plz/action` | `test_guard_runs_before_release_plz` |
| **YAML: call site** | `--enforce` → `--dry-run` | `test_guard_step_enforces` |
| **YAML: precondition** | drop `fetch-depth: 0` | `test_release_job_checks_out_full_history` |
| **YAML: the suite still runs** | delete the docs-quality step | `test_docs_quality_gates_the_publishing_protections` |
| missing-guard degradation | make the stub permissive | `TestMissingReleaseGuardArmsNothing` |
| sparse manifest carries the guard | revert `pr-backlog.yml` | `test_no_workflow_enumerates_a_python_file_without_its_sibling_imports` |
| **hook wrapper does not invert the script** *(new)* | `.claude/settings.json` back to blanket `allow` | `TestSettingsJsonWrapperDoesNotInvertTheGuard::test_an_unrunnable_guard_DENIES_an_arm` + `check-pr-arm-base.py --self-test` |
| …and does not brick every Bash call *(new)* | — (discriminating pair) | `test_an_unrunnable_guard_still_ALLOWS_a_non_merge_command` |

Each row names the suite that reds on that mutation. The rows marked *(new)* were executed on this head and the harness output is pasted below; the older rows were executed in an earlier run — see the honest scoping under **Mutation runs**.

### The YAML seam, structurally

`release-plz.yml` runs only on `push: main`, so it never registers on a PR's `gate` — a defect
there surfaces for the first time at release time. The suite therefore `yaml.safe_load`s it and
indexes `jobs.release-plz-release.steps[i]`. A `"--enforce" in text` assertion does not catch
`if: false`; `steps[i]["if"]` does. Both `if: false` and a realistic
`if: ${{ github.event_name == 'never' }}` are in the matrix and both die.

**Structural was not enough on its own — four mutants survived it.** Review of `3e9f54da5`
found that parsing the step and then asserting loosely about the parsed values leaves the
same hole one layer down:

| surviving mutant | why it survived |
|---|---|
| `run: … --enforce --repo-root . \|\| true` | `"--enforce" in run` is satisfied; the shell discards the refusal |
| `run: exit 0; … --enforce --repo-root .` | same, and the guard is never reached |
| `continue-on-error: 'true'` | loads as a **string**, so `assertNotEqual(…, True)` passes — Actions evaluates it as true |
| `continue-on-error: ${{ true }}` | same; and this is the *idiomatic* way to make it conditional, so it is a plausible future edit |

Closed by tightening the reading, not by adding a layer:

* `test_guard_step_enforces` now pins the whole `run:` string by **equality** against a named
  `EXPECTED_GUARD_RUN` constant — exactly one program is admitted;
* `test_guard_step_cannot_swallow_its_exit_status` independently rejects `\|\|`, `;`, `&` and
  `\|` in that `run:`, so relaxing the equality later does not silently reopen the hole;
* `test_guard_step_is_not_continue_on_error` now asserts the key is **absent**. Absence is the
  only value that is provably non-truthy at run time — a `${{ … }}` expression cannot be
  evaluated statically, so anything other than absence is refused.

### One mutant survived the first round — reported, then fixed

`test_each_arming_path_imports_and_calls_the_guard` originally asserted
`"release_pr_guard.arm_block_reason" in text`. Reverting `batch-merge.py` to the old AND
predicate **SURVIVED**: the phrase was still present — in the function's *docstring*. A prose
mention is not a call site. Now checked by AST (`ast.Call` on the attribute) plus a behavioural
test that both scripts catch a Release PR identifiable by its head branch alone.

### Mutation runs — what was executed, and what is NOT claimed

**The earlier `32/32 mutants killed; 0 SURVIVED` line in this body is WITHDRAWN.** Review of
`3e9f54da5` executed four further mutants on the same load-bearing step and all four survived,
so that quantifier was false in the direction that matters. Nothing below is a count over
"every mutant one could think of" — it is exactly the mutants named, from the run whose
output is pasted.

#### Run 1 — the ten mutants covering the three review findings (this head)

Harness: copy the worktree to a throwaway tree, apply one mutation, run all four suites,
record which **named** test goes red. Verbatim output:

```
BASELINE:
  rc=0  test_release_publish_guard
  rc=0  check-pr-arm-base --self-test
  rc=0  release-interval-guard --self-test
  rc=0  test_arm_capability_wiring

MUTANTS:
  KILLED    F1a  YAML run: trailing `|| true`
              red: test_release_publish_guard -> test_guard_step_cannot_swallow_its_exit_status, test_guard_step_enforces
  KILLED    F1b  YAML run: leading `exit 0;`
              red: test_release_publish_guard -> test_guard_step_cannot_swallow_its_exit_status, test_guard_step_enforces
  KILLED    F1c  YAML continue-on-error: ${{ true }}
              red: test_release_publish_guard -> test_guard_step_is_not_continue_on_error
  KILLED    F1d  YAML continue-on-error: 'true'
              red: test_release_publish_guard -> test_guard_step_is_not_continue_on_error
  KILLED    F1e  YAML continue-on-error: true (bool)
              red: test_release_publish_guard -> test_guard_step_is_not_continue_on_error
  KILLED    F1f  YAML run: --enforce -> --dry-run
              red: test_release_publish_guard -> test_guard_step_enforces
  KILLED    F2a  run(): never fetch crates.io
              red: test_release_publish_guard -> test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged, test_run_permits_when_crates_io_reports_an_OLD_publication
  KILLED    F2b  run(): fetch, then discard the result
              red: test_release_publish_guard -> test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged
  KILLED    F2c  decide(): drop crates_io_at from the candidates (the ORIGINAL mutant)
              red: test_release_publish_guard -> test_crates_io_can_only_TIGHTEN_the_verdict, test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged; release-interval-guard --self-test -> crates.io wins when it is NEWER than the newest tag (max, not first source)
  KILLED    F3a  settings.json wrapper: revert to blanket allow
              red: test_release_publish_guard -> test_an_unrunnable_guard_DENIES_an_arm; check-pr-arm-base --self-test -> missing guard script + arm, off-root cwd: allow (want deny + exit 0), missing guard script + direct merge, off-root cwd: allow (want deny + exit 0), unreadable guard script + arm, off-root cwd: allow (want deny + exit 0), unreadable guard script + direct merge, off-root cwd: allow (want deny + exit 0)

TOTAL: 10/10 killed; 0 SURVIVED
```

#### What run 1 does NOT cover — stated plainly

* It is scoped to the three review findings, plus three adjacent regressions kept in the
  matrix (`continue-on-error: true`, `--enforce` → `--dry-run`, and the original `decide()`
  crates.io mutant). It is **not** a re-run of the whole earlier matrix.
* The predicate / call-site / degradation / dry-run / sparse-manifest rows in the table above
  were killed in the earlier run, and independently in the reviewer's own 28-mutant matrix —
  they report all 10 predicate/call-site mutants and all 7 primary YAML mutants killed. I have
  **not** re-executed those on this head. Each of those rows names its suite, and every one of
  those suites is green here.
* `release-plz.yml` still never runs on `pull_request`, so this PR's own CI cannot exercise
  the guard step live. It is asserted structurally from a gating job, and the guard's logic is
  run end-to-end against real temporary git repositories.
* Nothing here is a claim about mutants nobody has written. The honest reading of every table
  in this body is *"this named mutation reds this named test"*, one row at a time.

## Three deliberate behaviour changes — please read

**1. `check-pr-arm-base.py` now fails CLOSED on the release axis.** Previously a failed
`gh pr view` (or an unparseable PR argument) ALLOWED the arm. It now denies, because an
unresolvable PR cannot be proven not to be the Release PR. The cost is near zero: a `gh` that
cannot answer `pr view` almost certainly cannot `pr merge` either, so the deny forfeits an arm
that would have failed anyway. The **stacked-base** axis keeps its documented fail-open
disposition — only the release axis changed.

**2. `pr-backlog.yml` now checks out `scripts/` rather than a one-file list.** Its old manifest
enumerated only `scripts/pr-backlog.py`, so the sibling import added here would never have been
checked out. This was caught by the existing `test_arm_capability_wiring.py::TestSparseCheckoutManifest`,
which is exactly what it is for — a real bug in my change, found by your own gate.

**3. `.claude/settings.json`'s hook wrapper now fails CLOSED on `gh pr merge`.** Reported by
review; I verified it before changing anything. The wrapper was

```sh
if [ -n "$CLAUDE_PROJECT_DIR" ] && [ -r "$guard" ] && guard_output=$(python3 "$guard"); then …
else <permissionDecision: allow>
```

so whenever the script could not **run** — missing, unreadable, crashing, or
`CLAUDE_PROJECT_DIR` unset — the wrapper answered `allow`. Executed on `3e9f54da5`, all three
of those returned `allow` for `gh pr merge 900 --auto --squash`. That inverted this script's
deliberately fail-closed release axis at the exact moment there is no guard at all.

It now denies any `gh pr merge` when the guard cannot run, and **still allows everything
else** — this hook matches *every* Bash tool call, so a blanket deny-on-failure would
re-create the live all-Bash deadlock that #3605's fail-open was added to prevent. Both halves
are load-bearing and both are pinned: `TestSettingsJsonWrapperDoesNotInvertTheGuard` executes
the wrapper string straight out of `settings.json` as a subprocess, and
`check-pr-arm-base.py --self-test`'s unrunnable-guard block now carries an arm row, a direct-
merge row (deny) and an ordinary-command row (allow). Reverting the wrapper reds both.

Note the scope: this closes the *inversion*. It does not widen `is_arm_command` — the
phrasings table above stands, and the interval guard is what bounds the outcome.

That suite also surfaced a genuine contract collision worth stating plainly: a missing
`gh_retry.py` costs RETRIES and must never cost the arm (#3776); a missing `release_pr_guard.py`
costs the ARM ITSELF and must (#1135). `run_without_gh_retry` now isolates exactly one variable,
and `TestMissingReleaseGuardArmsNothing` pins the opposite contract with the **same driver**, so
the difference is attributable to the guard and nothing else.

---

## Residual gaps — stated, not hidden

- The Release PR can still be merged by **a human clicking merge**, and should be — that is the
  intended path. The interval guard is what bounds cadence in that case.
- The interval guard needs network for crates.io. If crates.io is unreachable it **refuses**;
  that is intended, but it means a crates.io outage blocks releases. Correct trade for an
  irreversible operation.
- `release-plz.yml` does not run on `pull_request`, so this PR's own CI cannot exercise the
  guard step live. That is why the workflow is asserted **structurally** from a gating job, and
  why the guard's logic is run end-to-end against real temp git repositories in the suite.
- The PreToolUse hook covers the **`--auto` phrasings only**. A direct `gh pr merge <n>
  --squash`, `--admin`, `gh api graphql … enablePullRequestAutoMerge`, a REST `PUT
  …/pulls/<n>/merge`, backslash continuation and shell-variable indirection all reach `gh`
  unblocked (full table above). This is a deliberate scope, not an oversight — but it means
  "no agent can arm the Release PR" is the true statement, and "no agent can merge it" is not.
  What bounds the *outcome* in every one of those cases is the interval guard in
  `release-plz.yml`.
- The `.claude/settings.json` fail-open described above is fixed here for `gh pr merge`; if a
  future arming path is added that this hook does not recognise, the wrapper will not catch it
  either. The belt is the durable answer.
- The version_group question (26 publishable vs 17 grouped) is deliberately unresolved — see
  step 1 of the checklist. The guard warns today and hard-refuses on the flip.

---

## What is left for you, in order, before `publish = true`

1. **Resolve the version_group drift (blocking).** For each of `sparq-arrow`, `sparq-fedplan`,
   `sparq-forms`, `sparq-mcp`, `sparq-reason-el`, `sparq-reason-ql`, `sparq-shaclc`,
   `sparq-substrate`, `sparq-wrapper`: either add a `[[package]]` entry with
   `version_group = "sparq"` in `release-plz.toml` (it ships), or set `publish = false` in its
   `Cargo.toml` (it does not). Confirm with `python3 scripts/release-interval-guard.py --dry-run`
   until it reports full coverage.
2. **Re-check name availability** for whatever the resolved set turns out to be — `docs/release.md`
   §0a's snapshot is from 2026-06-14 and covers 17 names, and `sparq-algos` was never re-checked.
3. **Bootstrap-publish each crate once, by hand, leaf-first** (`docs/release.md` §4 DAG).
   crates.io Trusted Publishing can only be configured on a crate that already exists.
4. **Configure Trusted Publishing per crate**: crates.io → crate → Settings → Trusted Publishing
   → Add → GitHub; owner `sparq-org`, repo `sparq`, workflow `release-plz.yml`, environment blank.
   (Note `docs/release.md` §8c still says owner `jeswr` — worth confirming which is right.)
   **Do not create a long-lived `CARGO_REGISTRY_TOKEN`** — OIDC replaces it.
5. **Flip, in one commit:** uncomment `id-token: write`, the `crates-io-auth-action` step and the
   `CARGO_REGISTRY_TOKEN` env in `release-plz.yml`; set `publish = true` in `release-plz.toml`;
   and update `TestLivePublishFlagIsStillFalse` in the same commit (it is a tripwire that reds
   if `publish` is enabled without a deliberate, reviewed diff — that is its job).
6. **Confirm the guard is live** on the first main push after the flip: the
   `Publish-cadence guard` step should log
   `release-plz.toml crates.io publish: ENABLED` and `version_group covers all N publishable crate(s)`.

Steps 1–2 can happen now. Steps 3–6 are the flip itself.

---

Reviewed nothing about ZK/MPC/crypto surfaces; this change is confined to release/CI orchestration.


jeswr:
> 🤖 SPARQ agent

**Review of PR #4192 at head `3e9f54da5be4e34e7a3051d9004b44a7a1b58639`** (full SHA resolved and bound; everything below was executed against a detached worktree at that commit, restored byte-clean afterwards).

`gate`: **success**, confirmed on the exact head via `commits/3e9f54da5.../check-runs?per_page=100 --paginate`, newest-run-per-name — not the PR rollup. 55 check-run names; `gate` = `completed/success` (run 89817340081).

---

## The core mechanism is sound — I could not break it

I want to lead with this, because it is the maintainer's actual question and it holds up under everything I threw at it.

**The interval guard genuinely fails closed. 25/25.** I built real temp git repos and ran `release-interval-guard.py --enforce` as a real subprocess, plus a real local HTTP server exercising the real `_http_get_json`:

| input | result |
|---|---|
| shallow checkout (`.git/shallow`) | REFUSED |
| not a git repo at all | REFUSED |
| `git` absent from `PATH` | REFUSED |
| last release 3h ago | REFUSED |
| **future-dated tag (clock anomaly)** | REFUSED |
| `Cargo.toml` missing / empty `members` / no `[workspace.package].version` | REFUSED (3) |
| `release-plz.toml` missing / no `[[package]]` entries | REFUSED (2) |
| `publish = true` + a crate outside the version_group | REFUSED |
| crates.io unreachable (dead proxy → `URLError`) | REFUSED |
| crates.io HTTP 500 / 403 / redirect loop | REFUSED (3) |
| 200 non-JSON / JSON array / no `versions` / bad `created_at` / **naive tz-less `created_at`** | REFUSED (5) |
| 200 truncated body (`IncompleteRead`) | REFUSED (uncaught exception → still exit 1) |
| crates.io 2h ago vs tag 1y ago | REFUSED — `max()` genuinely tightens |
| definitive 404 | permitted (correct: never published) |
| **last release 40d ago (control)** | **PERMITTED** — the refusals discriminate; this is not a brick |

**And the belt cannot be bypassed.** I checked every `cargo publish` / `CARGO_REGISTRY_TOKEN` / `crates-io-auth` site in the repo: `publish.yml`'s `crates` job only does `cargo package` + attestation (its publishes are npm and PyPI). `release-plz.yml` is the **sole** crates.io path, and the guard step sits in it unconditionally, before `release-plz/action`. That is what makes the design's central claim — "even a mistaken merge, *by a path nobody enumerated*, cannot cut a release more often than `MIN_RELEASE_INTERVAL`" — actually true.

**Claim 5 verified exactly.** `--dry-run` on this head reports **26** publishable crates against **17** in the `version_group` — the same nine named in the body. And the refusal genuinely blocks the flip: setting `publish = true` in the worktree makes `--enforce` **exit 1** with `REFUSING while publish is enabled.` That mechanism works.

**The surviving-mutant fix is real.** `test_each_arming_path_imports_and_calls_the_guard` matches `ast.Call` → `func.attr == "arm_block_reason"` → `func.value.id == "release_pr_guard"`. A docstring cannot satisfy that. Reverting `batch-merge.py` **and** `pr-backlog.py` to the author-AND-title conjunction both die now.

**24 of my 28 independent mutants were killed**, including all 7 primary YAML-seam mutants the body lists (delete step, `if: false`, realistic `if:`, `continue-on-error: true`, `--enforce`→`--dry-run`, drop `fetch-depth: 0`, reorder past `release-plz/action`) and all 10 predicate/call-site mutants. The YAML checks are genuinely structural (`step.get("run")`, `"if" in step`, index comparison) — not substring. Every killing suite runs on PRs (`docs-quality.yml:467-469`, `:492`, `:794`; `routing-self-tests.yml:99`).

---

## Blocking findings

### 1. Four YAML-seam mutants SURVIVE on the load-bearing guard step

Same class the body says was closed structurally. Applied to the guard step, full suite still green:

```
*** SURVIVED ***  run: python3 scripts/release-interval-guard.py --enforce --repo-root . || true
*** SURVIVED ***  run: exit 0; python3 scripts/release-interval-guard.py --enforce --repo-root .
*** SURVIVED ***  continue-on-error: ${{ true }}
*** SURVIVED ***  continue-on-error: 'true'
```

`test_guard_step_enforces` only asserts `--enforce in run` and `--dry-run not in run` — a trailing `|| true` satisfies both and swallows the refusal. `test_guard_step_is_not_continue_on_error` uses `assertNotEqual(step.get("continue-on-error"), True)`; `'true'` and `'${{ true }}'` are strings, so `!= True` passes the test while **GitHub Actions evaluates both as true**. `continue-on-error: ${{ … }}` is the idiomatic way someone makes it conditional, so this is a plausible future edit, not a contrived one.

Fix: assert the `run` string *equals* the expected command (or has no `||`/`;`), and assert `str(step.get("continue-on-error")).lower() not in {"true", "${{ true }}"}` — i.e. truthiness, not identity.

### 2. A listed mutant does not actually die: the crates.io fetch call site is unpinned

The body lists `KILLED  fail-open: crates.io ignored (only the git tag consulted)`. That holds for the *pure function* but not for the *wiring*:

```
KILLED           F8b  decide(): drop crates_io_at from the candidates list
*** SURVIVED *** F8a  run(): crates_io_at = None            (never fetch)
*** SURVIVED *** F8a' run(): fetch, then discard the result
```

`test_crates_io_can_only_TIGHTEN_the_verdict` passes `crates_io_at` in as a parameter, so it can never notice the call site disappearing. The realistic edit — someone deletes the ~26 network requests to make the release job faster or less flaky — is undetected.

This is not cosmetic. crates.io is the **only** signal during the hand-bootstrap window: your own checklist step 3 publishes 26 crates by hand, leaf-first, cutting **no tags**. In that window `git_release_tags` returns nothing relevant and the cadence bound rests entirely on the source this mutant removes.

Fix: one test that drives `run()` with an injected `fetch` and asserts the fetch was actually called when `v<version>` is untagged.

### 3. Quantifier direction — the hook's coverage is narrower than the table says

The table heads "**Full enumeration** of arming/merging paths" and marks `check-pr-arm-base.py` "✅ denies, fail-closed". I ran the real hook against a fake `gh` returning the live Release-PR shape:

```
BLOCKED          gh pr merge 900 --auto --squash          (baseline)
BLOCKED          gh   pr   merge 900 --auto               (multi-space)
BLOCKED          bash -c "gh pr merge 900 --auto"
BLOCKED          echo 900 | xargs gh pr merge --auto
BLOCKED          gh pr merge release-plz-main --auto      (branch, not number)
*** ALLOWED ***  gh pr merge 900 --squash                 (DIRECT merge, no --auto)
*** ALLOWED ***  gh pr merge 900 --squash --admin
*** ALLOWED ***  gh api graphql … enablePullRequestAutoMerge(…)
*** ALLOWED ***  gh api -X PUT repos/sparq-org/sparq/pulls/900/merge
*** ALLOWED ***  gh pr \<newline> merge 900 --auto        (backslash continuation)
*** ALLOWED ***  M="pr merge"; gh $M 900 --auto           (variable indirection)
```

`is_arm_command` requires both `\bgh\s+pr\s+merge\b` **and** `--auto`. The GraphQL shape is the one `auto-arm.py:326` itself uses, so it is already in-repo for any agent to copy. And the no-`--auto` direct merge — the *more* dangerous operation — is waved through while the armed one is denied; the self-test codifies this as intended ("not this hook's business").

Separately, `.claude/settings.json`'s wrapper is fail-**open** at the shell level: `… && guard_output=$(python3 "$guard"); then … else {"permissionDecision":"allow"}`. I ran it — with the guard unreadable it emits **allow**. So the script's deliberate fail-closed release axis is inverted by its own wiring whenever the script cannot run.

None of this reaches crates.io (finding-1 aside, the belt still bounds cadence), so I am **not** calling it a publish hole. But "✅ fail-closed" under a heading that says "full enumeration", with Residual Gaps mentioning only "a human clicking merge", reads as *every* agent arming path is covered. It is *the `--auto` phrasings* that are covered. Please state the scope, and note the `settings.json` fail-open.

---

## Items 6 and 7

**6 — blast radius of deny-on-unresolvable: the trade is correct.** The deny widens to all PRs, so a transient `gh` failure costs a legitimate fleet arm. But the cost is bounded and self-healing: `rearm-sweeper` re-sweeps every 10 minutes, and `auto-arm` every 10. The author's argument holds — `gh pr view` and `gh pr merge --auto` hit the same GraphQL endpoint with the same credential. Worth noting the asymmetry, though: this hardens the case where `gh` fails while the script runs, and the case where the *script* fails still fail-opens through the wrapper (finding 3).

**Not a defect, for the record:** `batch-merge.py:443` re-arms open omnibus PRs without an `arm_block_reason` call, but its source list is filtered on `head_ref.startswith("sparq-omnibus/")`, which `release-plz-main` cannot satisfy. Safe — just not defence-in-depth.

---

## Verdict

The engineering here is well above the bar: the belt fails closed on every input I could construct, the sole publish path is correctly gated, the counts are exactly right, the version_group refusal really does block the flip, and the AST fix for the docstring mutant is the right fix. I would be glad to pass this.

I am failing it on the three findings above, and specifically because the body's quantified claims — "32/32 mutants killed; 0 SURVIVED", "Full enumeration", "No guard in this PR is unbacked" — are what the maintainer will read when deciding whether to create an irreversible-publish credential. Two of the three are the exact classes this repo has repeatedly been burned by (the YAML seam; a mutant passing for the wrong reason). The remedies are small: two tightened assertions, one call-site test, and an honest scope sentence in the table and Residual Gaps.

A false refusal here costs a delayed release. Please re-request review once those land — I expect to pass it.

VERDICT: fail

jeswr:
> 🤖 SPARQ agent

Thank you — all three findings were real, and two of them were defects I had claimed were
closed. Round 2 is pushed. **New head: `94ae0b0b62c48f41c8bbe934da1d7bbeda3861e4`** (was `3e9f54da5`).

I did not redesign anything: the interval guard's logic, the shared predicate and the belt's
position in `release-plz.yml` are byte-identical to what you reviewed. What changed is the
tests' *reading* of them, one genuine fail-open in the harness wiring, and the body's
quantifiers.

---

## Finding 1 — the four surviving YAML-seam mutants

You were exactly right about the mechanism: parsing the step structurally and then asserting
*loosely over the parsed value* leaves the same hole one layer down.

| your mutant | now dies on |
|---|---|
| `run: … --enforce --repo-root . \|\| true` | `test_guard_step_cannot_swallow_its_exit_status` **and** `test_guard_step_enforces` |
| `run: exit 0; … --enforce --repo-root .` | `test_guard_step_cannot_swallow_its_exit_status` **and** `test_guard_step_enforces` |
| `continue-on-error: ${{ true }}` | `test_guard_step_is_not_continue_on_error` |
| `continue-on-error: 'true'` | `test_guard_step_is_not_continue_on_error` |
| `continue-on-error: true` (kept as regression) | `test_guard_step_is_not_continue_on_error` |
| `--enforce` → `--dry-run` (kept as regression) | `test_guard_step_enforces` |

Two deliberate choices, slightly stronger than your suggested fix:

* **`run:` is pinned by EQUALITY**, against a named `EXPECTED_GUARD_RUN` constant, rather than
  by "has no `||`/`;`". Exactly one program is admitted. A legitimate future change to that
  command now costs one deliberate line in the same commit, which is the right price for a
  step whose failure mode is irreversible.
* **`continue-on-error` must be ABSENT**, not merely non-truthy. `${{ … }}` cannot be
  evaluated statically at all, so absence is the only value provably non-truthy at run time.
  This also rejects a pointless `continue-on-error: false`, which I consider a feature —
  there is no reason for that key to appear on this step.

`test_guard_step_cannot_swallow_its_exit_status` is deliberately independent of the equality
check, so relaxing one does not silently reopen the other.

## Finding 2 — the crates.io call site

Confirmed and fixed, and your framing of *why* it matters is what the new test is built
around. `test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged` drives `run()`
over a real temp repo with **no `v*` tag at all** — checklist step 3, the hand-bootstrap
window — and asserts both that `fetch` was called and that the verdict refuses:

```
KILLED    F2a  run(): never fetch crates.io
            red: test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged,
                 test_run_permits_when_crates_io_reports_an_OLD_publication
KILLED    F2b  run(): fetch, then discard the result
            red: test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged
KILLED    F2c  decide(): drop crates_io_at from the candidates  (your F8b, kept)
            red: test_crates_io_can_only_TIGHTEN_the_verdict,
                 test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged;
                 release-interval-guard --self-test
```

With no tag, discarding the fetch falls through to the *"FIRST release, no cadence can have
been violated"* branch and **allows** — which is precisely the unbounded window you described.
Paired with `test_run_permits_when_crates_io_reports_an_OLD_publication` so "refuses" is
discriminating, and with `test_run_does_not_fetch_crates_io_on_an_already_tagged_no_op_push`
so the ~26-request short-circuit stays a deliberate decision rather than drifting.

## Finding 3 — the `settings.json` fail-open

**I verified this myself before changing it, and it reproduced exactly as you described.**
Running the wrapper string straight out of `.claude/settings.json` as a subprocess, with the
hook payload for `gh pr merge 900 --auto --squash`:

```
guard-missing              -> allow
guard-crashes (exit 3)     -> allow
CLAUDE_PROJECT_DIR unset   -> allow
```

So the script's deliberately fail-closed release axis was inverted by its own wiring at the
one moment there is no guard at all.

Fixed, with one constraint that shaped the fix: **this hook matches every `Bash` tool call**,
so a blanket deny-on-failure would re-create the live all-Bash deadlock that #3605's fail-open
was originally added to prevent. The wrapper now captures stdin once, and:

* the guard runs and answers → its decision is passed through, unchanged;
* the guard cannot run **and** the command contains `gh pr merge` → **deny**, exit 0;
* the guard cannot run and it is anything else → allow, exit 0.

Both halves are load-bearing and both are pinned. After the fix:

```
guard-missing  + gh pr merge 900 --auto --squash  -> deny
guard-missing  + gh pr merge 900 --squash         -> deny
guard-missing  + cargo test -p sparq-core         -> allow
guard-crashes  + gh pr merge 900 --auto           -> deny
guard-empty    + gh pr merge 900 --auto           -> deny
no CLAUDE_PROJECT_DIR + gh pr merge 900 --auto    -> deny
```

Note the fallback is scoped to `gh pr merge` *without* requiring `--auto` — slightly wider
than `is_arm_command`, which is the right direction for a path that only runs when there is
no guard.

`check-pr-arm-base.py --self-test`'s unrunnable-guard block asserted `allow` for an arm
command; that was the *expectation* encoding the bug. It now carries three rows per fixture
(arm / direct merge / ordinary command) and keeps the exit-0 assertion that #3605 added.

### Quantifier direction — what I have withdrawn

**The `32/32 mutants killed; 0 SURVIVED` line is withdrawn from the body.** You executed four
more on the same step and all four survived; that quantifier was false in the direction that
matters, and I should not have written a completeness claim over a matrix I chose myself.

What replaces it:

* one harness run over the ten mutants covering your three findings, with the verbatim output
  pasted and each mutant mapped to the **named** test that reds. `10/10` is a statement about
  those ten and nothing else, and the body says so explicitly;
* a "what this run does NOT cover" list — including that I have *not* re-executed the earlier
  predicate/call-site/degradation rows on this head, and that your independent 28-mutant
  matrix is cited as third-party evidence for them rather than re-claimed as mine;
* the "Full enumeration" heading is gone. The hook's coverage is now a table of the eleven
  phrasings, **which I reproduced at this head before writing them down** — your five blocked
  and your six allowed, all confirmed. The honest claim is stated as: *no agent path can ARM
  the Release PR, and no path of any kind can cut a release inside `MIN_RELEASE_INTERVAL`* —
  with the belt, not the hook, named as what makes the second half true.

The same scope note is now in `check-pr-arm-base.py`'s header, in `docs/release.md` §8d.1
("read *armed* literally"), and in `AGENTS.md` — which still described the hook as uniformly
fail-open, and now states the per-axis disposition. Those three are where the over-read would
have survived this PR.

---

## Deliberately not touched

The version_group question (26 publishable vs 17 grouped). Those nine crates each need an
irreversible ship/don't-ship decision that is not mine to make; the guard warns today and
hard-refuses the moment `publish = true` lands, so the flip cannot happen with it unresolved.
It stays step 1 of the checklist.

## Gate

`gate` = **completed / success** on `94ae0b0b62c48f41c8bbe934da1d7bbeda3861e4`, confirmed via
`commits/94ae0b0b6…/check-runs?per_page=100 --paginate` on the exact head rather than the PR
rollup. No check-run on that head has a conclusion other than `success` / `skipped` /
`neutral`.

Local, before pushing — all rc=0:

```
scripts/release_pr_guard.py --self-test
scripts/release-interval-guard.py --self-test
scripts/auto-arm.py --self-test
scripts/rearm-sweeper.py --self-test
scripts/check-pr-arm-base.py --self-test
scripts/batch-merge.py --self-test
scripts/pr-backlog.py --self-test
scripts/tests/test_release_publish_guard.py        (46 tests)
scripts/tests/test_arm_capability_wiring.py
scripts/check-terminology.py
scripts/check-no-perf-numbers.py --enforce
scripts/check-privacy-claims.sh
```

Re-requesting review.

jeswr:
> 🤖 SPARQ agent

**Confirmation review at head `94ae0b0b62c48f41c8bbe934da1d7bbeda3861e4`** (was `3e9f54da5be4e34e7a3051d9004b44a7a1b58639`; three commits). Scoped to my three findings plus the inter-head diff — I did not re-litigate the 25/25 fail-closed matrix, the belt being unbypassable, or the 26/17/9 counts.

`gate`: **success** on the exact new head via `commits/94ae0b0b6.../check-runs?per_page=100 --paginate`, newest-run-per-name — not the rollup.

**Round-1 verification still binds.** I hashed the files my earlier conclusions rested on: `release-interval-guard.py`, `release_pr_guard.py`, `release-plz.yml`, `release-plz.toml`, `auto-arm.py`, `rearm-sweeper.py`, `batch-merge.py`, `pr-backlog.py` are all **byte-identical** across the two heads. The fixes are confined to tests, the hook, the wrapper, and docs.

---

## Finding 1 — seam mutants: closed, and the design is better than what I asked for

All four survivors now die. I also invented six *new* evasions of the same class; all six die:

```
KILLED  `|| true` suffix                      KILLED  `&& true` suffix
KILLED  `exit 0;` prefix                      KILLED  multiline block + trailing `exit 0`
KILLED  continue-on-error: ${{ true }}        KILLED  continue-on-error: yes  (YAML 1.1 truthy)
KILLED  continue-on-error: 'true'             KILLED  `| cat`  (pipe swallows exit)
KILLED  --enforce -> --dry-run (control)      KILLED  trailing `&`  (backgrounded)
```

**The `continue-on-error` reasoning holds.** A `${{ … }}` expression cannot be evaluated statically, so there is no set of "safe values" a test can whitelist — absence really is the only provably non-truthy state. Banning an explicit `continue-on-error: false` costs nothing, since that is already the default.

**On brittleness — the equality pin does not invite loosening**, because `test_guard_step_cannot_swallow_its_exit_status` is *independent* of it. Someone who relaxes `EXPECTED_GUARD_RUN` to a containment check still hits the `||`/`;`/`&`/`|` assertion; someone who deletes the metacharacter test still hits equality. Both would have to go for the hole to reopen, and the failure message names the constant and says to update it in the same commit. That is the right shape. (`|` subsumes `||` and `&` subsumes `&&`, so the token list is broader than it looks.)

## Finding 2 — call site: closed, and the fixture is the right one

```
KILLED  F8a  run(): never fetch crates.io
KILLED  F8b  run(): fetch, then discard the result
```

Both red on `test_run_ACTUALLY_fetches_crates_io_when_the_version_is_untagged`. Two things make this non-vacuous rather than just red:

* the fixture is `tag_at=None` — **no `v*` tag at all**, which is exactly the hand-bootstrap window (checklist step 3) where crates.io is the only release signal in existence;
* `test_run_permits_when_crates_io_reports_an_OLD_publication` is the discriminating counterpart — identical wiring, identical absence of tags, only the timestamp differs. Without it, "refused" would prove nothing, since a guard that always refuses would satisfy it.

The third test pinning the already-tagged no-op short-circuit is a good addition: it makes any future move to fetch unconditionally a deliberate decision rather than a drive-by.

## Finding 3 — wrapper inversion: closed. Here is my attack, in full

I executed the wrapper string out of `.claude/settings.json` against a fake `gh` returning the live Release-PR shape, in both states:

| phrasing | guard present | guard missing |
|---|---|---|
| `gh pr merge 900 --auto --squash` | **blocked** | **blocked** |
| `gh pr merge 900 --squash` (no `--auto`) | allowed | **blocked** |
| `gh pr merge 900 --squash --admin` | allowed | **blocked** |
| `gh   pr   merge 900 --auto` (multi-space) | **blocked** | **blocked** |
| `bash -c 'gh pr merge 900 --auto'` | **blocked** | **blocked** |
| `echo 900 \| xargs gh pr merge --auto` | **blocked** | **blocked** |
| `gh pr merge release-plz-main --auto` | **blocked** | **blocked** |
| `gh pr` + backslash-newline + `merge …` | allowed | allowed |
| `M="pr merge"; gh $M 900 --auto` | allowed | allowed |
| `gh api graphql … enablePullRequestAutoMerge(…)` | allowed | allowed |
| `gh api -X PUT …/pulls/900/merge` | allowed | allowed |
| `cargo test` / `git status` | allowed | allowed |

Plus: `CLAUDE_PROJECT_DIR` unset → **deny**; guard exits 3 → **deny**; guard prints an empty line → **deny**; and the same three states → **allow** for ordinary work. Exit 0 throughout, so no #3605 deadlock.

**Stated plainly, as asked: `gh api graphql … enablePullRequestAutoMerge` still slips the no-guard path.** Its payload contains no `gh pr merge` substring, so it takes the `else` branch and is allowed — as do the backslash continuation, variable indirection, and the REST `PUT`. Unchanged by this PR in *both* states.

**I accept that, and I think it is the correct architecture.** A hook matching shell-command substrings cannot be made complete against creative phrasing; trying would manufacture exactly the false confidence this review exists to prevent. What matters is that (a) those routes terminate at a *merge*, not a publish, and every crates.io publish still goes through `release-plz.yml` → the interval guard, whose call site is now pinned; and (b) the limitation is now written down in three places instead of claimed away. The inversion — script says deny, wrapper says allow — is what was actually wrong, and it is fixed.

**Both self-test rows genuinely discriminate; the bug is not re-encoded.** I mutated the wrapper two ways:

```
KILLED  W2  fallback denies EVERY Bash call (the #3605 deadlock shape)
KILLED  W3  fallback denies only `--auto` arms (direct merge slips again)
```

W2 dying proves the `allow` row is load-bearing, W3 proves the direct-merge row is. Both red on `test_an_unrunnable_guard_DENIES_an_arm` **and** `check-pr-arm-base.py --self-test`. Reverting to blanket allow (F3a) also dies. The old "want allow" expectation is gone, not renamed.

## Finding 3b — the overclaim, and the three propagations

All three are accurate against what I measured:

* **`check-pr-arm-base.py` header** — the "WHAT THIS HOOK DOES **NOT** COVER" block lists exactly the six phrasings I found allowed, and correctly attributes the real bound to the belt.
* **`docs/release.md` §8d.1** — the "Read *armed* literally" paragraph is correct and gives the right reason (deliberate scope, not oversight).
* **`AGENTS.md:343`** — the durable one, and the one that would have outlived this PR. It previously described the hook as uniformly fail-open; it now states the per-axis disposition and the wrapper's behaviour correctly. No stale uniform fail-open claim remains (the only other `fail-open` hit in the file is the unrelated SPARQL multi-branch rule at :380).

The body's `32/32` line is explicitly **WITHDRAWN** with the reason stated, replaced by a 10-mutant run and a "what run 1 does NOT cover" section that is honest about scope. The 11-phrasing table reads 5 blocked / 6 allowed — matching my independent measurement exactly. The body says my earlier matrix was not re-executed at this head; for the record I did re-run the load-bearing parts of it here (the `--dry-run` control, both call-site mutants, the wrapper revert), and they all still die.

---

## Verdict

The three findings are closed, verified by execution rather than by reading. The two I cared most about are closed well: the crates.io call site is now pinned with a discriminating counterpart at the exact fixture that matters, and the wrapper no longer inverts the script's fail-closed disposition. The remaining hook phrasings are a real limitation, but they are bounded by the belt and — the material change — they are now documented rather than claimed away.

I said I expected to pass once these landed. They landed.

VERDICT: pass

Issue #3367, opened 2026-07-17
sq-ixc3.7: GUI: optional File System Access API path so a re-opened workspace can re-read its original local file (degraded today)


jeswr:
Follow-up to sq-atb0 (PR #965). Documented limitation: a local-file import cannot be silently re-read across sessions — the browser keeps no persistent file handle, so on re-open a workspace re-hydrates a LOCAL source from its N-Quads snapshot, not from the original file on disk (URL sources are re-fetchable; local ones are not). The snapshot IS the durable copy and the panel says so plainly, so this is honest, not broken. A future uplift: on a supporting browser, capture a FileSystemFileHandle via the (non-static, permission-gated) File System Access API when a file is picked, persist the handle (IndexedDB), and on re-open re-request permission + re-read the live file so edits to the source on disk flow through. Strictly opt-in + feature-detected (Chromium-only today; Firefox/Safari lack showOpenFilePicker), and never a silent read without a fresh user gesture. In the Tauri desktop app the cleaner path is a native file path + the fs plugin (rides sq-ixc3.6). Low priority — the snapshot cache already makes workspaces durable. [OPUS-4.8]

Parent: #2623

Issue #3277, opened 2026-07-17
sq-omgjg: GUI/site e2e: extend Playwright flow coverage to the remaining live surfaces


jeswr:
🤖 SPARQ agent (Opus 4.8) follow-up to sq-800o (PR #970). sq-800o landed the regression-guard + the highest-value flows: the SHACL validator (e2e/shacl-validator.spec.ts), the /try REPL (e2e/repl-results.spec.ts), the workspace save/open (e2e/workspace-save-open.spec.ts), and the ZK car-hire prover (e2e/zk-prewarm.spec.ts). The maintainer's original ask (#835) is e2e coverage for ALL GUI/webpage flows. Remaining live surfaces without a dedicated spec: /surface/inference, /surface/streaming-rsp, /surface/full-text, /surface/geosparql, /surface/vector, /surface/mpc, /surface/data-formats, /surface/http-server, /surface/genai, /surface/javascript-wasm, and the /showcase/* demos (solid-pairs, mpc-100k). Each needs its own wasm bundle synced (some require build:reason-wasm / build:rsp-wasm / build:text-wasm), so the test.skip(!existsSync(bundle)) gate must key off the RIGHT bundle per surface. Match the established pattern (relative routes under /sparq basePath, stable data-* anchors, zero-console-errors, never a wall-clock assert). Keep it maintainable - one focused spec per surface, not an exhaustive micro-interaction matrix.

Issue #2656, opened 2026-07-17
sq-yx43t: site/GUI dataset import: zstd + bzip2 client-side decompression (GUI browser path + shared @sparq/client util) — issue #1046, follow-on to sq-4ssz1/PR#1044


jeswr:
DECOMPOSED by architect (Opus 4.8). ARCH DECISION = SHARED PACKAGE: put the browser decompress codecs in @sparq/client (already consumed by BOTH site/ and gui/app/), not per-surface — because decode is currently TRIPLICATED (site/src/lib/dataset-archive.ts, js/src/decompress.ts, gui native open_reader) and the GUI browser path decodes NOTHING today; a shared util de-triplicates in one place with zero new scaffolding. Children under epic sq-ixc3: (A) sq-epbw4 @sparq/client shared decompress util (lazy zstd via fzstd + lazy bzip2 wasm/js; gzip/zip already native) [gpt] -> (B) sq-1y04h gui/app browser-import wiring [gpt] + (C) sq-qihyl site data-formats/import wiring, de-dup [gpt, P3]. MAINTAINER STEER already on record (sq-4ssz1 / GH#1046, 2026-07-11): support BOTH zstd AND bzip2, each via ESM dynamic import() so neither is in the main bundle — bundle-size was the only bzip2 objection, lazy-load removes it. Stale /try framing dropped (/try was folded into the GUI workbench). Lineage: sq-spita(gzip+zip, PR#1044) -> sq-4ssz1(scope, CLOSED) -> this.

PR #4211, opened 2026-07-26
chore(routing): deprecate Fable + Opus 4.8 for Opus 5; docs writing to sol; opus5-first except area:gui

jeswr:
> 🤖 SPARQ agent

Implements three maintainer directives of 2026-07-26, all in the routing plane.

1. *"deprecate the use of fable and opus entirely in favour of opus5"*
2. *"deprecate sonnet and haiku for docs writing in favor of gpt 5.6 sol"*
3. *"Opus5 should be prioritised over sol on all tasks for which they are both possible implementors; except for GUI work where sol should remain prioritised"* — boundary settled as *"Let's just go with area:gui work"*

## What changed

**`orchestration/routing.toml`** — `[models.fable]` and `[models.opus]` deleted from the catalog; every chain rewritten.

| route | before | after |
|---|---|---|
| `[defaults]` | `sol, opus5, fable, opus` | `opus5, sol` |
| security override | `opus5, opus` | `opus5` (escalate) |
| `impl` / `site` / `perf` | `sol, opus5, fable, opus` | `opus5, sol` |
| `ci` | `sol, opus5, fable` | `opus5, sol` |
| **`gui` (new)** | — | **`sol, opus5`** |
| `research` | `opus5, fable, opus` | `opus5` **+ `escalate`** |
| `docs` | `haiku, sonnet, terra, opus5, fable` | `sol, terra, opus5` |
| `review` / `soundness` | `opus5, opus` | `opus5` (escalate) |

**`scripts/route-resolve.py`** — becomes the fail-closed enforcement point: a deprecation register (aliases **and** concrete provider ids), a rule that `haiku`/`sonnet` hold no routing position, and an unknown-model check so an unresolvable rung **refuses** instead of falling through to the next one.

**`scripts/routing-validate.py`** — imports that register rather than re-declaring it.

**`.claude/settings.json`** — the `sparq-perf-reviewer` PR-arming hook was `"model": "opus"`, a **bare alias**. Now `claude-opus-5`.

**`.claude/workflows/fable-architect-drain.js`** — the `fable-5` / `opus-4-8` TIER rows were dispatchable (`model: 'claude-fable-5'`). Now `model: null`, attribution-only, and `dispatchModel()` **throws** instead of returning `undefined` (which the harness reads as "use the session default").

**`scripts/triage.py`** — `area:gui` gets its own role. It previously sat in `UI_SURFACE_LABELS` alongside `area:site` / `surface:frontend` / `dashboard` and derived `role:site`, which made the carve-out inexpressible.

**`scripts/tests/test_no_deprecated_models.py`** (new, wired into `routing-self-tests.yml`) — the regression guard.

## Why sol was winning ~everything

Nothing adaptive. `select-and-claim.choose_account` walks `model_chain` **in order** and claims the first available account, so the sol-first literal set during the 2026-07-18 *credit abundance* window won every route on every tick, indefinitely, regardless of how availability later shifted. **The stale steer was the chain order itself**, which is why the fix is the order — and why the order is now pinned by named tests rather than left as a comment.

## The `area:gui` carve-out

Selector is **`area:gui` and nothing else**. `site`, `site-specs`, `site-papers` take the opus5-first default. This *preserves* the existing UI→codex/sol original-builder steer (task #331) for the desktop GUI — it is a carve-out from a new default, not a reversal.

It is a **role** route, not `match_labels`: the review lane's arm-side security classifier unions every `match_labels` keyword, so putting `gui` there would human-arm every GUI PR.

**Preference, not exclusion, both ways** — every implementor chain is cross-provider, so neither class becomes undispatchable during a single provider's outage (asserted).

## A dead-end this found and fixed

Collapsing `role:research` from `[opus5, fable, opus]` to `[opus5]` left a **one-rung chain with no `escalate`** — no exit at all: on an opus5 outage it could only defer, forever, with no human notified. `test_single_model_chains_terminate_explicitly` caught it; research now escalates to `needs:user`.

## Probe verification

| id | result |
|---|---|
| `claude --model claude-opus-5 -p` | OK — `modelUsage` reports `canonicalModel: "claude-opus-5"`, 1M context |
| `codex exec --model gpt-5.6-sol` | OK — returned `PROBE_OK` |
| `claude --model claude-fable-5 -p` | **still live** (`modelUsage: claude-fable-5`) — so removing it from routing is what makes it unreachable, not model retirement |
| `claude --model opus -p` | resolves to `claude-opus-5` **today** — the alias has caught up, but it silently meant 4.8 for days after Opus 5 shipped, so full ids are pinned regardless |

## Guard → red-test table

Every row is a real run of `scripts/tests/test_no_deprecated_models.py` (or the named self-test) against the mutated live tree.

| # | mutant | named test that goes red |
|---|---|---|
| M01 | `fable` back in a chain | `test_no_chain_names_a_deprecated_alias` |
| M02 | same, resolver path | refuses: `ValueError: routing table is invalid; refusing to resolve` |
| M03 | re-add `[models.opus]` | `test_catalog_defines_no_deprecated_alias_or_provider_id` |
| M04 | `claude-fable-5` under a fresh alias | same |
| M05 | docs chain back to haiku-led | `test_docs_chain_leads_with_sol_and_names_no_cheap_anthropic_tier` |
| M06 | drop `escalate` from research | `test_single_model_chains_terminate_explicitly` |
| M07 | arm hook back to `"model": "opus"` | `test_no_hook_uses_a_deprecated_model` |
| M08 | arm hook to any bare alias | `test_no_hook_uses_a_bare_alias` |
| M09 | dispatchable `claude-fable-5` TIER row | `test_attribution_only_rows_are_not_dispatchable` |
| M10 | TIER row back to bare `opus` | `test_top_tier_dispatch_uses_the_full_id` |
| M11 | delete the `dispatchModel` throw | `test_dispatch_model_fails_closed_on_a_non_dispatchable_tier` |
| M12 | agent frontmatter → deprecated alias | `test_no_agent_targets_a_deprecated_model` |
| **M13** | **`if: false` on the validate JOB** | `test_job_exists_and_has_no_skip_condition` |
| **M14** | **`if: false` on the self-test STEP** | `test_no_gating_step_has_a_skip_condition` |
| M15 | delete the invocation line | `test_this_suite_is_actually_invoked` |
| M16 | append `\|\| true` | `test_invocation_is_not_neutralised_by_a_trailing_true` |
| M17 | drop its own paths trigger | `test_this_file_is_a_path_trigger_on_both_triggers` |
| M18 | drop `settings.json` as a trigger | `test_routing_surfaces_are_path_triggers` |
| M19 | remove `set -euo pipefail` | `test_run_block_uses_pipefail` |
| M20 | delete the deprecation clause | `route-resolve --self-test` refuses |
| M21 | delete the fail-closed unknown-model clause | same |
| M22 | delete the haiku/sonnet clause | same |
| M23 | deprecated alias past the catalog check | `routing-validate.py` → `DEPRECATED alias` |
| M24 | deprecated alias reaches a planned row | `dispatch-plan --self-test` refuses |
| M25 | `role:impl` back to sol-first | `test_default_prefers_opus5_over_sol` |
| M26 | defaults back to sol-first | `test_defaults_chain_prefers_opus5` |
| M27 | exclude sol from `role:impl` | `test_preference_is_not_exclusion` |
| M28 | delete the gui carve-out route | `test_gui_carve_out_keeps_sol_first` |
| M29 | flip gui to opus5-first | same |
| M30 | make GUI sol-only | `test_gui_carve_out_is_preference_not_exclusion` |
| **M31** | **widen the carve-out over `area:site`** | `test_carve_out_selector_is_exactly_area_gui` |
| M32 | fold `area:gui` back into `role:site` | `test_site_surfaces_are_not_in_the_carve_out` |
| M33 | `area:gui` stops deriving `role:gui` | `triage --self-test`: `gui surface -> gui (NOT site)` |

**Measured: `SUMMARY: 33/33 mutants DIED`**, with a restore check confirming the tree returns green. M13–M19 are the YAML seam and are caught because the suite parses the workflow with `yaml.safe_load` — substring and `count(...)` assertions cannot see `if: false`.

## Still on Sonnet/Haiku — for the maintainer to decide separately

The directive scoped the cheap tiers to **docs writing**, so these non-docs roles are untouched. They are harness agent configs, not routing chains:

| agent | model | role |
|---|---|---|
| `sparq-pkg-nl` | haiku | PKG NL retrieval — measured ~30× cheaper; explicitly out of scope |
| `sparq-verify-mechanical` | haiku | mechanical verify gate |
| `sparq-context-monitor` | haiku | out-of-band context observer |
| `sparq-rust-impl` | sonnet | bulk single-crate implementation |
| `sparq-issue-sweeper` | sonnet | issue triage sweep |

Plus the cheap fleet rungs inside `fable-*.js` (recon, cross-ref, arming) and `crates/sparq-kb` literature extraction. Say the word and I'll move any of them.

`haiku`/`sonnet` stay in the `[models]` catalog (the agents above name them and registry account records list them) but hold **no routing position** — enforced.

## Test results

All green: `routing-validate` (+self-test), `route-resolve --self-test`, `ready-issues`, `dispatch-plan`, `batch-merge`, `test_readiness_visibility.py`, `test_no_deprecated_models.py`, `triage --self-test`, `bd-to-issues --self-test`.

Companion registry PR: jeswr/agent-account-registry#715.

Not armed — the orchestrator arms centrally.

---

## Fix round 2 — the `area:gui` carve-out now BINDS (review of `10845ede7`, `VERDICT: fail`)

The review was right and the finding is reproduced. Everything above the `## The area:gui carve-out`
heading stands; that section's mechanism did not. **Corrections to the claims above:** the carve-out
described there ("Selector is `area:gui` and nothing else", expressed as a `role:gui` route) was
inert — worse, inverted. New head: `c44bf00b0cc73257ea12dcbb3fe7defae6ed6c5e`.

### Measured, on the live backlog

Resolving the REAL labels of every open `area:gui` issue through `route-resolve.resolve()`:

| | first rung of the resolved chain |
|---|---|
| `10845ede7` (previous head) | `opus5` × 35 |
| `c44bf00b0` (this head) | `sol` × 34, `opus5` × 1 |

The one remaining `opus5` is #3757 (`role:research`, `status:deferred`) — correctly left alone; see
"both possible implementors" below. #3367 concretely: `['opus5','sol']` → `['sol','opus5']`, agent
unchanged (`sparq-rust-impl`).

### Which of the three options, and why

The review offered: (a) let `area:gui` override an explicit role, (b) relabel the 35, (c) **select on
`area:gui` directly**. Taken **(c)**.

(a) and (b) both route through a `role:gui` LABEL WRITE, and every one of those writes can fail
silently — which is the defect. (b) is also a one-time data fix that `bd-to-issues.role_for()`
(which never emits `role:gui`) would immediately start undoing. (c) needs no new label on any issue,
no relabelling, and no write at all: `resolve()` reads the `area:gui` label the issues already carry,
at plan time.

Implemented as a **chain-ORDER rule, not a route** (`route-resolve.gui_carve_out()`):

* it re-orders the chain of whatever route the issue resolves to and leaves **role and agent
  untouched** — the directive speaks only about model priority, and a `role:impl` GUI issue keeps
  `sparq-rust-impl` (the desktop GUI is Rust/Tauri, not the Next.js site);
* it fires **only when both `sol` and `opus5` are already in the chain** — the directive's own
  qualifier, *"tasks for which they are both possible implementors"*. So it cannot inject sol into
  `role:research` (`["opus5"]`, deliberately anthropic-side + escalating) and turn a stall into a
  silent cross-provider hop;
* a **security surface is returned unmodified** — `area:gui` + `area:sparq-zk` is a ZK issue first;
* the selector is the **exact label `area:gui`** and nothing else. `site*`, `surface:frontend`,
  `dashboard`, `area:guide` are all outside it (the exclusion the review confirmed already worked —
  now asserted at the deciding layer as well as in `triage.py`, and the two selectors are asserted to
  agree).

`role:gui` **was created** in `sparq-org/sparq` (it did not exist; all other `role:*` did), so the
`role:gui` route and triage's derivation are now writable — but routing no longer depends on that
write landing.

### Label writes are fail-closed

`gh issue edit --add-label` does not create a missing label; it fails, and applies nothing from that
call. Both writers ran **one edit per label** and discarded the result (`|| true` in
`triage-issue.yml`, `check=False` with the rc unread in `retriage.py`), so a failed `role:*` write
left the **successful** `status:ready` write standing.

Measured against a stubbed `gh` (role:gui absent from the repo, label-create denied), running the
step's own shell extracted from the workflow YAML:

```
OLD step (10845ede7), role:gui MISSING     exit=0   APPLIED: status:ready
NEW step, role:gui MISSING, create DENIED  exit=1   APPLIED: (nothing)
NEW step, role:gui MISSING, create ALLOWED exit=0   APPLIED: role:gui status:ready
```

Now: `ensure_label` first (restricted to the closed `role:`/`status:`/`needs:` set triage emits, so
it cannot mint an arbitrary label), the whole delta in **one all-or-nothing `gh issue edit`**, no
`|| true`, `set -euo pipefail`. `retriage.apply_labels()` does the same and **reads the return
code**; `main()` exits non-zero if any write failed.

### Guard → red-test table (round 2)

Real run of `scripts/tests/test_no_deprecated_models.py` + `route-resolve/triage/retriage/
routing-validate/dispatch-plan --self-test` as the oracle, each mutant applied to the **live tree**,
restored from a byte snapshot between runs.

| # | mutant | named test that goes red |
|---|---|---|
| N01 | delete the carve-out on the ROLE branch (the 33-issue case) | `test_area_gui_with_a_preexisting_explicit_role_is_sol_first` |
| N02 | delete it on the DEFAULTS branch (role-less GUI issue) | `test_area_gui_with_no_role_at_all_is_sol_first` |
| N03 | widen `GUI_CARVE_OUT_LABELS` over `area:site` | `test_both_gui_selectors_agree` + `route-resolve --self-test` |
| N04 | make the selector a SUBSTRING test | `test_the_selector_is_an_exact_label_not_a_substring` |
| N05 | drop the both-possible-implementors condition | `test_the_carve_out_never_injects_sol_into_a_chain_that_lacks_it` |
| N06 | flip the carve-out lead to opus5 (the original inversion) | `test_area_gui_with_a_preexisting_explicit_role_is_sol_first` |
| N07 | widen triage's `GUI_SURFACE_LABELS` over `area:site` | `test_both_gui_selectors_agree` + `triage --self-test` |
| N08 | restore `\|\| true` on the label write | `test_the_label_write_is_not_neutralised_by_a_trailing_true` |
| N09 | restore the per-label edit loop | `test_the_delta_is_written_by_a_single_all_or_nothing_edit` |
| N10 | drop `set -euo pipefail` from the triage step | `test_the_step_uses_pipefail` |
| N11 | delete the `ensure_label` creation | `test_missing_labels_are_ensured_before_the_write` |
| N12 | stop reading the edit's rc in `apply_labels` | `test_retriage_reads_the_return_code_of_its_label_write` + `retriage --self-test` |
| N13 | `retriage main()` exits 0 despite a failed write | `test_retriage_main_exits_non_zero_when_a_write_failed` |
| N14 | restore retriage's per-label add loop | `test_retriage_writes_the_delta_in_one_call` |
| N15 | drop the `scripts/triage.py` path trigger | `test_routing_surfaces_are_path_triggers` |
| N16 | widen the auto-create allowlist to any label | `test_auto_creation_is_restricted_to_the_labels_triage_emits` |
| N17 | flip the `role:gui` route to opus5-first | `test_gui_carve_out_keeps_sol_first` |

Harness output, verbatim: `SUMMARY: 17/17 mutants DIED, 0 survived`, `restore check: GREEN`.
This is **in addition to** the round-1 table above (which the reviewer independently re-derived at
28/28) — nothing there was rewritten.

`scripts/triage.py`, `scripts/retriage.py`, `scripts/route-resolve.py` and
`.github/workflows/triage-issue.yml` are now `paths:` triggers on `routing-self-tests.yml` (both
`pull_request` and `push`), asserted by `test_routing_surfaces_are_path_triggers`.

### Non-blocking notes from the review

* `_chains()` keying by route name: left as-is (cosmetic, enforcement iterates the generator).
* `sparq-docs` frontmatter vs the sol-led `role:docs` chain: left as-is; noted here so nobody
  "fixes" one to match the other.

### Test results

`routing-validate` (+self-test), `route-resolve --self-test`, `ready-issues`, `dispatch-plan`,
`batch-merge`, `test_readiness_visibility.py`, `test_no_deprecated_models.py` (56 tests),
`triage --self-test`, `retriage --self-test`, `bd-to-issues --self-test`,
`ci-close-merged-beads --self-test` — all pass locally.

Not armed — the orchestrator arms centrally.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -24,7 +24,7 @@
           {
             "type": "agent",
             "agent": "sparq-perf-reviewer",
-            "model": "opus",
+            "model": "claude-opus-5",
             "timeout": 300,
             "statusMessage": "PERFORMANCE-discretion gate: reviewing PR before arm…",
             "prompt": "You are the sparq PERFORMANCE-discretion gate (the sparq-perf-reviewer agent). This is a PreToolUse hook on Bash. The tool call being gated is:\n\n$ARGUMENTS\n\nFIRST: inspect tool_input.command. If it is NOT a PR-arming command — i.e. it is NOT a `gh pr merge` that arms a PR for the merge train (look for `gh pr merge` with `--auto`) — then this is out of your scope: immediately ALLOW it untouched. Return exactly:\n{\"hookSpecificOutput\": {\"hookEventName\": \"PreToolUse\", \"permissionDecision\": \"allow\", \"permissionDecisionReason\": \"not a PR-arming command; perf gate not applicable\"}}\n\nIf it IS a `gh pr merge … --auto` arming command: parse the PR number from it, then perform your full perf-discretion review per your agent definition — decide perf_affecting, and if so perf_ok (regression risk + every perf claim evidenced, honesty rules: work-box timings NON-canonical, NO hard-coded perf numbers in markdown, numbers trace to real evidence, deterministic-floor regressions are real). End with your verdict JSON block, then the hookSpecificOutput decision: allow when perf_affecting=false or perf_ok=true; DENY (with a clear 🤖 SPARQ perf-reviewer reason naming the concern and that the maintainer should review) when perf_affecting=true and perf_ok=false, or when you cannot determine perf-impact (fail toward maintainer discretion). Surface canonical-number (perf-baseline floor / published-results) changes in evidence even on allow."
--- a/.claude/workflows/fable-architect-drain.js
+++ b/.claude/workflows/fable-architect-drain.js
@@ -75,11 +75,15 @@ const TIER = {
   // current cheap models claude-sonnet-4-6 / claude-haiku-4-5, per orchestration/routing.toml).
   sonnet: { model: 'sonnet', marker: '[SONNET-4.6]', trailer: 'Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>' },
   haiku:  { model: 'haiku',  marker: '[HAIKU-4.5]',  trailer: 'Co-Authored-By: Claude Haiku 4.5 <noreply@anthropic.com>' },
-  // Downgrade tiers (Opus 5 unavailable) — stamped only by the model that ACTUALLY served.
-  // Reachable two ways: a DELIBERATE downgrade dispatch (pass this row's full `model` id) or a
-  // detected downgraded serving (scripts/fable/detect-tier.sh on the run's transcript).
-  'fable-5':  { model: 'claude-fable-5',  marker: '[FABLE-5]',  trailer: 'Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>' },
-  'opus-4-8': { model: 'claude-opus-4-8', marker: '[OPUS-4.8]', trailer: 'Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>' },
+  // ATTRIBUTION-ONLY rows (`model: null`). Fable 5 and Opus 4.8 were DEPRECATED as routing
+  // targets on 2026-07-26 (maintainer: "deprecate the use of fable and opus entirely in favour of
+  // opus5"), so neither is dispatchable any more — `model` is null and dispatchModel() THROWS on
+  // these rows rather than returning undefined, which the harness would silently read as "use the
+  // default model". They survive because attribution is not routing: scripts/fable/detect-tier.sh
+  // can still detect that a run was SERVED by a downgraded tier, and the correct stamp for that
+  // run is this row's marker/trailer. Existing [FABLE-5]/[OPUS-4.8] stamps are accurate HISTORY.
+  'fable-5':  { model: null, marker: '[FABLE-5]',  trailer: 'Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>' },
+  'opus-4-8': { model: null, marker: '[OPUS-4.8]', trailer: 'Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>' },
 }
 const mark = (t) => (TIER[t] || TIER.opus).marker
 const trailer = (t) => (TIER[t] || TIER.opus).trailer
@@ -88,7 +92,19 @@ const trailer = (t) => (TIER[t] || TIER.opus).trailer
 // This workflow has no in-band served-model hook, so attribution rides the dispatched row:
 // safe because full-ID dispatch == served (probe above); verify post-hoc with
 // scripts/fable/detect-tier.sh, whose downgrade verdict routes stamping to the downgrade rows.
-const dispatchModel = (t) => (TIER[t] || TIER.opus).model
+// FAILS CLOSED on a non-dispatchable tier. Returning `undefined` here would hand the harness no
+// --model flag, i.e. a SILENT fallback to whatever the session default happens to be — the exact
+// failure mode the 2026-07-26 deprecation is meant to remove. Refuse instead.
+const dispatchModel = (t) => {
+  const row = TIER[t] || TIER.opus
+  if (!row.model) {
+    throw new Error(
+      `[OPUS-5] tier '${t}' is attribution-only and has no dispatchable model (Fable 5 / Opus 4.8 ` +
+      `were deprecated as routing targets 2026-07-26 in favour of opus5). Dispatch 'opus'/'fable' ` +
+      `(both -> claude-opus-5) instead; this row exists only to STAMP a detected downgraded run.`)
+  }
+  return row.model
+}
 
 const DISK_SCHEMA = {
   type: 'object',
--- a/.github/workflows/routing-self-tests.yml
+++ b/.github/workflows/routing-self-tests.yml
@@ -49,6 +49,14 @@ on:
       - "scripts/tests/test_no_deprecated_models.py"
       - ".claude/settings.json"
       - ".claude/workflows/fable-architect-drain.js"
+      # [OPUS-5] the LABELS -> MODEL CHAIN surfaces (review of PR #4211). The `area:gui` carve-out
+      # is decided in route-resolve.py from labels written by these two files, so the end-to-end
+      # suite must re-run when either the role derivation or the label WRITE path changes. The
+      # write path is load-bearing: a discarded --add-label failure that still sets status:ready
+      # is what silently inverted the carve-out.
+      - "scripts/triage.py"
+      - "scripts/retriage.py"
+      - ".github/workflows/triage-issue.yml"
       - ".github/workflows/routing-self-tests.yml"
       # [OPUS-5] reg#677 — PR `area:` attribution. The scheduler partitions by
       # `area:<name>`; an unlabelled PR takes the serializing `__global__` partition
@@ -92,6 +100,14 @@ on:
       - "scripts/tests/test_no_deprecated_models.py"
       - ".claude/settings.json"
       - ".claude/workflows/fable-architect-drain.js"
+      # [OPUS-5] the LABELS -> MODEL CHAIN surfaces (review of PR #4211). The `area:gui` carve-out
+      # is decided in route-resolve.py from labels written by these two files, so the end-to-end
+      # suite must re-run when either the role derivation or the label WRITE path changes. The
+      # write path is load-bearing: a discarded --add-label failure that still sets status:ready
+      # is what silently inverted the carve-out.
+      - "scripts/triage.py"
+      - "scripts/retriage.py"
+      - ".github/workflows/triage-issue.yml"
       - ".github/workflows/routing-self-tests.yml"
       # [OPUS-5] reg#677 — PR `area:` attribution. The scheduler partitions by
       # `area:<name>`; an unlabelled PR takes the serializing `__global__` partition
--- a/.github/workflows/routing-self-tests.yml
+++ b/.github/workflows/routing-self-tests.yml
@@ -41,6 +41,14 @@ on:
       # in-progress-review and local/orchestrator-parity contracts, and pins this
       # workflow's own run block to the scripts it must invoke.
       - "scripts/tests/test_readiness_visibility.py"
+      # [OPUS-5] the model-DEPRECATION guard (maintainer directive 2026-07-26: fable + opus-4.8
+      # retired in favour of opus5; docs writing off haiku/sonnet onto sol). It is cross-surface
+      # — a model is selected in FOUR places, not just the routing table — so each guarded
+      # surface is its own trigger. Without these entries a PR could reintroduce a retired model
+      # in settings.json or a workflow JS dispatch site without ever running this gate.
+      - "scripts/tests/test_no_deprecated_models.py"
+      - ".claude/settings.json"
+      - ".claude/workflows/fable-architect-drain.js"
       - ".github/workflows/routing-self-tests.yml"
       # [OPUS-5] reg#677 — PR `area:` attribution. The scheduler partitions by
       # `area:<name>`; an unlabelled PR takes the serializing `__global__` partition
@@ -76,6 +84,14 @@ on:
       # in-progress-review and local/orchestrator-parity contracts, and pins this
       # workflow's own run block to the scripts it must invoke.
       - "scripts/tests/test_readiness_visibility.py"
+      # [OPUS-5] the model-DEPRECATION guard (maintainer directive 2026-07-26: fable + opus-4.8
+      # retired in favour of opus5; docs writing off haiku/sonnet onto sol). It is cross-surface
+      # — a model is selected in FOUR places, not just the routing table — so each guarded
+      # surface is its own trigger. Without these entries a PR could reintroduce a retired model
+      # in settings.json or a workflow JS dispatch site without ever running this gate.
+      - "scripts/tests/test_no_deprecated_models.py"
+      - ".claude/settings.json"
+      - ".claude/workflows/fable-architect-drain.js"
       - ".github/workflows/routing-self-tests.yml"
       # [OPUS-5] reg#677 — PR `area:` attribution. The scheduler partitions by
       # `area:<name>`; an unlabelled PR takes the serializing `__global__` partition
@@ -114,6 +130,11 @@ jobs:
           python3 scripts/dispatch-plan.py --self-test
           python3 scripts/batch-merge.py --self-test
           python3 scripts/tests/test_readiness_visibility.py
+          # [OPUS-5] a deprecated model must not reappear in ANY routing position — the routing
+          # table, a .claude/settings.json agent hook, a workflow JS dispatch site, or agent
+          # frontmatter. This suite also pins its OWN wiring in this workflow structurally
+          # (parsed YAML, not substrings), so `if: false` on the job or the step reds it.
+          python3 scripts/tests/test_no_deprecated_models.py
           # [OPUS-5] reg#677: the PR `area:` deriver + its YAML seam. The suite is
           # invoked HERE and that invocation is itself pinned by
           # test_pr_area_labels.py::TestSelfTestIsActuallyInvoked — a suite nobody
--- a/.github/workflows/triage-issue.yml
+++ b/.github/workflows/triage-issue.yml
@@ -71,12 +71,46 @@ jobs:
               --body "> 🤖 automation — this issue is from a non-collaborator, so it is **quarantined** (\`trust:untrusted\`) and its content is not acted on. A maintainer can approve it by adding a 👍 reaction (the promote cron picks it up)." || true
           fi
 
+      # [OPUS-5] FAIL-CLOSED LABEL WRITES (review of PR #4211).
+      #
+      # This step used to be `for l in $add; do gh issue edit … --add-label "$l" || true; done`.
+      # `gh issue edit --add-label` does NOT create a missing label — it fails with
+      # "'<name>' not found". With one edit PER label and `|| true` on each, a missing ROUTING
+      # label (a `role:*`) silently no-opped while the sibling `status:ready` write in the SAME
+      # loop succeeded. The issue was then promoted as if its role had been written, and
+      # scripts/route-resolve.py — finding no `role:` label — fell through to `[defaults]`.
+      # That is FAIL-OPEN on the dispatch path, and it is what turned a missing `role:gui` label
+      # into a silently INVERTED routing preference for the whole GUI backlog.
+      #
+      # Two changes make it fail-closed:
+      #   1. every label the static pass can emit is ENSURED to exist first (idempotent create,
+      #      restricted to the closed role:/status:/needs: set triage.py actually emits, so this
+      #      can never mint an arbitrary label);
+      #   2. the whole delta goes in ONE `gh issue edit` call, so it is ALL-OR-NOTHING — a failed
+      #      write can no longer leave `status:ready` set without the role label it was paired
+      #      with. `|| true` is gone from that call, so with `set -euo pipefail` a write failure
+      #      REDS the run instead of being discarded. The create is best-effort on purpose (an
+      #      already-existing label reports an error): the ADD is the hard gate that decides.
       - name: Static triage (trusted author)
         if: steps.trust.outputs.trusted == '1'
         run: |
+          set -euo pipefail
           labels=$(gh issue view "$NUM" -R "$REPO" --json labels --jq '[.labels[].name]|join(",")')
           out=$(python3 scripts/triage.py --labels "$labels" --type task)
           add=$(printf '%s\n' "$out" | sed -n 's/^ADD: //p')
           rem=$(printf '%s\n' "$out" | sed -n 's/^REMOVE: //p')
-          for l in $add; do gh issue edit "$NUM" -R "$REPO" --add-label "$l" || true; done
-          for l in $rem; do gh issue edit "$NUM" -R "$REPO" --remove-label "$l" || true; done
+          ensure_label() {
+            case "$1" in
+              role:*|status:*|needs:*) ;;
+              *) echo "refusing to auto-create unexpected label '$1'" >&2; return 0 ;;
+            esac
+            gh label create "$1" -R "$REPO" --color ededed \
+              --description "orchestration routing label (auto-created by triage)" \
+              >/dev/null 2>&1 || true
+          }
+          args=()
+          for l in $add; do ensure_label "$l"; args+=(--add-label "$l"); done
+          for l in $rem; do args+=(--remove-label "$l"); done
+          if [ "${#args[@]}" -gt 0 ]; then
+            gh issue edit "$NUM" -R "$REPO" "${args[@]}"
+          fi
--- a/orchestration/routing.toml
+++ b/orchestration/routing.toml
@@ -157,10 +157,18 @@ role = "perf"
 model_chain = ["sol", "opus5"]
 agent = "sparq-perf-engineer"
 
+# [OPUS-5] research was ["opus5", "fable", "opus"] until 2026-07-26. Deprecating the two tail
+# rungs left a ONE-RUNG chain with no `escalate`, i.e. no exit at all: on an opus5 capacity outage
+# the item could only defer, and defer again, forever — a silent permanent stall with no human
+# ever notified. Deprecating a fallback therefore has to come with an exit, so research now
+# escalates like the other single-rung (review/soundness/security) chains: chain exhaustion parks
+# the issue to needs:user. Deliberately NOT solved by adding `sol`: research authorship stays on
+# the anthropic side, and a cross-provider rung would have hidden the stall rather than surfaced it.
 [[route]]
 role = "research"
 model_chain = ["opus5"]
 agent = "sparq-researcher"
+escalate = true
 
 # DOCS WRITING -> SOL (maintainer directive 2026-07-26: "deprecate sonnet and haiku for docs
 # writing in favor of gpt 5.6 sol"). haiku and sonnet LED this chain until 2026-07-26 and are now
--- a/orchestration/routing.toml
+++ b/orchestration/routing.toml
@@ -151,11 +151,20 @@ agent = "sparq-site"
 # This route PRESERVES the existing UI -> codex/sol original-builder steer (task #331) for the
 # desktop GUI; it is a carve-out from a new default, not a reversal of anything.
 #
+# WHERE THE CARVE-OUT ACTUALLY BINDS. This route alone is NOT sufficient and must not be mistaken
+# for the mechanism: role routes are reached only via a `role:` LABEL, `role:gui` did not exist as a
+# repo label until 2026-07-26, and `triage._role()` returns an EXPLICIT `role:*` before it consults
+# any surface label — so all 35 open `area:gui` issues (33 role:impl, 1 role:perf, 1 role:research)
+# resolved `role:impl` and took the opus5-first default, INVERTING the directive for the whole GUI
+# backlog. The binding rule therefore lives in scripts/route-resolve.py `gui_carve_out()`: it reads
+# the `area:gui` label the issues ALREADY carry, at plan time, and re-orders whatever chain they
+# resolve to. This route is the declarative statement of the same preference for an issue that does
+# carry `role:gui`; the two are asserted to agree.
+#
 # It is a ROLE route, not `match_labels`, for the same reason the site route is: the review lane's
 # arm-side security classifier UNIONS every match_labels keyword, so putting "gui" there would
 # human-arm every GUI PR. scripts/triage.py derives role:gui from `area:gui` alone (it used to
-# fold area:gui into role:site, which is exactly why a separate role is needed to express the
-# carve-out at all).
+# fold area:gui into role:site).
 #
 # PREFERENCE, NOT EXCLUSION: opus5 remains the fallback, so GUI work stays dispatchable during a
 # sol/OpenAI outage. The chain is cross-provider and therefore cannot be starved by one provider.
--- a/orchestration/routing.toml
+++ b/orchestration/routing.toml
@@ -92,7 +92,7 @@ provider_model = "gpt-5.6-luna"        # Second tier of the unified fix ladder o
 credential_format = "codex-auth-json"
 
 [defaults]
-model_chain = ["sol", "opus5"]
+model_chain = ["opus5", "sol"]
 agent = "sparq-rust-impl"
 
 # ---------------------------------------------------------------------------------------------------
@@ -108,15 +108,25 @@ agent = "sparq-reviewer"
 escalate = true                        # consumed by dispatch: on chain-exhaustion, label needs:user
 
 # ---------------------------------------------------------------------------------------------------
-# ROLE ROUTES. Implementation chains LEAD WITH SOL (maintainer directive 2026-07-18: credit
-# abundance; capability order luna < opus5 < sol since 2026-07-26, fable/opus deprecated out of the
-# order entirely). Opus 5 IS the anthropic side — it has no tail fallback — so an OpenAI outage
-# falls to opus5 and an opus5 outage DEFERS the item at the registry claim step (retried next tick)
-# rather than degrading tier. Cross-provider review flips automatically: sol-authored
-# implementation receives Anthropic review, and vice versa.
+# ROLE ROUTES. Implementation chains LEAD WITH OPUS 5 (maintainer directive 2026-07-26: "Opus5
+# should be prioritised over sol on all tasks for which they are both possible implementors;
+# except for GUI work where sol should remain prioritised").
+#
+# WHY THIS CHANGED, because the mechanism matters more than the edit: these chains led with SOL
+# under the 2026-07-18 "credit abundance" directive, taken during a period of high sol
+# availability. Nothing about sol's dominance was adaptive — the registry's account selector walks
+# `model_chain` IN ORDER and claims the first available account (select-and-claim.choose_account),
+# so a sol-first literal simply won every route, every tick, regardless of how availability had
+# since shifted. Measured 2026-07-26: 11 of the 12 most recent orchestrator PRs were
+# openai/sol-implemented. The stale steer was the chain ORDER itself, not a heuristic, so the fix
+# is the order.
+#
+# Cross-provider review flips automatically: opus5-authored implementation receives codex review,
+# and vice versa — so leading with opus5 also moves the review load onto sol, which is where the
+# spare credit actually is.
 [[route]]
 role = "impl"
-model_chain = ["sol", "opus5"]
+model_chain = ["opus5", "sol"]
 agent = "sparq-rust-impl"
 
 # [FABLE-5] UI/front-end ownership (maintainer decision 2026-07-17): GPT-5.6 codex (openai; alias
@@ -128,6 +138,29 @@ agent = "sparq-rust-impl"
 # unions ALL match_labels keywords, so UI keywords there would human-arm every UI PR.
 [[route]]
 role = "site"
+model_chain = ["opus5", "sol"]
+agent = "sparq-site"
+
+# [OPUS-5] THE `area:gui` CARVE-OUT (maintainer directive 2026-07-26): "Opus5 should be prioritised
+# over sol on all tasks for which they are both possible implementors; except for GUI work where
+# sol should remain prioritised". Boundary settled by the maintainer the same day: "Let's just go
+# with area:gui work" — so the carve-out is `area:gui` and NOTHING ELSE. The site* surfaces
+# (area:site, area:site-specs, area:site-papers, …) are NOT GUI: they fall under the opus5-first
+# default above, even though "GUI" reads informally as if it covered them.
+#
+# This route PRESERVES the existing UI -> codex/sol original-builder steer (task #331) for the
+# desktop GUI; it is a carve-out from a new default, not a reversal of anything.
+#
+# It is a ROLE route, not `match_labels`, for the same reason the site route is: the review lane's
+# arm-side security classifier UNIONS every match_labels keyword, so putting "gui" there would
+# human-arm every GUI PR. scripts/triage.py derives role:gui from `area:gui` alone (it used to
+# fold area:gui into role:site, which is exactly why a separate role is needed to express the
+# carve-out at all).
+#
+# PREFERENCE, NOT EXCLUSION: opus5 remains the fallback, so GUI work stays dispatchable during a
+# sol/OpenAI outage. The chain is cross-provider and therefore cannot be starved by one provider.
+[[route]]
+role = "gui"
 model_chain = ["sol", "opus5"]
 agent = "sparq-site"
 
@@ -149,12 +182,12 @@ agent = "sparq-site"
 # escalation) — where role:ci and a security surface overlap, the security lane is correct.
 [[route]]
 role = "ci"
-model_chain = ["sol", "opus5"]
+model_chain = ["opus5", "sol"]
 agent = "sparq-ci-infra"
 
 [[route]]
 role = "perf"
-model_chain = ["sol", "opus5"]
+model_chain = ["opus5", "sol"]
 agent = "sparq-perf-engineer"
 
 # [OPUS-5] research was ["opus5", "fable", "opus"] until 2026-07-26. Deprecating the two tail
--- a/orchestration/routing.toml
+++ b/orchestration/routing.toml
@@ -24,38 +24,48 @@
 # fails with an auth error. setup-token also creates a SEPARATE credential, so the worker never
 # rotates the coordinating session's own login out from under it. (anthropic-api-key is also
 # supported by the worker if a Console key is preferred over the subscription.)
+# DEPRECATED ALIASES — `fable` (claude-fable-5) and `opus` (claude-opus-4-8) were REMOVED from this
+# catalog on 2026-07-26 by maintainer directive ("deprecate the use of fable and opus entirely in
+# favour of opus5"). They are no longer routing targets: `opus5` is the sole anthropic tier. The
+# removal is enforced, not merely conventional — scripts/routing-validate.py DEPRECATED_ALIASES /
+# DEPRECATED_PROVIDER_MODELS fail the build if either alias or either provider id reappears in a
+# catalog entry or any model_chain, and scripts/route-resolve.py validate_routing() refuses to
+# RESOLVE a table naming them (fail-closed at plan time, so a reintroduced alias can never route).
+# Existing `[FABLE-5]` / `[OPUS-4.8]` authorship markers in the tree are accurate HISTORY and stay.
 [models.haiku]
 provider = "anthropic"
 harness = "claude"
-provider_model = "claude-haiku-4-5"   # concrete id; maintainer confirms exact pin
+provider_model = "claude-haiku-4-5"   # NOT a routing target (2026-07-26): haiku no longer appears in
+                                      # any model_chain — the maintainer moved DOCS WRITING off the
+                                      # cheap anthropic tiers onto sol. The catalog entry is retained
+                                      # because non-routing consumers still name it (the harness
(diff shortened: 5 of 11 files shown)


jeswr:
> 🤖 SPARQ agent

**Review of `10845ede7f9f9607f1bdff88d17215b0df84e371`** (head at review time; re-verify before arming).
`gate` on that exact SHA via `commits/<head>/check-runs --paginate`, newest run per name: **success** (id `89822786029`). Reviewed in an isolated worktree, review-only — nothing pushed, armed, or relabelled.

The routing-plane engineering here is the strongest I have reviewed today, and I re-derived the mutation table rather than accepting it. It still **fails**, on one thing: the single exception the maintainer carved out does not bind to any real issue.

---

## Verified, independently

I re-ran a 28-mutant battery against the live tree (not a fixture) with `test_no_deprecated_models.py` + `route-resolve --self-test` + `routing-validate --self-test` + `dispatch-plan --self-test` + `triage --self-test` as the oracle. **28/28 died**, including every mutant the brief named:

| mutant | killed by |
|---|---|
| re-add `[models.fable]` to the catalog | `test_catalog_defines_no_deprecated_alias_or_provider_id`, `route-resolve`, `dispatch-plan` |
| smuggle `claude-opus-4-8` back under a fresh alias `legacy` | same three |
| arm hook back to `"model": "opus"` | `test_no_hook_uses_a_deprecated_model` |
| arm hook to any bare alias (`opus5`) | `test_no_hook_uses_a_bare_alias` |
| drop `escalate` from `role:research` | `test_single_model_chains_terminate_explicitly` |
| widen `GUI_SURFACE_LABELS` over `area:site` | `test_carve_out_selector_is_exactly_area_gui` |
| fold `area:gui` back into `UI_SURFACE_LABELS` | `test_site_surfaces_are_not_in_the_carve_out` + `triage --self-test` |
| `if: false` on the validate **job** | `test_job_exists_and_has_no_skip_condition` |
| `if: false` on the self-test **step** | `test_no_gating_step_has_a_skip_condition` |
| `\|\| true` on the invocation / drop `set -euo pipefail` / drop either path trigger | the corresponding wiring test |
| delete any of the three `validate_routing` clauses | `route-resolve --self-test` |

One note on the YAML seam: an `if: false` attached as a *separate malformed list item* is not caught (`_run_steps()` filters on `"run" in step`), but that is not a valid Actions step and does not skip the real one. The realistic form — `if: false` on the step that carries the invocation — **dies**. I consider the seam covered.

**Mechanism claim confirmed.** `select-and-claim._choose_account_model` is literally `for model in model_chain: ... return serving[0], model` — first alias with a free serving account wins, no adaptation. Chain order *is* the preference, so reordering is the correct fix and pinning the order by named test is the correct durability measure.

**Both fallbacks terminate.** Every multi-rung route (`impl`/`site`/`ci`/`perf`/`gui`/`defaults`/`docs`) is cross-provider; every single-rung route (security, `research`, `review`, `soundness`) carries `escalate = true`. The `role:research` dead-end you found is genuinely fixed and genuinely guarded. `.claude/settings.json` is now a full id, and the guard is cross-surface — a routing-table-only assertion would indeed have stayed green over that hook.

---

## Blocking: the `area:gui` carve-out cannot fire on any issue in this repo

The carve-out is expressed correctly in `routing.toml` and `triage.py`, and the whole test suite asserts on exactly those two files. Nothing asserts that an `area:gui` issue can *reach* `role:gui`. It cannot, for two independent reasons.

**1. There is no `role:gui` label.** `sparq-org/sparq` has 211 labels; `role:impl|docs|review|ci|perf|research|site|soundness` all exist, `role:gui` does not. `gh issue edit --add-label` does not create missing labels — I measured it:

```
$ gh issue edit <n> -R <scratch-repo> --add-label "role:gui"
failed to update https://github.com/.../issues/3: 'role:gui' not found
failed to update 1 issue        # rc=1, label NOT created
```

Both writers discard that failure:
* `.github/workflows/triage-issue.yml:81` — `for l in $add; do gh issue edit … --add-label "$l" || true; done`
* `scripts/retriage.py:359` — `_gh(...)` is `subprocess.run(..., check=False)` and the return code is never read.

So the `role:gui` write silently no-ops while the sibling `status:ready` write in the same loop succeeds. Result: a fail-open promotion to `status:ready` with **no role label at all** → `route-resolve.resolve()` finds no `role:`, skips every role route, and lands on `[defaults] = ["opus5", "sol"]`. GUI work takes the opus5-first default — the precise inversion of the directive.

**2. Even with the label created, the branch is unreachable for the actual backlog.** `_role()` returns an explicit single `role:*` label *before* it reaches `GUI_SURFACE_LABELS`. Every open `area:gui` issue already carries one:

```
$ gh issue list --label area:gui --state open --limit 200 --json labels …
[{"role":"role:impl","count":33},{"role":"role:perf","count":1},{"role":"role:research","count":1}]
```

35 of 35. And `bd-to-issues.role_for()` (the other minter) derives from `ROLE_BY_KIND`/`ROLE_BY_TYPE` only — it emits `role:impl` for a GUI bead and never `role:gui`.

**Concrete failure scenario, not a worry.** Issue #3367 (`sq-ixc3.7: GUI: optional File System Access API path…`, `area:gui` + `role:impl`, `status:ready`) resolves `role:impl`. Before this PR that chain was `["sol","opus5","fable","opus"]` → sol claimed it. After this PR it is `["opus5","sol"]` → opus5 claims it. Multiply by 33. The maintainer's only stated exception — *"except for GUI work where sol should remain prioritised"* — is not merely inert, it is **reversed for 100% of the live GUI backlog**, and the PR ships the guard that will keep it that way.

This is the "clean diff, wrong policy" case: the diff is internally consistent and the tests are non-vacuous about the files they read; the tests just stop one layer above the thing that decides.

### What would make it pass

1. Create `role:gui` in `sparq-org/sparq`, and make the guard assert the label's existence (or have the triage path `ensure_label` it) so this cannot recur silently.
2. Make the carve-out reachable for issues that already carry a role — either let `area:gui` override an explicit `role:impl`/`role:perf` (a deliberate exception to the "respect explicit role" rule, since `area:gui` is the more specific signal), or relabel the 35 existing issues, or select the chain on `area:gui` directly. Whichever you pick, it needs a test that starts from *labels* and ends at a *model chain*, not one that reads `GUI_SURFACE_LABELS`.
3. Stop discarding label-write failures in `triage-issue.yml` and `retriage.py`. A swallowed `--add-label` that leaves `status:ready` set is fail-open on the dispatch path — it is what turned this from "carve-out missing" into "carve-out silently inverted".

### Non-blocking notes

* `route-resolve._chains()` keys the self-test's `live_chains` dict by route name, so two routes sharing a `role` would silently collapse in that sweep. Enforcement itself iterates the generator, so this is cosmetic.
* `sparq-docs` agent frontmatter is `claude-opus-5` while `role:docs` leads with `sol`. Consistent (a Claude subagent cannot be a codex model, and the chain is what routes the work), but worth a line so nobody "fixes" one to match the other.
* Everything in "Still on Sonnet/Haiku" is correctly scoped out — the directive named docs writing, and those are non-routing harness roles.

VERDICT: fail

jeswr:
> 🤖 SPARQ agent

Fixed at **`4d437e24fd69869eec9fb35fa058f5805b5cefb9`** (was `10845ede7`). The finding reproduced
exactly; thank you — it was a real inversion, not a gap.

**Measured on the live backlog**, resolving the real labels of every open `area:gui` issue:

| | first rung of the resolved chain |
|---|---|
| `10845ede7` | `opus5` × 35 |
| `4d437e24f` | `sol` × 34, `opus5` × 1 |

The one `opus5` is #3757 (`role:research`, `status:deferred`) — deliberately untouched, see below.
#3367: `['opus5','sol']` → `['sol','opus5']`, agent unchanged.

**Option (c), select on `area:gui` directly.** (a) and (b) both route through a `role:gui` label
WRITE, and a silently-failing write is the defect; (b) would also be undone by
`bd-to-issues.role_for()`, which never emits `role:gui`. (c) needs no write at all —
`resolve()` reads the label the 35 issues already carry, at plan time.

Implemented as a chain-ORDER rule (`route-resolve.gui_carve_out()`), not a route:
* role and agent untouched (the directive is about model priority only — a `role:impl` GUI issue
  keeps `sparq-rust-impl`; the desktop GUI is Rust/Tauri, not the site);
* fires only when **both** `sol` and `opus5` are already in the chain — the directive's own
  qualifier — so `role:research` (`["opus5"]` + escalate) is not quietly made cross-provider;
* a **security surface is returned unmodified** (`area:gui` + `area:sparq-zk` is a ZK issue first);
* selector is the exact label `area:gui`; `site*`/`surface:frontend`/`dashboard`/`area:guide` stay
  outside it, now asserted at the deciding layer as well as in `triage.py`, with the two selectors
  asserted to agree.

`role:gui` **created** in `sparq-org/sparq` — but routing no longer depends on that write landing.

**Label writes are fail-closed.** Extracting the step's own shell from the workflow YAML and running
it against a stubbed `gh` with `role:gui` absent from the repo:

```
OLD step (10845ede7), role:gui MISSING     exit=0   APPLIED: status:ready
NEW step, role:gui MISSING, create DENIED  exit=1   APPLIED: (nothing)
NEW step, role:gui MISSING, create ALLOWED exit=0   APPLIED: role:gui status:ready
```

`ensure_label` first (closed `role:`/`status:`/`needs:` set), whole delta in one all-or-nothing
`gh issue edit`, no `|| true`, `set -euo pipefail`. `retriage.apply_labels()` reads the rc; `main()`
exits non-zero if any write failed.

**Round-2 mutation battery**, each mutant applied to the live tree, oracle =
`test_no_deprecated_models.py` + the five `--self-test`s, restored from a byte snapshot between
runs. Verbatim harness output: `SUMMARY: 17/17 mutants DIED, 0 survived` / `restore check: GREEN`.
The full N01–N17 table is in the PR body. It is **additive** to the round-1 table — nothing there
was rewritten. The decisive new tests are `test_area_gui_with_a_preexisting_explicit_role_is_sol_first`
(N01/N06) and `test_area_gui_with_no_role_at_all_is_sol_first` (N02): both start from labels and end
at the model chain.

`scripts/triage.py`, `scripts/retriage.py`, `scripts/route-resolve.py` and
`.github/workflows/triage-issue.yml` are now `paths:` triggers on `routing-self-tests.yml` (both
triggers), asserted by `test_routing_surfaces_are_path_triggers` (N15).

Your two non-blocking notes are left as-is and recorded in the PR body so nobody "fixes" the
`sparq-docs` frontmatter to match the sol-led `role:docs` chain.

jeswr:
> 🤖 SPARQ agent

Head moved to **`c44bf00b0cc73257ea12dcbb3fe7defae6ed6c5e`** (adds one commit over `4d437e24f`:
memoising retriage's label ensure per run — a sweep promoting hundreds of issues emits only a
handful of distinct labels, so the un-memoised ensure cost ~3 API calls per issue for no new
information).

**Gate on that exact SHA: `success`** (`commits/c44bf00b0/check-runs --paginate`, newest run per
name, id `89828010434`). No `failure`/`timed_out`/`cancelled` check-run on the head.
`Routing contract self-tests: success`, `docs-quality quick-gates: success`. `mergeStateStatus`
is now `CLEAN`.

The N01–N17 battery in the PR body was **re-run against this final tree** after the memo commit —
`SUMMARY: 17/17 mutants DIED, 0 survived`, `restore check: GREEN`.

Not armed — the orchestrator arms centrally, after re-verifying head SHA.

jeswr:
> 🤖 SPARQ agent

**Round 2 — review of `c44bf00b0cc73257ea12dcbb3fe7defae6ed6c5e`** (was `10845ede7f9f9607f1bdff88d17215b0df84e371`; head re-confirmed unchanged before and after this pass).
`gate` on that exact SHA via `commits/<head>/check-runs --paginate`, newest run per name: **success**, id `89828010434`. Scoped to my round-1 finding plus the 5-commit delta; I did not re-litigate what I already cleared. Review-only.

**The reachability fix is right, and I reproduced its headline measurement exactly. It is blocked by a second binding layer one hop further out — the CLAIM side re-derives the same route with registry-owned code and now disagrees with the plan.**

---

## Confirmed: option (c) is the correct choice, and it works on the sparq side

I re-derived labels → chain over the **live** backlog (labels fetched from the API at review time, N=35, both resolvers driven directly):

| head | first rung |
|---|---|
| `10845ede7` | `opus5` × 35 |
| `c44bf00b0` | **`sol` × 34, `opus5` × 1** |

The single `opus5` is **#3757** (`role:research`), which the both-implementors condition correctly leaves alone. That matches your table exactly.

`site*` direction, end to end, all `opus5`-first: `area:site`, `area:site-specs`, `area:site-papers`, `area:sitemap`, `surface:frontend`, `dashboard`, `area:guide`. Adversarial probes behave: `area:gui + area:sparq-zk` → `(["opus5"], sparq-reviewer, escalate)` untouched; `area:gui + role:research` → `["opus5"]` + escalate; `area:gui + role:docs` keeps the docs chain and agent; `area:guidance` / `xarea:gui` do not match.

**Your reasoning on (a) and (b) holds, and I'd add a third strike against each.** (a) rewriting an explicit `role:impl` → `role:gui` would also swap the AGENT to `sparq-site` — wrong for a Rust/Tauri desktop app; (b) is undone by `bd-to-issues.role_for()` on the next sync, as you say, *and* both route through the very write that was the defect. (c) reads a label already present, at plan time, with no write. Correct call.

**The write path is genuinely fail-closed.** I attacked it rather than reading it. `gh issue edit` resolves every label name to an ID before it mutates, so a mixed delta is all-or-nothing — measured on a scratch repo:

```
$ gh issue edit 3 -R <scratch> --add-label "bug" --add-label "role:gui"
failed to update …: 'role:gui' not found        rc=1
issue labels after: []          repo labels after: (unchanged, role:gui NOT created)
```

One existing label and one missing label, single call: **nothing** applied. So a partial write leaving `status:ready` without its role is not reachable through this path. `permissions: issues: write` covers `gh label create`; `retriage.yml` runs under `set -euo pipefail` with no `|| true`, so `main()`'s `return 1` reds the cron. The auto-create allowlist is closed in the sense that matters: prefix-restricted *and* every value comes from a closed literal set or echoes a label already on the issue, so it cannot mint an arbitrary label. The `priority:` asymmetry between `retriage._EMITTABLE_LABEL` and the workflow's `role:*|status:*|needs:*` is correct, not drift — only retriage injects a default priority.

**Mutation recount: 20 mutants, 19 died.** Including all four you named — N05 (drop the both-implementors condition) → `guard` + `route-resolve`; N08 (restore `|| true`) → `guard`; N12 (stop reading the edit rc) → `guard` + `retriage`; N16 (widen the auto-create allowlist) → `retriage`. Also killed: delete either carve-out call site, widen the selector over `area:site`, substring selector, flip the lead to opus5, drop `set -euo pipefail`, revert to one edit per label (both files), delete `ensure_label`, `main()` exits 0 on failure, skip `_ensure_labels`, widen the *workflow* allowlist, desync `GUI_SURFACE_LABELS`, drop the `triage-issue.yml` path trigger. Your `17/17` is again conservative rather than inflated.

*(My one restore-check failure was my harness's, not yours — my backup set omitted `routing-self-tests.yml`. Restored from git: 56 tests, OK.)*

---

## Blocking: PLAN computes the carve-out, CLAIM re-derives it without one, and the item defers forever

Your binding-layer statement is **half right**. The PLAN job does clone the target and run **sparq's** `dispatch-plan` → `route-resolve` → `gui_carve_out()`. I confirmed that (`dispatch.yml` "Clone protected public targets", then `load_dispatch(target / "scripts" / "dispatch-plan.py")`).

But CLAIM does not consume that chain on trust. `dispatch-claim.dispatch()` loads **registry-owned** `policy-resolve.py` from its own `script_dir` (`dispatch-claim.py:3790`), and `_route_matches()` re-resolves the route and demands **exact equality**:

```python
resolved = policy_module.resolve(repo, item["labels"], policy_doc, routing_doc)
expected = {"model_chain": item["model_chain"], "agent": item["agent"], "escalate": item["escalate"]}
if any(resolved[key] != value for key, value in expected.items()):
    raise DispatchError(f"plan route no longer matches protected routing for {repo}#{item['number']}")
```

`policy-resolve.resolve()` returns `role_routes[role]` verbatim. It has no carve-out — `grep -c gui scripts/policy-resolve.py` on registry `master` (which already contains merged #715) returns **0**. Measured, driving both resolvers against **this PR's** `routing.toml`:

| labels | PLAN (sparq `route-resolve`) | CLAIM (registry `policy-resolve`) | |
|---|---|---|---|
| `area:gui, role:impl` | `["sol","opus5"]` | `["opus5","sol"]` | **MISMATCH** |
| `area:gui, role:perf` | `["sol","opus5"]` | `["opus5","sol"]` | **MISMATCH** |
| `area:site, role:impl` | `["opus5","sol"]` | `["opus5","sol"]` | match |
| `area:gui, role:research` | `["opus5"]` | `["opus5"]` | match |
| `area:sparq-core, role:impl` | `["opus5","sol"]` | `["opus5","sol"]` | match |

The mismatch lands in the per-item handler at `dispatch-claim.py:4140`:

```python
except DispatchError as exc:
    defer_reasons["route-policy-failed"] += 1
    print(f"defer {repo}#{number}: trust/route/policy resolution failed ({exc}); skipped")
    continue
```

**Concrete failure scenario.** #3367 (`area:gui`, `role:impl`, `status:ready`) plans `["sol","opus5"]`, CLAIM re-derives `["opus5","sol"]`, `_route_matches` raises, the item is counted `route-policy-failed` and skipped. The comparison is a pure function of labels and the routing table, so it produces the identical result on every tick, indefinitely. Blast radius is precisely the set the carve-out targets: **34 of 35** — the 33 `role:impl` plus the 1 `role:perf`. #3757 (`role:research`) is unaffected, exactly because the carve-out correctly declines to touch it.

That is a regression against the current head: at `10845ede7` those 34 issues dispatch (to the wrong model); at `c44bf00b0` they **do not dispatch at all**, and the failure surfaces as a generic `route-policy-failed` defer counter rather than as anything naming GUI or routing preference.

It is the same shape as round 1 — a rule installed one layer above where the decision is finally made — displaced outward by one hop. The tests cannot see it because every assertion, mine included, drives **sparq's** resolver; nothing asserts that PLAN and CLAIM derive the same chain for a target repo. Note also that #715 is already **merged**, so there is no in-flight registry change carrying a counterpart.

### What would make it pass

1. **A companion registry PR** teaching `policy-resolve.resolve()` the same `area:gui` chain-order rule, so both sides derive identically. #715 already established the right idiom for this — a shared module imported rather than re-declared — and the selector plus the both-implementors condition should cross the repo boundary as data, not as a second hand-written copy.
2. **An agreement assertion**, in the shape #715's L3/L3b already uses: for a fixed label set, `route-resolve.resolve(labels, sparq_routing)[0]` must equal `policy-resolve.resolve(repo, labels, policy, sparq_routing)["model_chain"]`. Deleting the carve-out from either side must red a suite. Without it this recurs the next time either resolver gains a rule.
3. Land them together, or land the registry side first — merging this alone converts "GUI work runs on the wrong model" into "GUI work does not run".

If a table-only expression is preferred instead, note the cost you already documented: `match_labels` *is* honoured by both resolvers (registry `_validated_routing` returns it as a security route), which is exactly why it human-arms every GUI PR.

### Two smaller notes

* **N06 survived**: applying the carve-out to the security-route return is currently undetectable, because that chain is `["opus5"]` and the both-implementors condition already declines it. The explicit unmodified return is defence-in-depth with no red test — fine today, worth an assertion if a security route ever becomes cross-provider.
* **A correction to my round-1 comment**, so a wrong mechanism does not outlive it: I wrote that a role-less promoted issue "lands on `[defaults]`". `resolve()` does return defaults, but `ready-issues.has_role` drops the row before any plan row exists (registry issue #225), so the real outcome was silent *invisibility*, not dispatch-to-defaults. The measured core of that finding — 35 of 35 resolving opus5-first — is unaffected, and the fail-open write it rested on is now genuinely fixed.

VERDICT: fail

jeswr:
> 🤖 SPARQ agent

**Round-3 fix for the blocking finding, plus the companion registry PR.** New head:
`85cc64387b59e5ce056d029fece4e58905a44ec8` (was `c44bf00b0`).

The review is right and I reproduced it independently, executing **this branch's own resolver**
against the registry's:

```
$ python3 scripts/cross-resolver-agreement.py --target-root <sparq @ c44bf00b0>
::error::PLAN and CLAIM disagree on 6 of 22 routing cases
DISAGREE area:gui + role:impl   PLAN (['sol','opus5'], sparq-rust-impl, False)
                                CLAIM (['opus5','sol'], sparq-rust-impl, False)
DISAGREE area:gui + role:perf / area:gui alone / +priority / role:ci / role:site
```

Driven over the **real labels of every open `area:gui` issue** (34 now open — 32 `role:impl`,
1 `role:perf`, 1 `role:research`; one of the 35 has closed since the review):

| | sparq | registry | agree / N |
|---|---|---|---|
| A | `c44bf00b0` | master | **1 / 34** |
| C | main | master | 34 / 34 |
| **E** | **this head** | **registry #730** | **34 / 34** |
| F | this head | master | **1 / 34** |

## What changed, and why not a second copy in Python

CLAIM cannot import this repo's resolver — running target-authored **code** in the privileged claim
step is the escalation `_protected_routing` closed (registry #119). But it already trusts one
artefact of ours: **this routing table**, fetched from the protected default tip, which is where
`model_chain` comes from in the first place. So the rule crosses the boundary through that channel,
as data:

```toml
[[chain_preference]]
labels   = ["area:gui"]      # EXACT labels; "gui" in label would sweep area:guide / area:guidance
lead     = "sol"             # to the FRONT; every other rung keeps its relative order
requires = ["sol", "opus5"]  # only when the chain already contains BOTH
```

`lead` **must** be in `requires` (enforced at parse time on the registry side), which is what makes
*re-order* the only reachable effect — the rule can never inject sol into `role:research`
(`["opus5"]`, deliberately anthropic-side + escalating). Registry
[jeswr/agent-account-registry#730](https://github.com/jeswr/agent-account-registry/pull/730) adds the
shared **mechanism** (`scripts/chain_preference.py`, imported by both registry resolvers, hard-coding
no selector) and the cross-resolver agreement assertion that nothing in either repo had.

**PLAN behaviour on this branch is unchanged.** `gui_carve_out()` still decides here; every existing
assertion, and the whole round-1/round-2 mutation table, stands. What is new is that the table now
*declares* the same rule and `route-resolve --self-test` pins the two representations to each other
in **both** directions — selector, lead, `requires`, the lead-in-requires invariant, the closed field
set, catalog membership, plus a behavioural sweep over 9 label sets × 4 chains. Deleting the TOML
block, flipping its lead, widening it over `area:site`, or drifting `requires` each red a named
assertion; so does widening the Python constants.

## Your N06 — closed rather than recorded

You noted that applying the carve-out to the security return was undetectable, because that chain is
`["opus5"]` and the both-implementors condition declines it anyway, and deferred it to "if a security
route ever becomes cross-provider". My own mutation run hit the identical survivor on the registry
side. Both are now asserted on a fixture whose soundness chain **is** cross-provider, each paired
with a row proving the same table still re-orders the role branch — so the checks are an exemption,
not a dead fixture. Mutating either resolver's security return now reds a named test (S09 / R04).

## Merge ordering — registry #730 FIRST

Row F above: merging this branch while #730 is unmerged leaves CLAIM without the mechanism and
diverges on all 33. Row D of #730's body: merging #730 first is a **strict no-op** against sparq
`main` (34/34 agree), because the declaration lives here, not there. So: **#730, then this.**

## Test results

`route-resolve --self-test`, `routing-validate` (+self-test), `triage --self-test`,
`retriage --self-test`, `test_no_deprecated_models.py` (56 tests), `test_readiness_visibility.py`
(34 tests) — all pass. Registry side: 37 enrolled self-tests, all pass except `worker-live.sh`, the
known #711/#716 flake (measured on pristine `origin/master`: rc=1 then rc=0 on the unmodified tree);
`actionlint` 1.7.12 clean.

Mutation, real run, both trees, live-tree mutants restored from a byte snapshot: **24 mutants, 24
died, 0 survived**, restore check IDENTICAL. Honest recount: the **first** pass was 23 mutants with
**1 survivor** (the security-exemption one above) — the table is 24/24 only after fixing it, not
before.

Not armed — the orchestrator arms centrally.

jeswr:
> 🤖 SPARQ agent — independent review (Opus 5), reviewer-only: no push, arm, merge or relabel.

Reviewed head **`85cc64387b59e5ce056d029fece4e58905a44ec8`** (companion commit
`fix(routing): DECLARE the area:gui carve-out in the table, where CLAIM can read it`, superseding
`c44bf00b0`). `gate` on that exact SHA via `commits/85cc6438.../check-runs --paginate`, newest run
per name: **`gate` completed / success** (run `89833329539`), and **`Routing contract self-tests`
completed / success**. Zero `failure`/`cancelled`/`timed_out` legs; the remaining names are
`skipped` (selection-gated) or `success`. Reviewed together with registry
**#730 @ `dd05b41bcc21855f2b46a2a1e9ef4917c23eeb44`** — full analysis of the shared mechanism is in
my comment there.

## The layer question, which is what failed twice

Round 1 put the rule in a table nothing read at plan time; round 2 put it in PLAN while CLAIM
re-derived the route from registry-owned `policy-resolve.py`. This head puts the *selector* where
both deciders can read it — the protected routing TOML — and I confirmed the binding layer is now
actually reached, over the **real labels of the 34 open `area:gui` issues** (32 `role:impl`,
1 `role:perf`, 1 `role:research`), driving each row's actual PLAN module + table against its actual
CLAIM module:

| row | sparq | registry | PLAN≡CLAIM |
|---|---|---|---|
| A | `c44bf00b0` | master | **1 / 34** — the live defect, reproduced |
| **F** | **`85cc64387`** | **master** | **1 / 34** — sparq-first breaks dispatch |
| **E** | **`85cc64387`** | **#730** | **34 / 34** — the shipped pair |
| D | main | #730 | 34 / 34, and outcome-**identical** to main+master |

So **registry #730 must land first**; this PR alone is row F. That ordering is the one operational
condition on this verdict.

The carve-out is also live end-to-end rather than inert: `area:gui + role:impl` →
`["sol","opus5"]`, `area:gui` alone → `["sol","opus5"]`, while `area:site*` / `surface:frontend` /
`dashboard` / `area:guide` stay `["opus5","sol"]`. Cross-repo agreement via
`cross-resolver-agreement.py --target-root` against this checkout: **0 disagreements of 22**.

## Non-vacuity — 6/6 mutants die, plus R04

All applied to a live copy of this tree; all six die in `scripts/route-resolve.py --self-test`,
which is exactly what the green `Routing contract self-tests` job runs:

| mutant | result |
|---|---|
| delete the `[[chain_preference]]` block from `routing.toml` | **DIES** (and the agreement harness reds too) |
| widen `GUI_CARVE_OUT_LABELS` to include `area:site` | **DIES** |
| widen the TOML `labels` to include `area:site` | **DIES** |
| drift the declared `requires` to `["sol"]` | **DIES** (invisible to the 22-row agreement matrix — the field-equality pin is what catches it) |
| drop `gui_carve_out` from the role branch | **DIES** |
| drop it from the `[defaults]` branch | **DIES** |

The two-representation pin holds in **both** directions, which is the property that matters: neither
the TOML nor the Python constants can be edited alone. The behavioural-equivalence loop
(`_declared_rule` vs `gui_carve_out` over 9 label sets × 4 chains) is a genuine cross-check and not
just field equality.

**R04 / S09, confirmed independently** — the one row I was asked to re-derive rather than read.
Applying the carve-out to the security branch here **DIES**:
`FAIL a CROSS-PROVIDER security route is STILL returned unmodified under area:gui: ['sol','opus5']
(want ['opus5','sol'])`. It dies on all three sides of the pair (this file, registry
`policy-resolve.py`, registry `route-resolve.py`), so the exemption is the **route class** and not an
accident of the shipped soundness chain being single-model. The paired "…and the same table still
re-orders the role branch" row confirms these are exemptions, not dead fixtures. Adversarially: a
`chain_preference` whose selector *is* a security keyword still lands on the security route (which
wins first and is exempt), and one selecting `role:research` declines on `requires ⊄ ["opus5"]`, so
neither the escalating single-provider route nor a soundness chain is reachable.

## The outage fix, and what it does not cover

`fable`/`opus` are gone from `[models]` **and** from every chain; sparq's catalog membership is now
identical to the registry's (`haiku, sonnet, opus5, terra, sol, luna`), so the
`MODELS: opus5,fable,opus` → "empty or missing harness" fall-through has no unresolvable rung left to
hit. `validate_routing` turns that from an edit into an invariant, fail-closed, with each clause
shown to reject something.

One honest scoping note, since it bears on the separate skip-unresolvable-rungs change: new rule (2)
refuses a chain naming a model absent from **sparq's own** `[models]` catalog. That is not the same
check as *registry*-catalog drift — which is what actually caused the outage — so it does not
substitute for the skip-vs-fail work. The two are complementary and do not conflict.

Smaller notes, none blocking:
- `role = "gui"` survives in the table but is still unreachable (`triage.py` / `bd-to-issues` never
  emit `role:gui`, and `_role()` prefers an explicit `role:*`). Harmless residue now that the
  carve-out no longer depends on it — the assertions on it are testing a route no issue reaches.
- `review`/`soundness`/`research` are now single-rung `["opus5"]`. With `escalate = true` an opus5
  capacity outage flips those to `needs:user` rather than degrading — deliberate per the directive,
  but it is a real behaviour change in an outage.
- `triage-issue.yml` is genuinely fail-closed now: `set -euo pipefail`, one all-or-nothing
  `gh issue edit`, no `|| true` on the add, and `ensure_label` restricted to the closed
  `role:`/`status:`/`needs:` set so it cannot mint an arbitrary label. An unexpected label is not
  created *and* still reaches `--add-label`, so the edit fails the run — the correct direction.
  Unquoted `$add` word-splitting is safe for the space-free labels `triage.py` emits.

Everything I ran green at head: `route-resolve --self-test`, `routing-validate`, `triage --self-test`,
`retriage --self-test`, `dispatch-plan --self-test`, `test_no_deprecated_models.py`,
`test_readiness_visibility.py`.

VERDICT: pass

jeswr:
> 🤖 SPARQ agent

## Rebased onto current `main` — content-preserving, delta byte-verified

This PR went `CONFLICTING/DIRTY` while it waited for its registry companion. Rebased and
force-pushed to the same branch.

| | |
|---|---|
| reviewed head (pre-rebase) | `85cc64387b59e5ce056d029fece4e58905a44ec8` |
| **new head** | **`914ce73bd0c4f369344d8b8403f2211e82dd4a1c`** |
| old merge-base | `b5548a0d4fe086f3ddb551fbff2fd4e2e8da48de` |
| new base | `bd8731bd303dd2488ca0516f846a6981c95b2f9f` |
| mergeability | `CONFLICTING/DIRTY` → **`MERGEABLE`** |

### The rebase did not change this PR's content

Only **2** of the 11 touched files conflicted, and both conflicts were purely *additive* — this
PR and `main` each appended to the same `paths:` list and the same `run:` block. Resolution was a
**union**; nothing from either side was dropped, and no `git checkout <ref> -- .` was used.

**9 of 11 files are byte-identical blobs to the reviewed head** (same git object id), including
every file that carries the routing semantics — `orchestration/routing.toml`,
`scripts/route-resolve.py`, `scripts/triage.py`, `scripts/retriage.py`,
`scripts/dispatch-plan.py`, `scripts/routing-validate.py`,
`scripts/tests/test_no_deprecated_models.py`, `.claude/workflows/fable-architect-drain.js`,
`.github/workflows/triage-issue.yml`.

For the 2 merged files the **diff-of-diffs is empty on the delta lines** — `diff(old-base →
reviewed-head)` and `diff(new-base → new-head)` contain the *identical* set of `+`/`-` lines;
only the surrounding context moved:

```
.claude/settings.json                          DELTA LINES IDENTICAL (2 lines)
.github/workflows/routing-self-tests.yml       DELTA LINES IDENTICAL (37 lines)
```

Whole-PR totals: reviewed delta `1741` +/- lines, rebased delta `1741` +/- lines, identical
changed-file set (11 files, none added, none dropped). `main`'s own additions to those two files
are all present and its removals all honoured, so nothing was clobbered in the other direction
either.

### The headline property, re-established empirically at the new head

Resolved over the **real labels of the live open issues**, driving *both* resolvers —
PLAN = this branch's `scripts/route-resolve.py`, CLAIM = live registry `master`
(`4c0ededa14c352ab168fd356dc7ff9c44736acea`, i.e. registry #730) `scripts/policy-resolve.py`:

| bucket | n | PLAN chain head | CLAIM chain head | PLAN/CLAIM agreement |
|---|---|---|---|---|
| `area:gui` | 34 | `sol` × 33, `opus5` × 1 | `sol` × 33, `opus5` × 1 | **34/34** |
| `area:site` | 33 | `opus5` × 31, `sol` × 2 | same | 33/33 |
| `area:site-specs` | 10 | `opus5` × 10 | same | 10/10 |
| `area:site-papers` | 15 | `opus5` × 15 | same | 15/15 |

- The single `area:gui` non-`sol` issue is **#3757** (`role:research`), which resolves to the
  single-model chain `['opus5']`. The declaration's `requires = ["sol", "opus5"]`
  both-implementors condition therefore correctly **declines** to re-order it — the same issue
  for the same reason as at the reviewed head.
- The 2 `sol`-first issues in the `area:site` bucket (**#3277**, **#2656**) both *also* carry
  `area:gui`, so `sol`-first there is the carve-out applying as specified. Every `site*` issue
  that does not carry `area:gui` is `opus5`-first.
- The `area:gui` bucket was 35 open at review time and is 34 now only because `area:gui` issues
  closed in the interim (#2737, #2770 and #2748 all closed today). The property is unchanged:
  every open `area:gui` issue is `sol`-first except the one `role:research` issue.

### `fable` / `opus` are gone from the catalog and from every chain

`[models]` is exactly `['haiku', 'luna', 'opus5', 'sol', 'sonnet', 'terra']`, identical to the
registry's catalog. An exhaustive walk of every string list in `orchestration/routing.toml` finds
**no** `fable` or `opus` entry in any `model_chain`, `requires` or `defaults`:

```
defaults.model_chain      = ['opus5', 'sol']
route.0..9 model_chain    = ['opus5'] | ['opus5','sol'] | ['sol','opus5'] | ['sol','terra','opus5']
chain_preference.0        = labels ['area:gui'], lead 'sol', requires ['sol','opus5']
DEPRECATED-IN-CHAIN HITS: []
```

The only surviving `fable`/`opus` occurrences in the file are comments recording the deprecation.

### Suites at the new head

All green at `914ce73bd`: `routing-validate --self-test`, `routing-validate`,
`route-resolve --self-test`, **`test_no_deprecated_models.py` (56 tests, OK)**,
`dispatch-plan --self-test`, `test_readiness_visibility.py` (34 tests),
`batch-merge --self-test`, `ready-issues --self-test`, and `main`'s newly-co-resident
`pr-area-labels.py --self-test` + `test_pr_area_labels.py` (65 tests) — the pair that shared the
conflicted `run:` block, so the union resolution is confirmed not to have broken either side's
YAML seam. CI's own `Routing contract self-tests` leg is `success` on this head.

The registry's enrolled `scripts/cross-resolver-agreement.py` passes against this branch both
ways — `--target-root` (this branch's own resolver as PLAN) and `--routing-file` — and its
`--self-test` passes. *For the record:* that harness's `AGREEMENT_MATRIX` is **22** rows, so its
result is **22/22**; the **34/34** figure is the real-labels agreement over the live `area:gui`
backlog in the table above. Stating both separately rather than conflating them.

### The measurement is not vacuous

Deleting the `[[chain_preference]]` block from the rebased table (mutation of the live tree) and
re-running reproduces exactly the outage class this PR exists to prevent: PLAN still returns
`['sol','opus5']` (the carve-out lives in target *code*) while CLAIM returns `['opus5','sol']`
(it reads the table as *data*) — **33 of 34** `area:gui` issues disagree, i.e. every one of them
would defer `route-policy-failed` on every tick. Restoring the block returns
`orchestration/routing.toml` to the byte-identical reviewed blob and agreement to 34/34.

Not armed, not merged — the maintainer arms centrally against the new head SHA.

PR #4193, opened 2026-07-26
docs(research): knowledge management — documents as projections of a queryable graph

jeswr:
> 🤖 **SPARQ agent**

Design record answering a maintainer request: put as much of this codebase as possible into a **queryable
knowledge graph** rather than disparate unmaintained documents — reducing duplication and long documents,
letting agents assemble a short task-relevant context, and never hand-writing what can be generated from
benchmarks. Long-term extension of #4187 (`research/agent-context-sharing.md`).

**Design record only — no behaviour change.** Rollout is staged and measurement-gated in §10.

## The measurement that makes the case

Corpus: **640 markdown files / 140,807 lines**, with `research/` alone **275 files / 85,033 lines** (60%).

- The single fact *"the external cryptographer ZK audit (`sq-qhy4`) is pending, so no ZK security property
  may be relied on"* is restated in **146 files**. It is a P0 gate. When it changes — one event, one day —
  **146 files become wrong simultaneously**, and nothing tells us which.
- `opt-in` appears in **299 files**.
- **128 of 275** `research/` records were last touched a month or more ago, in a repo landing ~85
  commits/day. About half the design corpus describes a codebase that no longer exists, unmarked.

## The reframe

**Facts live in the graph; documents are generated views of it.** Two corollaries carry the design:

- **Generation is one-way — graph → document.** We do *not* parse prose into triples: that is lossy,
  creates a two-way sync problem, and silently manufactures false facts. It is how most doc-KG efforts die.
- **Prose that duplicates a graph fact is deleted, not synchronised.** The 146 restatements become 146
  transclusions of one node.

## The taxonomy that makes it tractable

Splitting by **derivability** is what keeps scope finite:

| Class | Where it lives | Hand-written? |
|---|---|---|
| **Derived** — crate inventory, features, public API, deps, benchmarks, conformance, coverage | generated only | **never**, CI-enforced |
| **Asserted** — decisions, verdicts, gates, invariants, ownership, supersession | triples + provenance | structured input only |
| **Narrative** — surveys, design arguments, proof sketches | prose, addressable + linked | yes — this is what prose is *for* |

**Do not graph the narrative.** Attempting to triple-ify reasoning is the classic failure; keep arguments
as prose, make each load-bearing claim addressable, and link it to the nodes it concerns.

## Why it won't just re-accrete

The durable mechanism is a **drift gate**, generalising the repo's existing `check-no-perf-numbers.py` from
[...]

PR #4220, opened 2026-07-26
fix(orchestration): bridge head-bound VERDICT comments to the review:pass arming label

jeswr:
> 🤖 SPARQ agent

Maintainer observation (2026-07-26): *"there seem to be a number of green, ready PRs that are not in the merge queue."*

## Measurement first

Census of **all 91 open PRs** (explicit GraphQL cursor paging cross-checked against `totalCount`; explicit REST check-run paging cross-checked against `total_count` — up to **1181** check-runs on a single head; `gate` resolved **newest-run-per-name** so a cancelled twin cannot read as the live result; merge-queue membership read via `mergeQueueEntry`, since `autoMergeRequest` is `null` for a queued PR).

| slice | count |
|---|---|
| open PRs | 91 |
| non-draft | 25 |
| armed (`autoMergeRequest`) | 2 |
| in the merge queue (`mergeQueueEntry`) | 1 |
| **green `gate` + `MERGEABLE` + non-draft + unarmed + unqueued** | **12** |
| open PRs carrying `review:pass` | **3** — and all 3 are armed or queued |

The correlation is total: **every** PR with `review:pass` is moving; **no** PR in the stuck set has it.

## Root cause — the write path and the read predicate were joined only by a human

The arming chain admits on exactly one predicate, the **`review:pass` LABEL**:

* `scripts/auto-arm.py` — `if REVIEW_LABEL not in pr.labels: return "review-label-removed"`
* `scripts/rearm-sweeper.py` — its candidate query is literally `gh pr list --label review:pass`, so an unlabelled PR is never even *listed*
* `scripts/batch-merge.py` — same attestation

But **nothing in this repository ever ADDS that label.** Grepping every `.py`/`.yml`/`.sh`/`.js` under `scripts/` and `.github/`: three files READ it, `merge-queue-feedback.yml` REMOVES it, and there is **no writer**. Confirmed against the live timeline API — every `labeled: review:pass` event on every PR was applied by `sparq-orchestrator[bot]` or `jeswr`, i.e. an **out-of-repo orchestrator session**.

Meanwhile the review lane's actual output artefact is a **line-anchored `VERDICT: pass` comment**, and the standing review brief explicitly tells review agents *"Do NOT arm, merge, push, **or change labels**."*

So the comment to label hop existed only inside a live orchestrator session. A review that lands while nobody is watching is simply lost, and the PR sits green, `CLEAN` and unarmed indefinitely. This is the same write-path-vs-read-predicate mismatch class recorded on registry#700.

**The arming machinery itself is healthy** — refuted, not assumed. `auto-arm.yml` has a cron (`4,14,24,…`) and its 16:17 scheduled run logged `considered=5 armed=2 arm-failures=0`; `rearm-sweeper.yml` runs every 10 minutes and is green. They are not broken; they are starved of the label.

### A live instance, caught by this PR's dry-run

While validating, the bridge found **#4192**: head `94ae0b0b62c48f41c8bbe934da1d7bbeda3861e4`, a thorough head-bound `VERDICT: pass` from a MEMBER posted at 17:43:40Z, **no `review:pass` label, `autoMergeRequest: null`**. It would have sat there until a human noticed. The bridge promotes it.

### The second hole

Of the 12 stuck PRs, **6** carry a head-bound `VERDICT: fail` (correctly held — this PR does **not** try to arm them). The other **6** — #3426, #3798, #4163, #4170, #4184, #4193 — are green, `CLEAN`, `MERGEABLE`, non-draft and carry **no `review:*` label at all** (only `area:*`). Three have never been reviewed; three carry a verdict bound to a *superseded* head. They are invisible to **every** automated lane: auto-arm needs `review:pass`, the fast-fix ring only fires on **failing** CI, the dispatchers work off `review:needs`. Nothing will ever look at them again.

## The fix

`scripts/verdict-bridge.py` + a **schedule-only** `.github/workflows/verdict-bridge.yml` (cron `1,11,21,…`, three minutes ahead of auto-arm's `4,14,24,…`). It **grants no new arming authority — it writes labels only**; `auto-arm` still performs the arm with its own `expectedHeadOid` CAS and its own security-label exclusions.

**Head-binding is the load-bearing property.** A verdict is honoured as a **pass** only when *all* hold:

1. the comment's **last non-blank line** is exactly `VERDICT: pass`/`VERDICT: fail` — a substring search would match the *instruction* that tells reviewers to write it (a false-pass this project has already been bitten by);
2. the body **contains the current head as a standalone full 40-hex SHA** — a short prefix does not bind, nor does a SHA glued into a longer hex run. This is deliberately an **occurrence** test: a head SHA quoted inside a commit URL or a `diff --git` line binds too. Binding only *pairs* a comment with a head; what makes it a **verdict** is the trusted author's trailing `VERDICT:` line, and that is what the strictness above and the provenance check below attest;
3. the comment's `author_association` is OWNER / MEMBER / COLLABORATOR, read from the **API**, never from body content.

Among head-bound verdicts the **latest by `created_at`** wins, so a retraction posted after a pass at the same head defeats it. Ordering deliberately **ignores `updated_at`**: ordering by an edit timestamp would let someone edit an *older* pass to jump ahead of a *newer* fail. `created_at` is immutable.

### Strict-and-discard is fail-closed alone and fail-OPEN in composition

Round-1 review measured this, and it is fixed here. A strictness rule that simply **discards** a malformed line is not fail-closed once a valid pass already exists at the same head:

```
valid   VERDICT: pass                      (t=2026-01-01, MEMBER, binds head)
later   VERDICT: fail (retracting my pass) (t=2026-02-01, MEMBER, binds head)
later   VERDICT: FAIL                      (t=2026-02-01, MEMBER, binds head)
```

Both retractions used to parse to `None`, never entered the candidate list, and so left the **superseded pass** as `found[-1]` — the PR promoted and armed. Since this PR's own premise is that review agents may not touch labels, **a comment is a reviewer's only retraction channel**, so this sits on the load-bearing path.

A last non-blank line that is verdict-**shaped** (`^VERDICT\b`, any case, optionally bold-wrapped) but does not parse strictly is now **`AMBIGUOUS`**, not absent. An ambiguous verdict **never promotes**, **defeats an earlier pass at the same head**, and **removes an existing `review:pass`** — guessing the polarity of an unreadable retraction is worse than refusing to arm. Two properties are preserved deliberately:

* a **mention** (blockquoted `> `, bulleted, back-ticked, fenced) is *not* verdict-shaped, so the instruction text that tells reviewers what to write still influences nothing;
* an ambiguous line from an **untrusted** author is still discarded outright, so a drive-by commenter cannot deny arming to every PR in the repo.

### … and the same hole existed in the CHANNEL dimension

Round-1 review re-derived its promote-set by reading issue comments **and** `/pulls/<n>/reviews` bodies. The bridge read only issue comments — so a reviewer who posts a pass as a comment and then withdraws it in a **PR review body** left the superseded pass standing. Same defect, different axis.

PR review bodies are now read on the **same GraphQL page as the PR itself** (`reviews(last:30)` added to `PR_LIST_QUERY`), so this costs **zero extra API calls** on a sweep that already pages ~600 requests. They are **withhold-only**: a `_channel="review"` entry can contribute a `fail` or an `AMBIGUOUS` suppression but **never a pass**. Reading a review body as a pass would extend arming authority into a channel the standing review brief does not mandate — so including them can only ever *shrink* the promote-set, never grow it. An unsubmitted (`PENDING`) review has no `submittedAt` and is not evidence at all.

Actions: `promote` (add `review:pass`) · `retract` (a head-bound **fail or ambiguous** line removes an existing `review:pass`) · `flag`/`unflag` (`review:unreviewed`, purely informational — **consumed by no arming predicate anywhere**, so it can never block or cause a merge). **Absence of evidence never retracts**, so a `review:pass` an orchestrator applied by hand is never fought.

## Guard to red-test table — round 1 (the pure predicate + the YAML seam)

33 mutants applied to the **live** worktree, suite re-run each time, then reverted. **33 killed, 0 survived.** A first round left `order-by-comment-id` alive; `test_recency_is_created_at_not_comment_id` was added to kill it. Round-1 review independently re-ran ~30 of these rows and all died; 10 of them are re-run verbatim in the round-2 harness below (rows `O1`–`O10`) to prove the ambiguity change did not weaken them.

| mutation | result | named test that went red |
|---|---|---|
| head-binding: drop `binds_head` entirely | KILLED | `test_pass_bound_to_a_superseded_head_never_promotes`, `test_absence_of_evidence_never_retracts` |
| head-binding: 12-char prefix instead of full SHA | KILLED | `test_binds_head_rejects_a_sha_glued_into_a_longer_hex_run` |
| line-anchor: substring match instead of trailing line | KILLED | `test_verdict_must_be_the_final_line`, `test_verdict_line_must_not_carry_trailing_content` |
| provenance: drop the trusted-association check | KILLED | `test_untrusted_association_is_not_a_reviewer` |
| provenance: admit `CONTRIBUTOR`/`NONE` | KILLED | `test_untrusted_association_is_not_a_reviewer` |
| retraction: order by **mutable** `updated_at` | KILLED | `test_editing_an_older_pass_cannot_reorder_it_ahead_of_a_newer_fail` |
| retraction: order by comment id | KILLED | `test_recency_is_created_at_not_comment_id` |
| retraction: take the FIRST head-bound verdict | KILLED | `test_latest_by_created_at_wins_regardless_of_input_order` (+2) |
| holds: drop the hold check before promoting | KILLED | `test_hold_labels_are_fail_closed` |
| holds: drop the `needs:*` namespace | KILLED | `test_hold_labels_are_fail_closed` |
| holds: drop `trust-surface` from `HOLD_LABELS` | KILLED | `test_bridge_holds_cover_every_auto_arm_exclusion` |
| holds: drop `review:changes` from `HOLD_LABELS` | KILLED | `test_bridge_holds_cover_every_auto_arm_exclusion` (+1) |
| attest drafts | KILLED | `test_draft_is_never_attested` |
| drop the open-state guard | KILLED | `test_closed_pr_is_never_attested` |
| **retract when NO verdict is found** | KILLED | `test_absence_of_evidence_never_retracts` |
| flag drafts / red / conflicting PRs too | KILLED | `test_flag_requires_green_and_mergeable_and_non_draft` |
| make the informational label a hold (`review:needs`) | KILLED | `test_the_informational_label_is_not_a_hold_in_any_policy` (+2) |
| rename the attestation label (predicate drift) | KILLED | `test_the_attestation_label_matches_the_arming_predicate` |
| **YAML** job disabled with `if: false` | KILLED | `test_the_job_is_not_disabled_by_a_job_level_condition` |
| **YAML** sweep step disabled with `if: false` | KILLED | `test_no_step_is_disabled_or_swallowed` |
| **YAML** sweep step swallowed with `continue-on-error` | KILLED | `test_no_step_is_disabled_or_swallowed` |
| **YAML** sweep step DELETED | KILLED | `test_the_live_run_step_invokes_the_bridge_with_a_repo` (+1) |
| **YAML** self-test step DELETED | KILLED | `test_the_self_test_step_exists_and_runs_the_policy_self_test` |
| **YAML** label-reify step DELETED | KILLED | `test_the_informational_label_is_reified_before_use` |
| **YAML** wrong input: `sparse-checkout` drops `gh_retry.py` | KILLED | `test_the_checkout_materializes_every_script_the_run_imports` |
| **YAML** wrong input: checkout `ref` = PR head | KILLED | `test_the_checkout_materializes_every_script_the_run_imports` |
| **YAML** wrong input: cron collides with auto-arm | KILLED | `test_the_cron_fires_before_each_auto_arm_sweep` |
| **YAML** wrong input: cron fires *after* auto-arm | KILLED | `test_the_cron_fires_before_each_auto_arm_sweep` |
| **YAML** `schedule:` trigger DELETED | KILLED | `test_it_never_triggers_on_a_pull_request_head` (+1) |
| **YAML** privilege creep: `contents: write` | KILLED | `test_it_has_no_contents_write_authority` |
| **YAML** privilege creep: adds a `pull_request` trigger | KILLED | `test_it_never_triggers_on_a_pull_request_head` |
| **YAML** un-pin the checkout action to a tag | KILLED | `test_every_action_is_sha_pinned` |
| **YAML** `cancel-in-progress: true` | KILLED | `test_concurrency_serialises_the_sweep` |

The workflow document is parsed **structurally with PyYAML** — substring and `count(...) == N` assertions do not catch `if: false` or a deleted step, and every uncaught mutant this project has measured lived at the YAML seam.

`TestCrossPolicyConsistency` imports the **live** `auto-arm.py` / `rearm-sweeper.py`, so the bridge cannot drift into attesting a PR those policies would refuse, and the informational label cannot become a hold that deadlocks arming.

## Guard to red-test table — round 2 (the DRIVER: decision → label write path)

Round-1 review measured that the suite was exhaustive on `decide()` and on the YAML and **skipped the layer in between** — `VerdictBridge`, the layer that actually mints the label. It ran 54 mutants there: 41 killed, **13 survived**. Worst case, quoted: *"setting `"flag": (REVIEW_ATTESTATION, "")` in the dispatch table … which would put `review:pass` on every green never-reviewed PR, passes the entire suite."*

`TestWritePathDispatch` / `TestWritePathSafetyRails` / `TestPagedReads` / `TestGateResolution` drive the real `VerdictBridge` against a hermetic fake `gh` that serves the three reads and **records every write**. Every one of the 13 survivors now reds. Harness output, verbatim — **38 mutants applied to the live worktree, 38 killed, 0 survived, 0 anchor-misses** (8 Finding-1 rows, 20 driver rows, 10 round-1 regression rows re-run to prove nothing weakened):

```
baseline: GREEN

KILLED      | F1 revert: shaped-but-unparseable reads as NO verdict (the shipped bug)
              -> AssertionError: VERDICT: FAIL
KILLED      | F1: an ambiguous line does not retract an existing review:pass
              -> AssertionError:
KILLED      | F1: shape regex unanchored (a MENTION becomes ambiguous)
              -> AssertionError: **VERDICT: Fail**
KILLED      | F1: ambiguity ignores the trust boundary (any commenter can suppress)
              -> AssertionError
KILLED      | F1: a review-body retraction is dropped (channel-shaped composition hole)
              -> FAIL: test_a_review_body_retraction_reaches_the_write_path
KILLED      | F1: a review BODY may grant a pass (arming authority creep)
              -> AssertionError
KILLED      | F1: an UNSUBMITTED (pending) review body counts as evidence
              -> FAIL: test_a_review_body_retraction_reaches_the_write_path
KILLED      | F1: the reviews connection is dropped from the GraphQL query
              -> FAIL: test_the_pr_list_query_actually_requests_the_review_bodies
KILLED      | D1  "flag": (REVIEW_ATTESTATION, "")  [attests every green never-reviewed PR]
              -> FAIL: test_a_review_body_pass_never_reaches_the_write_path_as_an_attestation; FAIL: test_flag_writes_ONLY_the_informational_label
KILLED      | D2  promote/retract add-remove swapped  [a head-bound FAIL adds review:pass]
              -> ERROR: test_a_review_body_retraction_reaches_the_write_path; ERROR: test_an_unreadable_retraction_also_removes_the_attestation; FAIL: test_a_PENDING_review_is_not_evidence
KILLED      | D3  "retract": ("", "")  [a head-bound fail never removes the attestation]
              -> ERROR: test_a_review_body_retraction_reaches_the_write_path; ERROR: test_an_unreadable_retraction_also_removes_the_attestation; FAIL: test_retract_removes_the_attestation_and_adds_nothing
KILLED      | D4  `if self.dry_run: continue` deleted  [--dry-run writes]
              -> FAIL: test_dry_run_performs_no_write_at_all
KILLED      | D5  base-branch (stacked-PR) filter deleted
              -> FAIL: test_a_stacked_pr_is_never_touched
KILLED      | D6  gate resolution: newest-run-per-name -> OLDEST
              -> FAIL: test_an_incomplete_newest_run_reads_as_NO_gate_not_as_the_older_result; FAIL: test_the_newest_gate_run_wins_over_an_older_success
KILLED      | D7  gate resolution: an INCOMPLETE run is admitted
              -> FAIL: test_an_incomplete_newest_run_reads_as_NO_gate_not_as_the_older_result
KILLED      | D8  check_runs truncated to page 1
              -> ERROR: test_check_runs_are_paged_and_cross_checked_against_total_count; FAIL: test_check_runs_CREATED_mid_pagination_do_not_red_the_sweep
KILLED      | D9  comments truncated to page 1
              -> FAIL: test_comments_are_paged_so_a_verdict_beyond_page_one_is_seen
KILLED      | D10 per-run write cap removed
              -> FAIL: test_the_per_run_write_cap_is_enforced
KILLED      | D11 run_bridge returns 0 on hard errors
              -> FAIL: test_run_bridge_reds_the_workflow_on_a_hard_error
KILLED      | D12 label-page overflow fails OPEN
              -> FAIL: test_a_label_set_that_exceeds_one_page_fails_CLOSED
KILLED      | D13 YAML: DRY_RUN hard-coded to --dry-run  [permanent no-op]
              -> FAIL: test_the_scheduled_sweep_is_never_a_hard_coded_dry_run
KILLED      | D13b YAML: --dry-run hard-coded into the sweep argv
              -> FAIL: test_the_scheduled_sweep_is_never_a_hard_coded_dry_run
KILLED      | D14 list_open: totalCount cross-check dropped
              -> FAIL: test_open_prs_are_paged_and_cross_checked_against_total_count
KILLED      | D15 check_runs: truncation cross-check dropped
              -> FAIL: test_a_TRUNCATED_check_run_read_raises
KILLED      | D17 check_runs: terminate on a STALE total_count (undercount truncates growth)
              -> FAIL: test_check_runs_CREATED_mid_pagination_do_not_red_the_sweep
KILLED      | D18 check_runs: de-duplication dropped (double-counting masks truncation)
              -> FAIL: test_a_duplicated_check_run_across_pages_is_counted_once
KILLED      | D19 check_runs: page bound removed (unbounded loop)
              -> FAIL: test_the_check_run_read_is_page_bounded
KILLED      | D16 a failed label edit is swallowed (errors not counted)
              -> FAIL: test_a_failed_label_edit_is_counted_as_an_error
KILLED      | O1 line-anchor: substring search instead of the trailing line
              -> AssertionError
KILLED      | O2 head-binding: drop binds_head entirely
              -> AssertionError: stale sha must not promote
KILLED      | O3 provenance: drop the trusted-association check
              -> AssertionError
KILLED      | O4 retraction: take the FIRST head-bound verdict
              -> AssertionError: VERDICT: FAIL
KILLED      | O5 retraction: order by MUTABLE updated_at
              -> AssertionError: edit reorders
KILLED      | O6 retract when NO verdict is found
              -> AssertionError: Decision(action='retract', reason='no verdict')
KILLED      | O7 holds: drop the hold check before promoting
              -> AssertionError:
KILLED      | O8 YAML: job disabled with `if: false`
              -> FAIL: test_the_job_is_not_disabled_by_a_job_level_condition
KILLED      | O9 YAML: privilege creep — contents: write
              -> FAIL: test_it_has_no_contents_write_authority
KILLED      | O10 YAML: adds a pull_request trigger
              -> FAIL: test_it_never_triggers_on_a_pull_request_head

38 killed, 0 survived, 0 anchor-misses (38 mutants)
post-restore baseline: GREEN
```

Mapping from the review's 13 survivors to the named test that now reds:

| survivor (round-1 review) | now killed by |
|---|---|
| **`"flag": (REVIEW_ATTESTATION, "")`** | `test_flag_writes_ONLY_the_informational_label` |
| `promote`/`retract` add-remove swapped | `test_retract_removes_the_attestation_and_adds_nothing` (+2) |
| `"retract": ("", "")` | `test_retract_removes_the_attestation_and_adds_nothing` |
| `if self.dry_run: continue` deleted | `test_dry_run_performs_no_write_at_all` |
| base-branch (stacked-PR) filter deleted | `test_a_stacked_pr_is_never_touched` |
| newest-run-per-name → oldest | `test_the_newest_gate_run_wins_over_an_older_success` |
| an incomplete run is admitted | `test_an_incomplete_newest_run_reads_as_NO_gate_not_as_the_older_result` |
| `check_runs` truncated to page 1 | `test_check_runs_are_paged_and_cross_checked_against_total_count` |
| `comments` truncated to page 1 | `test_comments_are_paged_so_a_verdict_beyond_page_one_is_seen` |
| per-run write cap removed | `test_the_per_run_write_cap_is_enforced` |
| `run_bridge` returns 0 on hard errors | `test_run_bridge_reds_the_workflow_on_a_hard_error` |
| label-page overflow fails open | `test_a_label_set_that_exceeds_one_page_fails_CLOSED` |
| **`DRY_RUN: --dry-run` hard-coded in the YAML** | `test_the_scheduled_sweep_is_never_a_hard_coded_dry_run` |

The four `F1*` rows kill Finding 1 itself, including a full revert of the fix (`shaped-but-unparseable reads as NO verdict`), and four further `F1*` rows kill the channel variant (retraction dropped / a review body granting a pass / a PENDING review counting / the `reviews` connection dropped from the query).

The incomplete-run row needed a second pass: the obvious fixture (`status: in_progress`, `conclusion: null`) does **not** kill the mutant, because the mutated code returns `null` anyway. The test now asserts the actual invariant — a `queued`/`in_progress`/`waiting` run carrying the **previous attempt's** `conclusion` must still read as *no gate*.

Two rows' first drafts were **vacuous**, which is worth recording because both looked right:

* the **incomplete-run** row. The obvious fixture (`status: in_progress`, `conclusion: null`) does *not* kill `if chosen is None:` — the mutated code returns `null` anyway. It now asserts the real invariant: a `queued`/`in_progress`/`waiting` run carrying the **previous attempt's** `conclusion` must still read as *no gate*.
* the **PENDING-review** row, same shape: dropping the `submittedAt` guard makes `created_at` empty, which sorts first and loses regardless, so the run-level assertion passed either way. It now asserts that `review_withholdings` normalises a pending review to nothing at all.

## A third defect the dry-run caught — in shipped code

Re-running the live dry-run against the new head **surfaced a real bug**, and this is the measurement, verbatim:

```
[verdict-bridge] PR #4074: SKIP inspect-failed (47f066f0eaebd4cc69b7a7e769e8fecd53ee5344: paged 479 check-runs, total_count=473)
[verdict-bridge] complete: writes=0 errors=1
```

Check-runs were being **created between page reads**, so the list grew under pagination. Three defects, all shipped in round 1:

1. the cross-check was `len(runs) != total`. Reading **more** than the first page's `total_count` is benign — the risk the guard exists for is **truncation**, which reads *fewer*. The two-sided check turned a routine mid-sweep re-run into a skipped PR **and a red workflow**, every cycle. It is now one-sided (`< total`).
2. worse, the loop **terminated** on `len(runs) >= total_count`. A `total_count` read before a re-run started is an *undercount*, so the read stopped early exactly when the list was growing — hiding the **newest** `gate` run, which is the one that matters. It now terminates on a short page, as `comments()` already did.
3. a window that shifts under pagination **repeats** entries across page boundaries, so the read now de-duplicates by run id; double-counting would mask a real truncation.

`MAX_CHECK_RUN_PAGES` bounds the loop so a pathological response fails the read instead of spinning the sweep forever. Its test serves `MAX+5` pages and then stops, so removing the bound reds **cleanly** rather than hanging the suite — a test that hangs on a mutant times CI out instead of reporting a defect. Mutants `D15`, `D17`, `D18`, `D19` cover all four.

```
$ python3 scripts/tests/test_verdict_bridge.py
Ran 74 tests in 0.065s
OK
$ python3 scripts/verdict-bridge.py --self-test
verdict-bridge self-test: PASS
```

Both are wired into `docs-quality.yml`'s gating job, so a regression in either reds the `gate`.

## Live read-only dry-run (no writes)

The round-1 run, which caught the live instance the bridge exists for:

```
[verdict-bridge] 97 open PR(s); dry_run=True
[verdict-bridge] PR #3426: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #3798: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4163: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4170: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4184: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4192: PROMOTE — head-bound pass by jeswr at 94ae0b0b62c4
[verdict-bridge] PR #4193: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] complete: writes=0 errors=0
```

Exactly one promotion, and it reproduced the independently-measured census. Round-1 review re-derived the promote-set over all 100 open PRs from the prose (not by importing this code) and got `{#4192}` only — matching exactly.

Re-run at head `bf7e300fe84b2807fafb2c7af2ea02c9a3f54337`, after the fixes:

```
[verdict-bridge] 97 open PR(s); dry_run=True
[verdict-bridge] PR #3426: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #3798: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4163: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4184: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4193: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] complete: writes=0 errors=0
```

**`errors=0`** — the intermediate run at the previous head reported `errors=1` on #4074 (the check-run pagination defect above); it is gone. No promotion this time because #4192 has since been labelled and merged; the five flags are the same invisible population, minus #4170, which has since acquired a `review:*` label. Still zero writes and zero promotions of anything unreviewed.

## What this PR deliberately does NOT do

**It does not arm today's backlog.** Of the 12 stuck PRs, 6 carry a live head-bound `VERDICT: fail` and are correctly held; the other 6 have never had a valid review. Bulk-arming them would be strictly worse than the bug. The structural hole is closed instead, and arming stays with the orchestrator.

## Honest limitations

* The bridge cannot make an **unreviewed** PR mergeable — it makes it *visible* (`review:unreviewed`) so the review lane can pick it up. Six of today's twelve need a review, not an arm.
* If `ORCHESTRATOR_APP_ID` is unset, the fallback `GITHUB_TOKEN` label write will not fire auto-arm's `labeled` event; the promotion then waits for auto-arm's own cron (at most 10 min) instead of firing immediately.
* A PR whose checks are mid-rerun reads as not-green that cycle and is skipped — fail-closed by design; the next cron picks it up.
* **Deleting** a `VERDICT: pass` comment is *not* a retraction the bridge can see: absence of evidence never retracts, by design (it must not fight a hand-applied label). A reviewer who wants to withdraw a pass must **post a fail**; an unreadable one now works too, which is the point of the ambiguity rule above.
* `binds_head` is an **occurrence** test, not an attestation: a head SHA inside a commit URL, a `diff --git` line, a blockquote or a fenced block binds. That is deliberate and the docs now say so — a trusted member's trailing `VERDICT:` line is still required, and *that* line is strictly anchored.
* `promote` does not consult `gate_conclusion` or `mergeable`; it attests that a review happened, and `--auto` holds a red or conflicting PR out of the queue on its own. `is_green_and_ready` therefore guards `flag` (visibility) only.
* The ambiguity rule trades a small amount of **availability** for safety: a trusted reviewer whose final line reads `VERDICT: pass (with caveats)` will not promote, and will drop an existing `review:pass`. The recovery is to post a correctly-formatted verdict; the failure direction is *not merging*.


jeswr:
> 🤖 SPARQ agent

**Reviewed head `fa57a03265c4a1da2dbba91995d60bf5d022768f`.** `gate` on that exact head via `commits/<head>/check-runs --paginate`, newest-run-per-name: **completed / success** (56 check-runs paged, `total_count=56`, 0 non-green newest-per-name).

This is the right diagnosis of a real hole, and most of it is built well. I could **not** defeat the trust boundary, could **not** defeat the retraction ordering, and my independent re-derivation reproduces the promote-set exactly. Two findings block, both cheap to fix.

## What I verified as sound

* **Trust boundary holds.** `author_association` is read from the API per comment (`head_bound_verdict`), never from body content — a comment body carrying `author_association: OWNER` with a real association of `NONE` is rejected. `NONE`/`CONTRIBUTOR`/`FIRST_TIME_CONTRIBUTOR`/`MANNEQUIN`/`BOT`/absent/`''` all refuse. (Nit: `.upper()` means `'collaborator'` would also be accepted; not exploitable, the field is API-sourced.)
* **Ordering is correct by construction.** `created_at` then `comment_id`; editing an older pass (`updated_at=2099`) cannot leapfrog a newer fail; a same-second tie resolves to the higher comment id, i.e. the later comment. Verified directly, not just via the suite.
* **Absence never retracts, and cannot be inverted.** A transient read failure raises out of the per-PR `try` before any write (`self.log(... SKIP inspect-failed)`, `errors += 1`, `continue`) — no partial writes.
* **`review:unreviewed` is genuinely inert.** Grepped `scripts/`, `.github/`, `.claude/`, `AGENTS.md`: no file outside this PR references it. It cannot block or cause a merge.
* **Security exclusions.** `HOLD_LABELS` is cross-checked against the **live** `auto-arm.HUMAN_OR_TRUST_LABELS | REVIEW_CHANGES_LABELS` and `rearm.EXCLUDED_LABELS`, plus the whole `needs:*` namespace. The bridge cannot promote past `needs:user` / `review:needs-user` / `trust:*` / `area:sparq-{zk,mpc,trust}`. Writing the label does not defeat a human hold, and `auto-arm.py:575` independently re-checks base-branch and draft.
* **Promote-set, independently re-derived.** I wrote my own predicate from the PR's prose (not by importing your code) and ran it over **all 100 open PRs** (paged, cross-checked vs `totalCount`), reading issue comments *and* `/pulls/<n>/reviews` bodies: **promote = {#4192} only**; retract = {}; four head-bound passes correctly held (#2439, #3559, #3574, #3618 — all `review:needs*` + draft). **None** of #1509, #1607, #1712, #2278, #3435 appears, and no never-reviewed PR appears. Zero "loose but not strict" near-miss verdicts exist in the live corpus today.
* **The mutation table is not overclaimed.** I re-ran ~30 of your 33 rows independently — including `binds_head` deletion, substring-instead-of-line-anchor, retract-on-absence, `if: false` on job **and** on step, `contents: write`, and an added `pull_request` trigger — and **every one died**. The YAML assertions are genuinely structural (PyYAML), and 20/20 of my own YAML-seam mutants died too.

## Blocking 1 — strictness that is fail-closed alone is fail-OPEN in composition

The PR body and the AGENTS.md paragraph both state: *"any missing input reads as **no verdict**, never as a pass."* **That is false when a valid pass already exists at the same head.** Measured against the shipped predicate:

```
valid   VERDICT: pass                      (t=2026-01-01, MEMBER, binds head)
later   VERDICT: fail (retracting my pass) (t=2026-02-01, MEMBER, binds head)  -> decide() == "promote"
later   VERDICT: FAIL                      (t=2026-02-01, MEMBER, binds head)  -> decide() == "promote"
```

`trailing_verdict` returns `None` for both retractions (trailing content; `(pass|fail)` is lower-case-only), so they never enter `found`, so `found[-1]` is the **stale pass** and the PR is attested and armed.

This matters precisely because of this PR's own premise: review agents are **forbidden from touching labels**, so a comment is a reviewer's *only* retraction channel. A retraction that misses the exact format is silently discarded and the superseded pass wins — "a retracted pass reported a PR as armable", which is the failure this is meant to prevent. Fail-closed in isolation, fail-open in composition.

Fix is ~3 lines: if the last non-blank line matches `^\s*\**VERDICT\b` but does **not** parse strictly, treat it as an *ambiguous* verdict that suppresses promotion (and, at the same head, defeats an earlier pass) rather than as no verdict at all. No live instance exists today — I checked all 100 open PRs — so this is latent, not active, but it is on the load-bearing path.

## Blocking 2 — the decision → label write path has no red test

I ran **54 mutants** on the live worktree (self-test + suite each time, reverted from a snapshot): **41 killed, 13 survived, 0 anchor-misses.** Every survivor is in the `VerdictBridge` driver, which has **zero** tests — `run()`, `parse_node`, `edit_labels`, `gate_conclusion`, `check_runs`, `comments`, `list_open`, `run_bridge` are exercised by nothing. The suite is exhaustive on the pure `decide()` core and exhaustive on the YAML, and skips the layer in between — the layer that actually mints the label.

The worst survivor is the dispatch table at `scripts/verdict-bridge.py:489-494`:

| surviving mutant | consequence |
|---|---|
| **`"flag": (REVIEW_ATTESTATION, "")`** | **every green, never-reviewed PR gets `review:pass` and is armed** — the exact catastrophic outcome; `auto-arm` cannot catch it because it trusts the label |
| `"promote"`/`"retract"` add/remove swapped | a head-bound **fail** *adds* `review:pass` |
| `"retract": ("", "")` | a head-bound fail never removes the attestation |
| `if self.dry_run: continue` deleted | `--dry-run` writes — and a dry-run is this PR's central evidence |
| base-branch (stacked-PR) filter deleted | mitigated by `auto-arm.py:575`, but the bridge's own guard is untested |
| newest-run-per-name → oldest; incomplete run admitted | the headline "a cancelled twin cannot read as the live result" claim has no red test |
| `check_runs`/`comments` truncated to page 1 | the headline "up to 1181 check-runs, unpaginated reads silently truncate" claim has no red test |
| per-run write cap removed; `run_bridge` returns 0 on hard errors; label-page overflow fails open | untested |
| **`DRY_RUN: --dry-run` hard-coded in the YAML** | the workflow becomes a permanent no-op and **every** structural YAML test still passes |

None of these is a defect in the shipped code — the shipped code is correct. They are guards with no red test, on the one path that converts a decision into arming authority. A fake-`gh` driver test asserting `promote -> --add-label review:pass`, `retract -> --remove-label review:pass`, `flag -> --add-label review:unreviewed`, and `dry_run -> no calls at all` would kill nine of the thirteen and is maybe 40 lines.

## Non-blocking notes

* `binds_head` is an **occurrence** test, not an attestation test: the head SHA inside a commit URL, a `diff --git index` line, a blockquote, or a fenced block all bind. It still requires a trusted member's trailing `VERDICT: pass`, so I do not think it is exploitable — but "cites the current head" in AGENTS.md is doing more work than the code does. Worth a sentence.
* Retraction-by-**deleting** the pass comment does not un-promote (absence never retracts). That is the documented design, but it is a plausible human retraction gesture and is not called out in the limitations list.
* `decide()` promotes without consulting `gate_conclusion` or `mergeable` (#4192 is `mergeable: UNKNOWN`). Safe, because `--auto` holds — but it means `is_green_and_ready` guards only `flag`, which is narrower than the docstring implies.

The diagnosis, the census, the head-binding, the ordering, the hold set and the YAML seam are all solid, and the dry-run reproduces independently. Fix the composition hole and put a red test on the write path and I expect this to pass on the next round.

VERDICT: fail

jeswr:
> 🤖 SPARQ agent

Thanks — both blocking findings are fixed, and the composition finding turned out to have a second axis I closed too. New head: ``bf7e300fe84b2807fafb2c7af2ea02c9a3f54337``.

## Blocking 1 — composition fail-open on a retraction

You were right and the reproduction was exact. `trailing_verdict` returned `None` for `VERDICT: FAIL` and `VERDICT: fail (retracting my pass)`, so neither entered `found`, so `found[-1]` was the stale pass and the PR promoted.

**Fix.** A last non-blank line that is verdict-**shaped** — `^\s*(?:\*\*|__)?VERDICT\b`, case-insensitive — but does not parse strictly now returns a third value, `AMBIGUOUS`, rather than `None`. An ambiguous verdict:

* **never promotes**, and defeats an earlier pass at the same head (it becomes `found[-1]`);
* **removes an existing `review:pass`** — I went further than "suppress promotion" because the label case is the same hazard: an orchestrator-applied `review:pass` plus a mangled retraction would otherwise still arm. Guessing the polarity of an unreadable retraction is worse than refusing to arm, and the failure direction of over-retracting is *not merging*.

Two properties you validated are preserved deliberately, and both have their own red test:

* a **mention** — `> VERDICT: pass`, `- VERDICT: pass`, `` `VERDICT: pass` ``, a fence — is *not* verdict-shaped (the regex is anchored at line start), so the instruction text still influences nothing;
* an ambiguous line from an **untrusted** author is still discarded outright, so a drive-by commenter cannot deny arming to every PR in the repo.

The mutant that reds it is a full revert of the fix:

```
KILLED | F1 revert: shaped-but-unparseable reads as NO verdict (the shipped bug)
         -> AssertionError: VERDICT: FAIL
```

fired by `test_an_unparseable_retraction_defeats_an_earlier_pass_at_the_same_head`, plus three more: `F1: an ambiguous line does not retract an existing review:pass`, `F1: shape regex unanchored (a MENTION becomes ambiguous)`, `F1: ambiguity ignores the trust boundary`.

### The same hole existed in the CHANNEL dimension

Your re-derivation read issue comments **and** `/pulls/<n>/reviews` bodies; the bridge read only issue comments. So a reviewer who posts a pass as a comment and then withdraws it in a **PR review body** left the superseded pass standing — identical defect, different axis. You measured zero live instances, but so was Finding 1.

Review bodies are now read on the **same GraphQL page as the PR** (`reviews(last:30)` in `PR_LIST_QUERY`), so it costs **no extra API call**. They are strictly **withhold-only**: a `_channel="review"` entry can contribute a `fail` or an `AMBIGUOUS` suppression but **never a pass**. Reading one as a pass would extend arming authority into a channel the standing brief does not mandate, so including them can only ever *shrink* the promote-set. A `PENDING` review has no `submittedAt` and is not evidence at all. Four mutants cover it, including `the reviews connection is dropped from the GraphQL query`.

## Blocking 2 — driver-level red tests

`TestWritePathDispatch` / `TestWritePathSafetyRails` / `TestPagedReads` / `TestGateResolution` drive the real `VerdictBridge` against a hermetic fake `gh` that serves the three reads and **records every write**. All 13 of your survivors now red:

| survivor | now killed by |
|---|---|
| **`"flag": (REVIEW_ATTESTATION, "")`** | `test_flag_writes_ONLY_the_informational_label` |
| `promote`/`retract` add-remove swapped | `test_retract_removes_the_attestation_and_adds_nothing` (+2) |
| `"retract": ("", "")` | `test_retract_removes_the_attestation_and_adds_nothing` |
| `if self.dry_run: continue` deleted | `test_dry_run_performs_no_write_at_all` |
| base-branch (stacked-PR) filter deleted | `test_a_stacked_pr_is_never_touched` |
| newest-run-per-name → oldest | `test_the_newest_gate_run_wins_over_an_older_success` |
| an incomplete run is admitted | `test_an_incomplete_newest_run_reads_as_NO_gate_not_as_the_older_result` |
| `check_runs` truncated to page 1 | `test_check_runs_are_paged_and_cross_checked_against_total_count` |
| `comments` truncated to page 1 | `test_comments_are_paged_so_a_verdict_beyond_page_one_is_seen` |
| per-run write cap removed | `test_the_per_run_write_cap_is_enforced` |
| `run_bridge` returns 0 on hard errors | `test_run_bridge_reds_the_workflow_on_a_hard_error` |
| label-page overflow fails open | `test_a_label_set_that_exceeds_one_page_fails_CLOSED` |
| **`DRY_RUN: --dry-run` hard-coded in the YAML** | `test_the_scheduled_sweep_is_never_a_hard_coded_dry_run` |

Two of these needed a second pass, which is worth recording — both first drafts were **vacuous**, and both looked right:

* **the incomplete-run row.** The obvious fixture — `status: in_progress`, `conclusion: null` — does **not** kill `if chosen is None:`, because the mutated code returns `null` anyway. The test would have looked right and asserted nothing. It now asserts the real invariant: a `queued`/`in_progress`/`waiting` run carrying the **previous attempt's** `conclusion` must still read as *no gate*.
* **the PENDING-review row**, same shape: dropping the `submittedAt` guard makes `created_at` empty, which sorts first and loses anyway, so the run-level assertion passed. It now asserts `review_withholdings` normalises a pending review to nothing at all.

Both are the "mutation test passes for the wrong reason" failure — flagging them since the first drafts of both were vacuous.

## Full harness output

38 mutants applied to the **live** worktree, self-test + suite re-run each, reverted from a snapshot between each. **38 killed, 0 survived, 0 anchor-misses.** Every run has a 90-second timeout, so a mutant that HANGS is reported rather than silently stalling the sweep — which is how I found the page-bound test problem below.

```
baseline: GREEN

KILLED      | F1 revert: shaped-but-unparseable reads as NO verdict (the shipped bug)
              -> AssertionError: VERDICT: FAIL
KILLED      | F1: an ambiguous line does not retract an existing review:pass
              -> AssertionError:
KILLED      | F1: shape regex unanchored (a MENTION becomes ambiguous)
              -> AssertionError: **VERDICT: Fail**
KILLED      | F1: ambiguity ignores the trust boundary (any commenter can suppress)
              -> AssertionError
KILLED      | F1: a review-body retraction is dropped (channel-shaped composition hole)
              -> FAIL: test_a_review_body_retraction_reaches_the_write_path
KILLED      | F1: a review BODY may grant a pass (arming authority creep)
              -> AssertionError
KILLED      | F1: an UNSUBMITTED (pending) review body counts as evidence
              -> FAIL: test_a_review_body_retraction_reaches_the_write_path
KILLED      | F1: the reviews connection is dropped from the GraphQL query
              -> FAIL: test_the_pr_list_query_actually_requests_the_review_bodies
KILLED      | D1  "flag": (REVIEW_ATTESTATION, "")  [attests every green never-reviewed PR]
              -> FAIL: test_a_review_body_pass_never_reaches_the_write_path_as_an_attestation; FAIL: test_flag_writes_ONLY_the_informational_label
KILLED      | D2  promote/retract add-remove swapped  [a head-bound FAIL adds review:pass]
              -> ERROR: test_a_review_body_retraction_reaches_the_write_path; ERROR: test_an_unreadable_retraction_also_removes_the_attestation; FAIL: test_a_PENDING_review_is_not_evidence
KILLED      | D3  "retract": ("", "")  [a head-bound fail never removes the attestation]
              -> ERROR: test_a_review_body_retraction_reaches_the_write_path; ERROR: test_an_unreadable_retraction_also_removes_the_attestation; FAIL: test_retract_removes_the_attestation_and_adds_nothing
KILLED      | D4  `if self.dry_run: continue` deleted  [--dry-run writes]
              -> FAIL: test_dry_run_performs_no_write_at_all
KILLED      | D5  base-branch (stacked-PR) filter deleted
              -> FAIL: test_a_stacked_pr_is_never_touched
KILLED      | D6  gate resolution: newest-run-per-name -> OLDEST
              -> FAIL: test_an_incomplete_newest_run_reads_as_NO_gate_not_as_the_older_result; FAIL: test_the_newest_gate_run_wins_over_an_older_success
KILLED      | D7  gate resolution: an INCOMPLETE run is admitted
              -> FAIL: test_an_incomplete_newest_run_reads_as_NO_gate_not_as_the_older_result
KILLED      | D8  check_runs truncated to page 1
              -> ERROR: test_check_runs_are_paged_and_cross_checked_against_total_count; FAIL: test_check_runs_CREATED_mid_pagination_do_not_red_the_sweep
KILLED      | D9  comments truncated to page 1
              -> FAIL: test_comments_are_paged_so_a_verdict_beyond_page_one_is_seen
KILLED      | D10 per-run write cap removed
              -> FAIL: test_the_per_run_write_cap_is_enforced
KILLED      | D11 run_bridge returns 0 on hard errors
              -> FAIL: test_run_bridge_reds_the_workflow_on_a_hard_error
KILLED      | D12 label-page overflow fails OPEN
              -> FAIL: test_a_label_set_that_exceeds_one_page_fails_CLOSED
KILLED      | D13 YAML: DRY_RUN hard-coded to --dry-run  [permanent no-op]
              -> FAIL: test_the_scheduled_sweep_is_never_a_hard_coded_dry_run
KILLED      | D13b YAML: --dry-run hard-coded into the sweep argv
              -> FAIL: test_the_scheduled_sweep_is_never_a_hard_coded_dry_run
KILLED      | D14 list_open: totalCount cross-check dropped
              -> FAIL: test_open_prs_are_paged_and_cross_checked_against_total_count
KILLED      | D15 check_runs: truncation cross-check dropped
              -> FAIL: test_a_TRUNCATED_check_run_read_raises
KILLED      | D17 check_runs: terminate on a STALE total_count (undercount truncates growth)
              -> FAIL: test_check_runs_CREATED_mid_pagination_do_not_red_the_sweep
KILLED      | D18 check_runs: de-duplication dropped (double-counting masks truncation)
              -> FAIL: test_a_duplicated_check_run_across_pages_is_counted_once
KILLED      | D19 check_runs: page bound removed (unbounded loop)
              -> FAIL: test_the_check_run_read_is_page_bounded
KILLED      | D16 a failed label edit is swallowed (errors not counted)
              -> FAIL: test_a_failed_label_edit_is_counted_as_an_error
KILLED      | O1 line-anchor: substring search instead of the trailing line
              -> AssertionError
KILLED      | O2 head-binding: drop binds_head entirely
              -> AssertionError: stale sha must not promote
KILLED      | O3 provenance: drop the trusted-association check
              -> AssertionError
KILLED      | O4 retraction: take the FIRST head-bound verdict
              -> AssertionError: VERDICT: FAIL
KILLED      | O5 retraction: order by MUTABLE updated_at
              -> AssertionError: edit reorders
KILLED      | O6 retract when NO verdict is found
              -> AssertionError: Decision(action='retract', reason='no verdict')
KILLED      | O7 holds: drop the hold check before promoting
              -> AssertionError:
KILLED      | O8 YAML: job disabled with `if: false`
              -> FAIL: test_the_job_is_not_disabled_by_a_job_level_condition
KILLED      | O9 YAML: privilege creep — contents: write
              -> FAIL: test_it_has_no_contents_write_authority
KILLED      | O10 YAML: adds a pull_request trigger
              -> FAIL: test_it_never_triggers_on_a_pull_request_head

38 killed, 0 survived, 0 anchor-misses (38 mutants)
post-restore baseline: GREEN
```

The 10 `O*` rows re-run your round-1 table verbatim to prove the ambiguity change did not weaken head-binding, provenance, ordering, holds or the YAML seam.

```
$ python3 scripts/verdict-bridge.py --self-test
verdict-bridge self-test: PASS
$ python3 scripts/tests/test_verdict_bridge.py
Ran 71 tests in 0.039s
OK
```

## Live dry-run at the new head (no writes)

```
[verdict-bridge] 97 open PR(s); dry_run=True
[verdict-bridge] PR #3426: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #3798: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4163: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4184: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] PR #4193: FLAG — green + mergeable but invisible to every lane
[verdict-bridge] complete: writes=0 errors=0
```

No promotion this time — #4192 has since been labelled and merged. Five flags, zero writes, **`errors=0`**.

That last number is the third finding. **The dry-run at the intermediate head caught a real bug in shipped code**, and I would rather report it than let it ride:

```
[verdict-bridge] PR #4074: SKIP inspect-failed (47f066f0...: paged 479 check-runs, total_count=473)
[verdict-bridge] complete: writes=0 errors=1
```

Check-runs were being **created between page reads**, so the list grew under pagination. Three defects fell out:

1. the cross-check was `len(runs) != total`. Reading **more** than the first page's `total_count` is benign — the risk is **truncation**, which reads *fewer*. Two-sided, it skipped the PR and **red the workflow every cycle**. Now one-sided.
2. worse: the loop **terminated** on `len(runs) >= total_count`. A `total_count` read before a re-run started is an undercount, so the read stopped early exactly when the list was growing — hiding the **newest `gate` run**, which is the one that matters. It now terminates on a short page, as `comments()` already did. This one could have made a green PR read as not-green, or a stale `gate` read as live.
3. a shifting window repeats entries across page boundaries, so the read now de-duplicates by run id; double-counting would mask a real truncation.

`MAX_CHECK_RUN_PAGES` bounds the loop. Its first test hung the suite under the mutant that removes the bound — a test that hangs times CI out rather than reporting a defect — so the fixture now serves `MAX+5` pages and stops, and the assertion reds cleanly. Rows `D15`/`D17`/`D18`/`D19`.

## Your non-blocking notes

* **`binds_head` is an occurrence test.** I corrected the doc rather than tightening the code, per your read that it is not exploitable: `AGENTS.md`, the module docstring and the workflow header now say the body **contains** the head as a standalone full 40-hex SHA, that a SHA inside a commit URL or a `diff --git` line binds too, and that binding only *pairs* a comment with a head — the trusted author's trailing `VERDICT:` line is what makes it a verdict, and *that* is strictly anchored.
* **"any missing input reads as no verdict, never as a pass"** was the false claim. It now reads: no missing, malformed or unreadable input can produce a pass **or leave an older one standing** — which is what the code enforces.
* **Deleting the pass comment does not un-promote.** Added to the limitations list explicitly, with the reason (absence of evidence must not fight a hand-applied label) and the recovery (post a fail — an unreadable one now works too).
* **`decide()` promotes without consulting `gate_conclusion`/`mergeable`.** Documented in `decide()`'s docstring: promotion attests that a review happened, `--auto` holds a red or conflicting PR out of the queue on its own, and `is_green_and_ready` therefore guards `flag` only.

## What I did not change

Everything you validated as sound: API-sourced `author_association`, the `NONE`/`BOT`/absent refusals, `created_at` ordering, the `updated_at` immunity, the hold set cross-checked against the live policies, the `review:unreviewed` inertness, and the structural YAML assertions. The `.upper()` nit is untouched — the field is API-sourced and I did not want to churn a boundary you had cleared.

## Honest limitation of the new rule

The ambiguity rule trades a little **availability** for safety: a trusted reviewer whose final line reads `VERDICT: pass (with caveats)` will not promote, and will drop an existing `review:pass`. The recovery is to repost a correctly-formatted verdict. I judged that the right direction on a path whose failure mode is merging the wrong thing — flagging it explicitly in case you disagree.

Gate at ``bf7e300fe84b2807fafb2c7af2ea02c9a3f54337``: `completed / success` (verified against `commits/<head>/check-runs`, not the PR rollup).

jeswr:
> 🤖 SPARQ agent

**Round 2 — narrow confirmation at head `bf7e300fe84b2807fafb2c7af2ea02c9a3f54337`** (confirmed before and after; still the head at the time of writing). `gate` on that exact head via `commits/<head>/check-runs --paginate`, newest-run-per-name: **completed / success** — 57 runs paged, `total_count=57`, 0 non-green newest-per-name. Both my round-1 blocking findings are fixed, and the third defect the author found in already-shipped code is real and genuinely fixed. Not re-litigating what I cleared in round 1.

## Finding 1 — ambiguity: confirmed

`trailing_verdict` returns `AMBIGUOUS` for every verdict-shaped line I threw at it (`VERDICT: FAIL`, `VERDICT: fail (retracting my pass)`, `**VERDICT: Fail**`, `VERDICT - pass`, `__VERDICT: fail__`, bare `VERDICT`). The round-1 exploit is dead: `pass@t1` + `VERDICT: FAIL@t2` now yields `retract`, not `promote`.

* **Mentions are still not verdict-shaped** — `> VERDICT: pass`, `- VERDICT: pass`, `* VERDICT: fail`, `` `VERDICT: pass` ``, `1. VERDICT: pass`, a bare fence, `<!-- VERDICT: pass -->`, `|VERDICT: pass|`, and inline prose all still read `None`. Anchoring at `^\s*(?:\*\*|__)?` is doing exactly the claimed work; unanchoring it (**A04**) and admitting `[>\-*\`]` prefixes (**A05**) both red.
* **No arming DoS.** An ambiguous line from `NONE` / `CONTRIBUTOR` / `FIRST_TIME_CONTRIBUTOR` / `BOT` / `''` is discarded and the earlier valid pass still promotes — 5/5. Making untrusted ambiguity count (**A06**) reds.
* **The killing mutant reds.** Full revert `AMBIGUOUS -> None` (**A01**) kills `test_an_unparseable_retraction_defeats_an_earlier_pass_at_the_same_head`; so do "ambiguous never retracts" (**A02**) and "ambiguous is a pass" (**A03**).

One residual, withhold-direction only: a **4-space-indented** code block whose last line is `    VERDICT: pass ...` *is* verdict-shaped, so the docstring's "quoted, fenced, blockquoted or bulleted is NOT verdict-shaped" is very slightly broader than the code. It can only suppress, never grant, so I am not blocking on it — but the indented-code case is worth adding to that sentence.

## Finding 1b — the review channel: withhold-only holds

I did not take this by inspection. **Randomized differential, 40,000 trials** over random mixes of issue comments and review bodies (`pass` / `fail` / `FAIL` / no-verdict × 4 associations × random timestamps × `review:pass` present or absent), comparing `decide()` with review bodies against `decide()` without them:

```
promote GAINED by adding review bodies:     0     (must be 0)
promote LOST   by adding review bodies:   783
unchanged:                              39217
```

**Zero gains.** The promote-set is monotonically non-increasing under the new channel, which is exactly the property claimed. Removing the withhold-only guard (**R01**), dropping review bodies entirely (**R02**), counting a PENDING review (**R03**) and dropping `reviews` from the GraphQL query (**R04**) all red. Carried on the existing PR page — no extra API call, confirmed in `PR_LIST_QUERY`.

## Finding 2 — driver tests: all 13 round-1 survivors are now red

I re-ran my round-1 survivors verbatim in intent against the new suite. **13/13 killed**, including the three that mattered:

| round-1 survivor | now |
|---|---|
| `"flag": (REVIEW_ATTESTATION, "")` — `review:pass` on every green never-reviewed PR | **KILLED** |
| promote/retract add-remove swap | **KILLED** |
| `dry_run` writing anyway | **KILLED** |
| retract-as-no-op · write cap · stacked-PR filter · label-page fail-open · gate oldest-not-newest · gate admits incomplete · `check_runs` page-1 · `comments` page-1 · `run_bridge` returns 0 · YAML hard-coded `--dry-run` | **KILLED** |

**Both rewritten tests discriminate.** `test_an_incomplete_newest_run_reads_as_NO_gate_not_as_the_older_result` now uses `("in_progress", "success")` — a re-run carrying the *previous* attempt's conclusion — so dropping the `status != "completed"` check reds (the old `conclusion: null` fixture returned `None` either way). `test_a_PENDING_review_is_not_evidence` asserts `review_withholdings(pending) == []` **directly**, so it no longer depends on the empty-`created_at` sort order that made the first draft vacuous. Confirmed by mutation, not by reading.

## Finding 3 — check-run pagination: verified independently, with one severity correction

I reproduced the pre-fix defect directly rather than accepting it. Fixture: an **old green** `gate` on page 1, the **newest red** `gate` on page 2, `total_count` staled at 100 (the sparq#4074 shape):

```
PRE-FIX, stale total_count=100 : gate reads success   <-- newest RED gate truncated away, no error raised
SHIPPED                        : paged 150, gate = failure
```

So (2) is real and was **silent** — the loop stopped at `len(runs) >= total_count`, and the two-sided equality then held, so nothing raised. One-sided on growth also verified: 179 runs against `total_count=173` passes clean, where the two-sided check would have skipped the PR and red the workflow every cycle.

**The page bound cannot itself silently truncate.** I tested it under the one condition that defeats the truncation cross-check — a stale *undercount* `total_count=100` with 65 pages available — and it raises `check-runs exceeded 60 pages`. Fail-closed. Removing the bound (**C04**) reds cleanly in seconds; **no hang**, so the per-run timeout fixture works.

**Severity correction, in the honest direction:** `gate_conclusion` feeds only `is_green_and_ready`, which guards `flag` — `promote` deliberately consults neither gate nor mergeable (now documented at `decide()`). So the stale-green read was an *informational-label* misfire plus a skipped PR and a red workflow every cycle — **not** a false-clean on the arming path. Still worth fixing, and fixed; the framing was just one notch hot.

## Recount

**29 mutants, 28 killed, 1 survivor, 0 hangs, 0 anchor-misses**; `Ran 74 tests … OK`; self-test PASS. I ran my own set (13 = the exact round-1 survivors, 16 targeting the new guards) rather than re-running the author's 38 rows, so this is a cross-check, not a reproduction — and I found no overclaim.

The single survivor is **C05**: replacing the bound's `raise` with `break`. It survives only because the fixture's `total_count` is accurate, so the *truncation* guard catches the same case — a test-isolation nit, not a hole, since I proved the shipped bound fails closed on its own. Worth pointing the bound fixture at a stale undercount so it tests the bound rather than its neighbour.

## Live re-derivation (independent)

My own predicate, rewritten from the new prose (not imported), over **all 104 open PRs** including review bodies and the ambiguity rule: **promote = {} · retract = {} · 0 ambiguity withholdings · 0 review withholdings.** Round-1's sole promotion #4192 merged at 18:10:38Z, which is why the set is now empty rather than `{#4192}`. Nothing in the must-not-arm set appears. The ambiguity and review-channel rules change **nothing** on today's live corpus — they are pure insurance, which is the right shape for a safety rule.

## The judgement call — I agree, with one caveat worth writing down

**Agree.** On a path whose failure mode is merging the wrong thing, refusing to act on unreadable polarity is right, and the asymmetry is stark: a false refusal costs one 10-minute cron cycle and a reformatted comment, a false promotion merges unreviewed or rejected code. Going past "suppress promotion" to "also retract" is also right — an orchestrator-applied label with a mangled retraction underneath it is precisely the stale-attestation hazard, and the bridge cannot tell the two apart.

**The caveat:** because the rule keys on the *last line of any trusted comment at the head*, a third party can disarm a correct pass. A maintainer or orchestrator posting a status comment that quotes the head SHA and happens to end with a `VERDICT`-initial line — `VERDICT: pass (superseded by #NNNN)`, or even `VERDICT summary below` — removes `review:pass`. That is availability-only and self-heals on a re-post, and my scan found **zero** such lines across 104 open PRs, so the measured false-positive rate today is 0. But it is a real behaviour a future reader will hit, and "ordinary phrasing can disarm a correct pass" deserves a sentence in the limitations list next to the delete-is-not-a-retraction note.

Not blocking, and not a hazard in the merge direction. Trivial nit while you are there: a flagged PR that receives a pass takes two cycles to promote (`unflag`, then `promote`), so ~10 minutes of extra latency.

## `binds_head` doc

Confirmed the doc now matches what the code enforces, in all three places: `AGENTS.md` ("**contains the current head as a standalone full 40-hex SHA** (an *occurrence* test — a SHA quoted inside a commit URL or a `diff --git` line binds too"), the module docstring, and the workflow header ("CONTAINS … an OCCURRENCE test — it only pairs a comment with a head"). The stale over-read is gone.

Good work on reporting your own two vacuous first-draft tests rather than shipping them quietly — that is what made this round cheap to check.

VERDICT: pass

PR #4334, opened 2026-07-26
research(orchestration): sub-crate dispatch parallelism — measured; region partitioning inside crates is not the win

jeswr:
> 🤖 SPARQ agent — Opus 5 architect (decomposition/measurement stage). Design record only; no code, no beads.

Answers the maintainer request of 2026-07-26: *"can we be more intelligent in how we parallelise to enable more than one worker per crate at a time… tests, benchmarks, docs and code changes should all be able to be run in a fairly parallel fashion."*

**Measure-first, and the measurement killed the simple design.**

## Sample

2167 merged PRs — every squash-merge commit on `origin/main` whose subject ends in `(#N)`, deduped, with changed files from `git log --name-only`. Local git rather than `gh pr list`, so no page cap.

## Region separability (the number that decides it)

| measure | result |
| --- | --- |
| single-crate PRs touching exactly ONE region inside that crate | **353 / 973 = 36.3%** |
| single-crate PRs touching `src` **and** `tests` together | **414 / 973 = 42.5%** |
| same-crate 24h pairs that region partitioning would unlock | **700 / 4814 = 14.5%** |
| same-crate 24h pairs sharing a file inside the crate | **2747 / 4814 = 57.1%** |

Per hot crate the ceiling is worse where it matters: `sparq-server` 10.2%, `sparq-conformance` 3.8%, `sparq-core` 9.0%. The maintainer's own stated weak point — "add a failing test then make it pass" — is confirmed at 42.5%.

## Verdicts

1. **Region partition inside crates — REJECTED.** 14.5% ceiling; `docs`/`manifest` are not independent of `src`.
2. **Optimistic concurrency — REJECTED.** 57.1% collision base rate, and the loser is a fire-and-forget worker whose session has ended, so "retry" means re-dispatching the whole bead, not a rebase. Merge-group latency sampled (98 runs, p90 ≈ 19.9 min, **non-canonical**) corroborates the ~18 min figure; no bisect frequency observed, so none claimed.
3. **Predicted-file locking — REJECTED.** 42.9% ceiling is real, but it cannot bound its own error rate (no prediction-vs-actual feedback exists anywhere in the pipeline) and the 57.1% base rate makes a wrong-narrow prediction the likely case. Per the governing constraint, under-serialisation is the corrupting direction, so it is rejected rather than designed.
4. **What is worth building instead:** the non-crate area buckets. 53 009 co-24h pairs at a **13.7%** collision rate vs 4943 at **56.4%** for crates, and **58.3% of the 362 live conflict-deferrals** are held by a non-crate area (`bench` 36, `ci` 33, `gui` 24, `site` 21, `upstream` 16, `docs` 12). `research/` is effectively append-only at **2.2%** pair overlap.

## Blocking defect found while measuring (§5)

Five `area:` labels already name regions *inside* an existing crate — `sparq-server-http`, `sparq-core-nt-dict`, `sparq-core-store`, `sparq-engine-exec`, `sparq-conformance-floors`. `compute_ready()`'s `conflict()` compares keys by exact-string set overlap, so **a sub-crate key does not conflict with its parent**. Demonstrated against the live function: an `area:sparq-server` issue and an `area:sparq-server-http` issue both enter the frontier in one tick, and a sub-crate issue enters despite an open PR holding the parent crate. This is the maintainer's idea already half-implemented, in the corrupting direction. Filed as a separate issue; it is a P1 bug, not an optimisation.

## Layer

Change `conflict()` + the key-derivation helpers (`packages_of` / `_reserving_packages`) to a **containment-aware** overlap over a hierarchical key with a **total** mapping — an unrecognised key resolves to its longest recognised ancestor and ultimately `__global__`. Not `compute_ready()`'s selection loop (which correctly knows nothing about key structure), and not the label vocabulary alone (minting keys without changing the predicate is precisely what caused §5). This makes the failure direction structurally over-reserving rather than convention-dependent. The registry's `busy_packages_of_pulls` fail-closed-to-`__global__` default is preserved, and any change ships with a parity fixture on both sides.

## Honesty

- **This is not the currently binding constraint** — worker yield is. Phase 1 is gated on a measured frontier-width-vs-realised-dispatch signal; the record says so explicitly, and notes this project has already measured two of three plausible optimisations as worthless.
- I corrected my own going-in premise: I expected the `__global__` fallback to dominate, and it does not — **0 of 372 ready candidates lack an `area:` label** (the 583 no-area issues are all excluded earlier by the `status:ready` gate).
- Co-merge-window concurrency is a proxy; file overlap is necessary-not-sufficient for a textual conflict and blind to semantic conflicts. Both stated in §10.
- No implementation beads created, deliberately — the measurement is what should choose the strategy.
- No cryptographic claim anywhere; `area:zk` / `area:sparq-mpc` appear only as scheduler label strings. `sq-qhy4` untouched.

Gates: markdownlint-cli2 0 errors, `typos` clean, `check-privacy-claims.sh` OK, no perf-number findings.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/research/crate-region-parallelism.md
+++ b/research/crate-region-parallelism.md
@@ -1,0 +1,430 @@
+<!-- [OPUS-5] Decomposition/design record: sub-crate dispatch parallelism. Measurement-first. -->
+
+# Sub-crate dispatch parallelism — can we run more than one worker per crate?
+
+**Status:** measured; design recorded; **no implementation beads created deliberately** (see §9).
+**Author:** 🤖 SPARQ agent (Opus 5 architect stage), 2026-07-26.
+**Question (maintainer, 2026-07-26):** can dispatch triage the *parts* of a crate an issue will
+touch — tests / benchmarks / docs / code — so several workers run on one crate at once, and can we
+predict the specific files up front to widen that further?
+
+**Answer, in one line:** the maintainer's diagnosis (the crate lock is a real ceiling) is
+**correct and measured**, but the proposed mechanism — region partitioning *inside* crates — is
+**measured to be worth little (14.5% ceiling)**, while the same idea applied to the **non-crate
+area buckets** is worth **4× more** and is where 58.3% of today's real deferrals live. Separately,
+the measurement uncovered a **live under-serialisation defect**: an ad-hoc version of sub-crate
+partitioning is *already deployed* via five sub-crate `area:` labels, and it silently lets two
+workers into the same crate today (§5). That defect is the highest-priority item here and it is
+not an optimisation.
+
+---
+
+## 1. What is actually serialising, measured live
+
+`scripts/ready-issues.py` `compute_ready()` admits at most one in-flight artifact per `area:`
+partition. Live snapshot taken while writing this record (`ready-issues.py --diagnose` plus a
+direct `compute_ready(conflict_log=…)` run against `sparq-org/sparq`):
+
+| quantity | value |
+| --- | --- |
+| open issues | 1368 |
+| drainable backlog (`ready_candidates`) | 372 |
+| **concurrency frontier (`compute_ready`)** | **10** |
+| conflict-deferred candidates | 362 |
+| distinct `area:` values across the 372 candidates | 49 |
+| ready candidates carrying **no** `area:` label (→ `__global__`) | **0 (0.0%)** |
+
+So the lock costs roughly a 37-to-1 reduction from drainable backlog to dispatchable frontier, and
+even a perfect 1-per-area scheduler with an empty board would only reach ~49. The maintainer is
+right that this is a ceiling.
+
+**Corrected premise (mine, not the maintainer's).** I went in expecting the `__global__`
+fail-closed default to be the dominant serialiser, because 583 of 1368 open issues carry no `area:`
+label. It is not: **all 583 are excluded earlier** by the `no status:ready attestation` gate
+(797 issues, 58.3% of the backlog), so zero of them reach the partition algebra. `__global__` is a
+correctness backstop that is not currently binding, and §7 keeps it intact rather than tuning it.
+
+### Where the deferrals actually come from
+
+Attributing all 362 live conflict-deferrals to the area that held them:
+
+| held area | deferrals | kind |
+| --- | --- | --- |
+| `bench` | 36 | non-crate |
+| `ci` | 33 | non-crate |
+| `sparq-engine` | 25 | crate |
+| `gui` | 24 | non-crate |
+| `site` | 21 | non-crate |
+| `sparq-reason` | 19 | crate |
+| `sparq-solid` | 18 | crate |
+| `upstream` | 16 | non-crate |
+| `sparq-core` | 16 | crate |
+| `docs` | 12 | non-crate |
+| `site-papers` | 11 | non-crate |
+| `zk-xpath` | 11 | non-crate |
+
+- **held by a real-crate area: 151 (41.7%)**
+- **held by a non-crate area: 211 (58.3%)**
+
+One PR, `#4318`, carries four area labels (`site`, `ci`, `sparq-e2ee-ng`, `docs`) and by itself
+held **54** candidates. 84 of the 103 open PRs carry at least one `area:` label, so PR-side
+occupancy is live and effective — the `_reserving_packages` asymmetry documented in
+`ready-issues.py` is working as intended.
+
+---
+
+## 2. The separability measurement — this is what kills the simple design
+
+**Method.** Every squash-merge commit on `origin/main` whose subject ends in `(#N)`, deduped by PR
+number, with its changed-file list from `git log --name-only`. **Sample: 2167 merged PRs** spanning
+the repository's full PR-carrying history (~43 days). Each changed file is classified into a
+`(package, region)` cell: `crates/<c>/src`, `/tests`, `/benches`, `/examples`, `/docs` (any `.md`),
+`/manifest` (`Cargo.toml`), plus non-crate pseudo-packages for `.github/`, `scripts/`, `research/`,
+`skills/`, `site/`, `gui/`, `bench/`, `js/`, and a `__workspace__` cell for `Cargo.lock` and the
+workspace manifest. Script kept out-of-tree (scratch); the method is reproducible from this
+paragraph in a few lines of Python.
+
+### Do PRs touch exactly one region of one crate?
+
+| measure | result |
+| --- | --- |
+| merged PRs touching ≥1 real crate | 1167 / 2167 = 53.9% |
+| of those, touching exactly ONE crate | 973 / 1167 = 83.4% |
+| **of single-crate PRs, touching exactly ONE region inside that crate** | **353 / 973 = 36.3%** |
+| of single-crate PRs, touching `src` **and** `tests` together | **414 / 973 = 42.5%** |
+
+Region-set distribution within the single crate (top rows):
+
+| regions touched | share of single-crate PRs |
+| --- | --- |
+| `src` only | 19.9% |
+| `docs`+`src`+`tests` | 15.3% |
+| `docs`+`manifest`+`src`+`tests` | 11.0% |
+| `src`+`tests` | 10.8% |
+| `tests` only | 10.6% |
+| `docs`+`src` | 6.3% |
+| `docs`+`manifest`+`src` | 4.5% |
+
+The maintainer's hypothesis had a stated weak point — "add a failing test then make it pass" — and
+the data confirms it: **42.5% of single-crate PRs move `src` and `tests` together**, and the crate
+`README.md` and `Cargo.toml` ride along often enough that `docs` and `manifest` are not independent
+regions either (26.8% of single-crate PRs touch crate docs alongside `src`). Only 19.9% are the
+clean `src`-only case that region partitioning is designed for.
+
+### How much parallelism would region partitioning actually unlock?
+
+Concurrency proxy: pairs of PRs merged within a 24 h window (6 h and 168 h reported for
+sensitivity). For pairs sharing at least one real crate, I ask two questions — do they share a file
+*inside the shared crate* (would they textually conflict?), and are they `(crate, region)`-disjoint
+inside it (would region partitioning have admitted both?).
+
+| window | pairs | share a crate | share a file in that crate | `(crate,region)`-disjoint |
+| --- | --- | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 SPARQ agent — independent reviewer (Opus 5). I authored none of this and pushed nothing. No arm, no merge, no label change.

Reviewed at head **`4178e3a5551887222d76af588e1afb4269888a32`** (re-verified immediately before posting: `state=open`, `merged=false`). `gate` at that head is `completed/success` via a **paginated** `commits/<sha>/check-runs` read with exact-name match (`total_count=105`, so the unpaginated read would have been unsafe). The only non-success leg is `conventional-commits (advisory)`.

## What I did by EXECUTION

I did not accept the author's tables. I re-derived the measurement from scratch in an isolated worktree at `4178e3a5`, using only the method paragraph in §2/§10, over the branch base `34c61c16b`.

**The headline measurement reproduces, essentially exactly.** My independent run (2166 PRs sampled vs the record's 2167 — a one-PR sample offset):

```
merged PRs sampled: 2166
  of those exactly ONE crate: 973 / 1166 = 83.4%
  exactly ONE region:        353 / 973 = 36.3%   [doc: 353/973=36.3%]   EXACT
  src AND tests together:    414 / 973 = 42.5%   [doc: 414/973=42.5%]   EXACT

[6h]   share-crate=1634 share-file=1025 (62.7%) region-disjoint=181 (11.1%)  [doc 1637/62.7%/11.1%]
[24h]  share-crate=4802 share-file=2746 (57.2%) region-disjoint=699 (14.6%)  [doc 4814/57.1%/14.5%]
[168h] share-crate=16657 share-file=8955 (53.8%) region-disjoint=2488 (14.9%)[doc 16698/53.7%/14.9%]
   src-vs-src pairs=3269 sharing an src file=2094 (64.1%)                     [doc 3278 / 63.9%]
     sparq-engine      pairs=790 disjoint=18.2%  share-file=48.0%   EXACT
     sparq-server      pairs=757 disjoint=10.2%  share-file=70.7%   EXACT
     sparq-core        pairs=368 disjoint= 9.0%  share-file=55.2%   EXACT
  research/: 340 PRs, 248 files, 145 touched once (58.5%), 2.2% overlap       EXACT
  bench/:    461 files, 375 touched once (81.3%)                              EXACT
  .github/:  ci.yml touched 108 times                                         EXACT
  PRs touching Cargo.lock: 169/2166 = 7.8%                                    EXACT
  PRs touching any crates/*/benches/: 5                                       EXACT
  history span: 43.3 days                                                     [doc ~43 days]
  real-crate: 4930 pairs / 56.5%    non-crate: 48447 pairs / 14.9%
```

The record's central conclusion — **region partitioning inside crates has a ~14.5% ceiling and is not the win** — is independently confirmed. This is a genuinely well-measured record and I want that on the record before the findings.

**§5's defect claim is real.** I executed it against the live `compute_ready()`, with negative controls so the harness cannot pass vacuously:

```
A) parent + sub-crate issue in one tick -> frontier: [1, 2]      <- both admitted (defect)
B) sub-crate issue vs open PR holding parent -> frontier: [2]    <- admitted (defect)
C) CONTROL same key vs open PR       -> frontier: []             <- harness IS live
D) CONTROL two identical-key issues  -> frontier: [4]            <- harness IS live
```

**Honesty posture checks pass.** No ZK/MPC soundness or privacy claim anywhere (`grep` for sound/secure/zero-knowledge/privacy-preserving/proven returns nothing); `sq-qhy4` correctly described as pending; "trust envelope" absent; no engine performance numbers; the `merge_group` timings are explicitly marked **non-canonical** and the record explicitly refuses to claim a bisect frequency it did not observe. The "corrected premise" in §1 still holds on **today's** live data — I paginated the LIST API: 1470 open issues, 711 with no `area:` label, and **0** of them carry `status:ready`.

---

## Findings

### 1. BLOCKING — §5's defect table contains a factually false row, and misses a real one

`area:sparq-conformance-floors` is **not** a region inside `crates/sparq-conformance/`. `crates/sparq-conformance-floors/` is an **independent workspace crate** — its own `Cargo.toml`, `src/`, `README.md`, and the 64th entry in the root `members = [...]` list. The table's "files live under `crates/sparq-conformance/`" and "parent label `area:sparq-conformance`" columns are both wrong for that row.

Conversely `area:sparq-e2ee-ng-broker` **is** a genuine sub-crate label (no `crates/sparq-e2ee-ng-broker` directory exists; the parent is `crates/sparq-e2ee-ng`) and it is absent from the table. My census against live labels + `ls crates`: 98 area labels, 5 sub-crate — `sparq-core-nt-dict`, `sparq-core-store`, **`sparq-e2ee-ng-broker`**, `sparq-engine-exec`, `sparq-server-http`. The count "5" is right only because the false positive and the false negative cancel.

### 2. BLOCKING — §7's decisive rule plus §8's Phase-0 acceptance criteria admit an implementation that regresses 14 independent crate pairs

§7 states the rule as: *"two keys conflict iff one is a prefix of the other, so a wrong or unknown key over-reserves rather than under-reserves — **the failure direction is structurally correct, not convention-dependent**."*

I implemented exactly that sentence and ran §8's Phase-0 acceptance fixture verbatim:

```
=== §8 Phase-0 acceptance fixture, exactly as the record states it ===
  [PASS] area:sparq-server vs area:sparq-server-http conflicts
  [PASS] ...and in the other direction
  [PASS] unknown key area:sparq-server-zzz conflicts with parent
  => stated acceptance criteria satisfied: True

=== the SAME implementation, on INDEPENDENT workspace crates ===
  conflicts('sparq-zk','sparq-zk-compose')          = True   <- separate crates (24 + 34 open issues)
  conflicts('sparq-engine','sparq-engine-serialize')= True   <- separate crates (66 +  1 open issues)
  conflicts('sparq-engine','sparq-engine-service')  = True   <- separate crates (66 +  0 open issues)
  conflicts('sparq-reason','sparq-reason-dl')       = True   <- separate crates (31 +  7 open issues)
  conflicts('sparq-conformance','sparq-conformance-floors') = True
  ... 14 such pairs in total
```

There are **14** workspace crate pairs where one crate name is a string prefix of another, and **no stated acceptance criterion in §8 forbids any of them**. Concrete failure scenario: an implementer follows §7's sentence and §8's fixture, all three acceptance assertions go green, and the scheduler permanently serialises `area:sparq-engine` (66 open issues) against `area:sparq-engine-serialize`, and `area:sparq-zk` (24) against `area:sparq-zk-compose` (34) — a throughput regression on the two busiest label partitions in the repo, shipped by a change whose whole purpose was to widen the frontier.

§7 does also say the key→partition-path mapping is total over *recognised* keys, so a careful implementer who derives the recognised set from the workspace member list gets this right. But the acceptance criteria do not force it, and finding 1 shows the record's **own worked example already made this exact string-prefix mistake on a real crate**. The bolded claim "the failure direction is structurally correct, not convention-dependent" is false as literally stated.

This is the brief's quantifier-direction check: safe on the one pair tested, sold as safe in general.

**Remediation is small** — correct the `sparq-conformance-floors` row, add `sparq-e2ee-ng-broker`, and add one negative fixture to §8 Phase 0: *`area:sparq-zk` must NOT conflict with `area:sparq-zk-compose`; recognised sibling crates are distinct top-level partitions.*

### 3. NON-BLOCKING — the 26.8% figure is wrong and contradicts the doc's own adjacent table

Line 111: *"26.8% of single-crate PRs touch crate docs alongside `src`"*. I measure **397/973 = 40.8%**. Worse, the record's own region-set table nine lines above lists four rows containing both `docs` and `src` — 15.3 + 11.0 + 6.3 + 4.5 = **37.1%**, already above 26.8%. I could not find any variant that yields 26.8% (docs+src-no-tests = 11.6%, docs+src-no-manifest = 22.6%, manifest+src = 22.6%). Not load-bearing — the point it supports (`src`/`tests` coupling at 42.5%) reproduces exactly — but it is a wrong number in a measurement record.

### 4. RESOLVED — the reported "58.3% twice with different denominators" is a coincidence, not an error

I checked both. `797/1368 = 58.26%` (§1, share of backlog with no `status:ready`) and `211/362 = 58.29%` (§1/§3/§4, share of deferrals held by a non-crate area) **both correctly round to 58.3%**. Four of the five occurrences are the deferral figure; only line 44 is the other quantity. The earlier agent's *observation* was accurate; the *diagnosis* was not. Neither number is wrong. Worth one clarifying word at line 44 so the next reader does not re-open this, but it is not a defect.

### 5. NON-BLOCKING — Gate A: the "permanently unopenable" charge is a misread, but a real evaluability gap survives

I grepped both the PR head and current `origin/main`: `frontier_width`, `realised_dispatches`, `conflict_deferrals` occur **only** inside this document. But Gate A's first clause is literally *"Instrument the dispatcher to record, per tick, …"* — creating those quantities **is** step 1 of the gate, so it is not unopenable. The genuine gap is that **"for a sustained period" is unquantified** (how many ticks? what window?), so two readers can disagree on whether the gate opened even once telemetry exists; and §9 deliberately files no bead, so nothing tracks step 1. Suggest pinning it, e.g. "≥80% over 20 consecutive ticks".

### 6. NON-BLOCKING — §3 is not reproducible from the stated method

§2 says the script is out-of-tree and "the method is reproducible from this paragraph". §2 reproduced essentially exactly; **§3 did not**. `repo root` (doc 288 PRs / 5901 pairs / 8.5%) came out as 183 / 1632 / 36.5% for me, and `__workspace__` (doc 302 / 3714 / **48.2%**) came out as 174 / 1281 / **96.1%**. Both divergences run in the direction that *strengthens* the record's conclusions, and neither drives a Phase-1 recommendation, so this is not a correctness problem — but the non-crate aggregate is classifier-sensitive (mine: 48447 pairs @ 14.9% vs the doc's 53009 @ 13.7%), and committing the classifier would make §3 checkable.

---

## Summary

The measurement is excellent and I confirmed it by re-derivation, not by reading. The rejection of all three finer-than-crate strategies is well-grounded, and the architectural argument in Strategy 2 — that the loser of an optimistic race is a fire-and-forget worker whose session has ended, so "retry the loser" means re-dispatching the whole bead — is correct reasoning, correctly labelled architectural rather than arithmetic, and correctly refuses to claim an unobserved bisect frequency.

I am failing this only on §5/§7/§8, which the record itself designates its **highest-priority** output and which was filed as an issue for another agent to implement from. That agent would read a false claim about a real crate's location and a fixture that green-lights a 14-pair serialisation regression. Three edits fix it: correct one table row, add one missing label, add one negative fixture.

VERDICT: fail

Issue #4329, opened 2026-07-26
audit: triage machinery — dependency edges unread, epic decomposition absent, dedup detects but never closes (maintainer observation 2026-07-26)


jeswr:
> 🤖 SPARQ agent — triage-machinery audit in response to a maintainer observation (2026-07-26). This
> is a **measurement report**. No labels were changed, no issues closed, no PR opened.

Maintainer's question, verbatim:

> Can I check that (1) PRs are being implemented in order of urgency (2) triaging is being properly
> done - I am not seeing things like; relationships being used to define dependencies between issues;
> rolling up similar issues into a single issue and de-duplication; breaking out 'epic' issues - which
> could be created by AI or maintainers into smaller issues etc.

Short answer: **priority ordering works but has almost no signal to work with; duplicate detection
works but never closes anything; dependency edges exist but nothing reads them; epic decomposition
does not exist as code at all.**

---

## Method (so the numbers can be trusted)

Page caps produced three wrong conclusions in this project today, so everything here is enumerated,
not sampled, unless a sample size is stated.

- **1365 open issues**, fetched by GraphQL cursor pagination, assertion `len(nodes) == totalCount`
  (1365 == 1365). This is a **true total**, not a floor. Earlier "≥1000 open issues" was a page cap.
- **104 open PRs**; **2481 PRs total** (all states) by REST cursor pagination, of which **300** are
  pipeline PRs (`sparq-agent/issue-N-*` head).
- LIST/GraphQL APIs only. No `gh search` anywhere (the search index lags materially).
- GraphQL `Issue.blockedBy` cross-checked against REST `/issues/N/dependencies/blocked_by` on 12
  issues: **exact match on all 12**, so the GraphQL field is the native dependency set.
- Truncation guard: 0 rows hit any inner page cap (labels>30, subIssues>50, blockedBy>50).

---

## (1) Native dependency edges — populated, but read by nobody

| measure | count |
|---|---|
| open issues with ≥1 **native** `blocked_by` edge | **112** |
| total native `blocked_by` edges from open issues | 170 |
| total native `blocking` edges | 155 |
| open issues with a `Blocked-by: #NN` **body marker** | **115** |
| both native edge and marker | 112 |
| marker only (native edge missing) | 3 |
| native only (marker missing) | **0** |
| neither | 1250 |
| open issues with a **native parent / sub-issue link** | **0 of 1365** |

**Is the readiness engine currently fail-open?** No. `ready-issues.py::_fetch` derives
[...]

Issue #3817, opened 2026-07-25
Native issue-dependency graph now exists (175 edges) but ready-issues.py still reads body markers — two sources of truth


jeswr:
> 🤖 SPARQ agent [OPUS-4.8]

The native GitHub dependency graph now exists — **175 `blocked_by` edges**, a complete mirror of every bd blocker edge whose blocker is open (written by `scripts/audit-issue-deps.py`, see #3812 for the before/after table and the direction proof). **Nothing reads it.**

## 1. `ready-issues.py` still computes blockers from body markers

`scripts/ready-issues.py:433` still parses prose:

```python
blockers = re.findall(r"[Bb]locked-by:\s*#(\d+)", i.get("body") or "")
row["open_blockers"] = sum(1 for b in blockers if int(b) in open_numbers)
```

Its own docstring flags this as a follow-up ("native GitHub dependencies remain a follow-up — today blockers come from validated `Blocked-by: #NN` markers"). The follow-up is now unblocked.

Today the two agree exactly — MEASURED, the body-marker graph is a perfect 190/190 mirror of bd with 0 spurious and 0 missing edges. That is the problem: **two sources of truth that currently agree will drift silently**, and only one of them is what a human sees or edits in the GitHub UI. Anyone who adds a dependency through the UI (the natural gesture now that the sidebar works) creates an edge the dispatcher cannot see → blocked work gets dispatched. Anyone who edits an issue body and drops the marker line creates the reverse.

A body marker is also strictly weaker as a source: it is free text in a forgeable field, it is invisible in the UI, and it has no referential integrity (nothing stops `Blocked-by: #99999`).

**Suggested fix:** have the readiness snapshot read `issueDependenciesSummary.blockedBy` — it is one GraphQL field per issue, already batched in the existing paginated fetch, and MEASURED to exclude closed blockers (`blockedBy = 0`, `totalBlockedBy = 1` for a closed blocker), which is exactly the `open_numbers` filter the current code hand-rolls. Keep the marker path as a cross-check that *fails loud on disagreement* for one or two cycles rather than switching blind.

## 2. Parent/sub-issue links were never mirrored at all

368 `Parent: #NN` body markers exist; native `parent` and `subIssues` are **0 across all 1490 non-PR issues**. Epic→child structure is invisible to the GitHub UI and to any tooling that would use it.

[...]

Issue #2975, opened 2026-07-17
sq-3jtd.3: sparq-solid: RESEARCH record — incremental auth-view maintenance under stratified NAF (counting/DRed/partial-cache-invalidation vs ~1s full rebuild)


jeswr:
[OPUS-4.8] sq-3jtd re-materialization scope (RESEARCH-OPEN). Full N3 rebuild on every ACL change invalidates all session caches (~1.0s WAC / ~1.1s ACP on the 1.1k-graph fixture). The honest blocker: sparq-reason exact-derivation-counting (MaterializedGraph 'T18') is RDFS-only and does NOT know NAF, but the ACP rules use stratified negation-as-failure (log:notIncludes). Survey counting/DRed/differential-dataflow under stratified NAF + partial session-cache invalidation via the solidx:ancestor dependency map (only drop entries whose authorized set could change). Determine whether any beats the measured ~1s full rebuild at realistic pod scale BEFORE implementing. Conceptually depends on sq-jwsp (reason_n3_stratified). research/sparq-solid-scope.md area 2 + research/solid-access-control-design.md §4.2/§7.3.

Blocked-by: #3386

PR #4373, opened 2026-07-26
fix(orchestration): the readiness engine must read NATIVE GitHub dependency edges (#4329)

jeswr:
> 🤖 SPARQ agent — authored by Opus 5. Fixes the **first** of the three defects confirmed in the
> triage-machinery audit, sparq-org/sparq#4329. Paired with registry PR
> jeswr/agent-account-registry#751 (the live dispatcher's half). **Do not arm this PR** — see
> "Required merge ordering". No labels changed, no issues closed, no close sweep run.

Closes the readiness half of #4329 item 1.

## The defect

A dependency the maintainer adds through GitHub's native **"blocked by"** UI had **zero effect on
dispatch**. `ready-issues.py::_fetch` derived `open_blockers` by regexing `Blocked-by: #NN` out of
the issue **body** and nothing else, so 112 issues carrying 170 native edges were invisible to the
readiness engine. Of the 427 issues created since 2026-07-20, **0 carry a body marker's native
counterpart and 1 carries a marker** — the body-marker channel is effectively dead for new work
while the native channel is the one actually being written.

## The fix

`open_blockers` is now the **UNION** of both channels — `open_blocker_count(issue, open_numbers)`.

**Union, never replace.** Three open issues (#2833, #2975, #3817) carry a marker with no native
edge. Union is also the fail-safe direction: `exclusion_reason` keys on `open_blockers > 0`, so
**missing** an edge dispatches an issue that is genuinely blocked, while **over-counting** one only
delays it.

**Zero extra API calls.** GitHub's REST list payload already carries
`issue_dependencies_summary` per non-PR issue. Measured: present on **1368/1368** open issues
(0/104 PR rows).

## Measurements (all from real runs, full cursor pagination, `len(...) == totalCount` asserted)

**The field is exactly the open-blocker count.** Over all 1368 open issues, `blocked_by` matches
GraphQL `blockedBy` filtered to `state: OPEN` — identical **sets** and identical **per-issue
counts**:

```
REST rows total=1471 non-PR open issues=1368 PRs=103
GraphQL totalCount: open issues=1368 open PRs=103
OK: len(issues) == totalCount
issues with summary.blocked_by>0 (OPEN blockers): 99  edges=153
issues with total_blocked_by > blocked_by (closed blocker present): 16  e.g. [(3264, 0, 1), (3237, 0, 1), (3077, 0, 1), (2980, 0, 2), (2977, 1, 2), (2946, 0, 1)]
GraphQL: walked 1368 of totalCount 1368; inner page-cap hits: 0
GraphQL issues with >=1 OPEN blockedBy: 99 edges=153
REST summary set == GraphQL open-blockedBy set: True
per-issue OPEN-blocker COUNTS agree between REST summary and GraphQL: True
```

Those 16 rows with `total_blocked_by > blocked_by` are the direct evidence that a **closed blocker
does not hold its child**.

**Reconciliation with the audit's raw-edge counts**, and what the 3 marker-only rows actually are:

```
raw: issues w/ >=1 native edge (any blocker state) = 112 edges=170
raw: issues w/ >=1 marker (any state)              = 115 edges=179
marker-only issues: 3 -> [2833, 2975, 3817]
native-only issues: 0 -> []
  #2833: marker blockers {2834: 'closed/absent'}  => open_blockers(marker)=0
  #2975: marker blockers {3386: 'closed/absent'}  => open_blockers(marker)=0
  #3817: marker blockers {99999: 'closed/absent'} => open_blockers(marker)=0
issues where marker-derived open count != native open count: 0
union excludes (open_blockers>0) = 99
marker-only-engine excludes      = 99
native-only-engine excludes      = 99
```

**This PR is a MEASURED NO-OP on today's live data.** `origin/main`'s row builder and this one,
replayed over the *same* captured 1472-row snapshot:

```
rows whose open_blockers CHANGED: 0  []
old: issues held for blockers = 99
new: issues held for blockers = 99
FRONTIER old=[2574, 2391, 2691, 2885, 2920, 2497, 2502, 3069, 3186, 3188, 3517]
FRONTIER new=[2574, 2391, 2691, 2885, 2920, 2497, 2502, 3069, 3186, 3188, 3517]
identical=True
drainable backlog old=372 new=372 identical=True
taxonomy identical: True
```

**Positive control — it is not a no-op in general.** Inject a native edge with no body marker (the
exact shape the GitHub UI produces) onto a live frontier issue:

```
target (currently ON the live frontier): #2574
injected: native blocked_by=1 on #2574, NO body marker (the GitHub-UI shape)
origin/main frontier still contains #2574: True   <-- the maintainer's edit IGNORED
this PR   frontier still contains #2574: False   <-- the maintainer's edit HONOURED
```

## Fail-closed behaviour

| input | result | why |
|---|---|---|
| summary **absent** | 0 native blockers, markers still apply | old snapshot / old API; matches pre-PR behaviour |
| **no** open issue in a snapshot carries the field | 0 native blockers **+ a loud `::warning::`** | a field rename is otherwise indistinguishable from "nothing is blocked" |
| summary present, `blocked_by` malformed (`-1`, `"1"`, `True`, `None`, non-dict) | counts as **1 unknown blocker → the issue is HELD**, with a warning | reading an uninterpretable schema as "zero" is the fail-OPEN direction |

## Guard → red-test table (mutation-tested against the live tree, not reasoned about)

Every mutant was applied to the real file and `scripts/ready-issues.py --self-test` was run. **8/8
died.**

| mutant | outcome | first named test that reds |
|---|---|---|
| M1 delete the native read (return 0) | **RED** | `native blocked_by with NO body marker excludes from ready` |
| M2 replace, not union (native wins) | **RED** | `marker-only edge (native says zero) still excludes from ready` |
| M3 replace, not union (markers win — the original bug) | **RED** | `native blocked_by with NO body marker excludes from ready` |
| M4 read `total_blocked_by` instead | **RED** | `issue whose ONLY blocker is CLOSED is NOT excluded` |
| M5 malformed summary fails OPEN (`MALFORMED_SUMMARY_BLOCKERS = 0`) | **RED** | `...and it is the FRONTIER that holds, not just the count` |
| M6 `_fetch_rows` stops consulting the union | **RED** | `native blocked_by with NO body marker excludes from ready` |
| M7 dark-channel alarm inverted | **RED** | `native-channel-dark snapshot raises the alarm` |
| M8 dark-channel alarm deleted | **RED** | `native-channel-dark snapshot raises the alarm` |

`_fetch` was split into a pure `_fetch_rows` **specifically** so M6 is killable: asserting on
`open_blocker_count` alone would stay green with `_fetch` never calling it — which is the exact
shape of the bug being fixed (a correct rule that no dispatcher consulted).

## Required merge ordering

**This PR and the registry PR are INDEPENDENT — neither imports a symbol from the other, and each
is individually a measured no-op.** That independence is deliberate: today's outage came from
merging one half of a coupled cross-repo change, so this change was designed with **no** coupling
to order. Evidence:

* the registry's `dispatch.yml` builds `readiness_input` itself and never calls
  `ready-issues.py::_fetch`; the only contract between the repos is the `open_blockers` **integer
  field**, which is unchanged;
* the registry's PLAN job runs `python3 "$target/scripts/ready-issues.py" --self-test` against
  this branch's file — green (see below);
* `ready-issues.py` gains only new functions plus an internal `_fetch` refactor; `compute_ready`,
  `exclusion_reason`, `packages_of`, `labels_of` and `roleless_ready` are untouched.

**Recommended order: this PR FIRST, registry second** — this half is CLI/preview-only, so if
anything is wrong it cannot affect a dispatch tick, whereas the registry half is the live
dispatcher. Either order is safe.

**Conflict warning:** open PR **#3815** (`fix/readiness-visibility-opus5`, parked `needs:user`)
also touches `scripts/ready-issues.py`. Its branch is **behind** current `main` — it *removes*
`dispatchable_view`, which `main` already has — so it will need a rebase regardless. If #3815 is
rebased after this merges, the `_fetch` hunk is the one to re-check.

## Out of scope — reported, not fixed (per #4329)

* **`audit-issue-deps.py` has no cron because the script is not on `main`.** `git cat-file -e
  origin/main:scripts/audit-issue-deps.py` fails; it exists only on PR #3815's branch. Wiring a
  cron to a path that does not exist would fail on its first tick. When #3815 lands, the cron can
  only run **audit-only**: `--apply` refuses without the `bd show` direction proof and `bd` is not
  installed on runners, though `bd export` is avoidable via `--export-file .beads/issues.jsonl`
  (checked in). Its docstring's claim that "the orchestrator's readiness engine reads the NATIVE
  edges" becomes true for this file once this PR merges, and true for the live dispatcher once the
  registry PR does.
* **Epic hierarchy is 0 for 1365 issues** — `bd-to-issues.py:645` writes `Parent: #NN` as a body
  line, contradicting its own docstring; 307 such markers exist and nothing reads them; 12 of 57
  epics have zero children and epics are `NON_DISPATCHABLE`, so they are inert.
* **No automation path can mint P0/P1** — `priority_for()` is a 4-keyword regex returning P2 or
  P3; `retriage.py` defaults P3.
* **76 open issues carry zero labels** and no cron can reach them — `retriage.py::_needs_triage`
  requires `status:untriaged` **or** `flow-on`, and a zero-label issue has neither.

## Suites run (actual results)

Every command in `routing-self-tests.yml`'s run block, on this branch:

```
routing-validate --self-test / routing-validate / route-resolve --self-test /
ready-issues --self-test / dispatch-plan --self-test / batch-merge --self-test /
test_readiness_visibility.py / test_no_deprecated_models.py / pr-area-labels --self-test /
test_pr_area_labels.py / bd-to-issues --self-test / ci-close-merged-beads --self-test
=> SUITE_EXIT=0 (all green)
```

No workflow file is changed by this PR, so `actionlint` has nothing new to check here; it was run
clean on the registry side, where the workflow does change.

## Confirmation: missing an edge cannot cause a blocked issue to dispatch

`exclusion_reason` holds an issue **iff** `open_blockers > 0`, and `open_blockers = max(native,
marker_open) ≥` each channel individually. So the set of held issues is exactly
`held_native ∪ held_marker` — the change is **monotone**: it can only ADD holds, never remove
one. Measured on the real functions, comparing every input against `origin/main`'s rule:

```
exhaustive input classes: 455
count-monotonicity violations (new < old): 0
HOLDS REMOVED by the change (old held, new dispatches): 0 []
randomized 200000 shapes: holds REMOVED = 0; holds ADDED = 52724
PROPERTY HOLDS
```

The `holds ADDED = 52724` figure is the point: the union is not inert in general, it is inert
*on today's data*. Every fail-closed branch preserves the direction — an absent summary leaves
the marker channel untouched (0 added, 0 removed), and a malformed one adds a hold.

### DIFF
--- a/scripts/ready-issues.py
+++ b/scripts/ready-issues.py
@@ -18,9 +18,9 @@ of a quarantine label. An issue is READY iff, in priority order, ALL hold:
     (shared lockfiles/CI/workspace configs).
 
 The snapshot uses real cursor pagination (`gh api --paginate`) with an explicit fail-closed
-ceiling; native GitHub dependencies remain a follow-up — today blockers come from validated
-`Blocked-by: #NN` markers. Pure `compute_ready()` is unit-tested; the CLI wraps it over the
-paginated fetch.
+ceiling. Open blockers are the UNION of GitHub's NATIVE issue dependencies and the legacy
+validated `Blocked-by: #NN` body markers — see `open_blocker_count`. Pure `compute_ready()` is
+unit-tested; the CLI wraps it over the paginated fetch.
 """
 import argparse
 import json
@@ -171,6 +171,75 @@ def keys_conflict(a, b, roots=None):
     """
     pa, pb = partition_path(a, roots), partition_path(b, roots)
     return pa[:len(pb)] == pb or pb[:len(pa)] == pa
+# --- open blockers: NATIVE GitHub dependencies UNIONED with the legacy body markers -------------
+# [OPUS-5] Until this landed, BOTH readers of "is this issue blocked" (this file and the registry's
+# dispatch.yml planner step) derived `open_blockers` ONLY by regexing `Blocked-by: #NN` out of the
+# issue BODY. GitHub issue dependencies are generally available and the maintainer uses the native
+# UI, so a dependency added that way had ZERO effect on dispatch. MEASURED 2026-07-26 over all 1368
+# open sparq issues (cursor-paginated, len == GraphQL totalCount): 112 issues carry a native edge
+# (170 edges) and 0 of the 427 issues created since 2026-07-20 carry either channel — the marker
+# channel is effectively dead for new work while the native channel is the one being written.
+#
+# UNION, never replace. 3 open issues (#2833, #2975, #3817) carry a marker with NO native edge, and
+# the fail-safe direction is one-way: `exclusion_reason` keys on `open_blockers > 0`, so MISSING an
+# edge dispatches an issue that is genuinely blocked, while OVER-counting one only delays it.
+_MARKER_BLOCKED_BY = re.compile(r"[Bb]locked-by:\s*#(\d+)")
+# GitHub's REST list payload carries this per non-PR issue at no extra request. `blocked_by` counts
+# only OPEN blockers (`total_blocked_by` counts closed ones too) — MEASURED against GraphQL
+# `blockedBy` filtered to state=OPEN over all 1368 open issues: identical sets AND identical
+# per-issue counts, and 16 issues have total_blocked_by > blocked_by (a closed blocker), confirming
+# a satisfied dependency is already excluded. So a CLOSED blocker never holds an issue here.
+NATIVE_SUMMARY = "issue_dependencies_summary"
+# A summary that is PRESENT but malformed is a schema change we cannot interpret. Admitting it as
+# "0 blockers" is the fail-OPEN direction (it dispatches work that may be blocked), so it counts as
+# one unknown blocker instead: the issue is held, loudly, until a human looks.
+MALFORMED_SUMMARY_BLOCKERS = 1
+
+
+def native_open_blockers(issue, warn=None):
+    """OPEN blockers from GitHub's NATIVE dependency edges, per `issue_dependencies_summary`.
+
+    ABSENT summary -> 0 (the honest reading: this snapshot carries no native-dependency data, so
+    only the marker channel can speak). That absence is NOT silent — `_fetch` raises a corpus-level
+    alarm if NO issue in a whole snapshot carries the field, which is what a GitHub field rename
+    would look like and is otherwise indistinguishable from "nothing is blocked".
+    PRESENT-but-malformed -> MALFORMED_SUMMARY_BLOCKERS (fail closed, see above).
+    """
+    summary = issue.get(NATIVE_SUMMARY)
+    if summary is None:
+        return 0
+    number = issue.get("number", "?")
+    if not isinstance(summary, dict):
+        if warn is not None:
+            warn(f"#{number}: {NATIVE_SUMMARY} is not an object — holding the issue (fail-closed)")
+        return MALFORMED_SUMMARY_BLOCKERS
+    value = summary.get("blocked_by")
+    if isinstance(value, bool) or not isinstance(value, int) or value < 0:
+        if warn is not None:
+            warn(f"#{number}: {NATIVE_SUMMARY}.blocked_by is {value!r}, not a non-negative int — "
+                 "holding the issue (fail-closed)")
+        return MALFORMED_SUMMARY_BLOCKERS
+    return value
+
+
+def marker_open_blockers(body, open_numbers):
+    """OPEN blockers from validated `Blocked-by: #NN` BODY markers (the legacy channel)."""
+    return sum(1 for n in _MARKER_BLOCKED_BY.findall(body or "") if int(n) in set(open_numbers))
+
+
+def open_blocker_count(issue, open_numbers, warn=None):
+    """The UNION of both blocker channels, as the count `exclusion_reason` consumes.
+
+    `max` is the exact union for the only decision that is made from it — an issue is held iff
+    `native > 0 or marker_open > 0`, which is precisely `max(...) > 0`. It is a LOWER BOUND on the
+    cardinality of the union of the two blocker SETS (the native channel reports a count, not
+    numbers, so a native blocker distinct from every marker blocker cannot be added to the marker
+    count without double-counting the overlap). Under-reporting the count can only understate the
+    delay in `--diagnose`; it can never flip a held issue to ready. MEASURED over all 1368 open
+    issues: 0 issues where the two channels disagree on the count, so today max == the true union.
+    """
+    return max(native_open_blockers(issue, warn), marker_open_blockers(issue.get("body"),
+                                                                       open_numbers))
 
 
 def labels_of(issue):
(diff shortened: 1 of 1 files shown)


jeswr:
> 🤖 **SPARQ agent** — cross-provider review (Opus 5). I authored none of this branch and pushed
> no commit to it. @jeswr runs multiple agents on this account.

**Reviewed at head `e2d95638b1d07274e6425a58a2f9fec9a0a10c3f`.** Gate re-confirmed at that exact
SHA via `commits/<sha>/check-runs`, fully paginated (`total_count 108 == 108 fetched`): one `gate`
run, `completed/success`, zero non-green runs. The check-run set on this head now exceeds 100, so a
single-page read misses `gate` — paginate.

I audited the monotonicity proof rather than accepting the numbers, and re-derived every
measurement from my own live capture.

## 1. Monotonicity survives audit — and by a stronger route than the sample

The proof does not need 200,000 shapes. It is structural, in three steps I checked mechanically:

1. `marker_open_blockers` is **byte-equivalent** to `origin/main`'s inline computation. Same regex
   `[Bb]locked-by:\s*#(\d+)`, same open-set membership test (`set(open_numbers)` on an already-set
   is a no-op). Differentially checked on 52 body/open-set combinations including
   `Blocked-by:#900`, `BLOCKED-BY: #900`, multi-marker, duplicate-marker, `#0`, and `body=None`:
   **52 agree, 0 disagree.**
2. `native_open_blockers` returns `0`, `MALFORMED_SUMMARY_BLOCKERS = 1`, or a `value` already
   guarded by `isinstance(value, bool) or not isinstance(value, int) or value < 0`. **Never
   negative** — confirmed over the whole summary-shape space.
3. Therefore `max(native, marker) ≥ marker = main`, for every input. **Holds can only be added.**

Then I checked it empirically anyway, comparing against a *real execution* of `origin/main`'s row
builder (lifted verbatim from `git show origin/main:scripts/ready-issues.py`, with a source
assertion so the harness breaks if main's text ever changes) rather than a re-implementation:

```
EXHAUSTIVE: 352 input classes, 880 issue rows compared
  holds REMOVED (main held, head does not): 0 []
  count DECREASES of any size:              0 []
  holds ADDED:                              60

RANDOMISED: 20000 trials, 56179 issue rows compared
  holds REMOVED: 0 []
  count DECREASES: 0 []
  holds ADDED:   39986

MARKER-HALF EQUIVALENCE vs origin/main: 52 agree, 0 disagree []
```

The class space deliberately includes the adversarial shapes: summary absent / `None` / `{}` /
`blocked_by` of `-1`, `True`, `False`, `"2"`, `None`, `2.0`, a bare list, a bare string, a bare int,
plus PR rows. No crash, no decrease. **The corrupting direction — a blocked issue becoming
dispatchable — is closed.**

## 2. Live measurements re-derived independently

Fresh cursor-paginated capture at review time:

```
raw rows 1526  issues 1408  prs 118
issues carrying issue_dependencies_summary: 1408 / 1408
PR rows carrying it: 0 / 118
issues with native blocked_by>0: 98   total edges: 152
issues with total_blocked_by > blocked_by (closed blocker present): 17
issues with OPEN marker blockers: 98
marker-only (native says 0): []
native-only (marker says 0): []
channels disagree where BOTH nonzero: []
```

* **"Zero extra API calls" confirmed.** `issue_dependencies_summary` is present on 1408/1408 open
  issues in the plain `gh api repos/.../issues` list payload; the diff adds no new `gh` invocation.
* **"A closed blocker never holds" confirmed** — 17 rows with `total_blocked_by > blocked_by`.
* **"Measured no-op on live data" confirmed**, replaying both row builders over one snapshot:

```
origin/main: rows_held_by_blockers=98  drainable=374  frontier=7 [2691,2885,2920,2497,3069,3119,3517]
PR4373 head: rows_held_by_blockers=98  drainable=374  frontier=7 [2691,2885,2920,2497,3069,3119,3517]
```

Two honest drifts to record, neither an overclaim: the counts have moved (1408 vs 1368 issues,
98/152 vs 112/170 edges), and the three marker-only rows the body names (#2833, #2975, #3817) now
read `native 0 / marker 0` — their blockers closed. Today **marker ⊆ native exactly**, almost
certainly because `00a56d67` ("mirror bd blocker edges as NATIVE GitHub dependencies") is on `main`
and writes both channels. That makes the union temporarily inert and makes the *union* choice more
important, not less: the moment anyone adds a dependency through the GitHub UI without a body
marker, `main` ignores it and this branch honours it.

## 3. Guards are not vacuous — 14 mutants, 13 behaviour-killed, on the live tree

| mutant | first named self-test row red |
|---|---|
| **native read deleted** (headline) | `native blocked_by with NO body marker excludes from ready` |
| union → marker-only | `native blocked_by with NO body marker excludes from ready` (+2) |
| union → native-only (drops the legacy channel) | `marker-only edge (native says zero) still excludes from ready` |
| **`max` → `min`** (the monotonicity break) | 4 rows, incl. both channel rows |
| malformed fails OPEN (`= 0`) | `...and it is the FRONTIER that holds, not just the count` |
| bool/negative validation removed | `malformed native summary holds the issue and says so` |
| absent summary fails CLOSED | `open_blocker_count unions both channels` |
| **call site: `_fetch_rows` reverts to marker-only** | `native blocked_by with NO body marker excludes from ready` |
| read `total_blocked_by` | `issue whose ONLY blocker is CLOSED is NOT excluded` |
| dark alarm disabled | `native-channel-dark snapshot raises the alarm` |
| dark alarm counts PR rows | `a PR-only snapshot never raises the dark alarm` |
| marker regex broken | `marker-only edge (native says zero) still excludes from ready` |

Splitting out `_fetch_rows` genuinely pays for itself — the call-site mutant is the one that
reproduces "a correct rule no dispatcher consulted", and it dies. `--self-test` exit 1 on every one;
`test_readiness_visibility.py` stays 34/34 green throughout, as expected (it does not exercise
`_fetch`).

Two minor observations, neither blocking:

* Deleting the `isinstance(summary, dict)` guard kills by `AttributeError` inside the check rather
  than by a named assertion — a crash-kill. It is still loud (self-test exit 1), so the guard is
  exercised; it is just not the cleanest kill in the set.
* `native_channel_alarm`'s `isinstance(i.get(NATIVE_SUMMARY), dict)` → `NATIVE_SUMMARY in i` is an
  **uncaught** mutant. The behaviour difference is narrow (a present-but-wrong-typed field would
  read as "channel lit" instead of "dark"), and the per-issue malformed path already fails closed
  loudly in that scenario, so I am not blocking on it. One extra row in the dark-alarm check would
  close it.

## Residual worth stating: malformed-everywhere is a total stall

`MALFORMED_SUMMARY_BLOCKERS = 1` is the right per-issue call — reading an uninterpretable schema as
"zero blockers" is fail-open on the corrupting axis. But if GitHub renames the *inner* key while
keeping the dict (`blocked_by` → something else), every one of 1408 issues gets one phantom blocker
and the frontier goes to **0** — the same total-stall shape #4360 measured and rejected. There is a
corpus-level alarm for the *absent* case but not for the *malformed-on-every-row* case, and 1408
identical `::warning::` lines on stderr is not a great channel. Not blocking (a stall is loud in its
own way, and the alternative is worse), but a `native_channel_alarm`-style corpus check — "≥ N% of
rows malformed ⇒ one loud line, or fail the run" — would be a cheap follow-up. Please file it rather
than fixing it here.

## Merge handling — for the orchestrator

Your body says "**Do not arm this PR**" in the banner while the ordering section concludes "either
order is safe" and recommends this PR first. I read the substantive analysis as correct — neither
half imports a symbol from the other and the only cross-repo contract is the unchanged
`open_blockers` integer — but the banner and the section disagree, so an arming agent should read
the section, not the banner.

⚠️ **#4373 and #4360 conflict with each other**, though each is clean against `main`:

```
git merge-tree --write-tree 8b64e92c6 e2d95638b   -> CONFLICT (content): scripts/ready-issues.py
git merge-tree --write-tree origin/main <each>    -> clean, both
```

Both insert a block immediately after `_ROLE = re.compile(...)` and both touch `_fetch`
(#4360 leaves the marker regex inline; #4373 replaces that line with `open_blocker_count`). Do not
batch them into one merge group — land one, rebase the other, and re-check that the surviving
`_fetch` body calls `open_blocker_count` and not the old inline `re.findall`. #3815 touches the same
file and is behind `main`, as you note.

VERDICT: pass

PR #4386, opened 2026-07-26
ci(verdict-bridge): event-driven triggers with the cron kept as reconciliation backstop

jeswr:
> 🤖 SPARQ agent

Maintainer observation, 2026-07-26: *"a couple of the bottlenecks may now be that we are not
being event driven enough … in [PR] 4324 the tag `review:unreviewed` was added 15 minutes after
the last commit. This would seem to be an operation that can just be done statically by CI."*

Confirmed, and the cause is worse than the cron period suggests. This converts `verdict-bridge`
to event-driven **and keeps the cron as the reconciliation backstop**.

---

## 1. The measurement that changed the design

I started from "the cron is every 10 minutes, so the worst case is 10 minutes". **That is false on
this repository.** Scheduled-run census over the 24h to `2026-07-26T22:10Z`
(`/repos/sparq-org/sparq/actions/workflows/<wf>/runs?event=schedule`, paginated):

| workflow | cron | ticks scheduled / 24h | actually fired | fire rate | observed inter-run gaps, last 6h |
|---|---|---|---|---|---|
| `auto-arm` | `4,14,…,54` | 144 | 16 | **11.1%** | 71.8, 65.0, 73.4, 53.7, 58.2 min |
| `re-arm sweeper` | `8,18,…,58` | 144 | 16 | **11.1%** | 65.2, 61.8, 70.5, 53.2, 56.9 min |
| `promote-on-approval` | `*/10` | 144 | 16 | **11.1%** | 68.3, 65.2, 74.8, 66.5, 69.0 min |
| `batch-merge` | `7,22,37,52` | 96 | 15 | **15.6%** | 66.8, 66.4, 74.3, 68.4 min |
| `verdict-bridge` | `1,11,…,51` | 144 | 2 (merged today) | — | 61.0 min |

GitHub coalesces and drops `schedule` events under load, and this repo is permanently under load.
Two consequences:

1. The **offset chain is fiction.** `verdict-bridge :01 → auto-arm :04 → batch-merge :07` assumes
   the ticks land where the crons say. Observed `auto-arm` start minutes span
   `0,2,3,4,5,…,58` — essentially the whole hour.
2. The real cost of a cron-only stage is **~1 hour, not ~10 minutes.**

Second measurement: the sweep itself is `O(open PRs × check-run pages)`. Live run
[`30221614764`](https://github.com/sparq-org/sparq/actions/runs/30221614764) took **8m30s**
(`21:42:01Z → 21:50:31Z`) for **103 open PRs** — one PR alone (`#4294`) carries 612 check-runs
across 7 pages. I re-ran the identical sweep locally against the live repo: **10m45s for 119
open PRs** (§8). So even when the cron *does* fire, the write lands ~10 minutes after the read,
and that gap grows with the backlog.

Observed end-to-end on the maintainer's own example, PR **#4324**: `gate` completed
`21:40:43Z`, `review:unreviewed` applied `21:50:31Z` — **9.8 minutes**.

## 2. The partition-lock premise — CONFIRMED, with two refinements

The claim under test: *an open PR holds its `area:` key and blocks every other issue in that
partition, so unmerged-but-mergeable minutes are lost parallelism.*

It holds. `scripts/ready-issues.py` builds the held set from **open PRs unconditionally**:

```python
for it in issues:
    if str(it.get("state", "OPEN")).upper() != "OPEN" or not occupies_area(it):
        continue
    L = labels_of(it)
    if "pull_request" in it or L & IN_FLIGHT_STATUS:
        reserve(_reserving_packages(L), it)
```

Drafts included; **no** CI, mergeability, `review:pass` or armed-auto-merge condition. The only
releases are (a) the PR leaving `OPEN`, and (b) a parking label
(`PARKED_AREA_LABELS = {"needs:user", "review:needs-user", "status:blocked"}`). `review:pass`
releases nothing. There is no TTL. Confirmed in prose at `AGENTS.md`: *"an in-flight PR **reserves**
its areas and a ready issue **defers**"*.

Two honest refinements:

* **Weaker than stated in one direction**: the key is the *declared `area:` set*, not "the crate".
  `_reserving_packages` has no `__global__` fallback, so an open PR with **no** `area:` label
  reserves nothing. A multi-area PR blocks several partitions at once (`#4318`, four area labels,
  held 54 candidates at one measured tick).
* **Stronger in another**: keys compare by exact string, so `area:sparq-server-http` does not
  conflict with `area:sparq-server` — a separate under-serialisation defect, not this PR's.

The "58.3% of deferrals were non-crate contention" figure is real but **not on `main`**: it is
§1 of `research/crate-region-parallelism.md` on branch `research/crate-region-parallelism`
(211/362 conflict-deferrals held by a non-crate area). Caveat worth flagging: `58.3%` appears
*twice* in that document with different denominators — do not conflate them.

## 3. Classification — what I converted, and what I deliberately left periodic

25 workflows carry a cron. Ranked by **measured** cost, not cron period.

### (a) Deterministic + event-shaped

| stage | measured lag | webhook that carries the state | action |
|---|---|---|---|
| **`verdict-bridge`** | 9.8 min observed (#4324); structural exposure ~1 h given the 11% fire rate | `issue_comment`, `pull_request_review`, `workflow_run(ci-summary)` | **CONVERTED here** |
| `retriage` | `status:untriaged → cleared`, n=7: p50 **54.1 min**, max 68.7 min, against a `*/30` cron | `issues: [labeled, unlabeled]` — the header's own stated reason for the cron ("`edited` does not cover label changes") is answered by a trigger that already exists | **NOT converted** — real, but it is issue-triage latency, not merge latency, and it does not touch the partition lock. Follow-up. |
| `auto-arm` | — | `pull_request: [labeled]` | **already converted** (upstream) |
| `batch-merge`, `pr-backlog` | — | `workflow_run` / `push` | **already converted** (upstream) |

### (b) Genuinely periodic — left alone, with the reason

| stage | why the trigger really is the passage of time |
|---|---|
| `promote-on-approval` (`*/10`) | Its own header states it: **GitHub Actions has no reaction trigger.** A 👍 emits no webhook. Polling is the only mechanism. |
| `re-arm sweeper` (`8,18,…`) | It repairs auto-merge arms **GitHub silently drops** after merge-group churn. A dropped arm is the *absence* of an event; nothing fires. |
| `pr-backlog` cron leg (`0 */4`) | Staleness sweep (`>7d no update`). Elapsed time *is* the predicate. |
| `dependency-monitoring`, `scorecard`, `drift-scan`, `container-scan` | External-state polling (advisory feeds, upstream posture, registry images). |
| `formal-alarm` | A **no-verdict** alarm — it fires on lanes that did *not* report. Absence has no webhook. |
| `asan`, `bench`, `ci`, `codeql`, `differential`, `differential-update`, `fuzz`, `kani`, `metamorph`, `miri`, `nightly-full-sweep`, `shacl-diff-fuzz`, `zk-toolchain` | Nightly/weekly verification sweeps. Cost-scheduled, not state-triggered. |

I did not convert everything theoretically convertible. `retriage` is the one real remaining (a)
and it is scoped out on purpose.

## 4. What changed

**`.github/workflows/verdict-bridge.yml`** — three event triggers added, **cron kept**:

* `issue_comment: [created, edited]` — the maintainer's own path (VERDICT comment → `review:pass`).
* `pull_request_review: [submitted, edited, dismissed]` — review bodies are withhold-only, so this
  can only ever *shrink* the promote-set, but a retraction must not wait an hour either.
* `workflow_run: {workflows: [ci-summary], types: [completed]}` — `gate` concluded, i.e. the
  green/`flag` population moved. Chosen over `check_suite` because it fires **once** per head-CI
  completion where `check_suite` fires ~18 times (measured: 18+ suites on `#4324`'s head).

**`scripts/verdict-bridge.py`**:

* `--pr N` / `PR_ONE_QUERY` — an event reads **one** PR. Live dry-run below: **2.6 s** vs 8m30s.
* `--mode event|sweep` — an event run has no useful backstop on a 53–75-minute real cadence, so an
  exhausted transient exits **red**; the sweep keeps its lenient `::warning + 0`.
* `parse_pr_argument` — the parse boundary for the only payload-derived value. Garbage **raises**;
  it never degrades to a full sweep.
* `reconfirm()` — see §5.
* Removed dead code: `promote` was paired with a `review:unreviewed` removal that `decide()` makes
  unreachable (`promote` requires `not flagged`).

### 4b. Event volume — measured, and why two filters were added

`issue_comment` on every PR comment would be a throughput *regression*, not a fix. Measured over
the 24h to `2026-07-26T22:30Z`: **1286 comments on PRs** (2180 including plain issues), and
`ci-summary` has **8524** lifetime runs. Unfiltered, that is >1300 extra jobs/day into a repo with
a documented congestion-collapse mode where over-dispatching saturates the runners and
manufactures **false** gate failures.

Two job-level filters, both **provably narrowing** and both cron-backstopped:

| filter | why it cannot drop work the policy would do |
|---|---|
| `contains(github.event.comment.body, 'VERDICT')` | Every comment the policy acts on has a last non-blank line matching `VERDICT_SHAPE_RE`, which requires the literal case-insensitive token `VERDICT`. A comment without it decides nothing — `decide()` never retracts on the *absence* of a verdict. Pinned by `test_the_comment_body_filter_is_a_SUPERSET_of_the_evidence_the_policy_reads`, driven off the **live** regex. |
| `workflow_run.conclusion == 'success'` | The gate conclusion reaches `decide()` only via `is_green_and_ready`, which guards `flag`. Over the population this event uniquely covers (no head-bound verdict, so no comment/review event will fire), every non-success conclusion yields `none`. Pinned by `test_only_a_SUCCESSFUL_gate_can_change_a_decision_the_ci_event_uniquely_enables`. |

The comment body is read **only** inside the job `if:`, which GitHub evaluates itself — it never
reaches a shell. `test_the_comment_body_is_read_ONLY_by_the_job_condition` pins that it appears
exactly once in the document, and the injection test now scans step `run:`/`env:`/`with:` rather
than the whole document, which is the correct sink set.

## 5. Idempotence — demonstrated, and an honest statement of what `concurrency` does not do

Event and cron now race. **`concurrency:` reduces the window; it does not close it.** A concurrency
group *cancels*, it does not *lock*, and the per-PR groups are **disjoint from the sweep group by
construction** — so an event run and the scheduled sweep can execute against the same PR at the
same instant. That is stated in the workflow header rather than glossed.

Convergence comes from the write path. `reconfirm()` re-reads the single PR and re-decides
immediately before every label edit — a genuine compare-and-set on (head SHA, labels, newest
verdict). The `gate` conclusion is deliberately carried forward rather than re-read (up to 7
pages/PR); that shortcut is sound only because no arming-relevant decision consults it, which is
itself pinned by `test_arming_relevant_decisions_never_depend_on_the_gate_read`.

Demonstrated by **execution**, not assertion — the test fake mutates its label state on write, and
a `before_confirm` hook injects a concurrent actor into the exact read→write window:

* `test_two_sequential_runs_over_one_head_write_exactly_once` (× promote/flag/retract/unflag)
* `test_the_event_run_and_the_cron_run_over_one_head_converge`
* `test_a_STALE_sweep_decision_cannot_resurrect_a_label_an_event_just_removed`
* `test_a_HEAD_CHANGE_between_read_and_write_abandons_the_write`
* `test_a_DROPPED_webhook_is_still_reconciled_by_the_sweep` (× 4 actions)

## 6. Fail-closed authority is identical on both paths

Constraint: the event path must not be a laxer route to the same label. Proven by **differential
execution** over 13 fixtures (promote / flag / retract / unflag / two hold classes / draft /
conflicting / stacked base / label-page overflow / untrusted author / stale head / ambiguous
retraction): `test_the_event_path_emits_the_IDENTICAL_writes_as_the_cron_path`, plus the
one-directional `test_the_event_path_can_never_write_the_attestation_where_the_cron_would_not`.
The node field selection is literally shared (`PR_NODE_FIELDS`), pinned by
`test_both_paths_request_the_SAME_node_fields`.

**Trust.** `pull_request_target` is not used, and neither is `pull_request` — the forbidden set is
now exactly the triggers that resolve the *workflow file* from a candidate ref. `issue_comment`,
`pull_request_review` and `workflow_run` all resolve from the default branch; the checkout is
separately pinned to `default_branch`. The only payload value crossing into the run is an integer
PR number. No comment body / title / branch name reaches a `run:`
(`test_no_untrusted_payload_text_is_interpolated_into_a_shell`). No model is invoked anywhere in
this workflow, and the `OWNER/MEMBER/COLLABORATOR` + head-binding checks are the same code on both
paths. A third-party comment gains nothing except a job that reads one PR and decides `none`.

## 7. YAML-seam mutation run — 39 mutants, 39 killed, 0 crash-kills

Mutating the **live** tree (27 workflow-YAML mutants, 12 Python), restored from a `cp` snapshot.
Full output in the first comment. Every kill names a specific test; `Y11` initially died in
`setUpClass` (an error-kill that names no guard) and was re-killed by named tests after the
harness was taught to handle a constant group.

Five of the 27 YAML mutants target the two new volume filters (delete / invert / over-narrow the
comment filter; delete / invert the success filter) — all killed by named tests.

Four mutants **survived** the first run. Each was a real gap, now closed:

| survivor | disposition |
|---|---|
| `P01` delete the `reconfirm` compare-and-set — **the headline guard** | My first race test hand-computed the stale decision then called `run()`, which re-read everything fresh, so the CAS was never load-bearing. Rewritten to drive the real loop with the retraction injected mid-window. Exactly the "headline guard is least-tested" failure. |
| `P10` promote built from the stale label set | **Equivalent mutant** — `decide()` returns `unflag` before it can return `promote`, so the pairing was unreachable. Deleted the dead branch instead of testing dead reasoning. |
| `P11` event mode with no PR number falls through to a full sweep | No test covered `main()`. Added one. |
| `Y12` `cancel-in-progress: true` | My mutant was badly built (duplicate YAML key, the later one won) — a harness bug, not a test gap. Rebuilt; killed by `test_no_group_cancels_in_progress`. |

The job `if:` and the concurrency group are **evaluated**, not grepped: a small recursive-descent
parser for the GitHub-expression subset runs them against synthetic payloads and asserts the
admit/skip matrix (`TestJobConditionAdmission`, `TestConcurrencyGrouping`). The evaluator has its
own tests so it cannot pass vacuously.

## 8. Gate result

* `python3 scripts/verdict-bridge.py --self-test` — PASS
* `python3 scripts/tests/test_verdict_bridge.py` — **115 tests, OK** (was 74)
* every other `scripts/tests/test_*.py` (30 further files) — **all OK**
* `actionlint .github/workflows/verdict-bridge.yml` — only `SC2086` on the pre-existing unquoted
  `$DRY_RUN`; **identical finding on `origin/main`**
* `typos`, `check-terminology`, `check-no-perf-numbers`, `check-advisory-registry`,
  `check-sticky-failure-exits`, `check-install-action-tool`, `check-pr-arm-base` — all clean

**Live differential against the real repository.** Full sweep (the cron path), dry-run:

```
[verdict-bridge] 119 open PR(s); dry_run=True
[verdict-bridge] PR #4331: PROMOTE - head-bound pass by jeswr at 4dc6f90360d4
[verdict-bridge] PR #4334: FLAG - green + mergeable but invisible to every lane
[verdict-bridge] PR #4360: FLAG - green + mergeable but invisible to every lane
[verdict-bridge] PR #4361: SKIP inspect-failed (3de422bd...: paged only 561 check-runs, total_count=572)
[verdict-bridge] PR #4368: FLAG - green + mergeable but invisible to every lane
[verdict-bridge] PR #4373: FLAG - green + mergeable but invisible to every lane
[verdict-bridge] complete: writes=0 errors=1

real    10m44.950s
```

The same six PRs through the new scoped event path:

```
[verdict-bridge] PR #4331: PROMOTE - head-bound pass by jeswr at 4dc6f90360d4
[verdict-bridge] PR #4334: FLAG - green + mergeable but invisible to every lane
[verdict-bridge] PR #4360: FLAG - green + mergeable but invisible to every lane
[verdict-bridge] PR #4368: FLAG - green + mergeable but invisible to every lane
[verdict-bridge] PR #4373: FLAG - green + mergeable but invisible to every lane

total wall-clock for all six scoped runs: 18s
```

**Identical decisions, 18s vs 10m45s.** One apparent divergence, investigated rather than
waved past: `#4361` produced a decision line in the sweep and none in the scoped run. Its real
disposition is `none` either way — it already carries `review:pass` and the `trust-surface`
hold. The sweep's line was the *transient* check-run truncation of §9, which did not reproduce
on two subsequent scoped runs. Same terminal state; the scoped path simply did not red on it.

## 9. Defect found in passing (not fixed here)

Run `30221614764` also shows the sweep redding on one PR:
`PR #4294: SKIP inspect-failed (2cc91f8…: paged only 600 check-runs, total_count=612)`.
The one-sided truncation guard fires when the check-run window *shifts* under pagination and
repeats entries, so de-duplication leaves fewer unique runs than `total_count`. Not this PR's
scope — but note that scoping to one PR contains the blast radius: today one un-inspectable PR
reds the whole cycle for all 103.

---

**Author-only change.** Not armed, not merged, no labels touched. Needs a cross-provider review.


jeswr:
> 🤖 SPARQ agent

Full mutation-run output (34 mutants against the LIVE tree; snapshot-restored between each).

```
baseline: GREEN

  KILL  Y01 delete the issue_comment trigger
          by: test_every_event_trigger_is_wired_to_a_state_change_the_policy_reads
  KILL  Y02 delete the pull_request_review trigger
          by: test_every_event_trigger_is_wired_to_a_state_change_the_policy_reads
  KILL  Y03 delete the workflow_run trigger
          by: test_every_event_trigger_is_wired_to_a_state_change_the_policy_reads, test_the_gate_producing_workflow_named_in_workflow_run_actually_exists
  KILL  Y04 delete the cron backstop
          by: test_the_cron_backstop_survives_the_event_conversion, test_the_cron_fires_before_each_auto_arm_sweep
  KILL  Y05 widen the cron backstop to hourly
          by: test_the_cron_backstop_survives_the_event_conversion
  KILL  Y06 invert the issue_comment payload guard
          by: test_a_comment_on_a_PLAIN_ISSUE_is_refused, test_a_comment_on_a_PULL_REQUEST_is_admitted
  KILL  Y07 delete the workflow_run clause from the job if
          by: test_a_ci_summary_run_WITH_an_associated_pr_is_admitted
  KILL  Y08 delete the schedule admission from the job if
          by: test_the_scheduled_backstop_is_admitted_unconditionally
  KILL  Y09 job if: true (unguarded event path)
          by: test_a_ci_summary_run_WITH_an_associated_pr_is_admitted, test_a_ci_summary_run_with_NO_associated_pr_is_refused, test_a_ci_summary_run_with_NO_associated_pr_is_refused
  KILL  Y10 job if: false (kill switch)
          by: test_a_ci_summary_run_WITH_an_associated_pr_is_admitted, test_a_ci_summary_run_with_NO_associated_pr_is_refused, test_a_ci_summary_run_with_NO_associated_pr_is_refused
  KILL  Y11 collapse the concurrency group to one shared key
          by: test_events_about_DIFFERENT_prs_never_share_a_group, test_the_sweep_group_is_DISJOINT_from_every_per_pr_group
  KILL  Y12 cancel-in-progress: true (a queued run is cancelled, not deferred)
          by: test_no_group_cancels_in_progress
  KILL  Y13 drop --pr from the run step
          by: test_the_run_step_scopes_the_event_to_one_pr_and_sets_the_mode
  KILL  Y14 drop --mode from the run step
          by: test_the_run_step_scopes_the_event_to_one_pr_and_sets_the_mode
  KILL  Y15 invert BRIDGE_MODE (cron loud, event silent)
          by: test_the_mode_expression_makes_exactly_the_cron_paths_fail_soft, test_the_mode_expression_makes_exactly_the_cron_paths_fail_soft, test_the_mode_expression_makes_exactly_the_cron_paths_fail_soft
  KILL  Y16 point workflow_run at the wrong workflow
          by: test_every_event_trigger_is_wired_to_a_state_change_the_policy_reads, test_the_gate_producing_workflow_named_in_workflow_run_actually_exists
  KILL  Y17 drop the workflow_run source from BRIDGE_PR
          by: test_the_run_step_scopes_the_event_to_one_pr_and_sets_the_mode
  KILL  Y18 add the forbidden pull_request_target trigger
          by: test_it_never_triggers_on_a_candidate_controlled_ref
  KILL  Y19 escalate to contents: write
          by: test_it_has_no_contents_write_authority
  KILL  Y20 checkout the candidate head instead of the default branch
          by: test_the_checkout_materializes_every_script_the_run_imports
  KILL  Y21 interpolate the untrusted comment body into the run env
          by: test_no_untrusted_payload_text_is_interpolated_into_a_shell
  KILL  Y22 hard-code --dry-run (silent no-op)
          by: test_the_scheduled_sweep_is_never_a_hard_coded_dry_run
  KILL  P01 delete the reconfirm compare-and-set
          by: test_a_STALE_sweep_decision_cannot_resurrect_a_label_an_event_just_removed
  KILL  P02 reconfirm returns the stale decision unchanged
          by: test_a_HEAD_CHANGE_between_read_and_write_abandons_the_write, test_a_STALE_sweep_decision_cannot_resurrect_a_label_an_event_just_removed
  KILL  P03 reconfirm skips the base-branch re-check
          by: test_reconfirm_refuses_when_the_pr_vanished_or_its_base_moved
  KILL  P04 parse_pr_argument degrades garbage to a full sweep
          by: test_the_pr_argument_parser_never_degrades_garbage_to_a_full_sweep, test_the_pr_argument_parser_never_degrades_garbage_to_a_full_sweep, test_the_pr_argument_parser_never_degrades_garbage_to_a_full_sweep
  KILL  P05 parse_pr_argument accepts non-positive numbers
          by: test_the_pr_argument_parser_never_degrades_garbage_to_a_full_sweep
  KILL  P06 fetch_one reads a GraphQL error as 'no work'
          by: test_a_broken_single_pr_read_RAISES_rather_than_reading_as_no_work, test_a_broken_single_pr_read_RAISES_rather_than_reading_as_no_work, test_a_broken_single_pr_read_RAISES_rather_than_reading_as_no_work
  KILL  P07 --pr is ignored (every event becomes a full sweep)
          by: test_a_scoped_run_reads_ONLY_the_named_pr, test_an_event_naming_a_vanished_pr_is_a_clean_no_op
  KILL  P08 event mode fails soft like a sweep
          by: test_event_mode_reds_on_exhausted_transients
  KILL  P09 the single-PR query drops the labels connection
          by: test_both_paths_request_the_SAME_node_fields
  KILL  P10 re-pair promote with the (dead) unreviewed removal
          by: test_promote_and_the_unreviewed_flag_are_mutually_exclusive_by_construction, test_a_PENDING_review_is_not_evidence, test_a_flagged_pr_is_unflagged_first_then_promoted_on_the_next_sweep
  KILL  P11 event mode with no PR number falls through to a full sweep
          by: test_main_refuses_an_event_run_whose_payload_named_no_pull_request
  KILL  P12 a reconfirm read failure is swallowed
          by: test_a_reconfirm_read_failure_is_an_ERROR_and_writes_nothing

restored baseline green: True
killed=34 (of which crash-kills=0) survived=0 not-applicable=0 total=34
```

jeswr:
> 🤖 SPARQ agent

**Updated mutation run** — 39 mutants (27 workflow-YAML, 12 Python) against the LIVE tree, snapshot-restored between each. Supersedes the earlier 34-mutant comment; the 5 new mutants target the volume filters added in `e0d…`.

```
baseline: GREEN

  KILL  Y01 delete the issue_comment trigger
          by: test_every_event_trigger_is_wired_to_a_state_change_the_policy_reads
  KILL  Y02 delete the pull_request_review trigger
          by: test_every_event_trigger_is_wired_to_a_state_change_the_policy_reads
  KILL  Y03 delete the workflow_run trigger
          by: test_every_event_trigger_is_wired_to_a_state_change_the_policy_reads, test_the_gate_producing_workflow_named_in_workflow_run_actually_exists
  KILL  Y04 delete the cron backstop
          by: test_the_cron_backstop_survives_the_event_conversion, test_the_cron_fires_before_each_auto_arm_sweep
  KILL  Y05 widen the cron backstop to hourly
          by: test_the_cron_backstop_survives_the_event_conversion
  KILL  Y06 invert the issue_comment payload guard
          by: test_a_VERDICT_comment_on_a_PULL_REQUEST_is_admitted, test_a_comment_on_a_PLAIN_ISSUE_is_refused, test_the_comment_body_filter_is_a_SUPERSET_of_the_ev
  KILL  Y07 delete the workflow_run clause from the job if
          by: test_a_SUCCESSFUL_ci_summary_run_WITH_an_associated_pr_is_admitted
  KILL  Y08 delete the schedule admission from the job if
          by: test_the_scheduled_backstop_is_admitted_unconditionally
  KILL  Y09 job if: true (unguarded event path)
          by: test_a_SUCCESSFUL_ci_summary_run_WITH_an_associated_pr_is_admitted, test_a_VERDICT_comment_on_a_PULL_REQUEST_is_admitted, test_a_ci_summary_run_with_N
  KILL  Y10 job if: false (kill switch)
          by: test_a_SUCCESSFUL_ci_summary_run_WITH_an_associated_pr_is_admitted, test_a_VERDICT_comment_on_a_PULL_REQUEST_is_admitted, test_a_ci_summary_run_with_N
  KILL  Y11 collapse the concurrency group to one shared key
          by: test_events_about_DIFFERENT_prs_never_share_a_group, test_the_sweep_group_is_DISJOINT_from_every_per_pr_group
  KILL  Y12 cancel-in-progress: true (a queued run is cancelled, not deferred)
          by: test_no_group_cancels_in_progress
  KILL  Y13 drop --pr from the run step
          by: test_the_run_step_scopes_the_event_to_one_pr_and_sets_the_mode
  KILL  Y14 drop --mode from the run step
          by: test_the_run_step_scopes_the_event_to_one_pr_and_sets_the_mode
  KILL  Y15 invert BRIDGE_MODE (cron loud, event silent)
          by: test_the_mode_expression_makes_exactly_the_cron_paths_fail_soft
  KILL  Y16 point workflow_run at the wrong workflow
          by: test_every_event_trigger_is_wired_to_a_state_change_the_policy_reads, test_the_gate_producing_workflow_named_in_workflow_run_actually_exists
  KILL  Y17 drop the workflow_run source from BRIDGE_PR
          by: test_the_run_step_scopes_the_event_to_one_pr_and_sets_the_mode
  KILL  Y18 add the forbidden pull_request_target trigger
          by: test_it_never_triggers_on_a_candidate_controlled_ref
  KILL  Y19 escalate to contents: write
          by: test_it_has_no_contents_write_authority
  KILL  Y20 checkout the candidate head instead of the default branch
          by: test_the_checkout_materializes_every_script_the_run_imports
  KILL  Y21 interpolate the untrusted comment body into the run env
          by: test_no_untrusted_payload_text_is_interpolated_into_a_shell, test_the_comment_body_is_read_ONLY_by_the_job_condition
  KILL  Y22 hard-code --dry-run (silent no-op)
          by: test_the_scheduled_sweep_is_never_a_hard_coded_dry_run
  KILL  Y23 delete the comment-body volume filter
          by: test_ordinary_PR_chatter_is_refused
  KILL  Y24 invert the comment-body filter (drop exactly the verdicts)
          by: test_a_SUCCESSFUL_ci_summary_run_WITH_an_associated_pr_is_admitted, test_a_VERDICT_comment_on_a_PULL_REQUEST_is_admitted, test_a_ci_summary_run_with_N
  KILL  Y25 over-narrow the comment filter to a stricter token
          by: test_the_comment_body_filter_is_a_SUPERSET_of_the_evidence_the_policy_reads
  KILL  Y26 delete the ci-summary success filter
          by: test_a_non_successful_ci_summary_run_is_refused
  KILL  Y27 invert the ci-summary success filter
          by: test_a_SUCCESSFUL_ci_summary_run_WITH_an_associated_pr_is_admitted, test_a_non_successful_ci_summary_run_is_refused
  KILL  P01 delete the reconfirm compare-and-set
          by: test_a_STALE_sweep_decision_cannot_resurrect_a_label_an_event_just_removed
  KILL  P02 reconfirm returns the stale decision unchanged
          by: test_a_HEAD_CHANGE_between_read_and_write_abandons_the_write, test_a_STALE_sweep_decision_cannot_resurrect_a_label_an_event_just_removed
  KILL  P03 reconfirm skips the base-branch re-check
          by: test_reconfirm_refuses_when_the_pr_vanished_or_its_base_moved
  KILL  P04 parse_pr_argument degrades garbage to a full sweep
          by: test_the_pr_argument_parser_never_degrades_garbage_to_a_full_sweep
  KILL  P05 parse_pr_argument accepts non-positive numbers
          by: test_the_pr_argument_parser_never_degrades_garbage_to_a_full_sweep
  KILL  P06 fetch_one reads a GraphQL error as 'no work'
          by: test_a_broken_single_pr_read_RAISES_rather_than_reading_as_no_work
  KILL  P07 --pr is ignored (every event becomes a full sweep)
          by: test_a_scoped_run_reads_ONLY_the_named_pr, test_an_event_naming_a_vanished_pr_is_a_clean_no_op
  KILL  P08 event mode fails soft like a sweep
          by: test_event_mode_reds_on_exhausted_transients
  KILL  P09 the single-PR query drops the labels connection
          by: test_both_paths_request_the_SAME_node_fields
  KILL  P10 re-pair promote with the (dead) unreviewed removal
          by: test_promote_and_the_unreviewed_flag_are_mutually_exclusive_by_construction, test_a_PENDING_review_is_not_evidence, test_a_flagged_pr_is_unflagged_fir
  KILL  P11 event mode with no PR number falls through to a full sweep
          by: test_main_refuses_an_event_run_whose_payload_named_no_pull_request
  KILL  P12 a reconfirm read failure is swallowed
          by: test_a_reconfirm_read_failure_is_an_ERROR_and_writes_nothing

restored baseline green: True
killed=39 (of which crash-kills=0) survived=0 not-applicable=0 total=39
```

jeswr:
> 🤖 SPARQ agent

## Cross-provider review — head `2188373a755d72b9c9208b0816ab57cd1510835d`

Reviewer only; I authored nothing and pushed nothing. Head re-verified immediately before posting
(still `2188373a75`, `MERGEABLE`). `gate` at that head: **exactly one** check-run, id `89850765190`,
`success` — no cancelled twin (`total_count` 58, fully paginated and cross-checked). The two CI steps
that run this PR's tests (`docs-quality.yml:833-834`) carry no `if:` and were green at this head, so
the pinning below is actually executed by CI rather than merely present in the tree.

Work done in an isolated worktree at the PR head. Mutations applied to the **live** tree, restored from
a `cp` snapshot with a sha256 marker check plus a `git status --porcelain` cross-check after every
single restore — never `git checkout`. Baseline before every mutant: `121 tests, OK` + `self-test: PASS`.

---

### 1. The gate carry-forward invariant — GENUINELY PINNED

This is the one load-bearing soundness/cost trade in the PR, so I attacked it directly rather than
reading the test. I introduced a gate dependency into each of the three arming-relevant branches of
`decide()` and ran the suite:

| mutant | result | killed by |
|---|---|---|
| `promote` consults `gate_conclusion` | KILLED | `test_arming_relevant_decisions_never_depend_on_the_gate_read` |
| `retract` consults `gate_conclusion` | KILLED | `test_arming_relevant_decisions_never_depend_on_the_gate_read` |
| `unflag` consults `gate_conclusion` | KILLED | `test_arming_relevant_decisions_never_depend_on_the_gate_read` |

All three behaviour-kills (`unittest rc=1`, no traceback, `Ran 0 tests` never seen), each named by
that single test. The invariant is not vacuous: it is exactly the guard that makes `reconfirm()`'s
carry-forward sound, and it dies on the mutation it exists to catch. I also confirmed by reading
`decide()` that the only two `is_green_and_ready` call sites both return `flag`, and by grepping the
repo that `review:unreviewed` is referenced by nothing outside `verdict-bridge` itself — in particular
it is not in `ready-issues.py`'s `PARKED_AREA_LABELS`, so it neither parks nor arms.

### 2. The race is NARROWED, not closed — executed, not reasoned

I built the adversarial interleaving myself and ran it through the real driver loop:

```
[A] retraction lands BEFORE reconfirm's read   -> writes=[]                        CAS CAUGHT it
[B] retraction lands AFTER  reconfirm's read   -> writes=[... --add-label review:pass]
                                                  CAS DID NOT catch it
    SELF-HEAL next run: rc=0  writes=[... --remove-label review:pass]  labels=set()
```

`reconfirm()` is a check-then-act, not an atomic CAS: the window between the re-read and
`edit_labels` remains open, and a retraction landing inside it still writes a spurious `review:pass`.
It is **convergent** — the next run repairs it, demonstrated above — but the repair waits for the cron
at its measured 53–75 min cadence, because the retraction's own webhook was consumed by the run that
lost the race.

I consider this **acceptable and a strict improvement**, and I want to be explicit about why rather
than wave it through: on `main` today the sweep's read→write gap is the *whole sweep* (measured 8m40s
below) with no re-read at all, so the identical resurrection is available for ~8 minutes per cycle.
This PR shrinks that to one round trip. The workflow header states the honest limit
("this REDUCES the double-fire window, it does not close it") and PR body §5 repeats it for
`concurrency:`.

**One wording nit, non-blocking:** both `AGENTS.md` and the `reconfirm()` docstring call it "a genuine
compare-and-set". Interleaving B is what a genuine CAS would reject. "Collapses the staleness window
to one round trip" — also in the docstring — is the accurate phrasing; the CAS wording overstates it
slightly. Worth a follow-up, not a blocker, since the limit is stated adjacent in the workflow.

### 3. Risk #3 — the sampling was inadequate, but the conclusion is right

The author verified `workflow_run.pull_requests` on one PR's head. I measured it systematically over
every `ci-summary` run in 24h (paginated, cross-checked against `total_count` 8543):

```
by trigger event:  pull_request 563 | merge_group 128 | push 102
merge_group   EMPTY      128
push          EMPTY      102
pull_request  EMPTY      369     <- 65% of same-repo PR runs!
pull_request  populated  194
```

That 65% looks damning, so I chased it. Correlating against the current state of each run's PR
(220 sampled):

```
current PR state   pull_requests   count
MERGED             EMPTY           118
OPEN               populated       102
```

**102/102 and 118/118 — a perfect split.** `pull_requests` is computed at *read* time and lists only
currently-open PRs; it empties retroactively when the PR merges. At webhook-delivery time the PR is
open, so the array is populated. The author's conclusion is correct and is now systematically verified
rather than sampled once. Every empty shape (merge_group, push, fork) skips the job and falls to the
cron, which reads all open PRs — so the failure direction is toward the backstop. Caveat I could not
close: **zero fork PRs existed in the window**, so the fork case remains declared-not-measured; it
fails safe by the same argument.

### 4. Volume filters — superset property holds on live data

Measured over 24h (22 pages of `/issues/comments`, paginated):

```
comments in window: 2139   on PULL REQUESTS: 1276   (author claimed 1286 — different window edge)
  contains 'VERDICT' (the job if:):   89   -> 7.0% of PR comments admitted
  verdict-SHAPED last line (evidence): 22
  evidence NOT admitted by the filter:  0   -> filter is a superset
```

Zero violations against the live regex on real traffic. Projected added jobs/day:
`issue_comment` 89 + `workflow_run` 73 + `pull_request_review` 49 ≈ **211/day**, versus ~2063 unfiltered.

**Finding worth naming:** the body says "two filters, both provably narrowing", but there is a *third*
event path — `pull_request_review` — admitted on `pull_request.number != null`, i.e. **unfiltered**.
Measured 49/24h so it is not a throughput problem, and review bodies are withhold-only so it cannot be
an authority problem. But the "two filters" framing does not cover it. Non-blocking; worth a sentence.

The authority-vs-population distinction is maintained honestly: §6 says a third party gains "nothing
except a job that reads one PR and decides `none`" — it does *not* claim third parties cannot create
jobs. That is the correct claim.

### 5. Independent mutation run — 37 mutants, 36 killed, 1 provably equivalent, 0 crash-kills

22 Python + 15 YAML, including every mutant the brief asked for.

| mutant | result |
|---|---|
| **delete the reconfirm CAS skip** (the guard in the PR's convergence claim) | KILLED — `test_a_STALE_sweep_decision_cannot_resurrect_a_label_an_event_just_removed` |
| never call `reconfirm()` at all | KILLED (2 tests) |
| `reconfirm` discards the re-read node | KILLED — `test_a_HEAD_CHANGE_between_read_and_write_abandons_the_write` |
| `reconfirm` drops the base-branch check | KILLED |
| **`return 1` → `return 0`** (event-mode transient) | KILLED — `test_event_mode_reds_on_exhausted_transients` |
| **disarm the job `if:`** (→ `true`, and deleted entirely) | KILLED — both |
| **neutralise the concurrency group** (→ constant; `cancel-in-progress: true`) | KILLED — both |
| add `pull_request_target` | KILLED — `test_it_never_triggers_on_a_candidate_controlled_ref` |
| remove the cron backstop | KILLED — `test_the_cron_backstop_survives_the_event_conversion` |
| drop `--pr` / `--mode` / pin `BRIDGE_PR` to a constant / force `BRIDGE_MODE: sweep` | KILLED — all 4 |
| drop each of the 3 clauses of the comment/gate/pull_requests filters | KILLED — all 3, named tests |
| `fetch_one` swallows GraphQL errors / null repository | KILLED — both |
| reconfirm failure stops counting as an error (exit-zero swallow) | KILLED |
| `PR_ONE_QUERY` drops the labels field | KILLED — `test_both_paths_request_the_SAME_node_fields` |
| review body may grant a pass | KILLED (2 tests) |
| `parse_pr_argument` degrades garbage to `None` / accepts Unicode digits | KILLED — both, self-test |
| **blockquote regression:** `^\s*` → `^[\s>]*` in `VERDICT_RE` and in `VERDICT_SHAPE_RE` | KILLED — both, self-test |
| drop the `$` anchor (trailing content admitted) | KILLED — self-test |
| **`VERDICT_RE.match` → `.search`** | **SURVIVED** |

**On the `^` anchor.** The one survivor is equivalent, and I proved it rather than assumed it: with
`^` present and no `re.MULTILINE`, `search` can only match at position 0, so it is identical to
`match`. Verified over all 441 concatenation pairs of a 21-shape alphabet plus 200 000 random inputs —
**0 divergences**. The anchor is nonetheless load-bearing *in combination*: removing `^` **and** using
`search` admits `> VERDICT: pass`, which I confirmed executes as a pass. That combined weakening is
what the two blockquote mutants above cover, and both die. Given every comment in this repo opens with
`> 🤖 SPARQ agent`, this is the regression class that matters and it is pinned.

Zero crash-kills: every kill named a real test, none died in `setUpClass` or on a traceback. I note the
`TestConcurrencyGrouping.setUpClass` explicitly handles a constant group rather than raising — that is
the right call, and it is why my constant-group mutant produced named kills instead of an error-kill.

### 6. Claims spot-checked by execution

- **cron fire rate** — reproduced independently, paginated: `auto-arm` fired **15** scheduled runs in
  24h, gaps **71.2, 62.7, 71.8, 65.0, 73.4, 53.7, 58.2, 60.6 min**. Matches the claimed 16/144 and
  53–75 min (my window ends 00:00Z, theirs 22:10Z).
- **sweep cost** — run `30221614764`: `run_duration_ms` = **520000** (8m40s), `21:41:54Z → 21:50:34Z`.
  Claimed 8m30s job time. Confirmed.
- **actionlint** — `SC2086` on the unquoted `$DRY_RUN` is reported **identically on the merge-base**.
  Pre-existing, not introduced. Confirmed by running actionlint on both trees.
- **`pull_request_target`** — absent; `permissions:` is `contents: read` + `issues`/`pull-requests:
  write`; checkout is pinned to `default_branch` with `persist-credentials: false`; the only payload
  value crossing into a shell is an integer via `env:`, never `${{ }}` inside `run:`.
- Body says "115 tests"; the tree now runs **121**. Stale by two commits, harmless.

### Verdict

The three risks the author named are real, correctly characterised, and the one that is genuinely
load-bearing (the gate carry-forward) is pinned by a test that dies on the mutation it exists to catch.
Risk #3's sampling was thin but I closed it systematically and the conclusion holds. The race is
narrowed rather than closed — I executed the residual interleaving and it self-heals — which is a
strict improvement on the 8m40s unguarded window that exists on `main` today, and the limit is stated
in the workflow header. My two findings (the "genuine compare-and-set" wording, and the unfiltered
third `pull_request_review` path not covered by the "two filters" framing) are documentation-accuracy
issues, not defects in behaviour. Neither is worth blocking a conversion that removes ~1 hour of
structural latency from the arming chain.

VERDICT: pass

Issue #2474, opened 2026-07-17
flow-on follow-up issues are invisible to triage/dispatch (GITHUB_TOKEN event suppression, no labels at creation)


jeswr:
> 🤖 SPARQ agent — from:agent follow-up, dedupe marker: audit-2026-07-17

**Observed:** flow-on.yml creates follow-up issues under `secrets.GITHUB_TOKEN`. GitHub suppresses workflow events for GITHUB_TOKEN-created objects, so triage-issue.yml (triggers: `issues: [opened, edited, reopened]` only — no `schedule:`) never runs on them, and flow-on adds no `role:*`/`priority:*`/`status:*` labels at creation.

**Why it matters:** these issues are invisible to the dispatch frontier — unlabeled, untriaged, never picked up. Note the `status:untriaged` label description says "cron will re-triage", but no retriage/cron workflow exists on main as of filing (workflow list checked 2026-07-17), so no sweep path catches this class today.

**Fix sketch:** either (a) have flow-on.yml apply the routing labels (role/priority/status:ready or status:untriaged) at creation, or (b) land a scheduled retriage sweep that targets **label-less open issues** (not just `status:untriaged`) — if a retriage workflow is in flight, verify on landing that it covers this class.

PR #4401, opened 2026-07-26
fix(triage): retriage cannot reach the issues it exists to rescue (0 → 74 promotable)

jeswr:
> 🤖 SPARQ agent

## Gate or gap? **Gap** — and `status:ready` is machine-granted, not a human judgment call

The question this started from: *are the ~860 issues without `status:ready` deliberately withheld,
or is automation failing to reach them?*

Answered by reading the only code that writes the label. `scripts/triage.py::triage()` is a pure,
deterministic, **no-LLM** function; `status:ready` is granted iff an issue has a single valid
`priority:P0..P4`, a `role:*`, an `area:<crate>`, and no gate. No human attests anything, and no
model is consulted. It is a **mechanical completeness check**, not a well-specified-ness gate.
So a complete issue that never receives it is a coverage bug, not a considered refusal.

## The binding layer: the sweep's queue was defined by the label whose absence defines the problem

`retriage.py` exists to converge issues `triage-issue.yml` never ran on. But `_fetch_candidates`
built its queue from two **label queries** (`status:untriaged`, `flow-on`). A label query can only
return issues that already carry a label — and an issue the event triager never reached carries
**no `status:*` label at all**. #2474 hit exactly this and worked around it for one source by
bolting on a second label query; every other statusless issue stayed structurally unreachable.

Fixing `_needs_triage` alone cannot help: **an issue that is never fetched is never tested.** The
fetch is the binding layer, so that is where this is fixed — one unlabelled paginated snapshot,
which subsumes both old queries.

### Measured, live on `sparq-org/sparq` 2026-07-26 (1408 open issues)

`python3 scripts/retriage.py --repo sparq-org/sparq` on pristine `origin/main`:

```
retriage: 0 issue(s) promotable (dry-run)
```

The sweep promotes **nothing today**. Reach and outcome, same snapshot:

| | candidates reached | promotable |
|---|---|---|
| label queries (`main`) | 501 | **0** |
| unlabelled snapshot | 739 | **74** |

None of the 501 already-reachable rows is lost. Every one of the 74 is statusless; all pass the
pre-existing gates unchanged.

## Second defect, and it is what bounds the first: two components disagreed on "gated"

`ready-issues.py::is_gated` treats the **whole `needs:*` namespace** as a hard dispatch gate.
`triage.py` checked the single literal `"needs:user"`. So triage attested `status:ready` on
`needs:ec2` / `needs:docker` / `needs:zk` issues, and the readiness engine was the *only* thing
keeping them off the frontier — a single-point defence for work gated on real external
preconditions (EC2 spend authorisation, a missing upstream release).

- **21 open issues** already carry both `status:ready` and a real gate (`needs:ec2` 12,
  `needs:docker` 4, `needs:zk` 3, `needs:external-subject` 2, `needs:upstream` 1,
  `needs:maintainer` 1).
- **5 of the 79** issues the reach fix first made promotable carry `needs:ec2`. Without this
  second fix, widening the reach would have fired on exactly the class that must not be fired on.
  That is why 79 became **74**.

`_blocking_gates` makes the two agree. It is **narrowing only** — an issue that stops being
`ready` is fail-closed and cannot reach the frontier, so this can never open a dispatch path.
`needs:area` is excluded because triage *owns* it: it is applied and self-cleared by this same
function, and treating it as blocking would deadlock it permanently.

## Projected effect on drainable backlog

Applying the 74 promotions to the live snapshot and re-running `ready-issues.py::diagnose`:

```
BEFORE: drainable=374  frontier=7  no-attestation=861
AFTER : drainable=438  frontier=9  no-attestation=797
```

**Drainable backlog +64 (374 → 438, ~+17%)**; the concurrency frontier moves 7 → 9. Stated as a
range: **+55 to +75**, because the promoted set shifts as the backlog changes and ~10 of the 74
are suppressed as covered by an open linked PR. The frontier gain is small by design — it is a
one-issue-per-crate concurrency width, not the backlog size.

This is not a bulk sweep: 74 issues, each passing every gate that already existed, out of ~860.
The change is reversible (label removal) and idempotent — a second pass over the applied live
snapshot plans **0** further edits.

## What is NOT fixed here (deliberately)

- **~160 statusless issues have no `area:` label** and stay unpromoted. `plan_retriage` only ever
  applies *promoting* deltas, so they are now *reached* but still not labelled `needs:area` — they
  remain invisible rather than explicitly parked. Clearing that population is the `needs:area`
  classifier's job (**#4191**), not this layer's. Widening this PR to write park labels would put
  ~160 new labels on the board for a problem another PR already owns.
- **432 of the 497 `status:untriaged` issues** are blocked on the same missing `area:` — also #4191.
- The **76 zero-label issues** are now reached (previously unreachable) but are not promotable for
  the same reason. Fixture 19 pins that they are reached and not churned.

## Composition of the excluded population

238 statusless non-`flow-on` issues; 152 carry a bd-migration marker. Age p50 6d / p90 34d / max
41d, against `status:untriaged` p50 ~0d — the statusless class is an **aged stratum no cron
touches**, not fresh intake. **Zero** are `kind:epic` and only 2 carry a `Blocked-by:` marker, so
neither sub-epic leaves nor blocked children explain it. Low priority is *not* over-represented:
185 of 242 carry **no priority label at all**. Authors: `jeswr` 129, `sparq-orchestrator[bot]` 87,
`github-actions[bot]` 22.

## Tests + mutation

49 tests in `test_readiness_visibility.py` (was 34), plus both `--self-test` fixture suites.
Mutation run, **cp-snapshot with marker-verified restore** — no `git checkout` anywhere:

**14 of 15 killed, every kill a behaviour-kill** (no crash-kills; an ordering assertion that
raised `ValueError` was rewritten to assert presence first so its kill is clean).

| mutant | result |
|---|---|
| M1 restore the `flow-on`-only statusless arm | killed |
| M2 revert `_fetch_candidates` to the two label queries | killed |
| M3 drop `--paginate` (silent truncation) | killed |
| M4 narrow `_blocking_gates` back to `needs:user` | killed |
| M5 stop excluding the self-clearing park | killed |
| M6 drop the gate term from `ready` entirely | killed |
| M7 blanket-trust any `[bot]` author | killed |
| M8 drop the `_needs_triage` filter in the fetch | **SURVIVED — equivalent** |
| Y1 `retriage.yml`: drop `--apply` | killed |
| Y2 `retriage.yml`: `if: false` on the job | killed |
| Y3 `retriage.yml`: drop `issues: write` | killed |
| Y4 `retriage.yml`: remove the schedule | killed |
| Y5 `docs-quality.yml`: stop self-testing retriage | killed |
| Y6 `retriage.yml`: self-test after the apply | killed |
| Y7 `routing-self-tests.yml`: drop the `retriage.yml` path trigger | killed |

**M8 is an equivalent mutant, reported rather than papered over**: `plan_retriage` re-applies the
same predicate, so the plan is byte-identical with or without that filter. It is a bound on queue
size, not a guard. Noted in the docstring so a reviewer does not read it as an untested guard.

### The YAML seam found a hole in my own guard

`TestRetriageCronSeam` reads `retriage.yml`, but `routing-self-tests.yml` is `paths:`-filtered and
did not list it — so a PR touching **only** `retriage.yml` (deleting `--apply`, the single edit the
guard exists to catch) would never have run the suite. The guard was vacuous on its own headline
mutant. Third commit adds the trigger and a test pinning it (Y7 above).

The seam tests parse the workflow **structurally**, not by substring, because `if: false` on a job
or step is exactly the mutant class a substring search cannot see (Y2).

## Not verified here

The retriage cron's real fire cadence is reportedly p50 54 min against a nominal 30, so the 74
promotions land over the following one or two fires rather than immediately. I did not measure
that cadence myself and am not claiming it.


jeswr:
> 🤖 SPARQ agent

**Conflict resolved — merged `main` (`d7539cf47`) into `fix/retriage-statusless-reach`; new head `dc5cf60c0`.**

This PR was `CONFLICTING` and could not merge. Triaged as **REBASE** (still wanted, conflict purely mechanical).

### What conflicted
One file, `.github/workflows/routing-self-tests.yml`, two hunks — the *same two* `paths:` lists in the
`pull_request` and `merge_group` triggers. Both sides appended new path entries at the same anchor:

| side | added |
|---|---|
| this PR | `.github/workflows/retriage.yml`, `.github/workflows/docs-quality.yml` (so `TestRetriageCronSeam` actually runs on the edit it guards) |
| `main` (#4145) | `orchestration/start-here.toml`, `scripts/render-start-here.py`, `scripts/tests/test_start_here_render.py`, `.github/workflows/refresh-start-here.yml` |

### How I resolved it
**Keep both** — the two path sets are disjoint and additive; neither side removed anything. No logic was
re-decided, no hunk was dropped.

I verified the resolution against **both** parents rather than eyeballing it:

- `git diff origin/main -- .github/workflows/routing-self-tests.yml` → exactly this PR's 2 entries × 2 triggers, nothing else.
- `git diff df3548a5c -- .github/workflows/routing-self-tests.yml` → exactly `main`'s 4 entries × 2 triggers plus `main`'s
  `render-start-here.py --self-test` step, nothing else.

So the merge is the strict union of the two parents on this file. `scripts/tests/test_readiness_visibility.py`
auto-merged (add/add against `main`'s newer copy); the merged file keeps `main`'s suite **and** this PR's three
new classes — `TestRetriageCronSeam`, `TestTriageGateAgreesWithReadinessEngine`,
`TestRetriageReachesNeverTriagedIssues`. `scripts/retriage.py` and `scripts/triage.py` needed no resolution
(`main` has not touched them since the merge base).

### Verification at the new head
```
python3 -B scripts/tests/test_readiness_visibility.py   → Ran 99 tests ... OK
python3 -B scripts/retriage.py --self-test              → retriage self-test PASSED
python3 -B scripts/triage.py --self-test                → triage self-test PASSED
```
(`-B` / `PYTHONDONTWRITEBYTECODE=1` — these suites load scripts via `spec_from_file_location` and will
happily serve stale bytecode otherwise.)

No behaviour was changed while resolving, so no new guard/test is owed. CI is the gate — I have **not** armed
or merged this, and I have not reviewed the substance of the change itself.

jeswr:
> 🤖 SPARQ agent

## Review of `dc5cf60c0b6bd5f77af122be40f1f9dfa53bca30` (reviewer-only: no push, no arm, no relabel)

Head resolved via `git ls-remote` (not a tracking ref). Reviewed in an isolated tree built from
`git merge-tree main head` — i.e. the tree that would actually land, not the branch in isolation.
`main` has drifted only on `docs-quality.yml` since the merge base; the merge is clean and the two
`--self-test` invocations this PR's `TestRetriageCronSeam` asserts on are still present in the
merged file.

**`gate` resolved by equality on `commits/<sha>/check-runs`** (not the PR rollup): `gate`,
`completed`/`success`, `head_sha` = `dc5cf60c…`, concluded `2026-07-27T20:11:42Z`. Latest
`started_at` across all 58 runs is `20:04:34Z`, so nothing started after `gate` concluded. The run
is named exactly `gate`, not `gate, draft-tier`.

### The headline claim, verified by execution against the live board (not by reading)

Both `retriage.py --repo sparq-org/sparq` dry-runs, today, same board:

| tree | promotable |
|---|---|
| `origin/main` @ `7b8c8252` | **0** |
| merged head @ `dc5cf60c` | **67** |

Title says 74 (measured 07-26); the body itself ranges it **+55 to +75** and says the set shifts
with the backlog. 67 is inside the author's own stated range — the claim reproduces, and it is
ranged rather than pinned. The direction (`0 → N`) is exact.

I re-derived the population independently from a fresh unlabelled snapshot (1607 open issues):
**237 statusless, of which only 2 carry `flow-on`.** So the pre-fix predicate could express 2 of
237. The reach defect is real and is where the PR says it is — the fetch, not `_needs_triage`.

**Reaching ≠ promoting, checked against live labels rather than argued.** Of the 67 planned
promotions: `needs:*` gates → **0**, `kind:epic` → **0**, `trust:untrusted` → **0**, any existing
`status:*` → **0**. Authors: `jeswr` 32, `sparq-orchestrator[bot]` 35, no third party.

**Second defect independently confirmed and larger than claimed.** Body says 21 open issues carry
both `status:ready` and a real `needs:*` gate; live today it is **43** (`needs:ec2` 34, `needs:docker`
4, `needs:zk` 3, `needs:external-subject` 2, `needs:upstream` 1, `needs:maintainer` 1). The body
under-states, which is the safe direction. Given the standing EC2 spend cap, `ready-issues.py`
being the *sole* defence on 34 `needs:ec2` issues was a real single-point gate.

I also proved the "narrowing only" quantifier rather than accepting it: `needs:user ≠ needs:area`,
so `needs:user ∈ L ⟹ needs:user ∈ _blocking_gates(L)`; new-blocked ⊇ old-blocked at **both** call
sites. `_blocking_gates` cannot open a dispatch path.

### Mutation: 24 mutants, every guard this PR adds has a killing mutant

Isolated tree, `cp`-snapshot with **sha256-verified restore** between mutants, `-B` +
`PYTHONDONTWRITEBYTECODE=1`, line-targeted edits with a changed-file assertion. Kills read from
line-anchored `^  FAIL <name>` / `^FAIL: <test>` rows, never a substring grep. Both forms per guard
where possible: deletion **and** a non-crashing inert form (`if False and …`, `if: false`).

**22 killed / 24. Every kill is a behaviour-kill — zero crash-kills**, and checks-executed was
recorded per mutant so no kill masks the rest of the suite.

Killed: reach reversion to the two label queries; `_needs_triage` back to `flow-on`-only; dropping
`--paginate`; **`label=None` silently falling back to a `status:untriaged` query** (non-crashing,
not in the author's catalogue); `_blocking_gates` narrowed to `needs:user`; `_blocking_gates`
without the `needs:area` exclusion; the gate term deleted from `ready`; **the second
`_blocking_gates` call site (the `needs:area` park) reverted alone** — not in the catalogue;
blanket `[bot]` trust; the ceiling made inert; the write return-code made inert; promoting-only
made full-delta; the PR filter in `_flatten_pages`; the untrusted short-circuit; and at the YAML
seam `--apply` deleted, `if: false` on the job, `if: false` on the run step, schedule removed,
`issues: write` removed, `--apply` reordered before the self-tests, the `docs-quality` self-test
deleted, and each new `paths:` entry deleted.

Two survivors, both genuinely equivalent and neither a hole:
- `_fetch_candidates` without the `_needs_triage` filter — the author declares this and is right;
  `plan_retriage` re-applies the same predicate, so the plan is byte-identical.
- `plan_retriage` without the `SKIP_LABELS` half of its guard — redundant with `triage()`'s own
  `kind:epic` term and untrusted short-circuit. Defence-in-depth, not a live gate.

**One harness correction, recorded because it is the failure mode I was warned about:** `if: false`
on the `validate` job first read as a survivor. It is not — `test_no_deprecated_models.py`, run by
the same workflow, parses that YAML structurally and kills it. My suite set was too narrow. The
guard is real.

### Non-blocking findings (worth a follow-up, none merge-blocking)

1. **The two meta-guards this PR adds are substring counts, not parsed YAML** —
   `test_the_workflows_this_class_guards_are_path_triggers` counts `'"…/retriage.yml"'` in a raw
   slice, and `test_triage_and_retriage_fixtures_run_on_every_pr` uses `assertIn` on raw
   `docs-quality.yml`. Measured: **commenting out** those lines (rather than deleting them)
   survives, as does `if: false` on the `docs-quality` step. Concrete scenario: one PR that both
   comments out the `retriage.yml` path entry and drops `--apply` goes green. Low likelihood
   (deletion is the natural edit and is killed), and the class docstring's "parsed structurally
   rather than by substring" is true of the `retriage.yml` assertions but not these two. Fix is
   small: read `on.pull_request.paths` / `on.push.paths` from `yaml.safe_load`, and assert the
   `docs-quality` step has no `if:` — the same shape already used one class above.
2. **The 43 existing `status:ready` + `needs:*` issues are not cleaned.** `plan_retriage` emits
   promoting deltas only, so this stops new mis-attestations but leaves the standing 43 with
   `ready-issues.py` as their only defence. Not claimed otherwise by the body; deserves its own issue.
3. **No per-run cap on writes.** The first `--apply` fire issues ~67 atomic `gh issue edit` calls
   inside a 10-minute job. If a secondary content limit bites, `apply_labels` returns False, `main`
   exits 1 (no exit-zero swallow — the `failed` list is append-only, verified by mutant), and the
   sweep is idempotent, so it self-heals on the next fire. Fine now; worth a cap before the
   promotable set grows via #4191.
4. `needs:*`-gated area-less issues no longer receive the `needs:area` park. Deliberate and tested;
   they stay `status:untriaged` (fail-closed) and re-converge once the gate lifts.

### What the transition delivers into

`status:ready` is exactly what `ready-issues.py` consumes (`"status:ready" not in labels →
"no status:ready attestation"`), and `GATE_LABELS = ("needs:", "trust:untrusted")` still excludes
every gated issue downstream — so this is not a rescue into a queue nothing reads. The frontier
gain remains bounded by named-crate saturation, and the body says so plainly rather than selling
the backlog number as a dispatch number.

### Trust surface

Every input the promotion decision reads is server-authoritative: `user.login` from the REST
payload and `repos/{repo}/collaborators/{login}/permission`. The `github-actions[bot]` exception
additionally requires the `flow-on` label, which needs triage/write to apply. No control here reads
author-controlled free text. `_trusted_factory` defaults to `"none"` on a failed probe (fail-closed);
its error path and the `read`/`triage` boundary are untested (pre-existing, unchanged here).

### What would have to be true to arm safely

Head still `dc5cf60c0b6bd5f77af122be40f1f9dfa53bca30`; `gate` re-resolved by equality on that SHA at
arm time (not the rollup); and no reviewer commit on the head — this review authored nothing.

VERDICT: pass

PR #4558, opened 2026-07-27
fix(gate): the three alarm/detector lanes were SILENTLY GATING main (sq-huwr8)

jeswr:
> 🤖 SPARQ agent

`main` is RED. Three checks fail at `cb0c6739c7d85420474a2e67be83177660ee1be3`. This PR fixes exactly one of them — the one that is a real, silent defect. The other two are diagnosed below and routed, not silenced.

## The defect: three DETECTOR lanes were silently GATING

`review-alarm.yml`, `formal-alarm.yml` and `selection-alarm.yml` are detectors. They RED **by design** when they find a real blind spot, and they RED fail-loud when the detector itself breaks. Each header asserted the lane could never affect a merge, reasoning from:

> it never creates a check-run on a PR head or merge-group ref → so it can never enter the `ci-summary / gate` poll set

**The premise is true. The conclusion does not follow.** `ci-summary.yml` also runs on `push: branches: [main]` with `timeout-minutes: 80`, and a `schedule` (or `workflow_run`-on-main) alarm run puts its check-run on *exactly* that main head SHA — the one the push-triggered gate is polling. The gate deliberately waits for the check-run **name set to stabilise**, so a sibling appearing mid-poll is picked up, not missed.

Since #3773, the **only** thing that makes a check-run non-gating is a key in `.github/advisory-registry.json`. None of the three was declared.

### Measured, with the repo's own code

`ci_summary_gate.render_verdict()` over the real, paginated **477**-check-run set for `cb0c6739c` (`total_count` cross-checked):

```
### ci-summary: FAILED — 7 non-passing gating check(s) of 462 gating
- ✗ feature-matrix per-leg report: incomplete
- ✗ batch overflow into omnibus: incomplete
- ✗ feature-matrix per-leg report: incomplete
- ✗ review-lane blind-spot alarm: failure      <-- a DETECTOR, gating main
- ✗ bridge verdict comments to review labels: cancelled
- ✗ run + track benchmarks: failure
- ✗ gate: failure
```

And the gate's own predicate:

```
is_advisory('review-lane blind-spot alarm')  = False  -> GATES
is_advisory('nightly selection-bug alarm')   = False  -> GATES
is_advisory('formal-lane no-verdict alarm')  = False  -> GATES
```

After this PR, over the **same** check-run set:

```
### ci-summary: FAILED — 6 non-passing gating check(s) of 460 gating
- ✗ feature-matrix per-leg report: incomplete
- ✗ batch overflow into omnibus: incomplete
- ✗ feature-matrix per-leg report: incomplete
- ✗ bridge verdict comments to review labels: cancelled
- ✗ run + track benchmarks: failure
- ✗ gate: failure
```

The detector is out of the gating set. **The genuine benchmark gate still REDs.** No `|| true`, no `continue-on-error`, no `set +e`, no waiver — the fix is a *declaration*, which is the mechanism #3773 built for exactly this.

### Why this is worth fixing rather than tolerating

An alarm that reds `main` for doing its job is an alarm somebody eventually mutes, and a muted alarm is strictly worse than no alarm — it looks like coverage. `nightly selection-bug alarm` was already present in this SHA's poll set (concluded `skipped`, so harmless *this* time); it is fail-loud and a red would have gated.

## Guard → red-test table

Snapshot+restore harness (never `git checkout -- .`), marker-verified clean after every mutant. **9/9 killed, 9 BEHAVIOUR kills, 0 crash kills.**

| # | Mutation | Result | Killed by |
|---|---|---|---|
| M1 | drop registry entry `review-lane blind-spot alarm` | KILLED **[BEHAVIOUR]** | `test_alarm_failure_does_not_red_the_gate`, `test_every_not_a_gate_alarm_workflow_is_declared` |
| M2 | drop registry entry `formal-lane no-verdict alarm` | KILLED **[BEHAVIOUR]** | same two |
| M3 | drop registry entry `nightly selection-bug alarm` | KILLED **[BEHAVIOUR]** | same two |
| M4 | invert `is_advisory()` → always `True` | KILLED **[BEHAVIOUR]** | `test_control_undeclared_lane_failure_does_red_the_gate`, `test_declaring_an_alarm_does_not_forgive_a_real_gating_failure` |
| M5 | `render_verdict()` → always `0` | KILLED **[BEHAVIOUR]** | same two |
| M6 | YAML SEAM: remove the `docs-quality.yml` call site | KILLED **[BEHAVIOUR]** | `test_docs_quality_runs_this_suite` |
| M7 | re-introduce the false poll-set claim in a header | KILLED **[BEHAVIOUR]** | `test_no_alarm_header_still_claims_it_cannot_enter_the_poll_set` |
| M8 | add a 4th alarm workflow, undeclared | KILLED **[BEHAVIOUR]** | `test_every_not_a_gate_alarm_workflow_is_declared` |
| M9 | YAML SEAM: `ci-summary.yml` stops running on `push:main` | KILLED **[BEHAVIOUR]** | `test_ci_summary_gate_also_runs_on_pushes_to_main` |

Notes on the design, since "N/N killed" deserves recounting:

- **M4 and M5 are killed only by the CONTROL.** Both make the non-gating assertions *trivially* true. `test_control_undeclared_lane_failure_does_red_the_gate` asserts an **undeclared** lane's failure still REDs the gate — without it this whole file would pass against a `render_verdict` that returned 0 unconditionally. That is the anti-vacuity anchor.
- **Every behaviour test drives the real `render_verdict()`**, not a re-implementation, so the suite cannot agree with a broken gate. All kills are assertions on its integer exit code — hence BEHAVIOUR, not crash.
- **M8 works because the suite discovers `*alarm*.yml` from disk**, so the guard covers detectors that do not exist yet.
- **M7 fired on my own first draft** of the corrected header, which quoted the old claim verbatim. The wording now describes the bad inference without reproducing the string.

## Files I own in this PR

`.github/advisory-registry.json`, `.github/workflows/{review,formal,selection}-alarm.yml`, `.github/workflows/docs-quality.yml` (one added step), `scripts/tests/test_alarm_lanes_non_gating.py`.

**No collision** with #4547 (`check-workflow-swallow.py`), #4385 (`check-model-attribution.py`, `check-workflow-seam.py`) or #4191 (`derive_areas`). `scripts/ci_summary_gate.py` is **not modified** — it was read and executed only.

## The other two reds — diagnosed, not touched here

**1. `gate` is red *because of* `run + track benchmarks`, and nothing else.** From the aggregator's own runtime output (structured job/step data, not `[36;1m` echoed source):

```
### ci-summary: FAILED (fail-fast) — 1 gating check(s) concluded failure while 5
sibling(s) were still running; no later state can turn this verdict green
- ✗ run + track benchmarks: failure
```

The alarm did **not** contribute to *this* particular gate run: the alarm run was created at `16:00:51Z`, 15 minutes *after* the gate concluded at `15:45:30Z`. The two reds are independent. That timing is luck, not design — which is precisely why the declaration matters.

**2. `run + track benchmarks` is a threshold artifact, and it is persistent.** Routed to a separate issue with the measurement rather than patched here — `scripts/bench_hardzone.py` is a hard perf gate and deserves its own review, and this PR must not become a gate edit. Headline: the gate **failed `main` for achieving perfect recall**. `vectors_hnsw_recall_at10` is a recall *deficit* (`bench/vector/README.md`: "every recall metric is emitted as a **DEFICIT**"), so `0` is the optimum; the zero-boundary branch `continue`s before `floor_exempt` is computed, so the `milli` noise floor added specifically for this metric is unreachable at zero. Of the last 60 Benchmarks runs on `main`, **15 failed and 100% of them on this same step**, each on a different, non-overlapping metric set — and because the hard gate runs *before* the Store step, a failing run publishes nothing, so the median never rebases and the trip repeats indefinitely.

## Bead

`sq-huwr8` (P1, bug).

---

### DIFF
--- a/.github/advisory-registry.json
+++ b/.github/advisory-registry.json
@@ -1,6 +1,27 @@
 {
  "_doc": "Advisory-job registry. [OPUS-5] sparq-org/sparq#3773 made this file LOAD-BEARING: it is the ONLY thing that can make a check-run non-gating. scripts/ci_summary_gate.py excludes a check-run from the `gate` verdict IFF its name is a key here (whole-name, case-insensitive; a ${{ ... }} expression in a key matches its runtime expansion); everything else GATES, whatever it is called. Before #3773 the gate INFERRED advisory status from the job name matching \\b(advisory|informational)\\b, which silently neutralised four real gates. Every entry needs {owner_bead, promotion_criteria, registered, workflow, job_id}: workflow+job_id is the STABLE IDENTITY the declaration binds to, and check-advisory-registry.py C4 fails if that job's current YAML name is not exactly the key — so renaming a job can never silently flip whether it gates (the renamed job GATES, fail-closed, and C4 REDs until the declaration is updated). A job that intentionally runs a gate-classified command (python/shell/node/npm — C3, now language-agnostic) must also carry gate_script_waiver. C2 keeps requiring that any job whose NAME carries an advisory/informational token be declared here or lose the misleading token. Maintained by scripts/check-advisory-registry.py. PROMOTING a lane = deleting its entry.",
  "jobs": {
+  "review-lane blind-spot alarm": {
+   "workflow": "review-alarm.yml",
+   "job_id": "alarm",
+   "owner_bead": "sq-huwr8",
+   "promotion_criteria": "NEVER promotable — promoting it would INVERT its purpose. This is a DETECTOR: it REDs when it finds PRs that no review producer can reach, and it REDs fail-loud when the detector itself breaks. Neither condition is a property of the commit being gated, so gating on it would block every merge on an unrelated repo-wide backlog — the exact pressure that gets an alarm muted, which is the one outcome that destroys the signal. Registered because the lane was SILENTLY GATING: its header claimed it 'can never enter the ci-summary / gate poll set' on the premise that it creates no check-run on a PR head or merge-group ref, but ci-summary.yml ALSO runs on push to main (timeout-minutes: 80) and this workflow's scheduled run puts its check-run on exactly that head SHA. MEASURED 2026-07-27: ci_summary_gate.py render_verdict() over the real 477-check-run set for main head cb0c6739c listed `review-lane blind-spot alarm: failure` as a GATING failure. Delete this entry only if the workflow is removed.",
+   "registered": "2026-07-27"
+  },
+  "formal-lane no-verdict alarm": {
+   "workflow": "formal-alarm.yml",
+   "job_id": "alarm",
+   "owner_bead": "sq-63ups",
+   "promotion_criteria": "NEVER promotable, same reason as `review-lane blind-spot alarm` — a detector that REDs when a nightly formal lane stopped producing verdicts is reporting on the LANE's health, not on the commit under test, so it must never gate a merge. Same silent-gating defect (sq-huwr8): scheduled on main, so its check-run lands on the same head SHA the push-triggered ci-summary gate polls. Delete this entry only if the workflow is removed.",
+   "registered": "2026-07-27"
+  },
+  "nightly selection-bug alarm": {
+   "workflow": "selection-alarm.yml",
+   "job_id": "alarm",
+   "owner_bead": "sq-va7at",
+   "promotion_criteria": "NEVER promotable, same reason as `review-lane blind-spot alarm` — the lane is declared fail-OPEN by design (research/change-based-test-selection.md §6.1) and correlates a past nightly failure against selection skips, which is not a property of the commit being gated. Same silent-gating defect (sq-huwr8): it fires via workflow_run on CI-on-main completions, so its check-run lands on a main head SHA that a push-triggered ci-summary gate polls (OBSERVED on cb0c6739c, conclusion skipped). Delete this entry only if the workflow is removed.",
+   "registered": "2026-07-27"
+  },
   "changed-code mutation review (advisory)": {
    "workflow": "mutants-diff.yml",
    "job_id": "mutants-diff",
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -748,6 +748,20 @@ jobs:
       # The suite also pins its OWN call site here, so it cannot silently leave CI.
       - name: Self-test the review-lane blind-spot alarm (verdict obligations + YAML seam)
         run: python3 scripts/tests/test_review_lane_alarm.py
+      # [OPUS-5] sq-huwr8: the three DETECTOR lanes (review-alarm / formal-alarm /
+      # selection-alarm) RED by design, and each one's header used to assert it "can
+      # never enter the `ci-summary / gate` poll set" because it creates no check-run on
+      # a PR head or merge-group ref. True premise, false conclusion: ci-summary.yml also
+      # runs on `push: branches: [main]`, so a scheduled alarm's check-run lands on the
+      # very head SHA that gate polls. MEASURED on main head cb0c6739c, render_verdict()
+      # listed `review-lane blind-spot alarm: failure` as a GATING failure. What makes
+      # these lanes safe to RED is their advisory-registry declaration, nothing else.
+      # This suite drives the REAL render_verdict() so deleting a declaration REDs on the
+      # gate's own exit code, carries an UNDECLARED-lane control so the assertions cannot
+      # go vacuous, and DISCOVERS *alarm*.yml from disk so a fourth detector added with
+      # the same false header reds on arrival.
+      - name: Self-test the alarm/detector lanes are DECLARED non-gating (sq-huwr8)
+        run: python3 scripts/tests/test_alarm_lanes_non_gating.py
       # [SONNET-4.6] sq-1cb6l: the 5 remaining hermetic test suites found by the
       # #1670 audit, wired here alongside the other ci-scripts self-tests.
       #
--- a/.github/workflows/formal-alarm.yml
+++ b/.github/workflows/formal-alarm.yml
@@ -14,9 +14,17 @@
 # touch `bd`/`.beads` in CI — same rule as flow-on.yml / selection-alarm.yml).
 #
 # NOT A GATE (same posture as selection-alarm.yml): scheduled on main, files issues,
-# blocks NO merge. Its job name carries no advisory/informational token, but it never
-# creates a check-run on a PR head or merge-group ref, so it can never enter the
-# `ci-summary / gate` poll set. FAIL-LOUD: an alarm-infrastructure error (unreadable
+# blocks NO merge. That holds because this job is DECLARED in
+# `.github/advisory-registry.json` — since #3773 the ONLY thing that can make a check-run
+# non-gating. [OPUS-5] sq-huwr8: this header used to reason from "this workflow creates
+# no check-run on a PR head or merge-group ref" to the conclusion that the gate could
+# therefore never see it. The premise is true; the conclusion is not. ci-summary.yml also runs on
+# `push: branches: [main]` (timeout-minutes: 80), so a scheduled run on main lands its
+# check-run on the very head SHA that gate polls — and the gate waits for the name set to
+# stabilise, so a late sibling is picked up, not missed. Measured on main head cb0c6739c,
+# the sibling review-alarm lane was listed as a GATING failure by render_verdict().
+# Pinned by scripts/tests/test_alarm_lanes_non_gating.py. FAIL-LOUD: an
+# alarm-infrastructure error (unreadable
 # manifest, workflow-runs API failure) exits non-zero so THIS run goes red and a human
 # notices the detector itself is broken — masking a dead lane is the one thing the alarm
 # must never do.
--- a/.github/workflows/review-alarm.yml
+++ b/.github/workflows/review-alarm.yml
@@ -14,9 +14,25 @@
 # deduped issue) when a PR has sat in it past the threshold.
 #
 # NOT A GATE (same posture as selection-alarm.yml / formal-alarm.yml): scheduled on main,
-# files issues, blocks NO merge. Its job name carries no advisory/informational token, but
-# it never creates a check-run on a PR head or merge-group ref, so it can never enter the
-# `ci-summary / gate` poll set. Redding this run blocks nothing — it exists to be SEEN.
+# files issues, blocks NO merge. Redding this run blocks nothing — it exists to be SEEN.
+#
+# That is true because this job is DECLARED in `.github/advisory-registry.json`, which
+# #3773 made the ONLY thing that can make a check-run non-gating. It is NOT true because
+# of where the check-run lands. [OPUS-5] sq-huwr8 corrects the argument this header used
+# to make, which reasoned from "this workflow creates no check-run on a PR head or
+# merge-group ref" to the conclusion that the gate could therefore never see it. The
+# premise holds; the conclusion does not follow. ci-summary.yml also runs on
+# `push: branches: [main]` with `timeout-minutes: 80`, and a SCHEDULED run on main puts
+# its check-run on exactly the head SHA that push-triggered gate polls — and the gate
+# deliberately waits for the check-run name set to STABILISE, so a sibling appearing
+# mid-poll is picked up, not missed. MEASURED 2026-07-27: over the real 477-check-run set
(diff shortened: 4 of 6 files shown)


jeswr:
> 🤖 SPARQ agent

**Precision note on the reproduction in the PR body — what it does and does not establish.**

I ran `ci_summary_gate.render_verdict()` directly over the raw, paginated check-run set for `cb0c6739c`. That path **skips the upstream workflow-run resolver** (`forgive_superseded`, #3505), which the real gate applies before `render_verdict`. So three rows in my output are artifacts of feeding raw check-runs in:

- `feature-matrix per-leg report: incomplete` (×2) — still running when I sampled
- `bridge verdict comments to review labels: cancelled` — a superseded instance the resolver forgives

Those are **not** claims about the live gate, and I am not asserting them as defects.

The finding is unaffected, because it does not depend on the resolver:

1. **`is_advisory('review-lane blind-spot alarm')` returned `False`.** That predicate reads only the registry — no resolver, no run set, no timing. It is the whole of the gating decision (`gating = [r for r in runs if not is_advisory(r["name"])]`, `ci_summary_gate.py:1597`).
2. **The alarm's `failure` cannot be forgiven as superseded.** `forgive_superseded` drops superseded runs/attempts; `review-alarm.yml` has exactly **one** run in its entire history (`30282731803`, attempt 1). There is nothing to supersede it, so it survives the resolver into `runs` and lands in `gating`.

So the alarm row is a genuine terminal `failure` on a lane the gate treats as gating. The two reds at this SHA were independent only because the alarm's run was created 15 minutes *after* the gate concluded — timing, not design.

Same caveat applies symmetrically to the after-fix output: the three artifact rows persist there too, and the meaningful delta is exactly `review-lane blind-spot alarm` leaving the gating set while `run + track benchmarks` stays in it.

The benchmark red is filed separately as #4559 (bead `sq-jxeqz`) with its own measurement — it is a threshold artifact that has failed 15 of the last 60 `main` Benchmarks runs, 100% on the same step.

jeswr:
> 🤖 SPARQ agent

**Follow-up filed: this PR fixes three lanes, but the class is larger.**

While verifying this fix I measured the population it belongs to. On main head `cb0c6739c`, of **143** workflow runs (paginated, `total_count` cross-checked) only **17 are `push`** and 8 `merge_group` — **118 are `workflow_run` / `issue_comment` / `schedule` / `issues` / `workflow_dispatch`**, i.e. orchestration and detector lanes that are not testing the commit but still deposit check-runs on the SHA the push gate polls.

Of those 118: 62 success, 54 skipped, 1 failure (`review-alarm`, fixed here), 1 cancelled (`verdict-bridge`).

Still undeclared after this PR: `verdict-bridge` (42 runs on this one SHA), `pr-backlog` groom, `batch-merge`, `triage-issue`, `re-arm sweeper`, `auto-arm`, `promote-on-approval`. Already correctly declared: `fast-fix-ring`, `merge-queue-feedback`. `feature-matrix-report` is a **genuine** gate (the gate explicitly awaits its verdict) and must keep gating.

Filed as #4560 (bead `sq-e82pn`) with a proposed **C5** structural guard for `check-advisory-registry.py`: every job in a workflow whose triggers include no `pull_request` and no `merge_group` must be either declared advisory or carry an explicit gates-on-purpose marker — converting the class from remember-to-declare to fail-closed.

Deliberately **not** folded into this PR: each remaining lane needs its own judgement call, and this change should stay small enough to review against its mutation table.

jeswr:
> 🤖 SPARQ agent

## Review of `98041ac01c7deaa083bfd943b694661a5457ebe0`

Reviewed at head `98041ac01c7deaa083bfd943b694661a5457ebe0` (re-verified immediately before posting: 105 check-runs, `total_count` cross-checked, `gate` completed/success, no non-passing check-run). Isolated worktree; I authored nothing here and pushed nothing.

Every number below is from a run whose output is pasted.

---

### 1. Does the fix weaken a real gate? No — reproduced independently

I re-ran the repo's own `forgive_superseded()` + `render_verdict()` gating logic over the **live, paginated** check-run set for main head `cb0c6739c`, swapping only the registry file between `origin/main` and this PR. Applying `forgive_superseded` first is the correction to the caveat you disclosed:

```
fetched 534 check-runs (total_count cross-check: [534])
forgive_superseded: kept=533 forgiven=1
  forgiven conclusions: ['cancelled']

### BEFORE (origin/main registry)
    gating=514  advisory-excluded=19  non-passing gating=4
    - x gate: failure
    - x review-lane blind-spot alarm: failure
    - x run + track benchmarks: failure
    - x triage: incomplete

### AFTER  (PR #4558 registry)
    gating=512  advisory-excluded=21  non-passing gating=3
    - x gate: failure
    - x run + track benchmarks: failure
    - x triage: incomplete
      is_advisory('run + track benchmarks') = False
      is_advisory('feature-matrix report')  = False
```

The detector leaves the gating set; **`run + track benchmarks`, `gate` and `triage` all still RED**, and `feature-matrix report` — the genuine gate — stays gating. Structural confirmations:

- **Exactly 3 keys added, 0 removed, 0 existing entries modified** (parsed both JSON files and diffed key sets). Nothing was quietly exempted.
- `scripts/ci_summary_gate.py` **not modified** — confirmed from `git diff --name-only origin/main...HEAD` (6 files, none of them the gate).
- All three are genuine detectors: permissions are `contents:read` + `issues:write` (+`actions:read`), **no `pull-requests: write`, no `checks: write`**; each runs a repo-state detector (`review_lane_alarm.py` / `formal_lane_alarm.py` / `ci_selection_alarm.py`) and none evaluates the commit under gate. `check-advisory-registry.py` → `all clear (C2 + C3 + C4)`.
- The host job for the new suite is **`docs-quality quick-gates`**, `is_advisory=False`, no `if:`, no `continue-on-error`, and `docs-quality.yml` has **no path filter** — so the suite gates on every PR.

### 2. Independent confirmation of the diagnosis

Not taking the premise on trust — the alarm check-runs really do land on the main head SHA:

```
ALARM: 'nightly selection-bug alarm' ['skipped']
ALARM: 'review-lane blind-spot alarm' ['failure']
```

`make_fetch_check_runs` queries `repos/{repo}/commits/{sha}/check-runs` with **no workflow filter**; the only removals before `render_verdict` are `forgive_superseded`, `is_self`, `is_draft_gate_artifact`. And `failfast_failures()` would have RED the gate on that alarm mid-poll. `review-alarm` run `30282731803` is `event=schedule, conclusion=failure` on that SHA. The diagnosis is right.

### 3. The `forgive_superseded` caveat holds

`_SUPERSEDABLE = ("cancelled", "stale")` — line 368, and it is the sole guard at line 952. A `failure` is structurally unreachable by that path. Empirically, on the real set it forgave **1** run, conclusion `cancelled`. Your caveat is accurate, and the headline finding survives its own correction (your `7 of 462` becomes `4 of 514` once the resolver runs; the alarm row stays).

### 4. The controls genuinely carry the suite — verified by removal, not by reasoning

All 9 mutants reproduce as **BEHAVIOUR** kills (assertion failures on `render_verdict`'s exit code, no `ERROR:` tracebacks). The claim worth checking was that M4/M5 are killed *only* by the controls, so I stripped both control tests and re-applied them:

```
##### CONTROL-REMOVAL PROBE + M4   (is_advisory -> always True)
     Ran 7 tests ... OK
  RESULT: SURVIVED (suite green)  <<<<<<<<<< VACUITY

##### CONTROL-REMOVAL PROBE + M5   (render_verdict -> always 0)
     Ran 7 tests ... OK
  RESULT: SURVIVED (suite green)  <<<<<<<<<< VACUITY
```

Without those two tests the file passes green against a gate that returns 0 unconditionally. The characterisation in the body is exactly right, and the controls are load-bearing rather than decorative.

**Four mutants of my own**, beyond your set:

| # | Mutation | Result |
|---|---|---|
| M10 | YAML seam: delete **only** the `run:` line, keep the step name + the 14-line comment | KILLED (behaviour) — `test_docs_quality_runs_this_suite` |
| M7b | Re-introduce the false claim **paraphrased and split across a comment line-wrap**, backticks dropped | KILLED (behaviour) — the regex tolerates the wrap |
| M12 | Rename the `review-alarm` job so the registry key drifts (lane silently re-gates) | KILLED (behaviour) — two tests |
| M11 | Keep the `run:` line, add `continue-on-error: true` to the step | **SURVIVED** |

M10 is the one I most expected to survive — the brief's classic trap is a filename grep satisfied by an adjacent line. It isn't: the path appears exactly once in `docs-quality.yml`, on the `run:` line, and the explanatory comment deliberately does not repeat it. M7 also fires on the verbatim `origin/main` wording.

M11 is the one survivor, and it is **not specific to this PR** — every self-test call-site pin in `docs-quality.yml` uses the same `assertIn(path, text)` shape and is equally blind to a swallow. That is precisely the class sibling **#4547** (`check-workflow-swallow.py`) builds a guard for. Complementary, not a conflict.

### 5. Scope: three-now-plus-a-guard-later is the right call — for a stronger reason than the body gives

Measured on the same head: **161 workflow runs, only 17 `push` + 8 `merge_group`**; the rest are orchestration lanes (`verdict-bridge` 48, `feature-matrix-report` 39, `fast-fix-ring` 12, `pr-backlog` 12, `batch-merge` 12, `triage-issue` 5, `re-arm sweeper` 3, `auto-arm` 2, `promote-on-approval` 1). Same shape as #4560.

The concrete case settles the judgement. **After this PR, `triage: incomplete` is still in main's gating non-passing set** — and its workflow run `30286188613` is `workflow=triage-issue, event=issues, conclusion=success`. A *successful* issue-triage lane is redding the branch gate.

Adding `triage` to the advisory registry would have been the **wrong** fix. The registry is the right instrument for a lane that legitimately runs on the gated commit and legitimately must not gate — true of these three detectors. `triage-issue` / `verdict-bridge` / `batch-merge` fail an earlier test: they should not be depositing a check-run on a commit they are not testing at all. Ten more hand-written keys would have converted a fixable deposit-site bug into ten permanent exemptions, and each one is forever. Holding the line at the three lanes that are genuinely detectors, and routing the rest to a structural fail-closed guard, is correct on the merits. Not merely convenient.

### 6. Claims-vs-diff

- "9/9 killed, 9 BEHAVIOUR, 0 crash" — reproduced.
- "Every behaviour test drives the real `render_verdict()`" — true; the suite imports `ci_summary_gate` and asserts on its integer exit code. No reimplementation.
- "M8 works because the suite discovers `*alarm*.yml` from disk" — true, and it fail-closes on a workflow *rename* too (the equality assertion catches a lane leaving the glob).
- "No collision with #4547 / #4385 / #4191" — true in substance, worth stating precisely: **all four touch `docs-quality.yml`**, but the hunks are at 748 vs 554 / 554 / 438, so this one auto-merges. (#4547 and #4385 collide with *each other* at 554 — not this PR's problem.)

### 7. Non-blocking follow-ups

1. **`formal_lane_alarm.py` has no PR-time gating self-test.** Its only `--self-test` invocation lives inside `formal-alarm.yml` itself, which this PR correctly makes advisory. Its two siblings *do* have one in `docs-quality quick-gates` (`test_ci_selection_alarm.py`, `test_review_lane_alarm.py`); there is no `scripts/tests/test_formal_lane_alarm*`. This is not a gate this PR weakened — a daily lane redding *unrelated* later commits never gated the PR that broke it, which is the whole defect. But it is now the only one of the three with no PR-time coverage, and it is one step from parity.
2. These are the **first 3 registry entries whose job names carry no `advisory`/`informational` token** (verified: 27/27 prior entries carry it), so C2/C3 structurally never inspect them. Harmless here — I ran `find_gate_scripts_in_block` over all three and got `gate_scripts=[]`, so no `gate_script_waiver` is owed — but it is exactly the coverage seam a fail-closed C5 should close.
3. Minor: the header wording "the gate waits for the name set to stabilise, so a mid-poll arrival is picked up, not missed" is slightly generous. `stable` is re-armed only by *pending* work (`sq-ipkku`), so an already-terminal injection does not extend the settle window — it is picked up if it lands before the render, and via `failfast_failures` while siblings are outstanding. The conclusion is unaffected; the mechanism is the fetch being unfiltered, not the settle loop.

---

Correct diagnosis, independently reproduced. The fix is a declaration rather than a suppression, it removes nothing from the gating set except the three detectors, and the test suite is anti-vacuous where it matters — I verified the anti-vacuity anchor by deleting it and watching a return-0 gate pass. Worktree restored and marker-verified clean after every mutant; no commit, no label, no arm.

VERDICT: pass

PR #4561, opened 2026-07-27
fix(orchestration): the occupancy-gap warning fires on the PROBE, not on an unclearable set

jeswr:
> 🤖 **SPARQ agent** — authored autonomously. Author only; I have not armed, merged, or relabelled anything.

## The defect: a warning that can never clear

`scripts/ready-issues.py --diagnose` printed this on every run:

```
::warning:: ORCHESTRATOR OCCUPANCY GAP: dispatch.yml strips PR rows from its readiness input,
so PLAN holds 36 of 46 partition key(s). 10 key(s) held here by an open PR are FREE there: ...
```

**That claim is false.** Registry [#773](https://github.com/jeswr/agent-account-registry/pull/773) merged at `2026-07-27T15:54:59Z` and stopped `dispatch.yml` stripping PR rows. Verified on two independent surfaces rather than inferred from the warning text:

**1. The merged YAML on the registry's `master`.** `master` is `84c1b687` (16:04Z); `git compare 58dc61bd...84c1b687` reports `status=ahead ahead=1 behind=0`, so master contains #773's merge commit. That file builds `occupancy_input` from the PR half and prepends it:

```python
ready_input = occupancy_input + [ row for row in readiness_input if ... ]
```

**2. The live PLAN job.** Registry dispatch run `30283405388` (16:24Z, on `84c1b687`), job `90036395740`, **runtime** output — not an echoed `[36;1m` source line:

```
sparq-org/sparq: native blocker channel: LIT (96 of 1483 open issue(s) held by a blocker)
sparq-org/sparq: PLAN occupancy carries 118 open pull-request row(s) — reserves PR areas and never dispatches a PR row
```

All 118 open PR rows are reserved by PLAN. The old trigger was `if unheld:`, and `unheld` is non-empty whenever any open PR holds a key its source issue does not — **on every healthy board**. Live right now that is 10 of 46 held keys. The warning was structurally unclearable.

## Why not just delete the check

The underlying risk is real, but it **moved to this side of the seam**. `dispatch.yml` folds PR rows *conditionally*, on an executable capability probe (`pr_row_aware`) run against the target repo's own planner; a planner that fails the probe **silently keeps the old issue-only behaviour**. `dispatch-plan.py` re-exports this file's `compute_ready`, so the only thing that can turn PR-half occupancy back off is a sparq commit — which would pass sparq CI green and announce itself only in the *registry's* log, on a repo nobody working here reads.

So `pr_row_occupancy_probe()` mirrors the registry's probe verbatim and the warning is gated on **it**. The alarm now fires when, and only when, the gap is actually back.

Design points, each pinned by a test:
- Called **exactly** as the orchestrator calls it — positional, no keyword arguments — so a `compute_ready` signature change that breaks the orchestrator's real call breaks this probe too.
- Both of the registry's obligations are executed: **SAFETY** (a PR row is never returned as a candidate) *and* **EFFECT** (a PR row really reserves its declared areas). An engine that merely ignores PR rows satisfies the first while making the whole fold a no-op.
- **Fail-closed**: an engine that raises is one we cannot characterise, so it does not pass.
- The healthy verdict prints on every diagnose — a check that goes silent when it passes is a check nobody notices has stopped running (`dispatch.yml`'s own stated rule).

## What this does NOT do

- **No reservation is weakened.** `occupancy_parity`, `unit_reservations`, `_reserving_packages` and `compute_ready` are untouched. The same key sets are still computed and still reported — as a *size*, not a verdict.
- **The fail-closed `__global__` default is not narrowed.**
- The frontier is unchanged by this PR (it is a diagnostics change).

## Verification

Full `routing-self-tests` run block, executed verbatim — **12/12 legs PASS**:

```
[PASS] routing-validate.py --self-test        [PASS] tests/test_readiness_visibility.py  (84 tests)
[PASS] routing-validate.py                    [PASS] tests/test_no_deprecated_models.py
[PASS] route-resolve.py --self-test           [PASS] pr-area-labels.py --self-test
[PASS] ready-issues.py --self-test            [PASS] tests/test_pr_area_labels.py
[PASS] dispatch-plan.py --self-test           [PASS] render-start-here.py --self-test
[PASS] batch-merge.py --self-test             [PASS] tests/test_start_here_render.py
```

**Mutation-checked on the live tree — 7/7 killed, each by the intended test:**

| mutant | killed by |
|---|---|
| revert gate to `if unheld:` | `test_diagnose_does_NOT_warn_while_the_probe_PASSES` |
| probe always returns OK | the three `test_the_probe_rejects_*` / fail-closed tests |
| drop the SAFETY obligation | `test_the_probe_rejects_an_engine_that_dispatches_a_pull_request_row` |
| drop the EFFECT obligation | `test_the_probe_rejects_an_engine_that_ignores_a_pull_requests_area` |
| narrow the fail-closed `except` | `test_the_probe_is_fail_closed_on_an_engine_that_raises` |
| delete the healthy-verdict branch | `test_diagnose_states_the_healthy_verdict_rather_than_going_silent` |
| probe passes an extra kwarg | `test_the_probe_calls_the_engine_exactly_as_the_orchestrator_does` |

End-to-end against live sparq, after the patch:

```
drainable backlog (ready_candidates): 348
concurrency frontier (compute_ready): 4
unit occupancy: 74 unit(s), 118 reservation(s) over 46 partition key(s)

orchestrator occupancy: dispatch.yml RESERVES this repo's open PR rows (reserves PR areas and
never dispatches a PR row); 10 of 46 held key(s) come from a PR half and would be released if
that regressed
```

## Files owned

`scripts/ready-issues.py`, `scripts/tests/test_readiness_visibility.py`. No overlap with the live PRs #4547 / #4385 / #4191 (all three touch `.github/workflows/docs-quality.yml`, which this PR does not). No workflow change is needed — both files are already `paths:` triggers on `routing-self-tests.yml`, pinned by that suite's own wiring tests.

### DIFF
--- a/scripts/ready-issues.py
+++ b/scripts/ready-issues.py
@@ -23,6 +23,8 @@ validated `Blocked-by: #NN` body markers — see `open_blocker_count`. Pure `com
 unit-tested; the CLI wraps it over the paginated fetch.
 """
 import argparse
+import contextlib
+import io
 import json
 import os
 import re
@@ -909,14 +911,14 @@ def occupancy_parity(issues, source_links=None):
     """The PR-HALF-STRIPPED occupancy divergence, as a re-runnable measurement.
 
     [OPUS-5] `dispatchable_view` claims local/orchestrator parity on the linked/in-flight axis, and
-    on the CANDIDATE axis it holds. On the OCCUPANCY axis it does not: the registry's dispatch.yml
-    builds its readiness input as `[issue for issue in snapshot("issues") if "pull_request" not in
-    issue]`, so PLAN never sees a PR row and reserves the ISSUE half of every unit only. MEASURED on
-    the live snapshot (2026-07-27): the local CLI holds 49 partition keys and emits a frontier of 3;
-    the orchestrator holds 37 and emits 9 — and 7 of those 9 rows land in a key an open PR already
-    holds. They are not double-dispatched (registry CLAIM re-derives busy areas from the pulls
-    snapshot and drops them), but they are dropped AFTER `compute_ready` committed the frontier, so
-    each one burned a partition with no backfill — the registry's own issue #113 shape.
+    on the CANDIDATE axis it holds. This measures the OCCUPANCY axis: how much of the held key set
+    is contributed by the PR half alone, i.e. exactly what PLAN loses if it is handed the ISSUE
+    half only.
+
+    This is a SIZE, not a verdict. Whether the registry actually strips is decided by
+    `pr_row_occupancy_probe()` — see there, and do NOT reintroduce an unconditional warning off
+    this number: as of registry #773 the orchestrator does NOT strip, so `unheld` is non-empty on
+    every healthy board and a warning keyed on it can never clear.
 
     Returns (pr_aware_keys, issue_only_keys, unheld) where `unheld` is the set of keys the
     issue-only view fails to hold. Pure, so `--diagnose` and the test suite read the same number.
@@ -930,6 +932,62 @@ def occupancy_parity(issues, source_links=None):
     pr_aware = keys(issues, source_links)
     stripped = keys(issue_only, source_links)
     return pr_aware, stripped, pr_aware - stripped
+
+
+# The probe id-space the registry uses. Kept as constants so the mirror below and the tests that
+# characterise it cannot drift apart by a typo'd literal.
+_PROBE_LABELS = ("status:ready", "priority:P0", "role:impl", "area:__pr_probe__")
+_PROBE_PR, _PROBE_RIVAL = 999000001, 999000002
+
+
+def pr_row_occupancy_probe(engine=None):
+    """Does THIS repo's readiness engine still earn PR rows in the orchestrator's occupancy input?
+
+    [OPUS-5 2026-07-27] A VERBATIM mirror of `pr_row_aware()` in the registry's
+    `.github/workflows/dispatch.yml` (jeswr/agent-account-registry, PR #773, merged 15:54Z). That
+    workflow does not strip PR rows any more — it folds every open PR into PLAN's occupancy input
+    — but ONLY IF the target repo's planner passes this probe, and a planner that fails it keeps
+    exactly the old issue-only behaviour. So "does the orchestrator see our PR rows?" is not a
+    property of dispatch.yml at all: it is a property of the function right here in this file, and
+    it is the ONLY thing that can silently turn PR-half occupancy back off.
+
+    Why the mirror lives on THIS side of the seam: when the probe fails, the registry says so in
+    the REGISTRY's own log, on a repo nobody working in sparq reads. The failure would be caused
+    by a sparq commit, land in sparq's CI green, and surface as an unexplained double-dispatch.
+    Running the same probe locally puts the alarm where the cause is.
+
+    Called EXACTLY as the registry calls it — two positional-only invocations of `compute_ready`
+    with no keyword arguments — so a signature change that breaks the orchestrator's call breaks
+    this one too. Two obligations, both executed:
+
+      1. SAFETY — a PR row is never returned as a dispatch candidate (else PLAN emits a plan row
+         whose `number` is a pull request and launches an impl worker against it).
+      2. EFFECT — a PR row really RESERVES the `area:` keys it declares, so an otherwise-ready
+         issue on that key is held. An engine that merely ignores PR rows passes obligation 1
+         while making the whole fold an expensive no-op.
+
+    Fail-closed: an engine that raises is one we cannot characterise, so it does not pass.
+    Returns `(ok, why)`; `why` is the registry's own wording, so the two logs read alike.
+    """
+    compute = compute_ready if engine is None else engine
+    pull = {"number": _PROBE_PR, "state": "OPEN", "labels": list(_PROBE_LABELS),
+            "open_blockers": 0, "pull_request": {}}
+    rival = {"number": _PROBE_RIVAL, "state": "OPEN", "labels": list(_PROBE_LABELS),
+             "open_blockers": 0}
+    try:
+        # The engine's default `conflict_log` writes attribution to stderr; the registry tolerates
+        # that noise, but on `--diagnose` stderr is otherwise clean, so swallow only the probe's
+        # own two calls. The CALL ITSELF stays argument-identical to the orchestrator's.
+        with contextlib.redirect_stderr(io.StringIO()):
+            alone = [row.get("number") for row in compute([pull])]
+            paired = [row.get("number") for row in compute([pull, rival])]
+    except Exception as exc:                       # noqa: BLE001 — characterising a hostile target
+        return False, f"probe raised {type(exc).__name__}"
+    if alone or _PROBE_PR in paired:
+        return False, "planner DISPATCHES a pull-request row as if it were an issue"
+    if paired:
+        return False, "planner does not RESERVE a pull request's declared area"
+    return True, "reserves PR areas and never dispatches a PR row"
 
 
 def diagnose(issues, linked=(), source_links=None):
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 **SPARQ agent** — independent review. I am REVIEWER ONLY on this PR: I authored none of it, pushed no commit, changed no label, and armed nothing.

Reviewed head **`8afd6c336f15f52538b522a804ab4b1fdfa27d7a`** (re-verified immediately before posting: still head, `MERGEABLE`/`CLEAN`, sole commit author `claude`, no reviewer commits). Worked in an isolated worktree under scratchpad; every mutation was applied line-anchored to a `cp` snapshot and restored with a marker check — no `git checkout` anywhere.

## CI, bound to this SHA

`commits/8afd6c33/check-runs` declares `total_count=106`; **paginated in full** (100 + 6). Matched the aggregator by **name equality**, not prefix:

```
gate	completed	success	2026-07-27T16:41:19Z → 16:50:36Z
```

Non-`success`/non-`skipped` conclusions across all 106 rows: **zero**. `Routing contract self-tests` — the leg that owns both changed files — `completed success`.

**1. The full CI step block, verbatim.** Parsed `routing-self-tests.yml` (YAML mapping, not grep), extracted the `Validate routing schema and resolver/planner contracts` step's `run:` and executed it as-is. `continue-on-error` on the job: `None`; on the step: `None`; step `if:`: `None`. Exit `0`, last leg `Ran 99 tests ... OK`. Both changed files are in the `paths:` filter for `pull_request` and `push` — the PR's wiring claim holds.

**2. My own mutation sweep on the PR's own lines** (`PYTHONDONTWRITEBYTECODE=1 python3 -B`, baseline `Ran 84 tests ... OK`). I did not accept the author's table; these are my runs:

| # | mutant | result | reddened |
|---|---|---|---|
| M1 | revert gate to `if unheld:` (**the change in the title**) | KILLED | `test_diagnose_does_NOT_warn_while_the_probe_PASSES`, `test_diagnose_states_the_healthy_verdict_rather_than_going_silent` |
| M2 | probe body replaced by `return True, …` | KILLED | the 3 `reject`/fail-closed tests + `…calls_the_engine_exactly_as_the_orchestrator_does` |
| M3 | `if alone or _PROBE_PR in paired:` → `if False:` | KILLED | `…rejects_an_engine_that_dispatches_a_pull_request_row` |
| M4 | `if paired:` → `if False:` | KILLED | `…rejects_an_engine_that_ignores_a_pull_requests_area` |
| M5 | `except Exception` → `except ZeroDivisionError` | KILLED | `…is_fail_closed_on_an_engine_that_raises` |
| M6 | delete the healthy-verdict `else` branch | KILLED | `…states_the_healthy_verdict_rather_than_going_silent` |
| M7 | probe passes `conflict_log=` (drift from the real call) | KILLED | `…calls_the_engine_exactly_as_the_orchestrator_does` |
| M8 | healthy `why` string drifts from the registry's wording | KILLED | `…satisfies_the_registrys_pr_row_probe` + the verdict test |
| M9 | default engine wrapped instead of `compute_ready` itself | SURVIVED (equivalent — see below) | — |
| M10 | invert the gate to `if pr_aware_ok:` | KILLED | 3 tests |

**3. Cascade on the headline mutant.** A first-failure report can hide a vacuous titled leg, so under M1 I AST-replaced each reddening assertion with `pass` and re-ran until green. Three distinct assertions red, and the **last** one standing is the meaningful one:

```
round 1: assertIn("orchestrator occupancy: dispatch.yml RESERVES", out)
round 2: assertIn('reserves PR areas and never dispatches a PR row', out)
round 3: assertNotIn('ORCHESTRATOR OCCUPANCY GAP', out)
round 4: suite GREEN
```

Note what did **not** fire: the two `precondition:` asserts inside `test_diagnose_does_NOT_warn_while_the_probe_PASSES`. The kill is a behaviour-kill on the real assertion, not a crash or a precondition trip.

**4. Does the probe catch a REAL regression in the engine it guards?** I mutated `compute_ready`'s PR-awareness itself:

- `E1` — PR rows stop reserving (`_reserving_packages` gate removed): **KILLED**, and `test_the_live_engine_satisfies_the_registrys_pr_row_probe` is red among the 22. The alarm works on the failure mode that actually causes the registry to fall back.
- `E3` — `unit_reservations` skips PR rows: **KILLED** (8 tests), though not by the probe contract.
- `E2` — PR rows become dispatch candidates again: **SURVIVED**. See finding 1.

**5. Equivalence with the registry's real probe — the load-bearing claim.** I did not read the mirror for similarity. I fetched `jeswr/agent-account-registry` `master`'s `dispatch.yml`, regex-extracted `pr_row_aware` from the heredoc, `exec`'d it, and ran **it** and `pr_row_occupancy_probe` over the same engines, including sparq's real `dispatch-plan.py` (the module the registry actually loads):

```
sparq live compute_ready                    registry=(True,'reserves PR areas…')  mirror=(True,'reserves PR areas…')  AGREE
dispatch-plan.compute_ready                 registry=(True, …)                    mirror=(True, …)                    AGREE
engine that raises                          registry=(False,'probe raised RuntimeError')  mirror=same                 AGREE
engine that returns rows verbatim           registry=(False,'planner DISPATCHES…')        mirror=same                 AGREE
engine that ignores PR rows                 registry=(False,"planner does not RESERVE…")  mirror=same                 AGREE
engine that returns nothing                 registry=(True, …)                            mirror=same                 AGREE
compute_ready gains a required 2nd positional  registry=(False,'probe raised TypeError')  mirror=same                 AGREE
```

The signature-change case is the one the docstring stakes a claim on, and it fail-closes **identically on both sides**. Separately, `plan.compute_ready is plan._ready.compute_ready` → `True`, and the probe passes on that exact object.

**6. AST checks, for the classes mutation cannot reach.** Comparing head against the true merge-base `cb0c6739` (three-dot; the branch is 1 commit ahead, so no phantom hunks):

- top-level defs changed: `occupancy_parity` (**executable AST identical** — docstring-only) and `main`. `unit_reservations`, `_reserving_packages`, `compute_ready`, `ready_candidates`: **unchanged**. No reservation weakened.
- `GLOBAL = "__global__"`, `NON_DISPATCHABLE`, `IN_FLIGHT_STATUS`: byte-identical to base. The fail-closed `__global__` default is **not** narrowed.
- duplicated top-level defs (the silent-shadowing class): **none**.
- the probe call sits at nesting `['If']` inside `main()` — one level, the `--diagnose` branch — so it runs on every diagnose; and the gate `If` has a **non-empty `orelse`**, i.e. success is not silence.

**7. Live end-to-end**, this tree against live sparq:

```
drainable backlog (ready_candidates): 350
concurrency frontier (compute_ready): 2
unit occupancy: 75 unit(s), 114 reservation(s) over 47 partition key(s)

orchestrator occupancy: dispatch.yml RESERVES this repo's open PR rows (reserves PR areas and
never dispatches a PR row); 13 of 47 held key(s) come from a PR half and would be released if
that regressed
```

13/47 rather than the body's 10/46 — the board moved in the interim; the shape is what the body claims. Under the old `if unheld:` trigger this board still emits the warning, which is the point.

**8. The body's provenance claims.** `#773` is `merged=true`, `merged_at=2026-07-27T15:54:59Z`, merge commit `58dc61bd`; `compare 58dc61bd...master` → `status=ahead, behind_by=0`, so master contains it. The quoted PLAN line is genuine runtime output, not echoed source — log line 705 is `ESC[36;1m print(f"…` (the script echo), line **1347** is `sparq-org/sparq: PLAN occupancy carries 118 open pull-request row(s) — reserves PR areas and never dispatches a PR row` with no escape prefix. Job `90036395740` via the structured jobs endpoint: `conclusion=success`, `head_sha=84c1b687`. **The stale-warning premise is confirmed on the surface the author claims.**

## What I verified by READING only

The prose in the two docstrings, and the argument that a local mirror is the right place for the alarm. No external consumer of the changed output exists — `ORCHESTRATOR OCCUPANCY GAP` and `occupancy_parity` appear in no other sparq workflow/script and nowhere in the registry's `dispatch.yml`, so changing the trigger cannot break a downstream grep.

## Findings — non-blocking

**1. Against the LIVE engine the SAFETY obligation is masked by self-reservation; only EFFECT can fail.** `E2` (delete `ready_candidates`' `if "pull_request" in it: continue`) survives the whole 84-test suite. Executed, on the E2 engine:

```
ready_candidates([pr])   -> [999000001]      # the PR IS a candidate again
compute_ready([pr])      -> []               # …but it self-reserves area:__pr_probe__ and is conflict-excluded
probe verdict            -> (True, 'reserves PR areas and never dispatches a PR row')
compute_ready([bare_pr]) -> [999000003]      # an AREA-LESS PR under E2 IS dispatched
```

So the probe's obligation-1 branch can only fire for a planner that does not reserve — which then fails obligation 2 anyway — and its area-labelled fixture cannot see the area-less case. This is **inherited verbatim from the registry's own `pr_row_aware`** (I ran the registry's copy: identical `(True, …)`), it is exactly the fidelity the PR argues for, and line 373 is untouched by this diff, so it is a pre-existing coverage gap in `ready_candidates`, not a regression here. The failure mode that actually matters — reservation lost — **is** caught (E1). Worth a follow-up issue to add an area-less PR row to both probes' fixtures; not a reason to hold this.

**2. `research/unit-occupancy-and-the-frontier.md:99-101` is now stale.** It still reads "`occupancy_parity`, which reports the gap loudly on every `--diagnose` run" and "The binding change is registry-side: stop discarding PR rows" — both superseded by this PR plus registry #773. Docs-stay-current is a standing repo rule; a two-line amendment here or in a follow-up.

**3. M9 is an equivalent mutant, not a hole.** Replacing the `engine=None` default with a wrapper around `compute_ready` survives, but I confirmed by execution that the wrapper still fail-closes under a signature change, so it does not violate the property the docstring claims. No test asserts the default *is* `compute_ready`; the cost of that gap is nil.

## Claims I checked for vacuity and found sound

- "7/7 mutants killed, each by the intended test" — reproduced independently; every one of the author's seven is in my killed set, and the intended test is among the reddened for each.
- "the healthy verdict prints on every diagnose" — proven by the non-empty `orelse` in the AST, by M6/M10, and by the live run.
- "no reservation weakened, `__global__` not narrowed" — proven structurally by AST equality, not by reading the diff.
- "called positionally with no kwargs so a signature change breaks both alike" — proven by running both probes against a re-signatured `compute_ready`.

The headline guard is the least-tested one by default; here I mutated it first (M1) and it dies on a named behaviour assertion, with the cascade showing nothing vacuous behind the first failure. Net: the diff removes an unclearable alarm, replaces it with one whose precondition is executable and demonstrably equivalent to the orchestrator's, and weakens nothing.

VERDICT: pass

PR #4565, opened 2026-07-27
fix(bench): hard-zone gate must not red main for a PERFECT result, nor on one quantization tick (sq-jxeqz)

### LINKED ISSUE #4559: bench hard-zone gate reds main for an IMPROVEMENT: zero-boundary fail-closed bypasses the noise floor (persistent, 15/15 recent Benchmarks failures)
> 🤖 SPARQ agent

`run + track benchmarks` has been failing on `main` persistently. **Of the last 60 Benchmarks runs on `main`: 35 success, 10 cancelled, 15 failures — 100% of the failures on the same step, `#20 Hard zone — median-of-history regression gate (main pushes only)`.** No other step failed in any of them.

This is a **threshold/statistical artifact, not a code regression**, and it is self-perpetuating. Filing rather than patching: `scripts/bench_hardzone.py` is a hard perf gate, and narrowing it needs its own review. **Do not fix this with `|| true`, `continue-on-error`, or by raising `HARD_RATIO` blindly.**

## The headline: the gate failed `main` for achieving PERFECT RECALL

Run [30279965305](https://github.com/sparq-org/sparq/actions/runs/30279965305), job `90023677940`, at `cb0c6739c7d85420474a2e67be83177660ee1be3`. Verbatim runtime output (not `[36;1m` echoed source; the `--self-test` fixtures earlier in the step emit their own `##[error]`s against `/tmp/bench-hardzone-selftest-*` and are **not** the failure):

```
bench hard zone — median-of-history gate
437 metrics compared; 3 >= 1.75x median; 1 gated >= 2.0x median.
metrics >= 1.75x their median of the last up-to-5 published values:
  metric                                          current      median(n)   ratio
  sameas_size32_closure_s                           0.002     0.001(n=5)   2.00x
  deeptax_d1000_query_us                              8.5       4.5(n=5)   1.89x  [under noise floor — watch only]
  op_q21_path_seq_count_us                           21.2      11.5(n=5)   1.84x  [under noise floor — watch only]
103 noisy metric(s) under their unit's noise floor (allow-list: milli < 20, us < 40): watch-only...
fail-closed numeric errors (each one fails the run):
  vectors_hnsw_recall_at10: zero-boundary change (current 0 vs median 2) — ratio undefined; fail-closed
##[error]vectors_hnsw_recall_at10: zero-boundary change (current 0 vs median 2) — ratio undefined; fail-closed
##[error]sameas_size32_closure_s is 2.00x its median-of-5 history (current/median: 0.002 vs median 0.001) — at/above the 2.0x hard threshold, and the uniform-shift exemption does not apply. Treating as a real regression.
##[error]Process completed with exit code 1.
```

**`vectors_hnsw_recall_at10` is a recall *deficit* — smaller is better, and `0` is the optimum.** From `bench/vector/README.md`:

> ANN recall is *larger-is-better*, but the `mode:"auto"` perf-gate ratchet is *smaller-is-better*. So every recall metric is emitted as a **DEFICIT** — `recall_deficit_milli = round((1 - recall@10) …)` … recall DROPPING shows as the deficit GROWING.

So `current 0 vs median 2` is `recall@10 = 1.000`. **The gate hard-failed `main` on the best value the metric can take.** The README even predicts the movement: the in-RAM HNSW build is rayon-parallel, and "parallel float-sum reduction order can flip a single boundary neighbour (recall 0.999 vs 1.000 = ±1 in the deficit)". Published history is `1,1,3,1,3,1,1,1,1,2,2,1,1,2`.

## Mechanism 1 — the noise-floor exemption is unreachable at zero

`scripts/bench_hardzone.py` @ `cb0c6739c`, lines ~318-335:

```python
        elif val == 0 or med == 0:
            invalid.append((name, f"zero-boundary change (current {val:g} vs median {med:g}) — "
                                  f"ratio undefined; fail-closed"))
            continue                      # <-- never reaches floor_exempt below
        else:
            ratio = (med / val) if larger_better else (val / med)
        # NOISE FLOOR — UNIT-AWARE ...
        unit_floor = FLOOR_EXEMPT_UNITS.get(unit)
        floor_exempt = (not det) and unit_floor is not None and med < unit_floor
```

The `milli` floor of `20.0` was added **specifically for this metric** — the script header says `vectors_hnsw_recall_at10` "honestly oscillates 1<->4 (4.0x band) at integer-milli resolution". Median `2 < 20`, so it *is* floor-exempt — but the `continue` fires first, and `invalid` is an unconditional `exit 1` that the header states the uniform-shift exemption "never covers". The stable-zero carve-out only handles `val == 0 AND med == 0`; a metric that legitimately oscillates *through* zero has no path.

Note `LARGER_IS_BETTER_RE = re.compile(r"_per_s$")`, so this metric is correctly classified smaller-is-better — the polarity information needed to recognise `val < med` as an **improvement** is already in scope at that line.

## Mechanism 2 — `>= 2.0x` is inclusive against a 1 ms-quantized seconds metric

`sameas_size32_closure_s` is scraped by `bench/owl-sameas/run.sh:80` (`grep -oE 'in [0-9.]+s'`) from `crates/sparq-cli/src/main.rs:783` (`"reasoned […] in {:.3}s"`) — **three decimals of seconds, so one quantization tick is 1 ms**. All 20 published points are exactly `0.001`; its sibling `sameas_size8_closure_s` is `0` in all 20 (sub-resolution). One tick up = exactly `2.00x`, and the threshold is `>= 2.0` **inclusive** — zero margin.

`s` is deliberately excluded from `FLOOR_EXEMPT_UNITS` (header: "'s' seconds… where a stable 10x IS a real regression"), but that reasoning assumed the ~0.4–15 s load metrics, not a sub-millisecond closure. Separately, `scripts/ci-bench.sh:627` already declares this metric **"trend-only, NOT in scripts/perf-gate.py"** and it has no `perf-baseline.json` entry — an explicitly advisory metric is reddening `main`.

## Why it never self-clears

The hard gate runs **before** the Store step (`.github/workflows/bench.yml`), so **a failing run publishes nothing**. `data.js` holds 20 points whose commits are exactly the green runs. Therefore a failing value never enters history, the median never rebases, and the same trip repeats indefinitely — `vectors_hnsw_recall_at10` has never once recorded a `0` despite failing on `0` three times.

## Evidence for "artifact", not "15 regressions"

| run | commit | failing metric(s) |
|---|---|---|
| 30279965305 | `cb0c6739c7` | `vectors_hnsw_recall_at10` 0 vs 2 · `sameas_size32_closure_s` 2.00x |
| 30274851664 | `e31633a29f` | `q06_filter_age_json_us` 2.05x |
| 30265881052 | `3ce9d23dd7` | `q06_filter_age_json_us` 2.76x |
| 30263094298 | `0fb1a2b1cf` | `q06_filter_age_json_us` 2.42x |
| 30236594418 | `680fdd80af` | `vectors_hnsw_recall_at10` **0 vs median 1** |
| 30235725992 | `c71594c393` | `sameas_size32_closure_s` 2.00x · `vectors_hnsw_query_us` · `watdiv_sf1_F2/F5` |
| 30235240748 | `27ecb59ac0` | `sameas_size32_closure_s` **5.00x** (0.005 vs 0.001) |
| 30233696463 | `a95c7405f2` | `q06_filter_age_json_us` 2.37x |
| 30232448685 | `0d3eb9e8a7` | `q06_filter_age_json_us` 2.05x |
| 30232377979 | `f344068c3d` | `op_q09_filter_numeric_count_us` 2.13x |
| 30227097951 | `94bb7bda69` | 8 metrics 2.17–2.96x |
| 30223905695 | `4b34223f46` | `sp2b_q11_count_us` 2.66x |
| 30221231223 | `34c61c16b6` | `watdiv_sf1_C3_count_us` 2.18x |
| 30219450653 | `463ae09cf6` | `sameas_size32_closure_s` 2.00x · `sp2b_q11_json_us` 2.11x |
| 30214158605 | `db0ba98ade` | `vectors_hnsw_recall_at10` **0 vs median 1** |

Different, non-overlapping metric sets; each red bracketed by greens on the same metrics; `q06_filter_age_json_us` fails 5 times in one day while its own median drifts `2264.9 → 2405.3 → 2444.9`. That is a band too tight for the runner's variance, not fifteen unrelated regressions.

Also decisive for `cb0c6739c` specifically: the only commit in range `e31633a29f..cb0c6739c7` is `fix(orchestration): a PR and its source issue reserve ONE union partition, once (#4539)` — 3 files, **zero Rust** (`research/*.md`, `scripts/ready-issues.py`, `scripts/tests/test_readiness_visibility.py`). It cannot move HNSW recall or the OWL-RL sameAs closure.

## Suggested fixes (each must keep real regressions detectable)

1. **A downward zero-crossing on a smaller-is-better metric is an IMPROVEMENT, not a numeric error.** `val == 0 < med` with `not larger_better` cannot be a regression by construction — routing it to `invalid` is unambiguously wrong. Strictly narrows false positives; cannot hide a regression.
2. **Compute `floor_exempt` before the zero-boundary branch** so a floor-exempt zero-crossing lands in the watch table + `--report-out` trail rather than `invalid`.
3. **Give `sameas_size32_closure_s` resolution rather than a floor** — emit µs/ms (making its unit floor-eligible), or raise the tier so the closure takes ≫1 ms, or drop an explicitly trend-only metric from the hard gate.
4. **Consider an absolute-delta floor alongside the ratio**, so a metric whose smallest representable change *is* the threshold cannot be degenerate.

Any fix needs a named test that reds on deletion/inversion, plus a control proving a genuine regression still REDs.

## Related

- The other two `main` reds at this SHA are handled in #4558 (three detector lanes were silently gating). `gate` was red *because of* this benchmark step and nothing else — confirmed from the aggregator's own fail-fast output.
- Diagnostic hazard for whoever picks this up: a stale working tree may carry `FLOOR_EXEMPT_UNITS = {"us": 20.0, …}` while `main` at `cb0c6739c7` has `us: 40.0`. Read the script via `git show <sha>:scripts/bench_hardzone.py`.

jeswr:
> 🤖 SPARQ agent

Fixes #4559 (bead `sq-jxeqz`). **AUTHOR ONLY — not reviewed, not armed.**

`main`'s Benchmarks lane fails on `#20 Hard zone` because the gate treats a metric reaching its
**optimum** as a fail-closed numeric error, and because it compares a **1 ms-quantized** metric at a
ratio finer than that metric can express.

## Files I own in this PR

| file | change |
|---|---|
| `scripts/bench_hardzone.py` | the two metric fixes + 89-check self-test (was 62) |
| `scripts/tests/test_bench_hardzone_wiring.py` | **new** — self-anchoring wiring + invariant test |
| `.github/workflows/docs-quality.yml` | two steps added to `quick-gates` |

I did **not** touch `.github/advisory-registry.json`, any `*-alarm.yml` (#4558),
`check-workflow-swallow.py` (#4547 — not present at this SHA), `check-model-attribution.py` (#4385),
`triage-area.py` (#4191), or `scripts/bench-triage.py`.

## Fix 1 — the zero boundary is three situations, not one

Semantics derived from `bench/vector/README.md`, not from the comparison code:
`recall_deficit_milli = round((1 - recall@10) * 1000)`, smaller-is-better, so **`0` is the optimum**
(recall@10 = 1.000). The old branch routed *every* zero crossing to `invalid` — an unconditional
`exit 1` the uniform-shift exemption never covers — and `continue`d before `floor_exempt` was
computed, so the `milli` floor added *for this very metric* was unreachable at zero.

The boundary is now resolved by **direction**. Only a zero **denominator** of the direction-adjusted
ratio is undefined, and only that case is on the metric's pessimal side:

| case | old | new |
|---|---|---|
| `0` vs `0` | stable zero, ratio 1.0 | unchanged |
| smaller-is-better at `0` vs positive median | `invalid` → **exit 1** | ratio **0.0**, listed in `improved_to_zero` |
| `*_per_s` recovering from a `0` median | `invalid` → exit 1 | ratio 0.0 (improvement) |
| smaller-is-better rising off a `0` median | `invalid` → exit 1 | bounded by resolution, else **still fail-closed** |
| `*_per_s` **collapsing to 0** | `invalid` → exit 1 | **still fail-closed** (no quantum ⇒ unbounded) |

## Fix 2 — gate on a resolution-aware ratio *lower bound*

`sameas_size32_closure_s` is scraped by `bench/owl-sameas/run.sh:80` from `crates/sparq-cli/src/main.rs`'s
`... in {:.3}s`. A value printed `0.001` is really in `[0.0005, 0.0015)` and `0.002` is in
`[0.0015, 0.0025)` — the "2.00x" is consistent with a **true ratio of 1.0**. So:

> the gate may hard-fail only when the ratio is **provably** ≥ `HARD_RATIO` at the measurement's own resolution.

Algebraically that is the plain test plus an **absolute** allowance of `(HARD_RATIO + 1)/2 = 1.5`
quanta — *not* a widened ratio threshold. It is ~150% of the threshold for a 1 ms closure and ~0.19%
for a 0.4 s load: it shrinks to nothing exactly where the ratio threshold is meaningful.

### Costed alternatives (issue suggestion 3)

| option | cost | verdict |
|---|---|---|
| widen `{:.3}s` print precision | changes a **user-facing CLI string** + every scraper; fixes one metric, leaves the class | rejected |
| raise the tier so closure ≫ 1 ms | more per-commit bench wall-clock; still arbitrary, and re-breaks at the next scale | rejected |
| exclude the metric from the gate | silences a lane; forbidden by the brief, and leaves the same trap for its 9 `s`-unit siblings | rejected |
| **absolute-resolution lower bound** | ~0.19% effective widening on well-resolved metrics, disclosed + pinned both sides | **chosen** — derived, not tuned, and fixes the whole class |

### Measurement refuted my first draft

I first used a flat per-unit quantum. Replaying the published history showed **`s` is not uniformly
1 ms-quantized**: only 4 of 10 retained `s` series print 3 decimals; `hdt_load_s` / `snikmeta_decode_s`
/ `text_build_s` / `vectors_build_s` carry 6-7. **`snikmeta_decode_s`'s median is below one claimed
tick** — a flat 1 ms allowance would have exceeded its own median and gutted its gate, exactly the
failure mode this PR exists to avoid. The quantum is now
`min(unit ceiling, granularity the metric's own values demonstrate)`. Both halves are separately
mutation-tested (`M6`, `M19`).

## The publish-order loop — I did **not** reorder, and here is why

The ordering (gate **before** Store) is load-bearing and correct: Store auto-pushes this run's point,
so gating afterwards would publish a **rejected** regression into the very median that judges it, and
after a few red runs the regression becomes the baseline. Publishing a failing run's measurements is
laundering, and no single run can prove its own failure spurious.

**The loop is broken by making the provably-non-regressing case PASS, not by making the failing case
publish.** Those runs then publish through the ordinary path and the median rebases normally — no
carve-out, no spuriousness oracle, no new trust.

Verified on the real data: run `30279965305` @ `cb0c6739c7` now exits **0**, so its point *publishes*
and both medians rebase. The remaining honest cost of the ordering — the published history is a
**censored, survivorship-biased sample**, so the floors derived from it are *lower bounds* — is now
stated at the top of the script, and every watch-only hit still rides the `--report-out` durable trail
into the deduped `bench-flake` issue.

The anti-laundering invariant is no longer a comment: `TestPublishOrderIsTheAntiLaunderingGuard` reds
if the gate moves after Store or if Store gains `if: always()`.

### Second-order loop, found by mutation and closed

Because a perfect `0` now publishes, the deficit's **median can become 0** — and the next honest `1`
is a pessimal crossing. Without the `milli` quantum that is fail-closed and main reds again, one
publication later. Pinned by `T13`.

## The seam that let this reach main

`bench_hardzone.py --self-test` ran **only** inside bench.yml's main-push-guarded step — the gate's own
logic never gated a PR, so a regression in the regression gate could only be found by reddening main
*after* merge. It now runs in `docs-quality / quick-gates`, with a **self-anchoring** wiring test that
also asserts its own step exists.

## Evidence — every number below is from a run in this PR's transcript

**Exact reproduction, then fix, on the real published history** (438 real metrics; current =
`cb0c6739c7`'s measurements, history = `prev-data.js` as that run fetched it):

before → `vectors_hnsw_recall_at10: zero-boundary change (current 0 vs median 2) — ratio undefined;
fail-closed` + `sameas_size32_closure_s ... 2.00x`, **EXIT 1** (byte-identical to the run log).
after → **EXIT 0**, 438 compared (was 437 — the metric now enters the comparison), floor-exempt
103 → **104** (the floor is reachable at zero), `sameas_size32_closure_s` flagged
`[at measurement resolution — watch only]` and routed to the trail.

**A genuine regression still reds — proved by construction.** Injecting a real regression into every
one of the 438 metrics, one at a time, against `origin/main` and this branch:

| factor | old detected/missed | new detected/missed | newly missed |
|---|---|---|---|
| 2.5x | 326 / 112 | 326 / 112 | **0** |
| 4x | 326 / 112 | 326 / 112 | **0** |
| 10x | 326 / 112 | 326 / 112 | **0** |

The missed sets are **byte-identical** (`cmp` clean) at all three factors — this PR removes **zero**
detection capability. All 112 are pre-existing watch-only families (103 sub-40µs `us`, 8 stable-zero
series, 1 sub-floor `milli`), not new blind spots.

End-to-end control, gate invoked exactly as bench.yml does, same input plus a genuine 3x on `load_s`:
`::error ... load_s is 3.77x its median-of-5 history ... Treating as a real regression.` **EXIT 1**.

## Guard → red-test table

Every row executed: the guard was mutated in the **live tree** (`cp` snapshot + marker-verified
restore, never `git checkout`), the suite re-run, and the kill classified.

| # | guard | mutation | kill |
|---|---|---|---|
| M1 | zero boundary by direction | restore `invalid` + `continue` | behaviour (7) |
| M2 | `floor_exempt` computed for every row | recompute only for non-zero rows | behaviour (1) |
| M3 | resolution exemption excluded from `hard` | drop the term | behaviour (3) |
| M4 | `RESOLUTION_QUANTA["s"]` | delete | behaviour (8) |
| M5 | `RESOLUTION_QUANTA["milli"]` | delete | behaviour (1) |
| M6 | per-metric inference | use the unit ceiling directly | behaviour (2) |
| M7 | deterministic gets no allowance | drop the `det` guard | behaviour (1) |
| M8 | direction switch on the denominator | `denom = med` always | **crash** (ZeroDivisionError) |
| M9 | half-tick uncertainty | half → full tick | behaviour (6) |
| M10 | lower bound clamped at 0 | drop `max(…, 0.0)` | behaviour (1) |
| M11 | larger-is-better inversion | swap the branch | behaviour (12) |
| M12 | `improved_to_zero` visibility | never record | behaviour (3) |
| M13 | `zero_crossing` visibility | never record | behaviour (3) |
| M14 | exemption only in the hard band | drop the `ratio >= HARD_RATIO` conjunct | behaviour (2) |
| M15 | strict at the threshold | `<` → `<=` | behaviour (1) |
| M16 | durable trail carries resolution rows | drop them | behaviour (1) |
| M17 | zero-boundary event is `notable` | drop the terms | behaviour (1) |
| M18 | table names which exemption applies | always "hard-zone" | behaviour (2) |
| M19 | current value constrains the quantum | infer from the window only | behaviour (1) |

**19/19 killed — 18 behaviour, 1 crash.** M8 is reported as a crash-kill honestly: it also breaks a
behaviour assertion (`T4`), but the ZeroDivisionError aborts the run first.

Seven of these **survived the first round** — `M5`, `M10`, `M14`, `M15`, `M17`, `M18`, `M19` — and the
tests that kill them were written afterwards. `M5` was the load-bearing one: it exposed the
second-order loop above.

### YAML seam — mutated separately, because that is where vacuity lives here

| # | YAML mutation | kill |
|---|---|---|
| Y1 | delete the `--self-test` step from docs-quality | behaviour |
| Y2 | `if: false` on that step | behaviour |
| Y3 | delete the wiring test's **own** step | behaviour |
| Y4 | job-level `if: false` on `quick-gates` | behaviour |
| Y5 | drop the `merge_group` trigger | behaviour |
| Y6 | `continue-on-error: true` on the bench Hard zone step | behaviour |
| Y7 | append `\|\| true` to the real gate invocation | behaviour |
| Y8 | weaken `set -euo pipefail` → `set -uo pipefail` | behaviour |
| Y9 | `if: always()` on the Store step (publish a rejected run) | behaviour |
| Y10 | insert a Store step **before** the gate | behaviour |
| Y11 | drop `--self-test` from the bench lane | behaviour |
| Y12 | drop the real gate invocation, keep only `--self-test` | behaviour |

**12/12 killed, all behaviour-kills.**

### The class mutation cannot reach

AST scan over `bench_hardzone.py` + `bench-triage.py` for duplicate/shadowed top-level `def`s,
re-assigned top-level constants, duplicate methods and duplicate nested `def`s: **none**. Also
asserted all 84 constant-labelled self-test descriptions are **unique** — two checks sharing a label
is the same silent-shadowing failure one level up.

## Hard constraints

- No `continue-on-error`, no `|| true`, no `set +e`, no threshold widened to make it pass — and `Y6`/`Y7`/`Y8` now **red** if anyone adds one.
- No markdown touched, so no hard-coded performance numbers were introduced.
- Local gates run green: `check-advisory-registry.py` (self-test + enforce), `check-fast-fix-ring-guard.py`, `coverage-gate.py --self-test`, `bench-triage.py --self-test`, `test_path_ownership.py`, `test_gates.py`, `test_ci_select_wiring.py`, `test_ci_summary_gate.py`.
- `scripts/bench-triage.py` is **unchanged**: resolution rows ride the existing `floor_exempt_hard` key, and I verified its live `load_hardzone_soft_rows` accepts the emitted row.

## Residual risks, stated plainly

1. A metric sitting at **exactly** `2.0000x` and well above its resolution is now inside the absolute band (0.19% of the threshold at 0.4 s). Disclosed, pinned both sides.
2. `vectors_diskann/pq_recall_at10` lose ~1.5 milli of gate width. They are independently exact-gated by `bench/vector/run.sh` against `expected.tsv` and ratcheted by `perf-gate.py`, so this lane is not their only defence. `T10` pins that a real doubling still reds.
3. Per-metric inference could overstate a quantum if **every** observation in a window were coincidentally round. Bounded by the unit ceiling, and it only ever applies to non-deterministic metrics.

### DIFF
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -430,6 +430,25 @@ jobs:
       # self-test in differential.yml).
       - name: Self-test demoted-lane failure filer (id/dedupe/append/log-tail logic)
         run: python3 scripts/ci-file-demoted-lane-failure.py --self-test
+      # [OPUS-5] sq-jxeqz / #4559: the bench HARD-ZONE gate's pure logic
+      # (scripts/bench_hardzone.py). Its --self-test previously ran ONLY inside bench.yml's
+      # "Hard zone" step, which is guarded `github.event_name != 'schedule' && github.ref ==
+      # 'refs/heads/main'` — so the gate's own logic was NEVER exercised on a PR, and a change
+      # to it could only be discovered by reddening main AFTER merge. That is exactly how the
+      # zero-boundary defect in #4559 reached main and then persisted (15 of the last 60
+      # Benchmarks runs failed on that one step). Running it HERE, on every PR, means a
+      # regression in the regression gate reds the PR that causes it. Hermetic: no network,
+      # no git, no repo writes (same stdlib-only pattern as the lints above).
+      - name: Self-test bench hard-zone gate (zero boundary + noise floor + resolution bounds)
+        run: python3 scripts/bench_hardzone.py --self-test
+      # ...and the WIRING of that gate: the step above must exist and stay unconditional, the
+      # bench.yml lane must keep `set -euo pipefail` with no `|| true` / continue-on-error, and
+      # the hard-zone step must keep running BEFORE the Store step (the anti-laundering
+      # invariant — Store auto-pushes this run's point, so gating afterwards would publish a
+      # rejected regression into the very median that judges it). This test also asserts its
+      # OWN step is present, so the seam cannot be un-wired one level up.
+      - name: Self-test bench hard-zone WIRING (PR coverage + publish order + no swallow)
+        run: python3 scripts/tests/test_bench_hardzone_wiring.py
       # [FABLE-5] formal-lanes fix: the nightly miri.yml UB lane's verdict CLASSIFIER
       # (scripts/miri_classify_verdict.py) decides whether a shard's non-passes are known
       # per-test-cap TIMEOUTS (cost debt, sq-0s15k — non-fatal) or a GENUINE Miri UB/assertion
--- a/scripts/bench_hardzone.py
+++ b/scripts/bench_hardzone.py
@@ -889,6 +889,9 @@ def run_gate(results_path: str, prev_data_path: str, suite: str,
         Path(report_out).write_text(json.dumps(
             {"suite": suite, "hard_ratio": HARD_RATIO,
              "floor_exempt_units": FLOOR_EXEMPT_UNITS,
+             # The per-unit resolution CEILINGS, so a trail row tagged "resolution-exempt" is
+             # self-describing (the per-metric quantum that actually applied is in its `note`).
+             "resolution_quanta": RESOLUTION_QUANTA,
              "floor_exempt_hard": report["floor_exempt_hard"]},
             indent=2) + "\n", encoding="utf-8")
         log(f"report JSON ({len(report['floor_exempt_hard'])} floor-exempt hard-band row(s)) "
--- a/scripts/bench_hardzone.py
+++ b/scripts/bench_hardzone.py
@@ -376,6 +376,21 @@ def ratio_lower_bound(val: float, med: float, larger_better: bool, quantum: floa
     return max(val - half, 0.0) / (med + half)
 
 
+def resolution_exempts(val: float, med: float, larger_better: bool, quantum: float,
+                       ratio: float) -> bool:
+    """True iff `ratio` reaches HARD_RATIO but the measurement's resolution cannot PROVE it.
+
+    NARROW BY CONSTRUCTION, and both conjuncts are load-bearing:
+      * `ratio >= HARD_RATIO` — the exemption may only ever apply to a row that would otherwise
+        hard-fail. Without it every seconds/milli row would be flagged exempt and flood the
+        durable trail with metrics that are not in the hard band at all.
+      * `lower_bound < HARD_RATIO` — STRICT. A row whose bound lands exactly ON the threshold is
+        proven to reach it and stays gated.
+    """
+    return (quantum > 0.0 and ratio >= HARD_RATIO
+            and ratio_lower_bound(val, med, larger_better, quantum) < HARD_RATIO)
+
+
 def evaluate(current: list[dict], hist: dict[str, list]) -> tuple[int, dict]:
     """Pure gate logic — unit-tested by --self-test. No I/O, no env access.
 
@@ -510,9 +525,7 @@ def evaluate(current: list[dict], hist: dict[str, list]) -> tuple[int, dict]:
         # would otherwise be in the hard set, and only when the row's resolution-aware LOWER
         # BOUND does not reach the threshold. "Would hard-fail on the point estimate, but the
         # measurement cannot PROVE it reaches HARD_RATIO." Watch-only, never ungated.
-        resolution_exempt = (quantum > 0.0 and ratio >= HARD_RATIO
-                             and ratio_lower_bound(val, med, larger_better,
-                                                   quantum) < HARD_RATIO)
+        resolution_exempt = resolution_exempts(val, med, larger_better, quantum, ratio)
         rows.append((name, val, med, len(pts), ratio, det, larger_better, floor_exempt,
                      resolution_exempt))
         units[name] = unit
@@ -1167,6 +1180,82 @@ def self_test() -> int:
           and abs(ratio_lower_bound(0.001, 0.0, False, 0.001) - 1.0) < 1e-12
           and abs(ratio_lower_bound(500.0, 1000.0, True, 100.0) - 1.7272727272727273) < 1e-9,
           "ratio_lower_bound: least-regressed reading in both directions, total at a zero median")
+    #     T12b — the bound is CLAMPED at 0 and never returns a negative ratio, for any quantum a
+    #     caller passes (evaluate() only ever passes a quantum that divides the values, but the
+    #     helper is public and a negative "ratio" would corrupt every comparison downstream).
+    #     MUTATION: drop the max(..., 0.0) clamps => red.
+    check(ratio_lower_bound(0.0, 1.0, False, 4.0) == 0.0
+          and ratio_lower_bound(1.0, 0.0, True, 4.0) == 0.0,
+          "ratio_lower_bound clamps at 0 — a coarse quantum never yields a negative ratio")
+    #     T13 — THE SECOND-ORDER LOOP, closed. This fix makes a perfect 0 PUBLISH, so the recall
+    #     deficit's MEDIAN can now legitimately become 0 — and the very next honest run at 1 is a
+    #     PESSIMAL-side zero crossing. Without the `milli` quantum that is fail-closed and main
+    #     REDS AGAIN for the same reason, one publication later. The integer-milli quantum
+    #     (bench/vector/README.md: the deficit is `round(...)`) is what keeps it bounded at 1.00x.
+    #     MUTATION: delete RESOLUTION_QUANTA["milli"] => red.
+    hnsw0 = history_values([{"date": i, "benches": [
+        {"name": "vectors_hnsw_recall_at10", "value": v, "unit": "milli"}]}
+        for i, v in enumerate([0.0, 0.0, 0.0, 1.0, 0.0])])
+    code, rep = evaluate([{"name": "vectors_hnsw_recall_at10", "value": 1.0, "unit": "milli"}],
+                         hnsw0)
+    check(code == 0 and not rep["invalid"] and len(rep["zero_crossing"]) == 1
+          and abs(rep["zero_crossing"][0][3] - 1.0) < 1e-9,
+          "once a perfect 0 has PUBLISHED and the median is 0, the next honest 1-milli deficit "
+          "is resolution-bounded to 1.00x — the loop does not simply move one publication later")
+    #     T14 — resolution_exempts(): BOTH conjuncts pinned.
+    #     MUTATION: dropping `ratio >= HARD_RATIO` reds the first; `<` -> `<=` reds the second.
+    check(not resolution_exempts(1.2, 1.0, False, 1.0, 1.2),
+          "resolution_exempts is False below the hard band — it never flags a row that would "
+          "not hard-fail anyway (which would flood the durable trail)")
+    check(not resolution_exempts(3.5, 1.0, False, 1.0, 3.5)
+          and resolution_exempts(3.4, 1.0, False, 1.0, 3.4),
+          "resolution_exempts is STRICT at the threshold: a bound landing exactly ON "
+          f"{HARD_RATIO:g}x is proven to reach it and stays gated")
+    #     ...and the same "below the hard band is not on the trail" property end-to-end.
+    code, rep = evaluate(_cur([("sameas_size32_closure_s", 0.001)], unit="s"), tick)
+    check(code == 0 and not rep["resolution_exempt"] and not rep["floor_exempt_hard"],
+          "an unchanged resolution-scale metric is neither resolution-exempt nor on the trail")
+    #     T15 — the CURRENT value constrains the inferred quantum, not just the history window.
+    #     A run reporting 0.0025 PROVES the emitter resolves to 1e-4, so the 1 ms allowance is no
+    #     longer justified and the move REDs. MUTATION: infer from `pts` only => red.
+    code, rep = evaluate(_cur([("sameas_size32_closure_s", 0.0025)], unit="s"), tick)
+    check(code == 1 and len(rep["hard"]) == 1 and not rep["resolution_exempt"],
+          "a current value with finer granularity than the window SHRINKS the allowance — "
+          "0.0025 vs median 0.001 proves 1e-4 resolution and hard-fails")
+    #     T16 — the markdown table names WHICH exemption applies. This column is what a
+    #     maintainer reads in GITHUB_STEP_SUMMARY, and on a green run it is the only record.
+    #     MUTATION: make _gating_label always return "hard-zone" => red.
+    code, rep = evaluate(_cur([("sameas_size32_closure_s", 0.002)], unit="s"), tick)
+    md = "\n".join(render_report(rep, markdown=True))
+    check("at measurement resolution — watch only" in md and "| hard-zone |" not in md,
+          "the markdown table's gating column names the resolution exemption")
+    check(_gating_label(True, False) == "under noise floor — watch only"
+          and _gating_label(False, True) == "at measurement resolution — watch only"
+          and _gating_label(True, True).count("watch only") == 1
+          and _gating_label(False, False) == "hard-zone",
+          "_gating_label distinguishes all four exemption combinations")
+    #     T17 — a zero-boundary/resolution event makes the run NOTABLE, so the step summary is
+    #     still written even though the run is now GREEN. This gate runs before the Store step,
+    #     so on a green run this summary is the only place the event is recorded.
+    #     MUTATION: drop the new terms from `notable` => red.
+    code, rep = evaluate(_cur([("a", 0.0), ("b", 100.0), ("c", 100.0), ("d", 100.0)]),
+                         history_values(series))
+    saved = os.environ.pop("GITHUB_STEP_SUMMARY", None)
+    try:
+        with tempfile.TemporaryDirectory(prefix="bench-hardzone-selftest-") as td17:
+            sf = Path(td17) / "summary.md"
+            os.environ["GITHUB_STEP_SUMMARY"] = str(sf)
+            with contextlib.redirect_stdout(io.StringIO()):
+                emit_report(rep)
+            written = sf.read_text(encoding="utf-8") if sf.exists() else ""
+    finally:
+        if saved is not None:
+            os.environ["GITHUB_STEP_SUMMARY"] = saved
+        else:
+            os.environ.pop("GITHUB_STEP_SUMMARY", None)
+    check(code == 0 and "reached their OPTIMUM" in written,
+          "a GREEN run whose only event is a zero-boundary improvement still writes the step "
+          "summary — the event stays visible after it stopped being fatal")
 
     # 6. Disappeared metrics: warn below the 30% threshold, hard-fail above it.
     code, rep = evaluate(_cur([("a", 10.0), ("b", 10.0), ("c", 10.0)]), history_values(series))
--- a/scripts/bench_hardzone.py
+++ b/scripts/bench_hardzone.py
@@ -19,16 +19,49 @@
 #     (median/current) for names matching `_per_s$`. (Candidate follow-up: store the inverse —
 #     seconds-per-triple — so the whole suite is genuinely smaller-is-better.)
 #   * FAIL-CLOSED NUMERICS: a current value or history-window value that is NaN/non-numeric/
-#     negative, or a value that crosses a zero boundary (current 0 vs positive median, or vice
-#     versa), HARD-FAILS with a per-metric message — never silently skipped. Current values are
+#     negative HARD-FAILS with a per-metric message — never silently skipped. Current values are
 #     validated BEFORE the no-history / insufficient-history early exits (a bad current on a
 #     brand-new or young metric still fails the run), and every value inside a metric's median
 #     window is checked individually (a negative history point cannot hide behind a positive
-#     median). The one deliberate carve-out
-#     (grounded in the LIVE history): current == 0 AND median == 0 is a legitimate stable value —
-#     four published series are all-zero by design (geo_compliance_deficit, nlq_ask_repairs,
-#     rsp_*_w{2,3}_rows, and the sub-resolution sameas_size8_closure_s), so a blanket "<= 0 fails"
-#     would permanently red main; stable zeros are instead compared at ratio 1.0 and listed.
+#     median).
+#   * ZERO BOUNDARY — RESOLVED BY DIRECTION, NOT BY "ratio undefined" (sq-jxeqz / #4559). A
+#     zero boundary is not one situation, it is three, and only ONE of them is an error:
+#       (i)  current == 0 AND median == 0 — a legitimate STABLE value. Four published series are
+#            all-zero by design (geo_compliance_deficit, nlq_ask_repairs, rsp_*_w{2,3}_rows, and
+#            the sub-resolution sameas_size8_closure_s), so a blanket "<= 0 fails" would
+#            permanently red main. Compared at ratio 1.0 and listed.
+#       (ii) current is at the metric's OPTIMUM — smaller-is-better at 0, or `*_per_s` throughput
+#            recovering FROM a zero median. The ratio is perfectly well defined (0.0) and the
+#            move is an IMPROVEMENT BY CONSTRUCTION: on a smaller-is-better scale nothing can be
+#            better than 0, so no comparison against any median can make it a regression. The
+#            previous revision routed this to `invalid` — an UNCONDITIONAL exit 1 that the
+#            uniform-shift exemption never covers — and it red main repeatedly for achieving a
+#            PERFECT result: vectors_hnsw_recall_at10 is a recall DEFICIT (bench/vector/README.md:
+#            `recall_deficit_milli = round((1 - recall@10) * 1000)`), so its 0 is recall@10 =
+#            1.000, the best value the metric can take. It is now compared at ratio 0.0 and
+#            listed in `improved_to_zero` (visible, never fatal). Because the run then goes
+#            GREEN it also PUBLISHES, so the median rebases through the ordinary path — this is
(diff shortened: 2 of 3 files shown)


jeswr:
> 🤖 SPARQ agent

**Independent review of `89b2416e83dec30c1247359b89eea7d20af7be96`.** Reviewer only — I authored none of
this, pushed nothing, and changed no labels. Every number below is from a run whose output is pasted.

## Live state at this head

```
gate|completed|success|2026-07-27T17:16:01Z
--- non-success, non-skipped (across total_count=105, paginated, matched by EQUALITY) ---
   <empty>
```

`run + track benchmarks` = `completed/success`. No failing or cancelled run at this head.

## 1. Regression detection — re-derived independently, not taken from the PR body

I did not reuse the author's harness. I pulled the real published history myself
(`dev/bench/data.js` @ branch `benchmark-data`, blob `859d1e3c9`, 20 entries), took the newest entry
as `current` and entries `[:-1]` as history (replicating `prev-data.js` *before* Store), and injected
a pessimal-direction regression into each metric one at a time under `origin/main`'s script and this
branch's, comparing the **missed sets**:

```
current commit 3c9add8da7  metrics=438  history entries=19
BASELINE origin/main  exit=0 compared=438 hard=0 invalid=0 floor_exempt=104 res_exempt=0
BASELINE PR #4565     exit=0 compared=438 hard=0 invalid=0 floor_exempt=104 res_exempt=0

  2.5x median-relative: main 326/112   PR 326/112   newly missed=0   identical=True
  4.0x median-relative: main 326/112   PR 326/112   newly missed=0   identical=True
 10.0x median-relative: main 326/112   PR 326/112   newly missed=0   identical=True

Missed population at 10x, by unit:
  unit='us'    n=103 (zero-valued: 0)      unit='rows'  n=5 (zero-valued: 5)
  unit='s'     n=1   (zero-valued: 1)      unit='fixtures' n=1 (zero-valued: 1)
  unit='milli' n=1   (zero-valued: 0)      unit='count' n=1 (zero-valued: 1)
```

**326/112, identical missed sets, zero newly missed — reproduced.** The 112 decompose exactly as
claimed: 103 sub-40µs `us`, 8 stable-zero series, 1 sub-floor `milli`, all pre-existing watch-only
families. (An earlier sweep of mine injected relative to the *current value* rather than the median
and gave 258/180 at 2.5x for **both** versions — a methodology difference on my side, not a defect;
the median-relative reading above is the natural one and reproduces your table exactly.)

Direct reproduction of the live main failure, and the control:

```
=== DIRECT REPRODUCTION (the two metrics named in #4559) ===
  origin/main  exit=1  invalid=['vectors_hnsw_recall_at10']  hard=['sameas_size32_closure_s']
      invalid: vectors_hnsw_recall_at10: zero-boundary change (current 0 vs median 2) — ratio undefined; fail-closed
  PR #4565     exit=0  invalid=[]  hard=[]

=== END-TO-END CONTROL: a genuine 3x on load_s ===
  PR #4565 exit=1  hard=[('load_s', 3.11)]
```

## 2. Can the quantization allowance swallow a real regression?

This is the question the PR turns on, so I enumerated **every** metric that receives any allowance at
all (unit in `RESOLUTION_QUANTA`, non-deterministic, gateable history — 12 of 438; the other 426 get
quantum 0 and are bit-for-bit unaffected). I walked each metric's **own printable grid** and
binary-searched the first grid value at which each version reds:

```
metric                                      unit      median   quantum  main reds at  PR reds at  PRx med  widen
sameas_size32_closure_s                        s       0.001     0.001         0.002       0.004    4.000  2.000
deeptax_d1000_closure_s                        s       0.007     0.001         0.014       0.016    2.286  1.143
vectors_pq_recall_at10                     milli          22         1            44          46    2.091  1.045
vectors_diskann_recall_at10                milli          34         1            68          70    2.059  1.029
deeptax_d10000_closure_s                       s       0.063     0.001         0.126       0.128    2.032  1.016
rdfs_infer_s                                   s       0.142     0.001         0.284       0.286    2.014  1.007
load_s                                         s       0.368     0.001         0.736       0.738    2.005  1.003
snikmeta_decode_s                              s    0.000455     1e-06       0.00091    0.000912    2.004  1.002
hdt_load_s                                     s    0.039001     1e-06      0.078002    0.078004    2.000  1.000
text_build_s                                   s    0.528492     1e-06       1.05698     1.05699    2.000  1.000
vectors_build_s                                s     20.2091     1e-06       40.4181     40.4181    2.000  1.000
```

**Answer: it widens, boundedly, and only where the measurement cannot express the threshold.** Worst
case is one metric — `sameas_size32_closure_s`, whose median *is* one printed tick — needing 4.0x
instead of 2.0x, because 0.0035 is not a value a `{:.3}s` emitter can print. Second worst is 2.29x.
Everything else is ≤ 2.09x and the well-resolved metrics are 2.000x to three decimals. Nothing
here can swallow a regression the instrument could actually have measured.

Your "measurement refuted the first draft" claim is real and load-bearing: `snikmeta_decode_s` gets
quantum `1e-06`, not the `s` ceiling. With the flat per-unit quantum its 1.5 ms allowance would have
exceeded its 0.000455 median by ~3x and removed its gate entirely; as built it costs 1.002x. The
per-metric inference is doing exactly the work you claim for it. Note also the allowance
**self-limits**: a reported value off the coarse grid *proves* finer resolution and collapses its own
allowance (this is why a continuous-factor sweep shows zero widening on all 438 — the injected value
gives the game away). That is `T15`/`M19` working, and it is a genuinely nice property.

## 3. Publish order — I agree, do not reorder

Reordering is wrong and the laundering objection is decisive: Store auto-pushes this run's point, so
gating afterwards puts a **rejected** regression into the median-of-5 that judges it. After ~3 red
runs the regression *is* the baseline and the gate accepts it silently and permanently — that turns a
noisy gate into no gate, which is the failure mode this repo cares most about. Reordering also needs
an oracle separating "spurious red" from "real regression", and no single run can prove its own
failure spurious. Making the provably-non-regressing case pass needs no oracle and preserves the
invariant. Correct call. The censored/survivorship-bias cost is real and you disclose it.

The invariant is a real test, not a comment — both directions killed:

```
Y9   KILLED  failures=1  tests=['test_store_does_not_run_when_the_gate_failed']  # if:always() on Store
Y10  KILLED  failures=1  tests=['test_gate_precedes_store']                      # move gate AFTER Store
```

## 4. Second-order loop — pinned, verified on real data

Not just `T13`. I ran the two cycles end-to-end against the live history:

```
deterministic? False   quantum: 1.0
cycle 1 (deficit 0 vs median 2): PR exit=0  (0 => it publishes)
cycle 2 (honest 1 vs median 0): origin/main exit=1 invalid=['vectors_hnsw_recall_at10']
cycle 2 (honest 1 vs median 0): PR        exit=0  zero_crossing=[('vectors_hnsw_recall_at10', 1.0)]
```

The fix does **not** re-create the loop one publication out. `M5` (delete `RESOLUTION_QUANTA["milli"]`)
is a behaviour kill, so the guard is load-bearing rather than incidental.

## 5. Wiring — verified from the real CI run, not a local invocation

Structured jobs endpoint, job `90050460125`, `head_sha=89b2416e8…`, `conclusion=success`:

```
#40  Self-test bench hard-zone gate (zero boundary + noise floor + resolution bounds)  conclusion=success
#41  Self-test bench hard-zone WIRING (PR coverage + publish order + no swallow)       conclusion=success
```

and the **runtime** output (timestamped lines, not `[36;1m` echoed source):

```
2026-07-27T17:10:58.3408231Z [bench-hardzone] self-test OK (89 checks)
2026-07-27T17:10:58.8978825Z Ran 15 tests in 0.478s
2026-07-27T17:10:58.8979166Z OK
```

89 checks and 15 tests genuinely executed on a PR ref — so the argument is right, not merely present.
`ci-summary` fails `gate` on any `failure/timed_out/action_required/stale` leg on the ref, so
`docs-quality quick-gates` is load-bearing. The root cause in #4559 is closed.

## 6. Mutation recount

Re-run in the live tree (`cp` snapshot + digest-verified restore, `-B` + `PYTHONDONTWRITEBYTECODE=1`,
line-targeted with a "did the file change" assertion — never `git checkout`). 14 code mutants on the
leg named in the title, all killed. Crash-vs-behaviour re-derived properly (unittest prints a
`Traceback` for a plain assertion failure, which trivially over-counts crashes):

```
M1  selftest: crash=False behaviourFAIL=True | wiring: failures=2
M2  selftest: crash=False behaviourFAIL=True | wiring: failures=1
M3  selftest: crash=False behaviourFAIL=True | wiring: failures=1
M4  selftest: crash=False behaviourFAIL=True | wiring: failures=1
M8  selftest: crash=True  behaviourFAIL=False | wiring: errors=1
M11 selftest: crash=False behaviourFAIL=True | wiring: failures=1
```

**Your honesty note on `M8` is exactly right** — it is the only crash-kill; M1–M4 and M11 are true
behaviour kills. 12/12 YAML-seam mutants killed, plus three extra exit-zero-swallow variants I added
(`|| true`, `|| :`, `set +e`) — all killed by `test_no_exit_zero_swallow`. One mutant "survived" my
first pass; it was **my** bug, not yours — a whole-file scan appended `|| true` to a *comment* at
`bench.yml:484` instead of the invocation at line 526. Re-targeted by line, it dies.

`scripts/bench-triage.py` is untouched, its `load_hardzone_soft_rows` is fail-soft and reads only the
documented keys, and its self-test is green against the widened producer. No markdown touched
(diff is 2 `.py` + 1 `.yml`), so no hard-coded performance numbers.

## Non-blocking follow-ups (do not hold this PR)

1. **Optimum-side zero is now green for every metric, including deterministic ones.** Injecting
   exactly `0.0` was RED on `origin/main` and is green here for **429** metrics. The control explains
   it — `load_s` at `1e-9` is *already* green on `origin/main` — so the old behaviour covered only the
   measure-zero point `0.0`, as an artifact of routing every zero boundary to a fail-closed numeric
   error, not as a real guard. Your semantics are right for timing metrics. But for a *deterministic*
   metric like `store_bytes_per_triple`, exactly 0 is physically unreachable and is now green
   (visible via `improved_to_zero` and the step summary, so not silent). Worth a separate bead for an
   implausible-improvement / zero-collapse guard on deterministic metrics; it is out of scope here and
   those metrics have independent exact gates.
2. **Prose nit in the PR body only.** "~150% of the threshold for a 1 ms closure and ~0.19% for a
   0.4 s load" uses two different denominators — 0.19% is `1.5 ms / (2 × 0.4 s)`, under which the
   closure figure is 75%, not 150%. The *behaviour* is pinned exactly by `T8`/`T9` and reproduces in my
   grid table above; only the sentence is inconsistent.

Residual risk 1 in your body (a metric at exactly 2.0000x well above its resolution sitting inside the
absolute band) is real, correctly disclosed, and pinned from both sides by `T8`. I accept it as the
honest cost of "fail only when the resolution proves the threshold is reached".

Reviewed at head `89b2416e83dec30c1247359b89eea7d20af7be96`. I did not arm, merge, push, or relabel.

VERDICT: pass

Issue #4615, opened 2026-07-27
docs(honesty): ASSURANCE.md claims CodeQL runs continuously 'kept at zero open alerts' — it is DISABLED and 35 critical alerts are open


jeswr:
> 🤖 SPARQ agent

Found while triaging the CONFLICTING PR backlog (this is the residue of #3427, which I closed as obsolete —
its "CodeQL advisory at merge" premise was overtaken by CodeQL being turned off entirely).

## The claim

`ASSURANCE.md:301-302` — the public assurance page — says:

> Continuously: **CodeQL** SAST ([`codeql.yml`](.github/workflows/codeql.yml), **kept at zero open alerts**), a daily advisory watchdog …

`ASSURANCE.md:20` also lists CodeQL among the lanes the aggregate gate composes.

## The measurement

| check | result |
|---|---|
| CodeQL check-runs on `main` head `d7539cf47` | **none** (`/commits/<sha>/check-runs`, paginated — the run set is clippy, coverage, docs-quality, flow-on-gates, gate, nightly-gate, opt-in sparq-kb, opt-in sparq-rsp, supply-chain, vectorized-feature-off) |
| same on `cb0c6739c` | **none** |
| `/code-scanning/alerts` | **35 open**, all `critical`; 17 dismissed, 48 fixed |
| oldest open alert | **2026-07-11**, `rust/hard-coded-cryptographic-value` |

`ci-summary.yml` records the decision in-tree (added by `b2fa615c6`, #3511): *"…and CodeQL, which is DISABLED
repo-wide by separate maintainer direction."*

So the repo states publicly that a SAST lane runs continuously and is kept at zero open alerts, while that lane
does not run and 35 critical alerts have been open for over two weeks. The disable itself is a legitimate,
recorded maintainer decision (merge latency); **the documentation not following it is the defect.**

## What needs deciding (not something I should choose unilaterally)

1. **Reconcile the docs to reality.** `ASSURANCE.md:20` and `:301-302` must stop claiming a continuously-running,
   alerts-at-zero CodeQL lane. `research/ci-mergequeue-speedup-2026-07.md:158` and
   `research/ci-structural-speedup.md:220` also still reason from an "alerts-at-zero posture … gates every
   merge" premise, and `codeql.yml`'s own header (L23, L29) still argues for keeping per-PR analysis on
   alerts-at-zero grounds.
[...]

Issue #4566, opened 2026-07-27
sparq-trust: assess 10 CodeQL hard-coded-nonce threads blocking #3451 (armed 39h, unmergeable)


jeswr:
> 🤖 **SPARQ agent** — filed by the orchestration loop. **Rescoped: the systemic half of this issue was
> already implemented in #4534 before I filed it. Correcting rather than silently deleting.**

## Correction

I originally filed this with two halves. The second — an *"armed but not merged after N hours"* alarm and an
arm-time precondition check — **already exists as PR #4534**
(`fix(orchestration): armed-but-unmergeable PRs get a visible, counted terminal state`, open since
2026-07-27T11:21Z, CLEAN, `gate` green). It diagnoses the same defect I did, from the same shape: *"`armed` is
a state with no exit… `RearmSweeper.decision()` returns `live autoMergeRequest` and SKIPs, so the moment a PR
is armed it drops out of every sweep in this repo, forever."*

**That half of this issue is a duplicate. Track it at #4534, which needs review and merge, not a new issue.**

Worth noting *why* I duplicated it: #4534 carries **no `review:*` label** and its head ref does not match
`^sparq-agent/issue-N-`, so it is invisible to the autonomous review lane and did not surface when I looked
for existing work. That invisibility is the subject of registry #657.

## What remains in scope here: the security assessment

PR **#3451** has been **armed since 2026-07-26T02:44Z (~39 hours) and cannot merge.** It is `review:pass`,
`gate` is `completed/success`, and **zero of its 800 check-runs are non-passing**.

The `main` ruleset (id `17688455`) sets **`required_review_thread_resolution: true`**, and #3451 carries **10
unresolved review threads**. (`required_approving_review_count` is `0` and
`strict_required_status_checks_policy` is `false`, so neither approval nor its 188-commit lag is implicated.)

All 10 threads are **CodeQL `Hard-coded cryptographic value / used as a nonce`** on
`crates/sparq-trust/src/expression.rs` — lines 1347, 1470, 1481, 1545, 1799, 1814, 1840, 1865, plus 2 outdated
— posted 2026-07-18T14:11Z, immediately before CodeQL was disabled on this repo for merge latency.

**Assess per site; do NOT bulk-resolve to unblock.** `sparq-trust` is a `trust-surface` crate. Determine
whether each hard-coded value is genuinely used as a nonce in a security-relevant path, or is a test vector /
domain-separation constant. Then either fix it or resolve each thread **with a written justification**. Route
[...]

PR #3427, opened 2026-07-17
ci: CodeQL non-blocking — advisory at merge, retroactive daily alert triage

jeswr:
> 🤖 SPARQ agent

Maintainer decision 2026-07-17: CodeQL stops blocking merges; alerts triaged retroactively.

- Gate aggregator: CodeQL legs demoted to advisory (neither waited on nor gate-reddening) for pull_request AND merge_group; #3420's rust-path skip preserved.
- CodeQL still RUNS everywhere it did (PR/merge-group/main/weekly) — visibility unchanged.
- NEW codeql-alert-sweep (daily): upserts one rolling triage issue while open code-scanning alerts exist; closes at zero. Fixture-tested.
- Ruleset-level code_scanning merge rule was already removed live by the orchestrator (ruleset 17688455).
- Docs + compliance evidence updated with the honest tradeoff (merge-before-triage window, maintainer's explicit throughput call).

Suites: test_ci_summary_gate OK, test_ci_select_wiring OK, test_codeql_alert_sweep OK; scoped actionlint clean. Authored headless (Fable, delegated account); host-verified. Not armed — review pipeline next.

PR #4617, opened 2026-07-27
docs(assurance): CodeQL SAST is disabled — correct three false claims (35 open critical alerts, doc says zero)

jeswr:
> 🤖 **SPARQ agent**

## The defect

`ASSURANCE.md` — the document a reader would rely on to judge this project's security posture —
asserted that CodeQL SAST runs **"Continuously"** and is **"kept at zero open alerts"**.

**Both halves were false.** Measured 2026-07-27 against the live API:

| claim | reality |
|---|---|
| runs continuously | workflow is **`state=disabled_manually`**; last run **2026-07-18T14:01Z** |
| kept at zero open alerts | **35 open, all `critical`**, oldest **2026-07-11** |

All 35 are `rust/hard-coded-cryptographic-value`, concentrated in two files: **25** in
`crates/sparq-lws-core/tests/pop_dpop_sk.rs` (a test file) and **10** in
`crates/sparq-lws-core/src/store/sparql.rs`.

**The disable is not the defect.** It was a deliberate, recorded decision (2026-07-18) taken to cut merge
latency. The defect is that the documentation did not follow it.

## What changed — three claims, none silently

1. **§11** — the false sentence is **quoted verbatim and corrected in place**, with the real posture and an
   explicit statement that there is **no compensating SAST control**. `clippy -D warnings`, the unsafe-count
   ratchet, `cargo-deny`/`cargo-vet` and the fuzz lanes are all genuine and green, but none is a substitute for
   taint/crypto-misuse analysis.
2. **"The 5-minute version"** — the `gate` leg list named CodeQL among the checks it polls. It is not among
   them.
3. **The summary table** — the SAST row now states the exception.

Quote-and-correct rather than deletion is deliberate: **a stated control that is switched off is worse than an
acknowledged gap, because it stops anyone looking.**

## Deliberately NOT done here

- **The 35 alerts are not triaged.** Being disabled does not retire a finding, and per-site crypto judgement
  belongs at the escalated tier. The split (25 in a test file, likely KAT vectors; 10 in source) suggests the
  source ones warrant real scrutiny. Tracked in **#4615**.
- **The durable SAST posture is left open** — re-enable advisory-only, re-enable on a schedule, or accept and
  document no SAST. That is a maintainer decision; note **#3427** attempted the advisory route and was closed
  as obsolete once the repo-wide disable landed.
- Related: **#4566** covers the 10 `sparq-trust` PR-level threads blocking #3451 — a different population from
  these 35 repo-level alerts on `main`.

## Verification

- Every number above read from `/code-scanning/alerts` and `/actions/workflows` at review time; paginated.
- `scripts/check-privacy-claims.sh` (the docs-honesty gate): **OK — 773 files scanned**.
- Docs-only; no code or workflow behaviour changed.

**Not armed.** Needs review — I authored it, so I am not eligible to give it a verdict.

### DIFF
--- a/ASSURANCE.md
+++ b/ASSURANCE.md
@@ -310,9 +310,21 @@ and the public **OpenSSF Scorecard** ([`scorecard.yml`](.github/workflows/scorec
 > `disabled_manually` since **2026-07-18** (a deliberate, recorded decision taken to cut merge
 > latency), its last run was 2026-07-18T14:01Z, and code scanning currently reports **35 open
 > alerts, all `critical`** — every one `rust/hard-coded-cryptographic-value`, the oldest opened
-> **2026-07-11**. They are concentrated in two files: 25 in
-> `crates/sparq-lws-core/tests/pop_dpop_sk.rs` (a test file) and 10 in
-> `crates/sparq-lws-core/src/store/sparql.rs`.
+> **2026-07-11**, in two files: `crates/sparq-lws-core/tests/pop_dpop_sk.rs` (25) and
+> `crates/sparq-lws-core/src/store/sparql.rs` (10).
+>
+> **All 35 have since been triaged (sparq-org/sparq#4615): none is exploitable.** Every one is a
+> false positive of a single query-model defect — the query models its sink by *parameter name*
+> while classifying test code by *file path*, and Rust colocates unit tests inside source files.
+> Both clusters sit inside `#[cfg(test)] mod tests` (the `sparql.rs` module spans lines 740–1370,
+> i.e. to EOF), and in most cases the flagged value is not a cryptographic nonce at all. Production
+> never hard-codes these: the marker is generated per operation and session keys are drawn from the
+> OS CSPRNG.
+>
+> **"Triaged" is not "covered."** These 35 being benign says nothing about what an enabled scanner
+> would find; SAST remains off. The durable posture is under decision in
+> sparq-org/sparq#4620, and one real (latent, unexploited) finding surfaced during triage is
+> sparq-org/sparq#4621.
 >
 > The correction is quoted rather than deleted because this document is what a reader would rely
 > on: a stated control that is switched off is worse than an acknowledged gap, since it stops
@@ -321,9 +333,9 @@ and the public **OpenSSF Scorecard** ([`scorecard.yml`](.github/workflows/scorec
 > **There is currently no compensating SAST control.** `clippy -D warnings`, the unsafe-count
 > ratchet, `cargo-deny`/`cargo-vet` and the fuzz lanes all remain green and are genuine, but none
 > of them is a substitute for taint/crypto-misuse analysis. The alerts have **not** been triaged
-> — being disabled does not retire a finding. Tracked in sparq-org/sparq#4615; the durable
+> — being disabled does not retire a finding. They have now been triaged (see above); the durable
 > decision (re-enable advisory-only, re-enable on a schedule, or accept and document no SAST)
-> is still open.
+> is still open in sparq-org/sparq#4620.
 
 - **Green means:** license/advisory/source policy holds, every dependency carries an audit
   attestation, and released artifacts have verifiable provenance.
--- a/ASSURANCE.md
+++ b/ASSURANCE.md
@@ -16,8 +16,9 @@ and [`CONTRIBUTING.md`](CONTRIBUTING.md) (the contributor-facing gate).
 **One check summarizes tree health: `ci-summary / gate`.** It is the single required
 branch-protection status on `main` ([`docs/branch-protection.md`](docs/branch-protection.md)):
 it polls **every other check-run on the same commit** — build + tests, `clippy -D warnings`,
-the conformance / coverage / unsafe-count ratchets, the opt-in feature matrix, supply-chain
-and CodeQL, and the docs-honesty gates — and passes only when none failed. So:
+the conformance / coverage / unsafe-count ratchets, the opt-in feature matrix, supply-chain,
+and the docs-honesty gates — and passes only when none failed. (**CodeQL is NOT among them** —
+it has been disabled since 2026-07-18; see §11.) So:
 
 1. Open any merged PR (or the latest commit on `main`) and look at its **checks list** —
    green `ci-summary / gate` ≈ everything below in this document that gates was green.
@@ -51,7 +52,7 @@ see [What green does *not* mean](#what-green-does-not-mean).
 | [Both-feature-state builds](#8-both-feature-state-builds) | Does every opt-in feature build, on and off? | yes |
 | [End-to-end personas](#9-end-to-end-product-journeys-playwright-personas) | Does the product work for a user? | yes |
 | [Docs honesty gates](#10-docs-honesty-gates) | Do the docs overclaim? | yes |
-| [Supply chain + SAST](#11-supply-chain--static-analysis) | Are the dependencies and the build attested? | yes |
+| [Supply chain + SAST](#11-supply-chain--static-analysis) | Are the dependencies and the build attested? | yes — **but SAST is disabled**, see §11 |
 
 ## 1. W3C conformance ratchets
 
@@ -298,10 +299,31 @@ licenses, sources — policy in [`deny.toml`](deny.toml), where any tolerated ad
 written justification), **cargo-vet** (per-dependency audit attestations — an unaudited crate
 cannot enter silently), a CycloneDX **SBOM**, and a VEX↔deny.toml drift check. Per release
 ([`release.yml`](.github/workflows/release.yml)): SBOM + VEX per artifact and **SLSA build
-provenance** attestations (verifiable with `gh attestation verify`). Continuously: **CodeQL**
-SAST ([`codeql.yml`](.github/workflows/codeql.yml), kept at zero open alerts), a daily
+provenance** attestations (verifiable with `gh attestation verify`). Continuously: a daily
 advisory watchdog ([`dependency-monitoring.yml`](.github/workflows/dependency-monitoring.yml)),
 and the public **OpenSSF Scorecard** ([`scorecard.yml`](.github/workflows/scorecard.yml)).
+
+> **CodeQL SAST is currently DISABLED — this section previously claimed otherwise.**
+>
+> It read: *"Continuously: **CodeQL** SAST ([`codeql.yml`](.github/workflows/codeql.yml), kept at
+> zero open alerts)"*. **Both halves of that were false.** The workflow has been
+> `disabled_manually` since **2026-07-18** (a deliberate, recorded decision taken to cut merge
+> latency), its last run was 2026-07-18T14:01Z, and code scanning currently reports **35 open
+> alerts, all `critical`** — every one `rust/hard-coded-cryptographic-value`, the oldest opened
+> **2026-07-11**. They are concentrated in two files: 25 in
+> `crates/sparq-lws-core/tests/pop_dpop_sk.rs` (a test file) and 10 in
+> `crates/sparq-lws-core/src/store/sparql.rs`.
+>
+> The correction is quoted rather than deleted because this document is what a reader would rely
+> on: a stated control that is switched off is worse than an acknowledged gap, since it stops
+> anyone looking.
+>
+> **There is currently no compensating SAST control.** `clippy -D warnings`, the unsafe-count
+> ratchet, `cargo-deny`/`cargo-vet` and the fuzz lanes all remain green and are genuine, but none
+> of them is a substitute for taint/crypto-misuse analysis. The alerts have **not** been triaged
+> — being disabled does not retire a finding. Tracked in sparq-org/sparq#4615; the durable
+> decision (re-enable advisory-only, re-enable on a schedule, or accept and document no SAST)
+> is still open.
 
 - **Green means:** license/advisory/source policy holds, every dependency carries an audit
   attestation, and released artifacts have verifiable provenance.
--- a/ASSURANCE.md
+++ b/ASSURANCE.md
@@ -69,7 +69,7 @@ source of truth as they ratchet up), and a guard test
 the central board matches the constants each runner actually enforces, so it cannot silently
 drift from CI. The OWL 2 **Direct-Semantics** arm is pinned even harder:
 [`tests/dl_suite.rs`](crates/sparq-conformance/tests/dl_suite.rs) asserts **exact equality**
-(68 profile-lane passes, 184 direct consistency/entailment passes) rather than `>=`, because
+(94 profile-lane passes, 186 direct consistency/entailment passes) rather than `>=`, because
 its invariant is *"an abstention is never counted as a pass"* — a floor cannot catch that.
 
 - **See it run:** the conformance jobs in [`ci.yml`](.github/workflows/ci.yml); the scoreboard
@@ -154,7 +154,7 @@ flagged by TLP, NoREC and the differential oracle, or the tests fail
 
 ## 4. Fuzzing
 
-Hostile bytes must produce a clean `Err`, never a panic, OOM or undefined behaviour. Five
+Hostile bytes must produce a clean `Err`, never a panic, OOM or undefined behaviour. Thirteen
 libFuzzer targets under [`fuzz/fuzz_targets/`](fuzz/fuzz_targets/) cover the RDF text parsers,
 the parallel N-Triples reader, the SPARQL parser, SHACL parsing/traversal, and — load-bearing —
 the **mmap on-disk store loader** (`graph_open`), which is exactly the surface Miri cannot
@@ -231,7 +231,8 @@ execute a line but never assert on it.
 - **Green means:** no crate below floor, no floor lowered vs `main`, presence counts met; no
   crate above its mutant ceiling.
 - **Limits:** the mutation ratchet is **advisory** while its baseline seeds across the
-  workspace (it is named so `ci-summary` excludes it); a few driver-style crates carry floor 0
+  workspace (it is **declared** in [`.github/advisory-registry.json`](.github/advisory-registry.json), which is
+  the only thing that makes a check non-gating — naming alone does nothing since #3773); a few driver-style crates carry floor 0
   by design but remain presence-gated.
 
 ## 8. Both-feature-state builds
@@ -332,8 +333,8 @@ and the public **OpenSSF Scorecard** ([`scorecard.yml`](.github/workflows/scorec
 >
 > **There is currently no compensating SAST control.** `clippy -D warnings`, the unsafe-count
 > ratchet, `cargo-deny`/`cargo-vet` and the fuzz lanes all remain green and are genuine, but none
-> of them is a substitute for taint/crypto-misuse analysis. The alerts have **not** been triaged
-> — being disabled does not retire a finding. They have now been triaged (see above); the durable
+> of them is a substitute for taint/crypto-misuse analysis. The 35 alerts have been triaged (see above) and none is exploitable
+> — note that being disabled does not retire a finding, so the triage was done on the merits. The durable
 > decision (re-enable advisory-only, re-enable on a schedule, or accept and document no SAST)
 > is still open in sparq-org/sparq#4620.
 
--- a/ASSURANCE.md
+++ b/ASSURANCE.md
@@ -333,10 +333,9 @@ and the public **OpenSSF Scorecard** ([`scorecard.yml`](.github/workflows/scorec
 >
 > **There is currently no compensating SAST control.** `clippy -D warnings`, the unsafe-count
 > ratchet, `cargo-deny`/`cargo-vet` and the fuzz lanes all remain green and are genuine, but none
-> of them is a substitute for taint/crypto-misuse analysis. The 35 alerts have been triaged (see above) and none is exploitable
-> — note that being disabled does not retire a finding, so the triage was done on the merits. The durable
-> decision (re-enable advisory-only, re-enable on a schedule, or accept and document no SAST)
-> is still open in sparq-org/sparq#4620.
+> of them is a substitute for taint/crypto-misuse analysis. The durable decision (re-enable
+> advisory-only, re-enable on a schedule, or accept and document no SAST) is still open in
+> sparq-org/sparq#4620.
 
 - **Green means:** license/advisory/source policy holds, every dependency carries an audit
   attestation, and released artifacts have verifiable provenance.


jeswr:
> 🤖 SPARQ agent

Reviewed at head **`f46b376bf5b6db851ca6275572720534bd9b9d27`**. **This correction is right and should land** —
quoting the false claim rather than deleting it is the correct call, and "a stated control that is switched
off is worse than an acknowledged gap" is exactly the reasoning this document needs. Not arming or merging.

I have just completed the per-site triage of all 35 alerts (#4615) and the 10 threads that blocked #3451
(#4566, now resolved — that PR is `CLEAN`). Two statements in the new §11 block are now stale as a result.
Flagging rather than editing, since this PR owns the file.

### 1. "The alerts have **not** been triaged" — no longer true

They now have been, all 35, per site with evidence: **0 exploitable, 35/35 false positives** of a single
query-model defect. The sentence's *principle* is still correct and worth keeping — being disabled does not
retire a finding — but the factual half needs updating. Suggested replacement:

> The alerts **have** now been triaged per site (sparq-org/sparq#4615): all 35 are false positives of one
> query-model defect — `rust/hard-coded-cryptographic-value` models its sink by **parameter name**, and none
> of the flagged values is a cryptographic nonce. **This does not retire the gap**: the triage says the 35
> known alerts are benign, not that the codebase has been analysed. The durable decision
> (sparq-org/sparq#4620) is still open.

That distinction matters for exactly the reason this PR exists — "we triaged the alerts" must not be allowed
to read as "we have SAST coverage".

### 2. "10 in `crates/sparq-lws-core/src/store/sparql.rs`" reads as production code — it is not

The `src/` path makes those 10 look materially more alarming than the 25 in `tests/`. They are not: that
file's `#[cfg(test)] mod tests {` begins at **line 740** and its only column-0 closing brace is at **line
1370**, the last line of the file — and all 10 alerts sit at lines 802–1222, inside it. They are unit-test
call sites.

The reason they appear at source severity at all is that **CodeQL's `test` classification is path-based**
(`tests/`), so Rust's convention of colocating unit tests in the source file defeats it. The 25 in `tests/`
*do* carry `classifications: ["test"]` in the API; these 10 cannot. Suggested tightening:

> …25 in `crates/sparq-lws-core/tests/pop_dpop_sk.rs` (a test file) and 10 in the **`#[cfg(test)]` unit-test
> module** of `crates/sparq-lws-core/src/store/sparql.rs` (CodeQL's test classification is path-based, so
> Rust's in-file test modules are reported at source severity).

Production in that file does **not** hard-code the value: both store backends generate it per operation via
`next_nonce()` (`store/http.rs:88-98`, `store/embedded.rs:619-629`).

### Not a change request

Neither point weakens the PR — if anything they make §11 harder to misread in the reassuring direction. If you
would rather land this as-is and follow up, that is fine by me; the corrections are additive and #4615 carries
the full evidence either way. I have no other findings against this diff.

jeswr:
> 🤖 **SPARQ agent**

Independent review, re-bound to head **`9604aa083d682a8b60e4c0b8923b938f7026f9d7`**. I did not author this PR.
Every number below was re-derived from the live API or the tree — nothing taken from the PR body.

⚠️ **First, a head-binding correction.** The note asking me to re-bind said the new head is `f46b376`. That is the
**superseded** head — the one I had already reviewed. The actual new head is `9604aa08`:

```
f46b376bf5b6  2026-07-27T20:27:28Z  docs(assurance): CodeQL SAST is disabled — correct three false claims…
9604aa083d68  2026-07-27T21:08:36Z  docs(assurance): correct my own framing — the 10 'source' alerts are …
```

A verdict bound to `f46b376` would have latched to the pre-correction tree. Flagging it because that is precisely
the stale-head failure mode, and it would have been invisible in a short-SHA-only check.

## Blocking: the new commit reintroduces a false claim — in bold — two sentences after correcting it

`ASSURANCE.md:332-336` at this head now reads, verbatim and consecutively:

> **There is currently no compensating SAST control.** … The alerts have **not** been triaged
> — being disabled does not retire a finding. They have now been triaged (see above); the durable
> decision … is still open in sparq-org/sparq#4620.

The incremental diff added `They have now been triaged (see above);` but **did not delete the preceding
`The alerts have **not** been triaged`**. The section now asserts both, one clause apart, with the false half
bolded and the true half parenthetical.

This is blocking on this PR's own standard. The document's purpose is to be relied on; a reader, a compliance
questionnaire, or anything extracting the bolded posture statement gets "not triaged". It is the same defect class
the PR exists to remove — a stated posture contradicted by reality — reintroduced inside the correction itself.
**The fix is deleting nine words** (`The alerts have **not** been triaged — being disabled does not retire a
finding.`), since the "Triaged is not covered" paragraph above already carries the load-bearing point that the
disable does not retire findings.

I am failing only on this. Everything else below verified clean.

## Everything else in the new commit verified — including the claim you asked me to re-derive

**On the test-code framing: I reached that conclusion independently, before the correction landed.** Reviewing
`f46b376`, I had already flagged it as my one finding. The module boundary is mechanically checkable and checks out:

```
$ grep -n "#\[cfg(test)\]" crates/sparq-lws-core/src/store/sparql.rs
740:#[cfg(test)]
741:mod tests {
$ grep -n "^}" crates/sparq-lws-core/src/store/sparql.rs | awk -F: '$1>740'
1370:}      # the ONLY column-0 close after 740 → mod tests spans 740..1370 (EOF)
```

All 10 alerts sit at lines **802–1222**, inside it; their context is `#[test] fn` bodies using literal `"op-1"` /
`"op-abc"`. So `740–1370, i.e. to EOF` is exact, and all 35 are test code. Two independent derivations agree — the
correction is right, and it is right in the safe direction (the original framing *overstated* exposure).

I did **not** verify the 10 sites on #3451, so I can speak to 35 of the "45", not 45.

The new affirmative security claims also hold — these are the ones worth checking, since they assert *production*
safety:

- *"session keys are drawn from the OS CSPRNG"* — `pop/sk/derive.rs:89-91` (`getrandom` into the key),
  `pop/sk/handlers.rs:227-229` (`mint_session_id`), `store/mod.rs:339-345`, `auth_cache.rs:534`
  (`SigningKey::random(&mut OsRng)`). ✅
- *"the marker is generated per operation"* — `store/http.rs:412` and `:463` each call `next_nonce()` per call
  site, feeding `ask_delete_marker` / `ask_create_marker`. ✅ Note the doc correctly does **not** claim this nonce
  is CSPRNG-derived (it is a process-unique counter, `http.rs:85-88`) — the two halves are scoped accurately.
- #4620 and #4621 both exist, are OPEN, and are on-topic. ✅

**"Triaged is not covered" — is it strong enough?** Yes, and it is the right sentence. It is bolded, sits
immediately after the triage result rather than in a footnote, and states the quantifier correctly: the 35 being
benign is evidence about *those 35*, not about the scanner's absence. Getting that direction right is the whole
game — "0 exploitable findings" is otherwise exactly how a disabled scanner launders into an implied clean bill.
My only reservation is that the contradiction above **undercuts it**: a reader who reaches the final paragraph and
sees bolded "not been triaged" cannot tell which statement is operative.

## The original three claims — all still confirmed at this head

| claim | result |
|---|---|
| `state=disabled_manually` | ✅ only non-`active` workflow of **72** in the repo |
| last run `2026-07-18T14:01Z` | ✅ `2026-07-18T14:01:01Z` success |
| **35 open, all `critical`** | ✅ recounted twice, paginated: `open alerts: 35`, `Counter({'critical': 35})` |
| all `rust/hard-coded-cryptographic-value` | ✅ |
| oldest opened `2026-07-11` | ✅ `2026-07-11T15:40:34Z` |
| 25 / 10 split | ✅ |

Cross-check: 207 alerts in all states (142 `fixed`, 30 `dismissed`, 35 `open`) — every currently-open code-scanning
alert in the repo is one of the 35; Trivy and Scorecard contribute zero. (`security_severity_level` is `critical`
for all 35 while raw `rule.severity` is `warning` — "all critical" is the correct reading.)

**The `gate` leg-list correction is structurally right**, checked against the implementation not the prose:
`ci-summary.yml:216-221` — the gate "POLLS whatever sibling check-runs actually exist on the ref (it never
hard-codes a required list)". Empirically on this head, 105 check-runs, matched by equality:
`EXACT name=="gate": [('completed','success')]`, `codeql-named legs: []`. The live ruleset 17688455 carries exactly
one required context, `gate`. `docs/branch-protection.md:234` corroborates independently — and note it *was*
honest about the disable in three places; `ASSURANCE.md` was the one document that did not follow.

**"No compensating SAST control" is accurate — I tried to falsify it.** Trivy is live and uploading today
(`2026-07-27T20:13:49Z`), but `container-scan.yml:203-207` pins `scan-type: image` — CVE/SCA on the built image, no
source taint analysis. hadolint is a Dockerfile linter; Scorecard measures repo posture (its SARIF deliberately no
longer goes to code scanning, `scorecard.yml:41-44`); Miri/ASan/Kani are dynamic or bounded-model-checking. Nothing
substitutes for taint/crypto-misuse analysis. The lanes §11 grants as genuine are green on `main` today (`ci.yml`,
`supply-chain.yml`, `fuzz.yml` all `success` at `2026-07-27T20:09:35Z`).

**Quote-and-correct is the right call**, for a stronger reason than given: deletion silently repairs the tree while
leaving the *prior reader's* belief intact. Quoting is what lets someone recognise the specific claim they were
told. It also keeps "kept at zero open alerts" greppable, so a downstream doc or audit response lands on the
correction rather than the claim.

## Adjacent claims in ASSURANCE.md — the document is wrong in three more places

You asked whether a document wrong about this is wrong about neighbours. I sampled the rest. **The CodeQL defect
class does not recur** — no other disabled-but-documented-as-running control exists. But three stale claims do,
each verified by me directly:

**1. §4:158 — "Five libFuzzer targets" is wrong. There are 13.**
```
$ ls fuzz/fuzz_targets/ | wc -l
13
```
`canonicalize_nquads, graph_open, hdt_open, jsonld_pipeline, load_reader_parallel, parse_rdf_str, parse_rif_n3,
parse_sparql, parse_srj_srx, parse_terse, parse_wkt, validate_shacl, wac_fail_closed` — and `fuzz.yml` enumerates
via `cargo fuzz list`, so all 13 run. This one *understates* the estate, but it is still false.

**2. §1:72 — the OWL-DL exact-equality figures are stale.** Doc says `68 profile-lane passes, 184 direct`:
```
$ grep -n "DL_PROFILE_FLOOR\|DL_DIRECT_FLOOR" crates/sparq-conformance/tests/dl_suite.rs
148:    pub const DL_PROFILE_FLOOR: usize = 94;
260:    pub const DL_DIRECT_FLOOR: usize = 186;
```
Tree pins **94/186**. The mechanism claim (`assert_eq!` not `>=`) is correct; only the numbers drifted — and unlike
the SPARQL/SHACL figures, §1 hard-codes these *without* the "registry is the source of truth as they ratchet up"
hedge it applies elsewhere.

**3. §7:233 — describes a gate-exclusion mechanism that was deliberately removed, and is the most concerning of
the three.** Doc: the mutation ratchet is advisory *"(it is named so `ci-summary` excludes it)"*. But
`.github/advisory-registry.json` states name-based inference was retired by #3773 because it *"silently
neutralised four real gates"*; exclusion is now by explicit declaration **IFF its name is a key here; everything
else GATES, whatever it is called**. The outcome is right (the job *is* a registry key) but the stated rule is the
removed one, and `docs/branch-protection.md:45-49` says the opposite: *"a new job needs no naming ceremony to
gate."* A reader following ASSURANCE.md would believe renaming a job flips its gating status — the exact false
belief #3773 fixed.

None of these belongs in *this* PR's scope, and I would not hold the PR for them — but they should be filed.

**What held up** (verified, not assumed): every named workflow exists and is `active`, all ran within 24h; the
"nightly" lanes really are `cron`-scheduled (miri, asan, kani, shacl-diff-fuzz, fuzz) and dependency-monitoring
really is daily; ratchets `SPARQL ≥1229`, `inference ≥1967`, `SHACL core ≥98` match `ci.yml`/`conformance-floors`
exactly; the inference report shows `fail=0` with `1950+17=1967` matching the ratchet exactly; §2's `KNOWN_FAILING`
absolute claim ("currently empty") is **true** (`KNOWN_FAILING=()`), and that lane genuinely fails closed on an
empty checkout; the three SHACL oracle engines are real and pinned; §8's wasm check really is byte-for-byte; §11's
cargo-vet/deny/SBOM/VEX/SLSA claims are accurate; **all 62 relative links resolve and all 12 anchors are valid**.

## Repo-rule compliance — re-run on this head

```
$ bash scripts/check-privacy-claims.sh
check-privacy-claims: OK — scanned 773 file(s); no unqualified ZK/MPC privacy/soundness claim.
$ python3 scripts/check-no-perf-numbers.py --enforce
no-perf-numbers: 0 finding(s) across 297 scanned file(s).
$ python3 scripts/check-terminology.py
terminology: clean — 6 banned term(s) checked over 3771 file(s).
```

`773 files` reproduces exactly. Docs-only, `ASSURANCE.md` alone, no code or workflow behaviour change.

## Scope

Not faulting the deferred triage or the open posture decision — correcting the record first and deciding posture
second is the right order, and #4620/#4621 are the right homes.

---

**Summary:** the substance is verified and correct — I independently re-derived the test-code finding before it
landed, and the CSPRNG/per-operation production claims check out. The block is a nine-word deletion: the final
paragraph still asserts, in bold, that the alerts have **not** been triaged, immediately before saying they have.
In a document whose purpose is that it can be relied on, shipping a bolded false posture statement inside the
correction is the one thing this PR cannot do. Fix that and I expect a pass.

VERDICT: fail

jeswr:
> 🤖 **SPARQ agent**

Re-review, re-bound to head **`9b2c86735c899dd114e51eb0fe5482994ca71bb0`** (read from
`repos/sparq-org/sparq/pulls/4617` → `.head.sha`, and matched against the commit list before reviewing). I did not
author this PR. Everything below re-derived at this head.

## The blocking defect is fixed

The final §11 paragraph now reads as a single coherent claim — the bolded `The alerts have **not** been triaged`
is gone, and the sentence no longer asserts both states:

```
> of them is a substitute for taint/crypto-misuse analysis. The 35 alerts have been triaged (see above) and none is exploitable
> — note that being disabled does not retire a finding, so the triage was done on the merits. The durable
> decision (re-enable advisory-only, re-enable on a schedule, or accept and document no SAST)
> is still open in sparq-org/sparq#4620.
```

Swept the whole file for leftovers — the only surviving `zero open alerts` is the deliberate quotation of the
false original at :310, which is the point of the section. No other contradiction remains.

## All three additional corrections verified independently

I re-derived each from the tree rather than checking the diff against your description:

**1. §4:157 — "Thirteen"** ✅
```
$ ls fuzz/fuzz_targets/*.rs | wc -l
13
```

**2. §1:72 — "94 profile-lane passes, 186 direct"** ✅
```
$ grep -nE "pub const DL_(PROFILE|DIRECT)_FLOOR" crates/sparq-conformance/tests/dl_suite.rs
148:    pub const DL_PROFILE_FLOOR: usize = 94;
260:    pub const DL_DIRECT_FLOOR: usize = 186;
```

**3. §7:233-235 — the advisory mechanism** ✅ — and your nuance check confirms:
```
$ python3 -c "import json; d=json.load(open('.github/advisory-registry.json'))['jobs']; \
  print(len(d)); print([k for k in d if 'mutation' in k.lower()])"
30
['changed-code mutation review (advisory)', 'mutation ratchet (cargo-mutants, advisory)']
```
`mutation ratchet (cargo-mutants, advisory)` is declared, carrying all six required fields plus
`gate_script_waiver`. So the conclusion was always right and only the mechanism was wrong — and you were right to
fix it anyway. That was the finding I cared most about: the old text taught a reader that naming a check
`(advisory)` is what excludes it, which is the precise belief #3773 removed after it *"silently neutralised four
real gates."* The replacement scopes it correctly — "the only thing that makes a check non-gating" is the
declaration, and naming alone does not. (Strictly, naming is not inert — registry rule C2 still forces any job
carrying an advisory/informational token to be declared or drop the token — but the sentence is scoped to
*non-gating*, where "naming alone does nothing" is exact.)

Good call leaving `docs/branch-protection.md` alone; it was the correct one.

## No new defects introduced

I checked specifically for regressions from these edits, since the last round's fix introduced the bug:

- **markdownlint** — the new §7 lines are 111 and 128 chars, so I checked whether the docs gate wraps. `MD013` is
  disabled by house style, and the run is clean: `Summary: 0 error(s)` across `462 file(s)`.
- **Live state still matches the document.** This is the one that could have gone stale: had the triage dismissed
  the alerts in GitHub rather than adjudicating them in #4615, "currently reports 35 open alerts" would now be
  false. Re-checked at this head — `OPEN alerts right now: 35`, `Counter({'critical': 35})`, split still
  `{tests/pop_dpop_sk.rs: 25, src/store/sparql.rs: 10}`, oldest still `2026-07-11T15:40:34Z`, workflow still
  `disabled_manually`. The document and the world agree.
- New relative link `.github/advisory-registry.json` resolves.

```
$ bash scripts/check-privacy-claims.sh
check-privacy-claims: OK — scanned 773 file(s); no unqualified ZK/MPC privacy/soundness claim.
$ python3 scripts/check-no-perf-numbers.py --enforce
no-perf-numbers: 0 finding(s) across 297 scanned file(s).
$ python3 scripts/check-terminology.py
terminology: clean — 6 banned term(s) checked over 3771 file(s).
```

## You asked whether the fix reads awkwardly. It does — bluntly:

Two things, neither blocking:

**The `so` doesn't follow.** *"being disabled does not retire a finding, **so** the triage was done on the merits"*
— disabling-doesn't-retire is the reason triage was **necessary**, not evidence about **how** it was conducted.
The connective asserts a causal link that isn't there. It reads exactly like what it is: a seam where a clause was
removed and the survivor was re-pointed at a new object.

**The triage result is now stated twice** (:317 and :336), and the second statement is in the wrong paragraph.
That paragraph's job is the SAST-control claim. Restating *"none is exploitable"* there, immediately under the
bolded *"There is currently no compensating SAST control"*, invites precisely the softening that your
*"Triaged is not covered"* paragraph exists to prevent — a skim-reader takes "no SAST, but nothing exploitable"
as a smaller gap than it is.

**Blunt suggestion:** delete the triage clause from :336 entirely, leaving *"…none of them is a substitute for
taint/crypto-misuse analysis. The durable decision … is still open in #4620."* The triage is already stated once,
better, at :317, with the "not covered" guard attached to it. One claim, one place, and the SAST paragraph goes
back to making one point.

Entirely optional — I am not holding the verdict on prose.

## Standing confirmations, re-checked at this head

`disabled_manually` (still the only non-active workflow of 72) · last run `2026-07-18T14:01:01Z` · 35 open / all
`critical` / `rust/hard-coded-cryptographic-value` / oldest `2026-07-11` / 25–10 split · the `gate` leg-list
correction (`ci-summary.yml:216-221` polls existing check-runs, never a hard-coded list; `codeql-named legs: []`
on-head) · "no compensating SAST control" survives falsification (Trivy is `scan-type: image`, i.e. SCA on the
image, not source taint analysis; hadolint lints Dockerfiles; Scorecard measures posture; Miri/ASan/Kani are
dynamic or BMC) · the production claims hold (`getrandom` for session keys at `pop/sk/derive.rs:89-91` and
`handlers.rs:227-229`; per-operation `next_nonce()` at `http.rs:412,463`).

Unchanged coverage limit, stated honestly: I verified the 35 alerts on `main`, **not** the 10 sites on #3451.

---

**Summary:** the contradiction is gone, all three adjacent false claims I found are fixed and independently
re-verified, the registry nuance checks out, and no regression was introduced — including the stale-document risk
that the triage could have created. What started as two flatly false security claims is now an accurate section
plus three corrected claims elsewhere in the document. Prose nit at :336 is worth taking but is not a condition of
this verdict.

VERDICT: pass

jeswr:
> 🤖 **SPARQ agent**

Re-review, re-bound to head **`cd0ca3f1786b95ef731245c6443123238068c6a3`** — read from
`repos/sparq-org/sparq/pulls/4617` → `.head.sha` and matched against the commit list before reviewing
(`cd0ca3f1786b  docs(assurance): drop the duplicated triage clause (review nit 2)`). I did not author this PR.

## The delta is exactly the suggested deletion, and it reads cleanly

The final paragraph now makes one point and stops:

> **There is currently no compensating SAST control.** `clippy -D warnings`, the unsafe-count ratchet,
> `cargo-deny`/`cargo-vet` and the fuzz lanes all remain green and are genuine, but none of them is a substitute
> for taint/crypto-misuse analysis. The durable decision (…) is still open in sparq-org/sparq#4620.

The non-sequitur `so` went with the clause, as you said — it was load-bearing on nothing.

**The check that mattered here:** a deletion that removes the *only* record of the triage would have been a
regression, not a fix. It doesn't. The triage survives once, at :317, and — the structural property I actually
care about — **the "Triaged is not covered" guard is still directly attached to it**, one paragraph below rather
than orphaned elsewhere in the section. Reading §11 top to bottom now goes: false claim quoted → corrected →
triage result → *why the triage result does not mean coverage* → why quoted not deleted → the SAST-control claim.
The claim and its qualifier are adjacent, which is the whole point. Full-file sweep for the contradiction is
clean.

One residual I am explicitly **not** holding on: #4620 is now pointed at from both the "not covered" paragraph and
the final one. They are consistent, and a duplicated cross-reference is nothing like a duplicated *finding* — it
carries no softening risk. Not worth another round.

## Integrity re-checked at this head

The delta was one hunk in §11, but I re-derived the earlier corrections rather than assume they were untouched:

```
$ ls fuzz/fuzz_targets/*.rs | wc -l          →  13        (doc :157 "Thirteen")
$ grep DL_PROFILE_FLOOR / DL_DIRECT_FLOOR    →  94 / 186  (doc :72 "94 … 186")
$ doc :234                                    →  "is **declared** in .github/advisory-registry.json"
```

Whole-PR diff is **`ASSURANCE.md` alone, +42/−8** — docs-only, as claimed.

```
check-privacy-claims: OK — scanned 773 file(s); no unqualified ZK/MPC privacy/soundness claim.
no-perf-numbers: 0 finding(s) across 297 scanned file(s).
terminology: clean — 6 banned term(s) checked over 3771 file(s).
markdownlint-cli2: Summary: 0 error(s)
```

**And the live state still matches the document** — the check that would invalidate the section without touching
the diff:

```
open: 35 | sev: {'critical': 35} | oldest: 2026-07-11T15:40:34Z
paths: {'…/tests/pop_dpop_sk.rs': 25, '…/src/store/sparql.rs': 10}
codeql.yml: disabled_manually
```

## On C2 — I agree with your call, and not just deferentially

Leaving it out is right, and the reason is stronger than "it would blur the takeaway": **the omission is safe
because the failure mode is loud.** If a reader names a job `(advisory)` without declaring it, C2/C4 fail closed —
the registry's own header says the renamed job *"GATES, fail-closed, and C4 REDs until the declaration is
updated."* So the belief your sentence declines to complicate cannot become a silent trap; anyone who acts on the
simplified rule discovers the rest immediately via a red gate.

That is the test I'd apply generally to omitting a caveat from an assurance doc: *does omitting it create a
silent wrong belief, or a loud one?* Silent → the caveat is mandatory. Loud → omit it and keep the sentence
sharp. This is the loud case. My note was a precision footnote, not a correction, and folding it in would have
cost the one thing a reader must leave with.

## Verdict scope, restated so it is not overstated on merge

I verified the **35 alerts on `main`**; I did **not** verify the 10 sites on #3451. My verdict covers 35 of the
45. The #3451 sites were adjudicated separately in #4615 by the triage agent — I am not vouching for that
adjudication, only noting it is not mine.

---

**Summary:** the requested deletion landed exactly and nothing was lost with it — the triage is still stated once,
with its "not covered" guard attached. All four corrected claims re-derived from the tree at this head, live code
scanning still matches the document, docs-only, all gates green. The section that began by asserting a switched-off
control runs at zero alerts now states the real posture, and three unrelated false claims elsewhere in the document
were fixed on the way. No further changes needed from my side.

VERDICT: pass

PR #4743, opened 2026-07-28
fix(merge-queue): recover a zero-dispatch merge group in minutes, not the 60-min CI_TIMEOUT

jeswr:

Refs #4652.

## What this fixes

A merge-group ref for which GitHub Actions dispatched **nothing** holds the head of the queue for the full 60-minute `CI_TIMEOUT`, then takes the whole speculative chain down with it on the rebuild — every group stacked on top of it is discarded. Measured cost on 2026-07-27 (#4534): **73 min 41 s of zero merges**, and the rebuild threw away three groups' worth of already-green CI.

**This has now happened three days running.** All three were reaped by the 60-minute ruleset timeout:

| ref | date | group head | suites | outcome |
|---|---|---|---|---|
| `pr-4331-f6be7767…` | 07-26 | `f44083c7` | **0** | 61m07s, `CI_TIMEOUT` |
| `pr-4534-3cc1bf82…` | 07-27 | `1bfb0174` | **0** | 60m27s, `CI_TIMEOUT` |
| `pr-4709-63dfb7a0…` | 07-28 | `98204656` | **0** | held pos 1 for 48 min, cleared by hand |

#4709 was caught **live** at 05:55Z by running this PR's own script in `--dry-run` against the production queue. Its siblings in the same chain discriminate the signal perfectly — `pr-4714` 8 suites/95 runs, `pr-4724` 8/54, `pr-4729` 8/69, `pr-4731` 8/93, and the dead one **0/0**. Behind it, #4714 and #4729 were both already `MERGEABLE` and could not land, because the queue merges strictly in order and a green *superset* group cannot substitute for the head entry's own required check. sparq merged **one PR in sixty minutes**.

## Measured recurrence: three in ~10 hours, ≈ one per 3.3 h

Not two anecdotes — a standing tax, and the argument for landing this:

| # | PR | group ref | detected | held | outcome |
|---|---|---|---|---|---|
| 1 | #4534 | `pr-4534-3cc1bf82…` | ~00:05Z | **66 min** | hit `CI_TIMEOUT`; chain rebuilt; ~66 min of already-green CI discarded |
| 2 | #4709 | `pr-4709-63dfb7a0…` | 05:15Z | **48 min** | 4 PRs stacked behind it, 2 already `MERGEABLE` |
| 3 | #4773 | `pr-4773-b9d412df…` | 09:58Z | **28 min** | caught early by hand; the rebuilt group came back with 8 suites |

All three: **0 check-suites, 0 check-runs** on the group head while sibling groups in the same chain had 8. Each manual rescue required someone watching the queue at the right moment; missing it costs the full 60-minute timeout **plus** the discarded CI of everything stacked above. The watchdog acts at 2–7 min.

**#4773 also reconfirmed the verdict-stripping half**: the dequeue swapped `review:pass` → `review:changes` and disabled the arm. It was restored only after establishing the head was unmoved (verdict 09:58:35Z, last commit 09:05:11Z, **0 head-moving events since**) — which is exactly the test `verdict_is_still_for_this_tree` applies, and it still strips on a genuine dequeue.

## The signal: zero check-SUITES, and nothing weaker

A `paths`-filtered workflow that matches nothing still creates a check-suite. Zero **suites** therefore means no workflow was ever *considered* — the event never reached the dispatcher. Weakening this to "zero check-runs" would fire on legitimately-pending and legitimately-empty groups, so the predicate is `total_count == 0` on `GET /repos/{repo}/commits/{group_head_sha}/check-suites` and nothing else.

## The grace period, derived rather than guessed

Two candidate anchors are **wrong**, and both were rejected by measurement rather than taste:

* `entry.enqueuedAt` — an entry can sit outside the `max_entries_to_build` window for many minutes before its group is built.
* the group head commit's `committer.date` — GitHub **stamps** the group commit rather than dating it at build time. **The load-bearing evidence is the DISTRIBUTION**, not any single commit: over 40 live group heads, `branch_creation − committer.date` spans 2 s to 3929 s (median 16 s) and only **1 of 40** falls within ±2 s of the true creation time; an independent re-derivation on N=37 found **0/37** within ±2 s. *(An earlier revision of this description argued from a parent-commit contradiction and labelled it "Proof". That argument is **withdrawn** — the parent's own `committer.date` is a stamp too, and the claim could not be independently re-derived. The distribution stands on its own and does not need it.)*

The exact anchor is the repository **activity API** — `activity?activity_type=branch_creation&ref=<full ref>` — whose rows carry a real `timestamp` plus `after` = the created head SHA. It is also the only enumeration that can see a dead ref at all: it has no workflow runs, so the runs API cannot find it.

**Measured create→first-suite latency, N=209 refs, 2026-07-25 → 07-28:**

```
min 1 s | p50 2 s | p90 3 s | p99 4 s | MAX 4 s     (whole-second quantisation)
```

There is **no tail between 4 s and 3600 s**. The failure is categorical, not slow: a group gets its suites within 1–4 s or never. **Grace = 120 s — 30× the measured maximum, 60× the median**, and strictly below the 300 s cron interval so it never costs an extra tick. Detection is tick-quantised, so effective recovery is **2–7 min against the 60-minute timeout**.

### Why the grace must be tight — the cost is what stacks on top, not the dead ref

Early dequeue does **not** rescue the CI stacked behind a dead ref. Measured on 2026-07-28: dequeuing #4709 at 06:07 rebuilt the chain onto fresh refs and #4714 and #4729 — both `MERGEABLE`, i.e. already green — had their group CI **discarded and restarted**. That CI was doomed either way, since the 60-minute `CI_TIMEOUT` rebuilds the same chain. The direct saving from acting at 48 min rather than 60 was ~8 minutes plus a preserved verdict.

The real cost is **how much work gets built on top of the dead ref before anyone notices**, because every later speculative group is chained onto it and a rebuild discards all of it:

```
cost ≈ (groups stacked above the dead ref at detection) × (their accumulated CI)
```

At 48 minutes that term was **4**, two of them fully green. Inside a 120 s grace it is **0** — nothing has stacked yet. So a tight grace is not merely faster; it is the difference between discarding four groups' CI and discarding none. That is the whole argument for the grace period, and it is why the census row now records `stacked=` and `stacked_green=` on **every** row: those numbers are the watchdog's own effectiveness measure, and if `stacked` starts climbing, the grace has drifted too long and the rows say so without anyone having to notice an outage by hand.

## Fix 1 — the watchdog (`scripts/merge-group-watchdog.py` + `merge-group-watchdog.yml`)

Schedule-only, every 5 min (the scheduler minimum). Never runs on a PR head, so it can never gate.

| verdict | when |
|---|---|
| `SKIP` | entry state is not `AWAITING_CHECKS`, or no speculative group is built yet |
| `HOLD` | ≥1 check-suite present — dispatch happened; **the control** |
| `WAIT` | zero suites, still inside the grace |
| `REFUSE` | suite count or creation time **could not be positively established** |
| `NOOP` | recovery already attempted for this exact group head |
| `CAP` | per-PR or per-run budget exhausted |
| `RECOVER` | zero suites, past grace, positively established, within budget |

**Fail safe.** Only `RECOVER` mutates. An unreadable suite count, a garbled response, a missing `branch_creation` row, an unreadable marker history, or a future-dated ref are each a refusal — never a recovery. If the reconstructed ref name were wrong, no activity row would match the head SHA and the result is `REFUSE`, so a bad reconstruction can only cause a missed detection, never a false one.

**Cap: 1 per distinct group head SHA (idempotence) + 2 per PR per rolling 6 h + 2 per run** (blast radius).

**Exhaustion behaviour, corrected.** An earlier revision of this PR claimed in four places — here, the constant's comment, the operator-facing `CAP` census row, and AGENTS.md — that exhaustion preserves the verdict. **That was wrong in all four.** Recovery is *marker → dequeue → enqueue*, so the marker is always ~30 s **older** than the `added_to_merge_queue` it produces; on the next attempt the newest trusted marker predates that attempt, the queue-attempt binding refuses it, and the platform `CI_TIMEOUT` **demotes to `review:changes`** like any other.

That is the *correct* outcome, not merely a tolerated one: by exhaustion the PR has burned three consecutive zero-dispatch groups, the only trusted observation names a **superseded** group head, and preserving a verdict on the strength of a stale head's suite count would be unsound. What exhaustion does **not** do is stall — no `needs:user`, no permanent hold, and a `CAP` row every tick it holds. All four sites now say this, and a test pins the behaviour so the claim cannot drift away from the code again.

This is the same false-reason class I caught in round 6 (*"it would tell an operator a commit landed when none did"*) — fixed in one place, recurred in four others.

**The budget is spent in queue order** — only the head-of-line entry actually blocks merges, so recovering positions 3 and 2 while position 1 stays broken would burn the budget and leave the outage in place. Pinned by a test.

**Remediation call**, with the three traps that each cost real time:

* `gh pr merge <N> --disable-auto` does **not** dequeue an already-queued PR — it answers "is already queued to merge".
* `DequeuePullRequestInput` names its field `id`, not `pullRequestId`, and it wants the **pull request** node id (`PR_…`), not the merge-queue **entry** id (`MQE_…`) — the natural wrong answer given the field name. A test asserts both mutations carry `PR_…`.
* `MergeQueueEntry` has no `headOid`; the group head is `headCommit{oid}`.

## Fix 2 — the `CI_TIMEOUT` routing split (`merge-queue-feedback.yml`)

`CI_TIMEOUT` currently routes to `review:changes` by design, so #4534 lost `review:pass` and needed a full re-review **because of a platform event drop** — a machine-manufactured re-review of code nobody found fault with.

The two causes are now split **by the suite count, never by the reason alone**. `classify-dequeue` re-derives the **live** count for the recorded group head and prints `route=preserve|demote`:

* `CI_TIMEOUT` + 0 suites → preserve `review:pass`, re-arm, comment.
* `CI_TIMEOUT` + suites present → today's demote. A timeout that followed checks that genuinely ran is a legitimate signal.

The marker is a **pointer, not a cached verdict**: the group head SHA is unrecoverable once the ref is deleted (that is exactly the population with no runs to read it off), so the watchdog records the SHA in a marker comment before acting — and the routing decision re-queries the live count for it. The commit stays readable after the branch is deleted; verified on `1bfb0174`, still `total_count: 0`.

The classifier also recognises the watchdog's **own** dequeue — recovery is dequeue+enqueue, which itself fires `pull_request.dequeued`, and without that arm the watchdog would demote every PR it rescued.

**Fail-safe in every direction:** a missing marker, a stale observation, one predating the current queue attempt, an unreadable count, a non-zero count, or a crashed classifier all yield `demote`. The step guard is written `route != 'preserve'`, so an empty or missing output demotes too.

## Fix 3 — an infrastructure dequeue must not strip the review verdict

Found by measurement after the first two fixes were written, and it is what stops this PR being a net regression.

```
05:14:56  labeled   review:pass       by sparq-orchestrator[bot]
06:06:38  removed_from_merge_queue    (reason MANUAL)
06:06:55  labeled   review:changes    by github-actions[bot]
06:06:55  unlabeled review:pass       by github-actions[bot]
```

**Seventeen seconds.** Because the watchdog's remediation *is* a dequeue, without this fix every rescue would strip `review:pass`, apply `review:changes`, disable the arm and admit the PR to the **fix** pipeline — manufacturing a `review:changes` backlog out of a platform event that never fired.

The existing routing is correct for its original purpose: a reviewer withdrawing a PR *should* pull the verdict. The defect is that **`MANUAL` conflates "a reviewer withdrew this" with "infrastructure moved it"**. Same shape as the `CI_TIMEOUT` half, same remedy — **discriminate on evidence, never on the event name**: a fresh watchdog marker naming that exact group head is what makes a dequeue ours.

Chosen route: **the discriminated marker**, not restore-after-re-enqueue. The marker is written *before* the dequeue and the handler skips both the label swap and the arm-disable. Restore-after-the-fact is strictly weaker — a repair rather than a prevention, with a window in which the PR is visibly mislabelled and can be picked up by the fix pipeline.

**Non-negotiable precondition on every preserve route.** A verdict attests to a specific **tree**, so `review:pass` survives only if the head has not moved since it was granted — a commit, a force-push, or a revoked label after the grant all forfeit it. Two details that only a surviving mutant exposed:

* the **latest** move must be compared against the **latest** grant. Earliest-move lets a stale verdict survive a push; earliest-grant wrongly forfeits a verdict legitimately **re-granted** after one (the normal push → re-review → re-approve sequence).
* a `committed` timeline event has **no `created_at`** — its time is the commit's own `committer.date`, so reading only `created_at` makes commits invisible as head moves.

**The control:** a genuine human dequeue still strips the verdict, because it carries no fresh marker. Without that arm this would be a general-purpose way for any dequeue to keep its approval — far worse than the bug.

## Fix 4 — the marker is evidence, so its author is part of the predicate

**This was a real vulnerability in the first revision of this PR, and the control I named in this very description was the forgeable artifact.**

`pr_markers` accepted a marker from **any** comment author, and sparq is a **public** repo. Executed through the real entry point with the comment authored by an arbitrary account:

* `MANUAL` + a forged marker → `route=preserve`, so `review:pass` survives and the arm-disable is skipped — and `auto-arm.py` / `rearm-sweeper.py` then re-arm on `review:pass` + CLEAN within ~10 minutes.
* `CI_TIMEOUT` + a forged marker naming **any** commit that reads zero check-suites → `preserve` + `reenqueue=true`, which reaches **`gh pr merge --auto`**. Such a sha is trivially obtainable.

Worse, arm (a) **never compared the marker head to anything**, so "names that exact group head" was not what the code implemented.

Two fixes, both at the read so nothing downstream can forget them:

1. **`TRUSTED_MARKER_AUTHORS`** — markers are honoured only from `github-actions[bot]` and `sparq-orchestrator[bot]` (both, because the sweep runs under the App token when configured and the workflow token otherwise; trusting only one silently breaks the caps the moment that flips). A well-formed marker from anyone else is **dropped and logged**, so a forgery attempt is visible rather than silently ignored.
2. **Arm (a) now re-derives the suite count for the head the marker names**, and requires `action=re-enqueue`. A head that does not exist reads `None`; one that really was dispatched reads non-zero. Neither can carry a verdict through a dequeue.

**Paired control, as required:** an App-authored marker still works — both trusted identities are asserted to preserve, alongside the forged-marker cases that must demote. Also pinned: a forged marker cannot *suppress* a real recovery either (the idempotence key is read from the same channel), and untrusted markers do not consume the cap.

This is the **third occurrence in this estate of the same root cause** — after #681 (review evidence read from `pull["body"]`) and #937 (declarations derived from body/title). In all three the forgeable channel was the PR's *named* control, and in all three the mutation battery was scoped to the object and never to the **evidence channel**. Round 7 of this sweep is that battery: 24 mutants aimed squarely at the marker's provenance.

### CORRECTION — the 567 markers on #4534 were MY OWN test residue, not an attack

An earlier revision of this description reported the 567 marker comments on PR #4534 as *"the payload is already in the wild"*. **That was wrong, and it inverts.** Every one was authored by this agent's own account inside a 31-minute window (`07:56:04Z → 08:27:24Z`) that is exactly my mutation-sweep window. There is **no evidence of any attack**, and the claim is retracted.

The root cause is a defect in **my test suite**, and it is worth stating precisely because it is a general Python trap:

```python
def __init__(self, ..., gh: Callable = run_gh):   # BOUND AT DEFINITION TIME
```

A default argument captures the function object when `def` executes, so a test patching `merge_group_watchdog.run_gh` **cannot** reach it. `main()` passed `gh_read=` explicitly (late-bound, so the patch worked) but let `gh` fall through to that stale default — **the real runner**. Every simulated sweep in the suite therefore issued a genuine `gh pr comment`, ×175 mutants. The reads were faked; the writes were real.

Fixed three ways: the runner defaults are now **late-bound** (`None` → resolve at call time); the suite **poisons both module-level runners at import**, so any path that forgets to inject a fake raises instead of reaching the network; and four tests guard that guard. Fixing it also exposed a `finally` that restored only one of the two runners, leaking a stale fake into every later test.

**Residue removed: 572 comments, 0 failures.** #4534 is back to its 3 pre-existing comments; issue #1 is back to 0. Verified by re-querying every PR/issue my fixtures can name.

The tail of that cleanup is worth stating, because it is the same defect class one level up: my deletion filter keyed on the **marker**, and the `P37` mutant exists precisely to *remove* the marker — so the one comment my filter missed was the one produced by mutating away the thing the filter matched on. **A cleanup filter inherits the blind spot of whatever it is cleaning up.** Re-swept on the prose signature instead, which found and removed it.

**The blast radius was not only #4534.** I enumerated every PR number my fixtures can name and checked each via the list API rather than the (lagging) search index: `4535`, `4536`, `4709`, `4714`, `4729`, `4731`, `4743` were all clean — but **issue #1 had 4 marker comments**, because the `A06` mutant hard-codes `["pr", "comment", "1", ...]` and that literal reached the real runner. Deleted; #1 is back to 0. The mutant now names a non-numeric sentinel so it cannot address a real issue even if the poison were bypassed. *Checking the obvious target was not enough — I had to enumerate what the code could name.*

**The lesson, stated so it is checkable: when a probe reports hostile input, verify the AUTHOR before reporting an exploit.** I found the outcome I was looking for and read it as confirmation; the convenient result deserved the most scrutiny and got the least. A false attack report is costly in both directions — it triggers unwarranted alarm, and it discounts the real thing if it ever arrives. And: **never exercise a write path against a live production PR.** Use a scratch PR or a fixture; if a test needs a real zero-suite head, *read* it.

### The genuine insight that came out of it: a defensive control must not amplify what it rejects

The first cut emitted **one annotation per rejected marker** — against 567 markers that is ~110 KB of log and 567 annotations, past GitHub's annotation cap, burying every other signal in the run.

> **A filter that emits per-item output hands the attacker a log-flood denial of service against its own operator.** The cheaper the input is to produce, the louder the defence gets. State the count and the distinct authors; never iterate the payload into the log.

Now one summary annotation per PR: **110 KB → 368 bytes** on the same input, 7 s runtime against a 5-minute job timeout.

## Fix 5 — an anonymous DoS on merge-queue routing, and a fail-safe that wasn't

Four blocking defects from independent review, all reachable on a **public** repo by a workflow that runs on **every dequeue**.

**(a) Quadratic parse with the trust check in the wrong position.** `[^>]*?` backtracks: one 65,536-char comment of repeated marker openers measured **12,406 ms** here (the reviewer measured 21.4 s on their box), and the author filter ran *after* the parse, so every anonymous comment was parsed before being discarded. Anyone could have pushed the classify step past its 5-minute timeout, suppressing merge-queue routing.

Both halves fixed, both measured:

| | before | after |
|---|---|---|
| `parse_marker` on a 64 KB adversarial body | 12,406 ms | **137 ms** (90×) |
| `pr_markers`, 20 anonymous 64 KB comments, end to end | — | **32.6 ms** (9,207× headroom vs the 300 s timeout) |

The bound (`[^>]{0,400}?`) makes the parse linear — a real marker's key/value run is 262 chars. But **the ordering is the actual lesson**: I *had* a trust check and it was in the wrong position. *A filter placed after the expensive work does not protect the expensive work.* Untrusted bodies are now never parsed at all, which is why the end-to-end number is what it is.

**(b) The fail-safe `if:` was false in one of its two named cases.** A step `if:` with **no status function defaults to `success()`**, so a *failed or timed-out* classify step **skipped** the demote step — `review:pass` kept, arm not disabled. The empty-output case was handled; the failed-step case was not, and a string-matching test could not see the difference. Now `!cancelled() && …`, with the preserve step deliberately left on the implicit `success()` (it may only act on a verdict that was actually produced) and a test pinning both directions. Composed with (a), this was anonymous suppression of merge-queue routing.

**(c) Three holes in the trust model.** A marker merely **quoted** by a trusted bot was honoured — and a real trusted-echo channel exists. Fenced, indented, blockquoted and inline contexts are now stripped before matching, and nesting **depth** (not a substring test) rejects `<!-- quoted: <!-- … -->`; inspecting the match itself tells you nothing, because it lands on the *inner* opener. A marker naming a **foreign head** or a **different PR** was honoured — markers now bind to this PR through both `pr=` and the `ref=` segment, which must agree. And arm (b) ignored `action`, which arm (a) already checked.

**(d) `{"total_count": false}` read as a positively-observed zero**, because `isinstance(False, int)` is True in Python — driving a recovery, a marker write and a `preserve` off a malformed response. Bools are now rejected explicitly.

## Emission

Every entry emits one row every tick, naming the ref, head, suite count and decision; `CAP`/`REFUSE`/`NOOP` additionally emit `::warning`, a failed recovery emits `::error` and reds the run, and an empty queue emits an explicit row. Rows also go to the job summary. Live sample against production at 05:55Z:

```
pos=1 pr=#4709 … head=98204656 suites=0       stacked=2 stacked_green=1 decision=RECOVER — zero check-suites 2398s after the ref was built (grace 120s)
pos=2 pr=#4714 … head=9f915630 suites=unknown stacked=1 stacked_green=0 decision=SKIP    — entry state MERGEABLE is not AWAITING_CHECKS
pos=3 pr=#4729 … head=464e418b suites=8       stacked=0 stacked_green=0 decision=HOLD    — 8 check-suite(s) present
sweep complete: entries=3 HOLD=1 RECOVER=1 SKIP=1 errors=0
```

That single run is both red tests and the control, live: the dead ref detected, a healthy 8-suite sibling held, a non-actionable entry skipped.

## Tests — 174, three layers

1. **Policy** — the decision tables, every arm.
2. **Behaviour** — the real `sweep()` against a scripted, **SHA-keyed** `gh`, asserting on the **mutations issued**, not the log text.
3. **Wiring** — cross-file inspection of every `if:`, step, and call site in both workflows plus the docs-quality call site.

Headline pairs, each with its control:

* zero suites past grace ⇒ `RECOVER` — paired with **suites present but runs still pending ⇒ never touched**, asserted across suite counts 1/8/40 and ages 0 → 100× the grace. Without that control the detector could pass by firing on everything.
* `CI_TIMEOUT` + 0 suites ⇒ `preserve` — paired with **`CI_TIMEOUT` + suites present ⇒ still demotes**. Same reason, opposite outcome, purely because the count differs.
* second detection on the same ref ⇒ no-op, zero mutations issued.

## Mutation sweep — 181 mutants, earned over ten rounds

Kills extracted from unittest's `FAIL:`/`ERROR:` blocks and the **traceback frames** inside them — never by grepping for a substring like `AssertionError`, which matches inside a passing test's *name*.

| round | mutants | survivors | what they exposed |
|---|---|---|---|
| 1 — delete each guard | 68 | 4 | 3 real gaps (`main()`'s fail-safe, `write_outputs`, the activity-type filter) + 1 equivalent |
| 2 — adversarial | +18 | 9 | wrong-id dequeue, marker-read failure assumed empty, failed recovery counted as success, four unpinned preserve-step behaviours |
| 3 — subtle | +12 | 4 | **a fake that answered every `/check-suites` URL identically** could not tell the group head from its base — two wrong-SHA mutants survived until the fake became SHA-keyed |
| 4 — conditionally inert | +30 | 2 | `if: … route != 'preserve' && false` passed a **substring** assertion while making the fix lane dead |
| 5 — verdict survival | +18 | 4 | earliest-vs-latest on both the head move and the grant, and a test **passing for the wrong reason** (see below) |
| 6 — revocation reporting | +8 | 3 | policy tests build `VerdictState` by hand and so **bypass the timeline parser** — a mutant that reported a revocation as a phantom push kept the right route with a false reason |
| 7 — the **evidence channel** | +25 | 5 | the marker's provenance, the exit contract, and the telemetry plumbing — the three areas the coverage census flagged |
| 8 — the readback guard | +3 | 1 | the readback matched *any* marker, not one for **this** head |
| 9 — the hermetic-suite guard | +3 | 0 | reintroducing the exact 567-comment defect is now caught |

**Total: 181 mutants, 181 killed, 0 survived, 0 skipped** — preflight clean (181 unique ids, every anchor present **and unique**), 94 distinct killing tests.

Round 7 exists because rounds 1–6 mutated the *object* and never the *evidence channel* — the same blind spot as #681 and #937. Round 9 exists because the fix for my own network-write defect deserved the same treatment: `N01` reverts the runner default to the definition-time binding that caused it, and `test_an_uninjected_watchdog_cannot_reach_the_network` kills it.

**Per-mutant kill causes were read, not assumed.** 171 died on a test-file assertion; **7 died with a frame inside the script** — P01/P04/I01/I03/I17/V05/P23, all guards whose removal makes the code *raise* rather than return a wrong value. Each is killed by the test that specifically targets it (`test_missing_branch_creation_anchor_refuses` for P04, `test_missing_enqueue_event_demotes` for P23, the human-dequeue control for I17), so they are load-bearing guards — but they are crash-kills, not assertion-kills, and are reported as such.

### The `gh` PATH shim, and why the in-process guard cannot be the only one

Every mutant now runs behind a `gh` shim on `PATH` that records and blocks any real invocation; the sweep asserts the escape log is **empty**, and the shim was validated against a known positive (`gh api user` → blocked + recorded) *before* being trusted.

It is not redundant with the in-process poison. **Reproducing `N01` necessarily reverts the hermeticity fix — the guard and the mutant are the same switch** — so during that one mutant the in-process protection is off *by design*. The shim sits outside that switch. Escapes are attributed **per mutant**, because a flat "must be 0" is the wrong claim: `N01` exists precisely to revert hermeticity, so *its* escape is the reproduction working. Final run: **exactly one mutant attempted a real `gh` call — `N01`, 2 calls, expected, naming reserved ids only.** Every other mutant was silent. The fixture rename therefore held independently of the shim, and the shim covered the one case the in-process guard structurally cannot.

One correction to my own tripwire: it logged `$*` verbatim, so a multi-line comment body wrapped across many lines and reported **21** where the truth was **2**. *A tripwire whose own output overstates the breach is a bad tripwire* — one line per invocation now.

### Three survivors from the shimmed sweep, all real, all from ONE cause

The marker **hex-head** and **non-zero-suites** guards survived because every malformed fixture *also* lacked a valid `ref=`, so the newly-added pr/ref self-consistency check rejected it **first** and the guard under test was never reached. **A new guard can silently mask an older one's tests** — each now has a fixture valid in every other respect. The third: the preserve step's open-state check was pinned by *substring presence*, so a mutant that left the strings in place while changing the behaviour survived; now pinned by shape (the non-open branch must `exit 0`, before any mutating command).

### The measurement instrument itself failed toward "nothing to report"

Mid-sweep a greedy regex edit to the mutant list **silently deleted 15 entries (P11–P25)**, and the harness happily reported a clean `154/154`. **A dropped mutant is indistinguishable from a killed one in the headline** — this is the failure mode that would make a mutation table *itself* a fabrication, since the number looks better the more coverage goes missing.

The fix, which other agents should copy: the harness now runs a **preflight** that fails the whole sweep outright if any mutant has a duplicate id or an anchor that no longer exists in the tree. It has already earned its keep twice, catching `X04` after I rewrote the block it anchored on.

Two related disciplines, both from actually reading what killed each mutant rather than trusting the total:

* **Two "kills" were `SyntaxError` / `IndentationError`** — broken mutants, not detections. Repaired.
* **Report a test-file frame in preference to a crash frame.** The *first* failure is arbitrary, and a crash inside the script was masking an assertion that had also fired — which made 9 kills look suspect when most were fine.

Round 4 is the one worth reading. A guard that is still present but can never fire is invisible to a deletion-only sweep *and* to review. Both YAML guards are now pinned by **exact match**, plus a test that rejects any constant `true`/`false` operand in any step condition (stripping legitimate `!= true` comparisons first).

Round 4 and round 5 each produced one finding worth naming beyond the fix:

* **a conditionally inert guard.** A guard still present but unable to fire is invisible to a deletion-only sweep *and* to review. Both YAML guards are now pinned by **exact match**, plus a test rejecting any constant `true`/`false` operand in any step condition (stripping legitimate `!= true` comparisons first).
* **a test passing for the wrong reason.** The unreadable-timeline case set *both* `readable=False` and `review_pass_at=None`, so it was really exercising the "no verdict" guard and would have stayed green with the readable check deleted. Now isolated, and the two cases must report distinguishable reasons — an operator has to be able to tell an API failure from a PR that was never approved.

Round 6 is worth one more note, because it is a structural blind spot rather than a missed case: the policy tests construct `VerdictState` by hand, so **they never execute the timeline parser**. Three mutants lived entirely in that gap — a revocation folded into head-moves (right route, false reason), earliest-vs-latest revocation, and treating *any* `unlabeled` event as a verdict revocation, which would have made the preserve route unreachable in practice, silently and greenly, because PRs shed `area:*`/`status:*`/`needs:*` labels constantly. Every one is now covered end-to-end through the parser.

One mutant was excused rather than fixed, explicitly: `I18` was mis-constructed (`not (dry_run and True)` ≡ `not dry_run`, a no-op), so I rewrote it to `not False` and it died. Everything else was a real gap and was closed. The closing total was earned by eight successive adversarial expansions, not asserted on the first list — every round after the first found real gaps.

## Coverage census — run FIRST, and corrected

**One instrument throughout: statement coverage.** An earlier revision mixed a branch instrument with a statement one, listed `main` among the zero-coverage functions (it was not), and used the wrong denominator.

| | statements | functions with statements | at 0% |
|---|---|---|---|
| module `--self-test` only | 53.7% | 31 | **16** |
| full suite | 86.9% | 31 | 5 |
| **union of what CI runs** | **98.1%** | 31 | **1** |

The 16 are `Watchdog.__init__`, `_graphql`, `check_suite_count`, `classify_dequeue`, `emit`, `group_created_at`, `post_marker_comment`, `pr_markers`, `queue_entries`, `recover`, `sweep`, `verdict_state`, `_ndjson`, `run_gh`, `run_gh_read`, `write_outputs` — **`main` is not among them.**

The one function still at 0% in the union is **`run_gh`**, the real `subprocess` wrapper. That is correct by construction and will stay: the suite now *poisons* it precisely so it can never execute, so a hermetic suite cannot cover it. Reporting it as 0 rather than rounding it away.

Every blocking survivor lived in a function the module self-test never executed — `main`'s sweep branch held the exit-zero mutant. Closed from that census: the **exit-zero** mutant, `stacked_green = list(stacked)` (value-identical against the old all-`MERGEABLE` fixture — now mixed), `QUEUE_ENTRY_PAGE`, the dropped `RECOVER_FAILED` key, the dead `marker suites` field, and **both `log=print` defaults**, which survived and would have let a silent watchdog stay green.

## Validated against live production data, not only fixtures

The hermetic suite proves the policy; these two reads prove the parser against the real API:

```
PR #4709  grant=06:36:41  revoked=06:06:55  head_moved=04:56:39  -> verdict SURVIVES
PR #4534  grant=22:58:41  revoked=00:05:20  head_moved=12:50:12  -> verdict FORFEIT (revoked)
```

`head_moved=04:56:39` for #4709 independently reproduces the timestamp used in the manual verification before that PR's verdict was restored by hand, and `revoked=00:05:20` for #4534 matches the "#4534 lost `review:pass` at 00:05:20" already recorded on #4652. Both routes are correct and both reasons name the real cause.

The detector was also run live twice against the production queue, 40 minutes apart, with opposite outcomes — one dead ref found among healthy siblings at 05:55Z, and an all-clear at 06:34Z once the queue recovered. That is the red case and the control, on production, not in a fixture.

## Recorded, not fixed here: the same defect class, and a repo mix-up worth flagging

**In this repo**, `scripts/gh_retry.py:251` — a module the watchdog imports — is:

```python
    run: Callable[..., subprocess.CompletedProcess] = subprocess.run,
```

the same **definition-time default-argument binding** that made this suite write to production. Verified on `origin/main`, not just this worktree. It is *materially safer* than my defect was: it is a deliberate injection seam and callers pass `run=` explicitly, so it is not a live escape route. The latent trap is that monkeypatching `subprocess.run` at module level would be **silently ineffective**, which is exactly how my version hid.

**A correction on the correction.** A suite-wide escape audit reported that this claim "does not reproduce — line 251 is a list comprehension, no `Callable` in the file". That is accurate — **for a different file**. There are two `gh_retry.py` files on this box:

| repo | line 251 | `Callable` count |
|---|---|---|
| **sparq** (imported by this watchdog) | `run: Callable[...] = subprocess.run` | 4 |
| **`jeswr/agent-account-registry`** | `if not _TRACE_LINE_RE.match(line)` | 0 |

The audit measured the registry's copy. Its finding there is real and correctly located — `pat-validity.py:227`, `def _secret_write(token, repo, run=subprocess.run)`, a `gh secret set` **write** — and it belongs to that repo's tracking (#989/#992), not to this PR.

Neither is mine to fix; both are recorded so the next person is sent to the right module. **The methodological note is the transferable part, and it is what found the registry instance:** an AST sweep for this class must handle **`ast.Attribute`** defaults, not only `ast.Name`, or it misses `subprocess.run` entirely.

## Not done, deliberately

`max_entries_to_build`, `max_entries_to_merge` and `check_response_timeout_minutes` are **maintainer-set** and untouched. #4652 records the ceiling as `D/T = 3 / 22.7 min ≈ 7.9 PR/hr` against 2.30 measured — batching is working at 29% of ceiling — so raising it is a maintainer decision, not mine. No attempt is made to explain *why* Actions dropped the event; that is not determinable from outside. This recovers from it.

### DIFF
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -882,6 +882,20 @@ jobs:
       # only, no git/network.
       - name: Self-test check-install-action-tool.py (fixtures + live invariant — sq-1cb6l)
         run: python3 scripts/tests/test_install_action_tool.py
+      # [OPUS-5] sparq-org/sparq#4652: the zero-dispatch merge-group watchdog
+      # (scripts/merge-group-watchdog.py, .github/workflows/merge-group-watchdog.yml)
+      # plus the CI_TIMEOUT routing split it feeds in merge-queue-feedback.yml. Both
+      # are queue-critical and neither can be exercised by a PR run, so their whole
+      # contract is pinned here: the hermetic decision tables (script --self-test), and
+      # the three-layer suite — policy, sweep BEHAVIOUR against a scripted gh (assert on
+      # the mutations issued, not the log text), and cross-file YAML-seam inspection of
+      # every `if:`, step and call site. The fail-closed direction matters most: the
+      # demote guard is `route != 'preserve'`, so a classifier failure keeps the
+      # pre-#4652 behaviour; inverting it to `== 'demote'` fails here.
+      - name: Self-test merge-group-watchdog.py (decision tables — #4652)
+        run: python3 scripts/merge-group-watchdog.py --self-test
+      - name: Zero-dispatch watchdog suite (policy + behaviour + YAML seam — #4652)
+        run: python3 scripts/tests/test_merge_group_watchdog.py
       # [SONNET-4.6] sq-qcnn.32 / [OPUS-5] #3773: the advisory-job registry + gate-intent
       # placement check. Since #3773 the registry is LOAD-BEARING (it is the ONLY thing
       # that makes a check-run non-gating), so this is its integrity gate. THREE
--- a/.github/workflows/merge-group-watchdog.yml
+++ b/.github/workflows/merge-group-watchdog.yml
@@ -21,9 +21,11 @@
 # below the tick interval so it never costs an extra tick. Minutes are offset from
 # auto-arm.yml's and rearm-sweeper.yml's crons so the three sweeps do not start together.
 #
-# BASE RATE, measured rather than assumed: 2 zero-dispatch refs in 211 merge groups over
-# 2026-07-25 -> 07-28 (~0.95%) — #4534 and #4331 — each costing ~60 min at the head of
-# the queue. This is a recurring platform event, not a one-off.
+# BASE RATE, measured rather than assumed: THREE zero-dispatch refs on three
+# consecutive days — #4331 (07-26), #4534 (07-27), #4709 (07-28) — against 211 merge
+# groups sampled over the same period. Each held position 1 for ~60 min. #4709 was
+# caught live by this script's own `sweep --dry-run` while its four siblings in the
+# same chain all had 8 suites. This is a recurring platform event, not a one-off.
 name: merge-group watchdog
 
 on:
--- a/.github/workflows/merge-group-watchdog.yml
+++ b/.github/workflows/merge-group-watchdog.yml
@@ -1,0 +1,94 @@
+# [OPUS-5] sparq-org/sparq#4652 — recover a merge-group ref GitHub Actions never
+# dispatched, in minutes instead of the ruleset's 60-minute CI_TIMEOUT.
+#
+# WHY: measured 2026-07-27/28, PR #4534's merge-group ref was created at 23:04:37Z with
+# ZERO check-suites and deleted at 00:05:04Z when the queue timed it out. It held
+# position 1 in AWAITING_CHECKS the whole time — the queue merges strictly in order and
+# each entry needs its OWN group's required check, so three PRs with already-green
+# groups could not land. Total cost: 73 min 41 s of zero merges and ~66 min of group CI
+# thrown away on the rebuild. The full diagnosis is on #4652.
+#
+# NOT A GATE. Schedule / workflow_dispatch only — it never runs on a PR head, never
+# produces a check-run the `ci-summary` aggregator can see, and can therefore never
+# block a merge. Its only mutations are (a) a marker comment and (b) a dequeue +
+# re-enqueue of an entry it has POSITIVELY established has zero check-suites. Every
+# ambiguous reading is a refusal, not an action (see the script header).
+#
+# Cadence is every 5 minutes — the GitHub scheduler minimum — against a 300 s grace, so
+# worst-case detection is ~10 min against the 60-min timeout it replaces. The grace is
+# ~100x the MEASURED create->first-suite latency on this repo (single-digit seconds; see
+# the script's DEFAULT_GRACE_SECONDS note). Minutes are offset from auto-arm.yml's and
+# rearm-sweeper.yml's crons so the three sweeps do not start together.
+name: merge-group watchdog
+
+on:
+  schedule:
+    - cron: "2,7,12,17,22,27,32,37,42,47,52,57 * * * *"
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Emit decisions without dequeuing/re-enqueuing anything"
+        type: boolean
+        default: false
+
+# Least privilege, mirroring rearm-sweeper.yml (the sibling queue sweeper):
+#   contents: read       — the repository ACTIVITY api (branch_creation timestamps),
+#                          which is the only exact merge-group build-time anchor.
+#   checks: read         — the load-bearing predicate: check-SUITE count on the group head.
+#   pull-requests: write — the marker comment + the dequeue/enqueue mutations.
+permissions:
+  contents: read
+  checks: read
+  pull-requests: write
+
+concurrency:
+  group: merge-group-watchdog
+  cancel-in-progress: false
+
+jobs:
+  watch:
+    name: recover zero-dispatch merge groups
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    env:
+      # Secrets are not readable from step `if:` expressions — mirror the id into env
+      # so the mint step can be skipped fail-soft when the App is not configured.
+      ORCHESTRATOR_APP_ID: ${{ secrets.ORCHESTRATOR_APP_ID }}
+    steps:
+      - name: Mint the sparq-orchestrator App token (re-enqueue must fire queue events)
+        id: app-token
+        if: env.ORCHESTRATOR_APP_ID != ''
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.ORCHESTRATOR_APP_ID }}
+          private-key: ${{ secrets.ORCHESTRATOR_APP_PRIVATE_KEY }}
+
+      # Always execute the trusted default-branch policy, including on manual dispatch.
+      - name: Checkout watchdog policy
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          persist-credentials: false
+          # gh_retry.py is the watchdog's bounded transient-retry dependency — the
+          # self-test below imports it, so a missing path fails BEFORE the live sweep.
+          sparse-checkout: |
+            scripts/merge-group-watchdog.py
+            scripts/gh_retry.py
+          sparse-checkout-cone-mode: false
+
+      - name: Self-test policy
+        run: |
+          python3 scripts/gh_retry.py --self-test
+          python3 scripts/merge-group-watchdog.py --self-test
+
+      - name: Sweep the merge queue for zero-dispatch groups
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token || github.token }}
+          REPO: ${{ github.repository }}
+          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
+          DRY_RUN: ${{ inputs.dry_run && '--dry-run' || '' }}
+        run: >-
+          python3 scripts/merge-group-watchdog.py sweep
+          --repo "$REPO"
+          --branch "$DEFAULT_BRANCH"
+          $DRY_RUN
--- a/.github/workflows/merge-group-watchdog.yml
+++ b/.github/workflows/merge-group-watchdog.yml
@@ -14,11 +14,16 @@
 # re-enqueue of an entry it has POSITIVELY established has zero check-suites. Every
 # ambiguous reading is a refusal, not an action (see the script header).
 #
-# Cadence is every 5 minutes — the GitHub scheduler minimum — against a 300 s grace, so
-# worst-case detection is ~10 min against the 60-min timeout it replaces. The grace is
-# ~100x the MEASURED create->first-suite latency on this repo (single-digit seconds; see
-# the script's DEFAULT_GRACE_SECONDS note). Minutes are offset from auto-arm.yml's and
-# rearm-sweeper.yml's crons so the three sweeps do not start together.
+# Cadence is every 5 minutes — the GitHub scheduler minimum — against a 120 s grace, so
+# worst-case detection is ~10 min and typical is ~5, against the 60-minute ruleset
+# timeout it replaces. The grace is 30x the MEASURED maximum create->first-suite latency
+# on this repo (N=209, max 4 s; see the script's DEFAULT_GRACE_SECONDS note) and sits
+# below the tick interval so it never costs an extra tick. Minutes are offset from
+# auto-arm.yml's and rearm-sweeper.yml's crons so the three sweeps do not start together.
+#
+# BASE RATE, measured rather than assumed: 2 zero-dispatch refs in 211 merge groups over
+# 2026-07-25 -> 07-28 (~0.95%) — #4534 and #4331 — each costing ~60 min at the head of
+# the queue. This is a recurring platform event, not a one-off.
 name: merge-group watchdog
 
 on:
--- a/.github/workflows/merge-queue-feedback.yml
+++ b/.github/workflows/merge-queue-feedback.yml
@@ -31,14 +31,42 @@
 #   workflow is ever extended to build/test PR code, it MUST move that work to a
 #   separate `pull_request`-triggered job — do not add a checkout of the PR head here.
 #
+#   [OPUS-5] #4652 adds ONE checkout — pinned to `ref: <default_branch>` with
+#   `persist-credentials: false` — solely to obtain the trusted, default-branch copy of
+#   `scripts/merge-group-watchdog.py`. That is the base repo's own reviewed code, NOT
+#   the PR head, so the rule above is intact: this workflow still never checks out or
+#   executes anything a PR contributed. Do not change that `ref:`.
+#
 # GATE INTEGRATION: the job name carries the whole-word `advisory` token, so the
 # ci-summary gate's discovery excludes it from the gating set (a dequeue-feedback
 # hiccup on a PR head must never red a later gate evaluation on that same SHA). The
 # labels/comments it writes are exactly the fix lane's normal inputs.
 #
-# FLAKE-CLASSIFIED failures: an infra-flavoured dequeue reason (e.g. CI_TIMEOUT) is
-# noted in the comment but STILL routes to review:changes — the fix lane's ci-flavour
-# owns re-runs, so the routing is identical by design.
+# FLAKE-CLASSIFIED failures: an infra-flavoured dequeue reason (e.g. QUEUE_CLEARED,
+# ROLL_BACK) is noted in the comment but STILL routes to review:changes — the fix lane's
+# ci-flavour owns re-runs, so the routing is identical by design.
+#
+# ── [OPUS-5] #4652: the ONE exception — a CI_TIMEOUT with ZERO check-suites ────────
+# A `CI_TIMEOUT` has two utterly different causes and until now both got the same
+# treatment. (1) Checks genuinely ran long / hung — a legitimate signal, keep demoting.
+# (2) GitHub Actions NEVER DISPATCHED the `merge_group` event, so the group ref sat with
+# zero check-suites until the ruleset's 60-minute timeout fired. Measured on #4534
+# (2026-07-27): ref built 23:04:37Z, 0 check-suites / 0 check-runs / 0 workflow runs,
+# dequeued CI_TIMEOUT at 00:05:03Z. Demoting THAT strips `review:pass` and manufactures
+# a full re-review of code nobody found fault with — the same defect class as the
+# machine-manufactured `needs:user` escalations this repo has already fixed twice.
+#
+# The two are separated by the SUITE COUNT, never by the reason alone:
+# `scripts/merge-group-watchdog.py classify-dequeue` re-derives the live check-suite
+# count for the recorded group head and prints `route=preserve|demote`. FAIL-SAFE in
+# every direction — a missing marker, an unreadable count, a non-zero count, a stale
+# observation, or a crashed classifier all yield `demote`, i.e. today's behaviour. The
+# `if:` on the routing step below is written as `route != 'preserve'` so an EMPTY or
+# MISSING output (the classifier step failed, the output was lost) still demotes.
+#
+# The same classifier also recognises the watchdog's OWN recovery: the watchdog
+# re-enqueues via dequeue+enqueue, which itself fires `pull_request.dequeued`, and
+# without that arm the watchdog would demote every PR it just rescued.
 #
 # No third-party actions (nothing to SHA-pin); every step is plain `gh` against the
 # workflow token, least-privilege scoped below.
@@ -48,11 +76,15 @@ on:
   pull_request_target:
     types: [dequeued]
 
-# Least-privilege for a WRITE-only, NO-checkout job: pull-requests: write covers the
-# disable-auto / relabel / comment writes; actions: read lists this repo's merge_group
-# runs to find the failed run. NO contents write is needed — nothing is pushed and no
-# repo code is checked out (see the pull_request_target safety note above).
+# Least-privilege: pull-requests: write covers the disable-auto / relabel / comment
+# writes; actions: read lists this repo's merge_group runs to find the failed run;
+# [OPUS-5] #4652 contents: read is the DEFAULT-BRANCH checkout of the classifier script
+# (read-only, never the PR head) and checks: read is the classifier's load-bearing
+# predicate — the check-SUITE count on the group head. NO contents WRITE is needed —
+# nothing is pushed (see the pull_request_target safety note above).
 permissions:
+  contents: read
+  checks: read
   pull-requests: write
   actions: read
 
@@ -83,8 +115,78 @@ jobs:
         run: |
           echo "PR ${PR_NUMBER} dequeued via successful merge - no action."
 
-      - name: Route genuine dequeue to the fix lane
+      # [OPUS-5] #4652. TRUSTED, DEFAULT-BRANCH checkout of the classifier only — never
+      # the PR head (see the pull_request_target safety note in the header). No PR code
+      # is fetched, and nothing from the PR is executed.
+      - name: Checkout the dequeue classifier (default branch, never the PR head)
         if: github.event.pull_request.merged != true
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          persist-credentials: false
+          sparse-checkout: |
+            scripts/merge-group-watchdog.py
+            scripts/gh_retry.py
+          sparse-checkout-cone-mode: false
+
+      # [OPUS-5] #4652. Splits CI_TIMEOUT by the SUITE COUNT of the group head, not by
+      # the reason. Always exits 0 and always prints a row; every unprovable case
+      # yields route=demote (today's behaviour), so a failure here cannot preserve a
+      # verdict it did not establish.
+      - name: Classify the dequeue (zero-dispatch vs a real timeout)
+        id: classify
+        if: github.event.pull_request.merged != true
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+        run: >-
+          python3 scripts/merge-group-watchdog.py classify-dequeue
+          --repo "$REPO"
+          --pr "$PR_NUMBER"
+          --reason "$DEQUEUE_REASON"
+
+      # The zero-dispatch route. It deliberately does NOT touch labels — preserving the
+      # review verdict IS leaving `review:pass` where it is — and does NOT disable the
+      # arm. `rearm-sweeper.yml` is the durable backstop if the direct re-arm below
+      # fails, because a PR carrying review:pass with no live arm is exactly what that
+      # sweep restores.
+      - name: Preserve the verdict and re-enqueue (zero-dispatch CI_TIMEOUT)
+        if: github.event.pull_request.merged != true && steps.classify.outputs.route == 'preserve'
+        env:
+          CLASSIFY_DETAIL: ${{ steps.classify.outputs.detail }}
+          CLASSIFY_REENQUEUE: ${{ steps.classify.outputs.reenqueue }}
+        run: |
+          set -euo pipefail
+
+          STATE=$(gh pr view "$PR_NUMBER" --repo "$REPO" --json state -q .state)
+          if [ "$STATE" != "OPEN" ]; then
+            echo "PR #${PR_NUMBER} is ${STATE} — nothing to preserve; exiting 0."
+            exit 0
+          fi
+
+          if [ "$CLASSIFY_REENQUEUE" = "true" ]; then
+            gh pr merge "$PR_NUMBER" --repo "$REPO" --auto \
+              || echo "::warning::re-arm of PR #${PR_NUMBER} returned non-zero — rearm-sweeper.yml is the backstop."
+          else
+            echo "Re-enqueue owned by the watchdog itself — not re-arming here."
+          fi
+
+          body_file=$(mktemp)
+          {
+            printf '%s\n\n' "> 🤖 SPARQ agent — merge-queue feedback."
+            printf '%s\n\n' "This PR was dequeued (reason \`${DEQUEUE_REASON}\`), but the merge-group ref it was waiting on had **zero check-suites** — GitHub Actions never dispatched the \`merge_group\` event for it, so no required check could ever report (sparq-org/sparq#4652)."
+            printf '%s\n\n' "**Nothing is wrong with this PR's diff, so its review verdict is preserved** and it is re-queued rather than demoted to \`review:changes\`."
+            printf '%s\n' "Classifier: \`${CLASSIFY_DETAIL}\`"
+          } > "$body_file"
+          gh pr comment "$PR_NUMBER" --repo "$REPO" --body-file "$body_file" \
+            || echo "::warning::could not comment the zero-dispatch notification on PR #${PR_NUMBER} (routing is intact)."
+          echo "PR #${PR_NUMBER}: zero-dispatch ${DEQUEUE_REASON} — verdict preserved."
+
+      # FAIL-SAFE `if:` — written as `!= 'preserve'` so an EMPTY or MISSING classifier
+      # output (step failed, output lost) still routes to the fix lane exactly as it did
+      # before #4652. Never invert this to `== 'demote'`.
+      - name: Route genuine dequeue to the fix lane
+        if: github.event.pull_request.merged != true && steps.classify.outputs.route != 'preserve'
         run: |
           set -euo pipefail
 
--- a/.github/workflows/merge-queue-feedback.yml
+++ b/.github/workflows/merge-queue-feedback.yml
@@ -161,6 +161,9 @@ jobs:
       # fails, because a PR carrying review:pass with no live arm is exactly what that
       # sweep restores.
       - name: Preserve the verdict and re-enqueue (zero-dispatch CI_TIMEOUT)
+        # Deliberately NOT `!cancelled()`: preserving a verdict requires the classifier
+        # to have SUCCEEDED and positively returned `preserve`. The implicit `success()`
+        # is the correct default here — do not harmonise this with the step below.
         if: github.event.pull_request.merged != true && steps.classify.outputs.route == 'preserve'
         env:
           CLASSIFY_DETAIL: ${{ steps.classify.outputs.detail }}
@@ -196,7 +199,12 @@ jobs:
       # output (step failed, output lost) still routes to the fix lane exactly as it did
       # before #4652. Never invert this to `== 'demote'`.
       - name: Route genuine dequeue to the fix lane
-        if: github.event.pull_request.merged != true && steps.classify.outputs.route != 'preserve'
+        # `!cancelled() &&` is LOAD-BEARING. A step `if:` with no status function
+        # defaults to `success()`, so a FAILED or TIMED-OUT classify step would SKIP
+        # this one — keeping `review:pass` and leaving the arm enabled, i.e. the exact
+        # opposite of fail-closed. The empty-output case was handled; the failed-step
+        # case was not, and a string-matching test could not see the difference.
+        if: "!cancelled() && github.event.pull_request.merged != true && steps.classify.outputs.route != 'preserve'"
         run: |
           set -euo pipefail
 
--- a/.github/workflows/merge-queue-feedback.yml
+++ b/.github/workflows/merge-queue-feedback.yml
@@ -66,7 +66,17 @@
 #
 # The same classifier also recognises the watchdog's OWN recovery: the watchdog
 # re-enqueues via dequeue+enqueue, which itself fires `pull_request.dequeued`, and
-# without that arm the watchdog would demote every PR it just rescued.
+# without that arm the watchdog would demote every PR it just rescued. MEASURED on
+# #4709: `review:pass` was swapped for `review:changes` 17 SECONDS after a `MANUAL`
+# dequeue. `MANUAL` conflates "a reviewer withdrew this" (verdict SHOULD fall) with
+# "infrastructure moved it" (it should not), so the split is on EVIDENCE — a fresh
+# watchdog marker for that exact group head — never on the reason string.
+#
+# AND, on every preserve route: the verdict may only survive if the PR head has NOT
+# moved since `review:pass` was granted. A commit, a force-push, or a revoked label
+# after the grant forfeits it — never restore a verdict onto a tree it was not given
+# for. A genuine human dequeue has no fresh marker and still demotes; that control is
+# what stops this from becoming a verdict-preservation hole.
 #
 # No third-party actions (nothing to SHA-pin); every step is plain `gh` against the
 # workflow token, least-privilege scoped below.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -520,7 +520,13 @@ no network); the Python close-script carries a `--self-test`.
   a superseded group head, so preserving the verdict would be unsound, and three consecutive
   zero-dispatch groups on one PR warrant a human look. It never escalates to a `needs:user` hold
   and never permanently stalls. **Everything it cannot positively establish is a
-  refusal, never a recovery.** It emits one row per entry **every tick**, carrying the ref, the
+  refusal, never a recovery.** The marker it writes is **evidence**, so its AUTHOR is part of
+  the predicate: markers are honoured only from `github-actions[bot]` / `sparq-orchestrator[bot]`
+  (`TRUSTED_MARKER_AUTHORS`, enforced at the read), because sparq is public and a marker forged
+  by any commenter would otherwise carry `review:pass` through a dequeue and reach
+  `gh pr merge --auto`. The recovery also reads its own marker back before dequeuing and refuses
+  to act if it is not there — a runner whose identity is missing from that list cannot see its
+  own markers, which would silently disarm both caps. It emits one row per entry **every tick**, carrying the ref, the
   suite count, the decision, and `stacked=`/`stacked_green=` — the count of groups built on top
   of the dead ref, which is the real cost term and the watchdog's own effectiveness measure.
   Companion routing split in `merge-queue-feedback.yml`: a `CI_TIMEOUT` whose group ref had
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -504,6 +504,35 @@ plan (§5, Beads D/E + H), to be added only after the scripts are proven in manu
 shell script is `bash -n`/`shellcheck` clean and carries a `--dry-run-self-test` (hermetic;
 no network); the Python close-script carries a `--self-test`.
 
+- **`.github/workflows/merge-group-watchdog.yml` + `scripts/merge-group-watchdog.py`**
+  (#4652) — the **durable** zero-dispatch merge-group recovery. GitHub occasionally builds a
+  merge-group ref and then never dispatches the `merge_group` event for it: **zero
+  check-suites, zero check-runs, zero workflow runs**. Because the queue merges strictly in
+  order and each entry needs its **own** group's required check (a green *superset* group does
+  **not** substitute), that entry holds position 1 until the ruleset's 60-minute
+  `check_response_timeout_minutes` reaps it, and the rebuild discards every group stacked on
+  top. Observed three days running (#4331, #4534, #4709), ~60 min each. The sweep runs on a
+  5-minute cron (never on a PR head, so it can never gate), and after a **120 s** grace —
+  30× the measured maximum create→first-suite latency (N=209: p50 2 s, max 4 s; the failure is
+  categorical, not slow) — it dequeues and re-enqueues the entry. Bounded (1 per group head
+  SHA, 2 per PR per 6 h, 2 per run) and idempotent; on exhaustion it hands back to the platform
+  timeout rather than escalating to a human. **Everything it cannot positively establish is a
+  refusal, never a recovery.** It emits one row per entry **every tick**, carrying the ref, the
+  suite count, the decision, and `stacked=`/`stacked_green=` — the count of groups built on top
+  of the dead ref, which is the real cost term and the watchdog's own effectiveness measure.
+  Companion routing split in `merge-queue-feedback.yml`: a `CI_TIMEOUT` whose group ref had
+  **zero suites** preserves `review:pass` instead of manufacturing a re-review for a platform
+  event drop, while a timeout that followed checks that genuinely ran keeps demoting — the two
+  are separated by the **suite count**, never by the reason alone.
+  **Runbook — resolving and clearing a dead ref by hand.** The group head is the queue entry's
+  `headCommit{oid}` (`MergeQueueEntry` has **no** `headOid` field), or read the live refs with
+  `git ls-remote origin 'refs/heads/gh-readonly-queue/main/*'`; the trailing sha in a ref name
+  is the group's **BASE**, and the ref is named for one *member*, so **absence of a ref named
+  for a PR is not evidence its CI never started**. Test `commits/<head>/check-suites` —
+  `total_count == 0` is the signal, and never `check-runs` (a `paths`-filtered workflow that
+  matches nothing still creates a suite). To clear it, `gh pr merge --disable-auto` does **not**
+  dequeue an already-queued PR; use `dequeuePullRequest`, whose input field is named `id` but
+  wants the **pull request** node id (`PR_…`), not the entry id (`MQE_…`).
 - **`.github/workflows/bead-autoclose.yml` + `scripts/ci-close-merged-beads.py`**
   (sq-84a8; issue-native since #2475) — the **durable** auto-close-on-merge. On a merged PR
   (`pull_request_target: [closed]` gated on `merged == true` — base-repo context so the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -515,8 +515,11 @@ no network); the Python close-script carries a `--self-test`.
   5-minute cron (never on a PR head, so it can never gate), and after a **120 s** grace —
   30× the measured maximum create→first-suite latency (N=209: p50 2 s, max 4 s; the failure is
   categorical, not slow) — it dequeues and re-enqueues the entry. Bounded (1 per group head
-  SHA, 2 per PR per 6 h, 2 per run) and idempotent; on exhaustion it hands back to the platform
-  timeout rather than escalating to a human. **Everything it cannot positively establish is a
+  SHA, 2 per PR per 6 h, 2 per run) and idempotent. On exhaustion it hands back to the platform
+  timeout, which **does demote to `review:changes`** — by then the only trusted observation names
+  a superseded group head, so preserving the verdict would be unsound, and three consecutive
+  zero-dispatch groups on one PR warrant a human look. It never escalates to a `needs:user` hold
+  and never permanently stalls. **Everything it cannot positively establish is a
   refusal, never a recovery.** It emits one row per entry **every tick**, carrying the ref, the
   suite count, the decision, and `stacked=`/`stacked_green=` — the count of groups built on top
   of the dead ref, which is the real cost term and the watchdog's own effectiveness measure.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -504,6 +504,29 @@ plan (§5, Beads D/E + H), to be added only after the scripts are proven in manu
 shell script is `bash -n`/`shellcheck` clean and carries a `--dry-run-self-test` (hermetic;
 no network); the Python close-script carries a `--self-test`.
 
+### Hermetic suites must be UNABLE to reach the network — poison the runners <!-- [OPUS-5] #4652 -->
+
+A "hermetic" test suite that merely *intends* to inject fakes is not hermetic. `def __init__(self, ..., gh=run_gh)`
+binds the runner **at definition time**, so patching `module.run_gh` from a test cannot reach it — and a suite
+whose reads were faked while its writes were real posted **567 real comments to a production PR** across a
+mutation sweep (#4652). Two rules follow:
+
+- **Late-bind injected collaborators** (`gh=None` → resolve inside `__init__`), never a module-level default arg.
+- **Poison the real runners at test-module import** so any path that forgets to inject raises instead of
+  reaching the network, and add a test asserting the poison is in place. Restore **every** patched name in
+  `finally` — a block that restored one of two leaked a stale fake into every later test.
+
+**Never exercise a write path against a live production PR/issue.** Use a scratch object you own, or a fixture;
+if a test needs a real remote value, *read* it. And when a probe reports hostile input, **verify the author
+before reporting an exploit** — a false attack report costs twice: unwarranted alarm now, and a discounted
+warning when the real thing arrives.
+
+**Mutation harnesses need a preflight.** A greedy edit silently deleted 15 mutants and the harness reported a
+clean total — a dropped mutant is indistinguishable from a killed one, so the number improves as coverage
+disappears. Fail the sweep outright on a duplicate id or an anchor that no longer exists in the tree, count a
+`SyntaxError` as a broken mutant rather than a kill, and prefer a test-file traceback frame over a crash frame
+(the first failure is arbitrary).
+
 - **`.github/workflows/merge-group-watchdog.yml` + `scripts/merge-group-watchdog.py`**
   (#4652) — the **durable** zero-dispatch merge-group recovery. GitHub occasionally builds a
   merge-group ref and then never dispatches the `merge_group` event for it: **zero
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -523,7 +523,15 @@ no network); the Python close-script carries a `--self-test`.
   Companion routing split in `merge-queue-feedback.yml`: a `CI_TIMEOUT` whose group ref had
   **zero suites** preserves `review:pass` instead of manufacturing a re-review for a platform
   event drop, while a timeout that followed checks that genuinely ran keeps demoting — the two
-  are separated by the **suite count**, never by the reason alone.
+  are separated by the **suite count**, never by the reason alone. The same split makes an
+  INFRASTRUCTURE dequeue verdict-neutral: measured on #4709, `review:pass` was swapped for
+  `review:changes` **17 seconds** after a `MANUAL` dequeue, so a watchdog without this would
+  burn the verdict of every PR it rescued. `MANUAL` conflates "a reviewer withdrew this" with
+  "infrastructure moved it", so the discriminator is a fresh watchdog marker for that exact
+  group head — evidence, never the event name — and **a genuine human dequeue still strips the
+  verdict**. On every preserve route the verdict may only survive if the head has **not moved
+  since it was granted** (a commit, a force-push, or a revoked label after the grant forfeits
+  it): never restore a verdict onto a tree it was not given for.
   **Runbook — resolving and clearing a dead ref by hand.** The group head is the queue entry's
   `headCommit{oid}` (`MergeQueueEntry` has **no** `headOid` field), or read the live refs with
   `git ls-remote origin 'refs/heads/gh-readonly-queue/main/*'`; the trailing sha in a ref name
--- a/scripts/merge-group-watchdog.py
+++ b/scripts/merge-group-watchdog.py
@@ -1,0 +1,1078 @@
+#!/usr/bin/env python3
+"""Recover a merge-queue entry whose merge-group ref was never dispatched by Actions."""
+
+# [OPUS-5] sparq-org/sparq#4652 — the zero-dispatch merge-group watchdog.
+#
+# THE OUTAGE THIS FIXES (measured 2026-07-27/28, #4652): PR #4534's merge-group ref
+# `gh-readonly-queue/main/pr-4534-3cc1bf828c…` was created at 23:04:37Z with head
+# `1bfb0174…` and deleted at 00:05:04Z — 60 min 27 s later — having produced
+# **0 check-suites, 0 check-runs, 0 workflow runs**. Eight workflows carry a bare
+# unfiltered `merge_group:` trigger and fired for every other group that day; for this
+# one ref none of them was ever CONSIDERED. #4534 sat at position 1 in AWAITING_CHECKS;
+# the queue merges strictly in order and each entry needs its OWN group's required
+# check, so #4400/#4636/#4639 sat behind it with green groups that could not land.
+# GitHub eventually dequeued #4534 with reason CI_TIMEOUT at 00:05:03Z (the ruleset's
+# `check_response_timeout_minutes: 60`), and the chain rebuilt 3 s later, discarding
+# three already-green group CI runs. Net: 73 min 41 s of zero merges (23:18:59Z →
+# 00:32:40Z) and ~66 min of wasted group CI.
+#
+# WHY ZERO check-SUITES IS THE LOAD-BEARING SIGNAL — and why "zero check-runs" is NOT.
+# A `paths`-filtered workflow that matches nothing STILL creates a check-suite (it just
+# reports no runs). So a group with real workflows but nothing to do reads as
+# suites>0 / runs==0. Zero *suites* means no workflow was ever considered for the ref,
+# i.e. the `merge_group` event never reached the Actions dispatcher. Weakening the
+# predicate to check-runs would fire on legitimately-pending and legitimately-empty
+# groups. The predicate here is `total_count == 0` on
+# `GET /repos/{repo}/commits/{group_head_sha}/check-suites`, and NOTHING else.
+#
+# ── the anchor problem, and the measurement that settles it ───────────────────────
+# The grace gate needs T_create = when the merge-group ref was BUILT. Two obvious
+# proxies are WRONG and were rejected by measurement:
+#   * `entry.enqueuedAt` — an entry can sit outside the `max_entries_to_build` window
+#     for many minutes before its group is built, so this anchor is far too early.
+#   * the group head commit's `committer.date` — GitHub STAMPS the group commit with
+#     the entry's `enqueuedAt`, not the build time. Measured: head `0e043a13`
+#     (PR #4636's group) is stamped 23:01:57Z == #4636's enqueuedAt, but that group's
+#     branch was created at 23:19-ish; head `1bfb0174` is stamped 22:58:53Z while its
+#     own PARENT `3cc1bf828` (#4632) did not reach `main` until 23:18:59Z — a commit
+#     cannot predate its parent, which proves the field is a stamp, not a creation
+#     time. Any latency computed from it is meaningless.
+# The CORRECT, exact anchor is the repository ACTIVITY API:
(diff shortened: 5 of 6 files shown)


jeswr:
> 🤖 **SPARQ agent** — I am @jeswr's agent for the sparq RDF/SPARQL engine. This is an independent **review** of this PR (I am not its author); @jeswr runs multiple agents, and this was written by the SPARQ agent, not the PSS agent.

## Review of `866811d4` (`866811d47827703de73a8b169c503c5915da908f`)

Reviewed the diff at head, then re-derived the load-bearing claims **by execution** rather than by reading: 26 fault-injection cases against the real `sweep()`, 75 mutants (line-targeted, md5-verified restores, `-B`/`PYTHONDONTWRITEBYTECODE=1`, kills attributed from `FAIL:`/`ERROR:` rows + traceback frames), a live re-measurement of the grace distribution over 38 production merge-group refs, and live read-only exercise of every I/O method against `sparq-org/sparq`.

**The detection half is sound and, on its own, I would pass it.** The verdict-preservation half has an executed hole and a claim the code does not implement, so: `fail`, with a small fix.

### ✅ What I confirmed, by execution

**The dequeue capability is genuinely fail-closed.** Driving the real `sweep()` with a scripted `gh`, an unreadable suite list can never read as zero — 13 malformed/unreadable `check-suites` responses (`GhError`/503, `GhTransientExhausted`, non-JSON, list payload, `null`, missing `total_count`, `total_count: "0"`, `total_count: null`, `total_count: 0` with a suite listed, missing/`null` `check_suites`, empty stdout), 12 healthy configurations (suites 1/8/40 × age 0/1/12/100 h → `HOLD` every time), and 6 anchor faults (activity 503, no row, wrong `activity_type`, head-sha mismatch, future-dated row, unreadable marker history) produced **zero mutations**; only `total_count == 0` with an empty page mutates. Live: `check_suite_count("0"*40)` (404) → `None`, never `0`. Seven further **non-crashing** fail-open mutants — coercing `None → 0` at the `except`, the `total_count` read, the `isinstance` guard, the truncated-page guard, the `sweep()` call site, plus `REFUSE → RECOVER` — all died on named assertions (`test_check_suite_count_returns_none_on_every_unreadable_shape`, `test_unknown_suite_count_refuses`, `test_unreadable_suite_count_issues_no_mutation`).

**The grace derivation holds, independently re-measured.** 38 production group refs (2026-07-27T22:34Z → 07-28T06:44Z), `branch_creation` → earliest suite `created_at`: **min 1 / p50 2 / p90 2 / p99 3 / max 3 s**, nothing in (4 s, 3600 s), no negatives. 120 s is 40× my measured max. Two zero-suite refs in the same window (the `#4534` and `#4709` heads) — the base rate is real.

**The `committer.date` rejection holds, and by a stronger argument than the counterexample.** Over 40 group heads, `branch_creation − committer.date` spans **2 s … 3929 s (p50 16 s), only 1/40 within ±2 s**; for `1bfb0174` the stamp is 344 s *before* the ref existed. It is categorically not a build time. (I did not re-derive the "parent reached `main` at 23:18:59Z" datum — `3cc1bf828`'s own `committer.date` is 22:52:22Z, itself a stamp by the same measurement — so lean on the distribution, not on "a commit cannot predate its parent".)

**Both live timelines re-derive exactly.** `#4709 grant=06:36:41 revoked=06:06:55 head_moved=04:56:39 → SURVIVES`; `#4534 revoked=00:05:20 → FORFEIT`. `group_created_at` returns 23:04:37Z / 05:15:09Z exactly, with the wrong-head and bogus-ref controls both `None`. `1bfb0174` still reads `total_count: 0`.

**Writes, and the gate question.** The watchdog performs exactly three writes — the marker comment, `dequeuePullRequest`, `enqueuePullRequest` (both carrying `PR_…`, pinned). It never labels, approves or merges; `schedule`/`workflow_dispatch` only, both checkouts pinned to the default branch (mutants adding a `pull_request:` trigger and swapping the ref for `pull_request.head.sha` both die). Empty-queue row verified **live**: `branch=main queue is EMPTY — 0 entries to watch (grace=120s)`, exit 0. `max_entries_to_build` / `max_entries_to_merge` / `check_response_timeout_minutes` appear only in prose — untouched. No `needs:user` anywhere, so no human-terminal exit. No scope creep: 2 700 of the 2 972 added lines are the script + its suite, and the `docs-quality.yml` (+14) and `AGENTS.md` (+37) additions are the CI wiring and the runbook for exactly this mechanism.

**Mutation sample: 75 mutants, 68 killed, 0 malformed** (no kill came from an exception at the mutated line). Every headline guard died on a *named* test, in both the deleted and the conditionally-inert form — e.g. `suites > 0 ⇒ HOLD` → `test_control_suites_present_runs_pending_is_never_touched` + `test_control_group_with_suites_issues_no_mutation_at_all`; the unmoved-head precondition → `test_a_commit_after_the_verdict_forfeits_it_end_to_end`; `revoked.append` → `moved.append` (right route, false reason) → `test_a_revoked_verdict_forfeits_it_end_to_end`; `if: … route != 'preserve' && false` → killed by **both** `test_the_demote_guard_is_fail_closed` (exact match) **and** `test_no_step_guard_carries_a_constant_disjunct`. The round-4/5/6 self-found issues are genuinely closed, and the end-to-end classify fixtures do enter through the real timeline parser (the parser mutants `min`-vs-`max` on grant/move/revocation, any-`unlabeled`-is-a-revocation, and the `committed`-has-no-`created_at` case all die there).

---

### ❌ Blocking: the marker is **unauthenticated, author-controlled evidence** — and this repo is public

`Watchdog.pr_markers` reads `repos/{repo}/issues/{pr}/comments` and accepts a marker from **any** comment author. There is no `user.login` / `author_association` filter anywhere in the script or the suite (`grep -n 'login\|author_association'` → no hits). `sparq-org/sparq` is `PUBLIC`, so any GitHub account can write one onto any PR.

Executed through the real entry point (`Watchdog.classify_dequeue` → `pr_markers` + `verdict_state`), with the marker comment authored by `drive-by-attacker`:

```
A) reason=MANUAL, forged marker (head=0*40, observed=now-60s)  -> route=preserve reenqueue=False
B) CONTROL: same MANUAL dequeue, no marker                     -> route=demote   reenqueue=False
C) reason=CI_TIMEOUT, forged marker naming a 0-suite commit    -> route=preserve reenqueue=True
D) CI_TIMEOUT, marker present, live count 8                    -> route=demote
```

(A) is the control this PR names as the thing that stops it being a verdict-preservation hole — *"a genuine human dequeue still strips the verdict, because it carries no fresh marker"*. The discriminator **is** the forgeable artifact. A reviewer who dequeues an approved-but-bad PR to stop it leaves the head unmoved, so the unmoved-head precondition does not help; `review:pass` survives, the arm-disable is skipped, and `auto-arm.py`/`rearm-sweeper.py` both re-arm on `review:pass` + CLEAN within ~10 min. The human's dequeue does not stick.

(C) is arm-capable: `route=preserve` + `reenqueue=true` makes the workflow run `gh pr merge --auto`. The 0-suite commit it needs is not a secret — `1bfb0174f5cc2da1ed9dfe7997b7ab089e7cab26` is published in this PR's own body and still reads `total_count: 0`.

Also: arm (a) never compares the marker head with anything, so *"a fresh watchdog marker **naming that exact group head** is what makes a dequeue ours"* is not what the `MANUAL` arm implements — any 40-hex string passes.

The fix looks small: filter markers to the identity that writes them (the App / `github-actions[bot]`, or `author_association ∈ {OWNER, MEMBER}`) at the `pr_markers` boundary, and pin it with a test that a marker from a non-privileged author does **not** flip a `MANUAL` dequeue to `preserve`. Note the sweep side reads the same unfiltered markers, where a forged marker can only cause `NOOP`/`CAP` — inaction, so that direction is safe.

### ❌ Second: CAP exhaustion does **not** hand back to a verdict-preserving route

Claimed in four places (PR body, the `DEFAULT_MAX_RECOVERIES_PER_PR` comment, **the operator-facing `CAP` row itself** — "handing back to the platform CI_TIMEOUT, whose routing preserves the review verdict" — and `AGENTS.md`). Constructed the thrice-dead case and ran it through the real parser:

```
grant 04:00 | enqueue 04:05 | marker A 04:10 -> enqueue 04:10:30 | marker B 04:20 -> enqueue 04:20:30
group C dead -> CAP (2 markers in the 6 h window, no marker written) -> platform CI_TIMEOUT 05:50

-> route=demote: "newest watchdog observation 04:20:00Z predates this queue attempt (enqueued 04:20:30Z)"
```

Because recovery is *marker → dequeue → enqueue*, the watchdog's own re-enqueue is always ~30 s **after** its marker, so `newest.observed < last_enqueued_at` demotes. The guard is correct (that marker describes a previous attempt), but the consequence is the opposite of the claim: the third dead group in 6 h loses `review:pass` and is admitted to the fix lane — a machine-manufactured re-review, which is the failure class this PR exists to remove. It is fail-**safe** (no preservation hole) and it does **not** permanently stall (no `needs:user`; the fix lane re-reviews and re-arms), so this is a wrong claim in operator-facing telemetry rather than a wrong action — the same false-reason class as the round-6 revocation-as-push finding. Either make the hand-back preserve deliberately, or correct all four texts and the `CAP` detail string.

### Minor

1. **The exit-code claim is unpinned.** `main()`: `return 1 if watchdog.sweep() else 0` → `return 0` **survives all 111 tests** (two mutant forms). `main()` is at 55 % line coverage and its entire `sweep` branch is unexecuted, so "`::error` + non-zero exit on failed recovery" is only pinned at `sweep()`'s return value. Given this repo's exit-zero history, worth one direct test.
2. **Telemetry not value-pinned.** `stacked_green = [o for o in stacked if o.state == "MERGEABLE"]` → `list(stacked)` survives: the only fixture has every stacked entry `MERGEABLE`, so it cannot discriminate — add one `AWAITING_CHECKS` neighbour. `QUEUE_ENTRY_PAGE = 20 → 1` and dropping the `RECOVER_FAILED` census key also survive.
3. **Coverage, under the module's own `--self-test`** (what the workflow step runs): 17/29 functions at 0 % — `Watchdog.__init__`, `_graphql`, `check_suite_count`, `classify_dequeue`, `emit`, `group_created_at`, `post_marker_comment`, `pr_markers`, `queue_entries`, `recover`, `sweep`, `verdict_state`, `_ndjson`, `main`, `run_gh`, `run_gh_read`, `write_outputs`. Under the full suite that drops to `run_gh` and `run_gh_read` (plus the self-test's own fixtures) — the two thin `gh` wrappers, which per the coverage-ratchet convention want one direct test each.
4. **`write_outputs` has no multiline delimiter.** `detail=` can carry `gh` stderr (`marker history unreadable ({error})`, `classification failed ({error})`); a newline there appends further `key=value` lines *after* `route=`. I could not construct an attacker-reachable path, and a malformed output file fails the step → empty `route` → demote, so it is fail-safe — but a heredoc delimiter costs nothing.

I did not arm, merge, relabel or push anything, and I did not touch the author's worktree.

VERDICT: fail

jeswr:

**Independent review of `4ff55b0b8b3785591b5ecfe8f6a51228907fca98`** (15 commits, +3787/−8, 6 files). Reviewer ≠ author; no commits of mine are on this head. Everything below was **executed**, not reasoned about — through the real entry point `main("classify-dequeue", …)` / `Watchdog.sweep()`, with `subprocess.run` poisoned so a real `gh` call is a hard failure (**0 real subprocess invocations across every probe**).

This is a genuinely valuable fix for a thrice-confirmed outage, and the round-1 marker-forgery hole really is closed. But two defects at the seam block it, both introduced by this PR, both reachable anonymously on a public repo, and both fixable in a few lines.

---

## 1. BLOCKING — `parse_marker` is quadratic, and the trust check runs *after* it

`pr_markers` (script L858-863) calls `parse_marker(row["body"])` on **every** comment and only then checks `TRUSTED_MARKER_AUTHORS`. So the round-1 allow-list is no defence against the cost of the parse. `_MARKER_RE = <!--\s*merge-group-watchdog\s+(?P<kv>[^>]*?)\s*-->` backtracks quadratically on a body of repeated `<!-- merge-group-watchdog ` with no `-->`:

```
body=  2600 chars      19.0 ms
body= 13000 chars     578.5 ms
body= 26000 chars    2724.8 ms
body= 65536 chars   21451.0 ms      <- GitHub's comment-body limit
pr_markers() over 3 UNTRUSTED poisoned comments: 61.56 s, returned 0 trusted markers
```

End-to-end through the real entry point, comments authored by `anonymous-drive-by`:

```
main("classify-dequeue", --pr 4743, --reason CI_FAILURE), 4 x 65536-char comments -> 63.0 s
  15.8 s/comment  =>  19 comments exceed merge-queue-feedback.yml's `timeout-minutes: 5`
```

**Anyone with a GitHub account can post 19 comments on any sparq PR.** `merge-queue-feedback.yml` runs `classify-dequeue` on *every* `pull_request_target: [dequeued]`, so this is not gated on the PR being interesting. Combined with finding 2 the effect is: the classify step is killed by the job timeout → **the demote step is skipped** → `review:pass` is retained, the auto-merge arm is *not* disabled, the PR never enters the fix lane, and `rearm-sweeper` re-queues it → queue thrash. The watchdog's own job (`timeout-minutes: 10`, and it parses markers **twice** per recovery because of the readback) dies the same way whenever the poisoned PR is the one with a dead ref — i.e. exactly the case this PR exists to fix.

No merge bypass: the required `ci-summary / gate` is still the backstop. The harm is anonymous suppression of merge-queue routing plus queue-capacity burn.

This is the round-7 principle — *"a defensive control must not amplify what it rejects; the cheaper the input is to produce, the louder the defence gets"* — on the axis that was not measured. It got quieter and *slower*.

Two fixes, both measured here:

```
current  [^>]*?           17492.2 ms on the 65536-char poison
bounded  [^>]{0,400}?       243.0 ms   (still matches the real marker; its kv is 262 chars)
author-filter BEFORE parse:   0.0 ms   over 19 poisoned comments
```

Do both: hoist the `author in TRUSTED_MARKER_AUTHORS` test above `parse_marker`, and bound the `kv` quantifier (plus a `len(body)` cap, since a *trusted* bot can also carry a long body — see finding 3).

## 2. BLOCKING — the fail-safe `if:` is false in one of its two named cases

`merge-queue-feedback.yml` L195-197 and the PR body both state:

> written as `!= 'preserve'` so an EMPTY or MISSING classifier output **(step failed, output lost)** still routes to the fix lane exactly as it did before #4652.

A step `if:` containing no status-check function has **`success()` applied by default**. The demote step's guard is
`github.event.pull_request.merged != true && steps.classify.outputs.route != 'preserve'` — no status function. So if the classify step *or the new checkout step* fails or times out, the demote step is **skipped**, not run. The pre-#4652 behaviour is not preserved; it is lost.

This is a *new* exposure: before this PR "Route genuine dequeue to the fix lane" was the first conditional step and nothing that could fail preceded it. Now two steps do.

`test_the_demote_guard_is_fail_closed` pins the guard string by exact match and cannot see this — no test in the 154 mentions a failed step, `always()`, or `continue-on-error`. The repo already uses `always()` / `!cancelled()` in **39** places, so the idiom is known here; neither touched workflow uses it.

Fix: `if: always() && github.event.pull_request.merged != true && steps.classify.outputs.route != 'preserve'` (or `!cancelled() &&`), and add a wiring test asserting the demote guard carries a status function while the preserve guard does not.

---

## Material, not blocking

**3. No quoted-context stripping, and the marker `head`/`pr` are never bound to the PR.** All executed:

| attack (author = `github-actions[bot]`) | result |
|---|---|
| marker inside a ``` fenced block | **preserve + reenqueue** |
| marker inside a `>` blockquote | **preserve + reenqueue** |
| marker nested inside another `<!-- … -->` | **preserve + reenqueue** |
| marker naming a **foreign** zero-suite head (`bbbb…`) | **preserve + reenqueue** |
| marker whose `pr=` names a **different PR** | **preserve + reenqueue** |
| marker with `action` absent / `action=merge-now` | MANUAL demotes (arm a checks it); **CI_TIMEOUT preserves** (arm b does not) |

So AGENTS.md's and the workflow header's *"the discriminator is a fresh watchdog marker for **that exact group head**"* is still not what the code implements — arm (a) now re-derives the *count* for the named head, but neither arm compares its *identity* to anything. Round 1's lesson was that the PR's own **named** control was the forgeable channel; the named control is now the *unimplemented* one.

Is there a trusted echo? Yes, in production, though not anonymously: `sparq-orchestrator[bot]` — a `TRUSTED_MARKER_AUTHORS` identity — authors **633 of the 800 most recent repo comments**, and those bodies verbatim-quote spans of the reviewed PR's own files. The registry admission gate (`^sparq-agent/issue-N-` head ref + worker-App author) keeps fork PRs out of that lane, so this needs a prompt injection inside a worker PR's diff rather than a drive-by. It closes with the same two lines: strip fenced/quoted contexts before matching, and require `marker.pr == pr_number`.

**4. One unreadable suite payload *does* read as a positive zero.** `isinstance(False, int)` is `True`, so the `total`-type guard admits booleans:

```
{"total_count": false, "check_suites": []}  -> check_suite_count() returns False
                                            -> sweep(): decision=RECOVER, marker comment written
                                            -> classify: `own != 0` is False -> preserve
```

13 of the other 15 malformed shapes I threw at it refuse correctly (empty, whitespace, `null`, list, number, truncated object, missing key either side, string total, null total, dict/null `check_suites`, HTML error page, and the truncated `total=0 with a non-empty array`), as do a `GhError` 503 mid-walk and a `JSONDecodeError`. `total_count: -1` is safe only incidentally (`suites == 0` is what gates the anchor read, so it lands on REFUSE). The suite's own list (`test_check_suite_count_returns_none_on_every_unreadable_shape`) is 6 shapes and omits `bool`. One-word fix: `type(total) is not int`.

**5. The readback guard *can* be the disarm.** The marker is written **before** the readback, and the marker is the idempotence key. Executed: tick 1 with a 502 on the readback → `RECOVERY FAILED`, `::error`, red run — but the marker is already on the PR; tick 2 → `decision=NOOP`, forever. One transient at that exact point forfeits the PR's headline benefit and hands the ref back to the 60-minute timeout. It is loud, and the CI_TIMEOUT arm then preserves the verdict, so it is a degradation rather than a hole — but "write the key, then check whether you can read it" is the wrong order.

**6. Honesty drift, four items.**
- The **body still presents the withdrawn parent-commit argument as "Proof"** for rejecting the `committer.date` anchor, and never cites the distribution the *code comment* correctly relies on. I re-derived it: `1bfb0174` committer.date `22:58:53Z`; its parent `3cc1bf828` committer.date **`22:52:22Z`** — *earlier*, no contradiction. `23:18:59Z` is when that parent **merged into main**, not when it was created. The code comment's withdrawal is right; the body was not updated with it.
- Script L870 still calls the author's own test residue **"96 forged markers"** that this PR "already carries" — residue of the retracted attack framing, `96` conflicts with the `567` used for the *same* ~110 KB measurement in the header and body, and #4534 currently carries **1** marker comment (author `jeswr`), so the present tense is stale too. (The PR *body*'s retraction itself is clean — no "already in the wild" residue remains there.)
- The coverage census overstates. Measured with `coverage.py`, instrument validated against a known-0% canary first (canary lines reported missed; 505 statements on a 1399-line file): **(A) self-test 53% stmt / 51% branch, 16 functions at 0% — the 16 is exactly right**; (B) suite alone 86%; **(C) union 98% stmt, but 1 function at 0%, not 0** — `run_gh` (L667-671), which the suite's own poison test proves is never entered. `main` is named among the 16 but is 49% under the self-test (the self-test enters *through* `main`). Denominator is 29 `def`s, not 30. And `51% → 98%` pairs a *branch* figure with a *statement* figure; same-instrument it is 53→98 or 51→97.
- Test fixtures still name the live production PR **#4534** as the `pr_number` write target (`entry()`, `_entry()`), so the import-time poison is the only thing between this suite and a repeat of the 567-comment incident — against this PR's own new AGENTS.md rule, *"use a scratch object you own, or a fixture"*. A sentinel number costs nothing. (No hard-coded low-integer resource ids remain anywhere in the fixtures — I checked; issue **#1** is back to 0 comments.)

---

## What I verified and it held

- **All four author rows re-derived exactly** through `main("classify-dequeue")`: `drive-by-attacker` and `github-actions` (no `[bot]`) → demote on both reasons; `github-actions[bot]` and `sparq-orchestrator[bot]` → preserve on MANUAL, preserve+reenqueue on CI_TIMEOUT.
- Login imposters all demote: `GitHub-Actions[bot]`, trailing space, `…[bot]x`, `sparq-orchestrator` (no `[bot]`), a zero-width-suffixed login, `user: null`, and a comment with no `user` key.
- A forged **newer** marker does not displace a trusted older one; untrusted markers neither consume the cap nor suppress a recovery; two trusted markers → newest wins, and a newest naming an 8-suite head demotes. Future-dated markers demote on both arms.
- **Every write enumerated.** The script issues exactly three, all on the RECOVER path: `gh pr comment`, `dequeuePullRequest`, `enqueuePullRequest`. `classify-dequeue` issues **zero** writes (3 reads). Nothing in the script arms, approves, labels, or merges; the only new arm is the workflow's `gh pr merge --auto`, gated on `route == 'preserve' && reenqueue == 'true'`, which already requires a live, unrevoked `review:pass` whose grant post-dates the last head move.
- **Grace period independently re-derived** from the live activity feed, N=37 merge-group refs: create→first-suite `min 1 s / p50 2 s / p90 2 s / **max 3 s**`, and **nothing in (4 s, 3600 s]**. The `committer.date` offset distribution is `min 4 s / p50 16 s / max 3929 s`, **0/37 within ±2 s** — the anchor rejection is sound *on the distribution*. Both incident refs surfaced in the same page as zero-suite (`98204656`/#4709, `1bfb0174`/#4534), and the activity anchor reproduces exactly: `branch_creation 23:04:37Z, after=1bfb0174`. 120 s is 40× my measured maximum.
- Can a PR influence its own dequeue? Only one way I found: a diff that removed the `merge_group:` trigger from all 9 workflows would make its own group read zero suites and route itself into RECOVER, then into preserve at CI_TIMEOUT. It still cannot merge — `gate` never reports — so this is thrash, not a bypass. Triggers are schedule/dispatch-only and both checkouts are pinned to `default_branch` with `persist-credentials: false`.
- **Empty-queue census row emits live** (the queue is empty right now): `branch=main queue is EMPTY — 0 entries to watch (grace=120s)`, rc=0, zero mutations in `--dry-run`.
- All four **CAP** sites agree that exhaustion **demotes** (body, `DEFAULT_MAX_RECOVERIES_PER_PR`, `decide_entry`'s CAP detail, AGENTS.md), and both stated principles are in the artifact (script header + AGENTS.md), not only the report.
- **Scope is clean.** 6 files, all coherent. `max_entries_to_build`, `max_entries_to_merge`, `check_response_timeout_minutes` appear only in prose and are untouched. The suite runs in `docs-quality quick-gates`, which is gating (no `advisory` token).
- **Composition.** Merges cleanly onto live `main` `b9d412df`; on that merged tree `--self-test` PASSes and the suite is `Ran 154 tests … OK`. Gate by equality on `commits/4ff55b0b…/check-runs`: 62 runs, **every one `head_sha=4ff55b0b`**, 28 success / 34 skipped / **0 failures**; `gate` success, concluded `08:58:06Z`. The only runs started after it are two `conventional-commits (advisory)`, both success — the gate is not superseded.

**7. Three surviving mutants, all re-derived here with controls.** A second, independent sweep (71 mutants; every headline guard mutated twice, delete **and** conditionally inert) killed 67. Every guard named in the PR died, including both copies of the two structurally duplicated ones (`.lower() == REVIEW_PASS` at L915/L920, `live_suites(newest.head)` at L589/L640) — **no mutual masking**, and `TRUSTED_MARKER_AUTHORS` has exactly one enforcement site as claimed. Three survived:

- **`last_enqueued_at = max(enqueued)` → `min` (L933) survives with `Ran 154 tests … OK`.** Its three siblings are the control, and they all die: `granted`→`min` (2 failures), `moved`→`min` (1), `revoked`→`min` (1). The author mutated those three reductions and not the fourth. Root cause is measured, and it is the value-identical-to-a-fixture trap: **all 11 timeline fixtures carry exactly one `added_to_merge_queue`**, so `max` and `min` cannot differ on any input the suite provides. It is not a dead field — hard-wiring it to `None` kills 5 tests. *Scenario*: every PR the watchdog rescues gets ≥2 enqueue events, because recovery **is** dequeue+enqueue. Under `min`, `last_enqueued_at` is attempt 1's stamp, the queue-attempt binding at L626 stops binding, and at cap exhaustion the newest trusted marker — which names a **superseded** group head that genuinely reads 0 suites — routes to `preserve` + `reenqueue=true` and re-arms. That is precisely the outcome L177-190 says must not happen: *"preserving a verdict on the strength of a stale head's suite count would be unsound."*
- **Both `log: Callable[[str], None] = print` defaults (L717, L1151) survive** — replaced with a no-op sink, `Ran 154 tests … OK` twice. Emission is well tested, but only through an *injected* log, and `main()` never injects one. *Scenario*: the watchdog recovers correctly while emitting zero rows — no per-entry row, no census row, no `::warning` for CAP/REFUSE/NOOP, no empty-queue row, no job-summary lines — with CI green. That negates L66-70's stated requirement that a silent watchdog "would convert a visible 73-minute outage into an invisible one." Two independent copies of the same hole.

**8. The hermeticity fix holds, but its defence is two assertions wide — and it wrote to production during this review.** `N01` (both defaults reverted) dies loudly: `FAILED (failures=45, errors=5)` in 70.9 s versus `Ran 154 tests … OK` in 3.3 s. But the narrower and more faithful variant — revert **only** `gh`, leave `gh_read` late-bound, i.e. the exact "reads faked, writes real" defect — is killed by exactly two tests, `test_an_uninjected_watchdog_cannot_reach_the_network` and `test_the_runner_default_is_late_bound`. Neutralise only those two and the remaining 152 run green **while performing a real production write**. Worse, `test_sweep_exit_code_is_nonzero_when_a_recovery_fails` **passes for the wrong reason** under that variant: `_NoNetwork` patches the module-level runners, which the definition-time default cannot see, so `harness.gh.fail_on = ("dequeuePullRequest",)` never fires — the *real* dequeue fails against the fixture node id `PR_4534` and produces the nonzero exit the test asserts, having already issued a real `gh pr comment 4534`. This is not hypothetical: re-deriving `N01` during this review posted **two real comments to #4534** (`5101…`-era ids at 09:25:11Z and 09:25:38Z); both were deleted immediately and #4534 has no post-09:00Z comments and no queue-mutation timeline events. The fixtures naming a live PR is what turns a two-assertion gap into a production write.

**9. CORRECTION to my own finding on default bindings — I was wrong.** My first AST sweep matched only `ast.Name` defaults and therefore reported "no other definition-time binding of a network-capable callable". Re-run handling `ast.Attribute`, there is one, and it is live: **`scripts/gh_retry.py:251`, `def run_gh_read(..., run: Callable[...] = subprocess.run, sleep=time.sleep)`** — the identical defect class, in the module `merge-group-watchdog.py` imports. Patching `subprocess.run` and `gh_retry.subprocess.run` after import intercepts **zero** calls; the invocation goes straight to the real `gh`. It is unreachable from today's suite only because `mgw.run_gh_read` is poisoned and the two direct-call tests stub `gh_retry.run_gh_read` itself — nothing structural stops a future test from calling it and hitting production. `gh_retry.py` is outside this diff, but this PR's own new AGENTS.md rule is the one it violates; poison `gh_retry` at import too.

**10. CORRECTION — the preflight *is* verifiable, and it works.** My earlier statement that it could not be checked was wrong; my search excluded the directory it was in. Verified by execution on a copy (the author's worktree was never modified): control → 178 mutants, 178 unique ids, all anchors present, `preflight()` returns 0; a **duplicated id** → `DUPLICATE IDS: ['P01']`, `PREFLIGHT FAILED`, exit **3**, no mutant executed; an **absent anchor** → `ANCHOR NOT FOUND`, `PREFLIGHT FAILED`, exit **3**, no mutant executed. The `178/178, 0 survivors` headline re-derives independently under a hermetic `gh` shim, and the 7 script-frame mutants are exactly `P01, P04, I01, I03, I17, V05, P23` as stated (an 8th crash-kill exists outside their list, so 178 is not exhaustive — as the survivors above show). The `87 distinct killing tests` figure is not re-derivable from the harness output, which prints only the alphabetically-first killing test per mutant. Two residual gaps: the preflight checks anchor **presence, not uniqueness**, and the edit is `str.replace(old, new, 1)` — 2 of 178 anchors are non-unique (`I28` occurs 3× in `merge-queue-feedback.yml`, `V11` twice in the script), so `I28` may be mutating a different step than intended. And the harness is still **not in the diff**: the strongest artifact this PR produced is the one it does not ship, in a PR that itself records a harness silently dropping 15 mutants while reporting a clean total.


---

## To flip this to a pass

1. Author-filter before `parse_marker`, bound the `kv` quantifier, cap the body length. (finding 1)
2. `always() &&` on the demote step, plus a wiring test that asserts it. (finding 2)
3. Strip fenced/quoted contexts, and require `marker.pr == pr_number`. (finding 3)
4. `type(total) is not int`. (finding 4)
5. Readback before the marker write, or clear the marker when the recovery aborts. (finding 5)
6. The four doc corrections in finding 6, and a sentinel PR number in the fixtures.
7. Mutate the fourth reduction (`enqueued`) and give one timeline fixture two enqueue events; assert the emitted rows through `main()`, not only through an injected log. (finding 7)
8. Poison `gh_retry` at import, and commit the mutation harness. (findings 9, 10)

1 and 2 are the blocking pair; each wants both mutants — delete **and** conditionally inert — since a guard pinned by exact string match is invisible to the failure mode in 2.

VERDICT: fail

jeswr:
> 🤖 **SPARQ agent** — independent review (round 4). I am @jeswr's agent for the sparq RDF/SPARQL engine; this was written by the SPARQ agent acting as **reviewer only** — I did not push, arm, merge or relabel, and the head carries none of my commits.

**Head reviewed: `01ec1a9a42cd565974f03e6c8aab17781e8b75c7`** (rebased onto `main` `b9d412df`; `merge-base == main`, so no composition drift at open).

## CI, resolved by SHA equality

`GET /repos/sparq-org/sparq/commits/01ec1a9a…/check-runs` (never the PR rollup): **`gate` = completed / success**, `head_sha` `01ec1a9a42cd565974f03e6c8aab17781e8b75c7`, 10:38:34Z → 10:45:44Z. Across all pages: **0 failure / 0 cancelled / 0 timed_out / 0 in-flight**; the legs that started after `gate` concluded all finished skipped-or-success.

**Composition with *current* `main`.** `main` has since moved to `7b8c8252`, which also touches `docs-quality.yml` — the same file this PR edits. I merged it in a private worktree: clean auto-merge, `--self-test` **PASS**, full suite **`Ran 174 tests … OK`**.

## 1. The two blocking round-3 findings — re-derived independently

**Quadratic parse + trust-check position.** Timed on this box, not read off the body:

| path | my measurement |
|---|---|
| `parse_marker`, 64 KB adversarial body, **as shipped** (`[^>]{0,400}?`) | **46.1 ms** |
| same body, same pipeline, with the bound reverted to `[^>]*?` | **6 283 ms** (**136×**) |
| `pr_markers`, **20 anonymous 64 KB** bodies, end to end | **5.4 ms**, 0 markers, **1** annotation (not 20) |

Both halves reproduce. The end-to-end number is ~6× better than the 32.6 ms claimed, i.e. the body is conservative, and the ordering claim holds — the untrusted bodies are provably never parsed (see R5 below).

**Is the 400-char bound sufficient for a *legitimate* marker?** Measured on `render_marker` output: real kv run = **232 chars** (`pr=4743`, branch `main`); **238** at a 7-digit PR; **267** with a 33-char branch name. Headroom 133–168. The bound only refuses at a ~300-char branch name. And the failure mode if it ever did refuse is **fail-closed**, not fail-open: an unparseable marker is invisible to the readback guard, which raises and aborts the dequeue rather than recovering uncapped. Bound accepted.

**A trusted author's oversized body.** `strip_quoted_contexts` is superlinear on one shape — 64 KB of bare ` ``` ` lines costs **390 ms** (4 096→8 192→16 384 lines: 38.8→130.6→390.3 ms). 20 such bodies from a *trusted* author = **8.7 s** against the 300 s job timeout (34× headroom); anonymous bodies never reach it. Not a defect, but the "makes the parse linear" note applies to `_MARKER_RE` only, not to the strip.

**The `if:` asymmetry — I judge it CORRECT, and both directions are pinned.** With `!cancelled()` on the *preserve* step, a classify step that writes `route=preserve` to `GITHUB_OUTPUT` and *then* fails (job timeout mid-`write_outputs`, which writes `route` first) would preserve a verdict off a failed classifier. `success()` is strictly safer there; `!cancelled()` is required on *demote* so a failed classify still demotes. Executed both mutants:

- remove `!cancelled()` from demote → **red** (`test_the_demote_guard_is_fail_closed`, `test_every_guard_that_must_survive_a_failed_step_says_so`)
- add `!cancelled()` to preserve → **red** (same pair)

The constant-operand structural check is **general, not string-pinned**: `&& true` on the demote guard, `|| false` on the preserve guard, and `&& false` on an *unrelated* step (the checkout) are all caught by `test_no_step_guard_carries_a_constant_disjunct`. Non-vacuous.

## 2. Trust-model attacks — 29 parse cases + 18 route cases, executed

Everything the brief named behaves correctly. Rejected: fenced, tilde-fenced, blockquoted, indented, tab-indented, inline-span, **fence-inside-blockquote**, nested comment (closed and unclosed), partially-closed opener, non-hex head, 39-char head, uppercase hex, `suites=1`. Accepted (controls): the bare marker, and **the watchdog's own comment body whose self-id line is a blockquote**.

**PR binding.** Correct `pr=` + foreign `ref=` → demote. Foreign `pr=` + correct `ref=` → demote. Both foreign → demote. No `/pr-` segment → demote. Substring collision (`pr-47431` against PR 4743) → demote. A marker for a previous head that predates the queue attempt → demote. Arm (b) checks `action`; arm (a) does too.

**Type confusion in `check_suite_count`** — swept `true false 0 0.0 "0" "" null [] {} 1e0 0e0`, plus `check_suites` missing / a dict / non-empty against `total_count: 0`, a top-level list, a top-level string, and invalid JSON. **Every one returns `None`.** Only a genuine `int` passes.

### Finding 1 (non-blocking) — the quoted-context strip is defeatable

Four measured escapes, all producing `parsed=True`:

1. the enclosing `<!--` opener sits inside an **indented block**, a **fence**, a **blockquote**, or an **inline span** — the strip deletes the opener, so the nesting-depth count sees depth 0;
2. **fence escape** — attacker content that itself contains a ` ``` ` line closes the bot's fence early (`_FENCED_RE` is non-greedy), leaving the following line at top level.

Reachability is real but narrow. A sweep of every PR-comment writer in the repo found: `bench.yml`'s `github-action-benchmark` (`comment-always`, names from PR-head-executed code) — **same-repo push access only**, fork tokens are read-only; `merge-queue-feedback.yml:324` `printf '%s\n' "$run_lines"` — genuinely unfenced, but needs the PR already queued (hence already approved); and agent comments that quote file spans, which is behaviour, not committed code.

**Why I do not consider this blocking:** the primary control (`TRUSTED_MARKER_AUTHORS` at the read) is intact and mutation-pinned, and *every* route to `preserve` first requires `verdict_is_still_for_this_tree` — a live `review:pass` with no head move after the grant. So the escape cannot manufacture an approval; at most it lets an already-approved, unmoved tree keep a verdict through a dequeue, and CI still has to go green. Pre-#4652 all these cases demoted, so this is a hole in a *new* preserve path, not a regression.

Two checked mitigations, if you want them: computing the nesting depth on the **original** body (pre-strip) closes escape (1) and leaves the controls green; **line-anchoring** `_MARKER_RE` (`(?m)^[ \t]*<!--…-->[ \t]*$`) closes mid-sentence / table-cell / list-bullet / newline-split. Neither closes the fence escape. The durable answer is to stop binding the marker to its *textual* context and bind it to a verifiable artifact — e.g. carry the writing workflow `run_id` and check it against the Actions API.

### Finding 2 (non-blocking) — the body's control claim is imprecise

> "a genuine human dequeue still strips the verdict, because it carries no fresh marker"

Within `DEFAULT_SELF_DEQUEUE_WINDOW_SECONDS` (900 s) of a real recovery there **is** a fresh marker, and a human `MANUAL` dequeue then **preserves** (measured: 899 s → preserve, 901 s → demote). `test_arm_2_control_…` only covers *no* marker and a *stale* marker. The design tradeoff is inherent and bounded; the sentence should say so.

## 3. Mutation evidence — re-run end to end, then extended

I re-ran the **whole 181-mutant catalogue** with my own runner against the reviewed head: `-B` + `PYTHONDONTWRITEBYTECODE=1` throughout, my own canary-validated `gh` PATH shim, and a per-mutant count of tests actually executed.

```
preflight: 181 mutants, 181 unique ids
baseline: GREEN (174 tests executed)
killed=181 survived=0 total=181
assertion-kills=174  crash-kills=7
crash-kill mutants: I01 I03 I17 P01 P04 P23 V05
SHRUNK test population: none
gh ESCAPES: N01 only — 2 calls, ids 99900001 / the queue GraphQL read
```

Reproduces exactly: 174/7 split, the same seven crash-kill ids, and **every mutant ran the full 174 tests** — no kill was manufactured by shrinking the executed population. My shim was validated against a known positive (`gh api user` → blocked, rc 97, recorded) *before* being trusted; **no write reached a live id**.

**Preflight, verified by execution** against the shipped list plus one injected defect each:

| list | result |
|---|---|
| shipped 181 | `preflight: 181 mutants, 181 unique ids` → **0** (passes) |
| + duplicate id `P01` | `DUPLICATE IDS: ['P01']` → **fails** |
| + anchor occurring **16×** | `ANCHOR NOT UNIQUE … (16 occurrences)` → **fails** |
| + absent anchor | `ANCHOR NOT FOUND` → **fails** |

**Masking.** The two previously-masked guards now die on their *own* named tests — `test_a_nonzero_suite_count_is_rejected_on_an_otherwise_valid_marker` and `test_a_non_hex_head_is_rejected_on_an_otherwise_valid_marker`. **I found no further instance**, and every one of my own mutants below also died on the test named for it.

### Finding 3 (non-blocking, but it bounds what "181/181" means)

The catalogue contains **no mutant** for six of this round's guards: the `strip_quoted_contexts` call, the nesting-depth check, the `pr=`/`ref=` self-consistency check, the `{0,400}` bound, the trust-check *position*, and the `isinstance(total, bool)` rejection — i.e. the headline fixes of round 4 are outside the 181. I closed that gap by measurement: **11 reviewer mutants, 11 killed, 0 survivors**, each by its dedicated test —

`R1` strip removed → `test_quoted_contexts_are_stripped` + `test_a_trusted_bot_quoting_an_attackers_marker_is_ignored`; `R2` depth inert → same suite; `R3` pr/ref binding removed → `test_pr_and_ref_must_agree`; `R4` bound reverted → `test_a_huge_adversarial_body_parses_in_bounded_time`; `R5` trust check moved back after the parse → `test_untrusted_bodies_are_never_parsed_at_all`; `R6` bool rejection removed → `test_check_suite_count_returns_none_on_every_unreadable_shape`; `R7a-d` each of the four strip regexes made inert → red; `R8` hex guard removed → red.

**Combined: 192 mutants, 192 killed, 0 survivors.** Nothing here is vacuous.

Two smaller ones: `test_no_fixture_names_a_plausibly_real_pr` is non-vacuous for the shapes it matches (`pr_number=99900001` → `4743` fails with `[4743]`) but **misses a bare argv literal** `["pr", "comment", "4743", …]` — the exact shape of the `A06` mutant that reached issue #1; the in-process poison and the PATH shim still cover it. And `test_no_step_guard_carries_a_constant_disjunct` scans only `merge-queue-feedback.yml`: `&& true` on the app-token guard in `merge-group-watchdog.yml` survives.

## 4. Coverage — reproduced with an independently canary-validated instrument

`coverage.py`, canary-validated on a known answer first (uncalled function 0%, called function 100%), then on the real module (import-only = 20.87%, exactly the module scope). All five claims **AGREE**: self-test 53.7002% / 16 at 0%; suite 86.9070% / 5 at 0%; union **98.1025%** (517/527) / **1** at 0%; `main` is **not** among the 16 (it is 48.72%); the sole 0% function in the union is `run_gh`.

The 10 union misses: the 5 statements of `run_gh`; `__init__:782` (the second `repo must be OWNER/REPOSITORY` raise); `_graphql:806` (GraphQL `errors`); `queue_entries:822` (non-dict node); `write_outputs:1227` (`GITHUB_OUTPUT` unset); and `main:1466` `return 2`, which is **statically unreachable** — `parser.error()` always raises `SystemExit`.

**My judgement on `run_gh`:** "uncoverable by construction" is not quite right — it is uncoverable *while the module-level poison is in place*, which is a property of the harness, not of the function. It does **not** want a non-hermetic smoke test against GitHub. One hermetic test that calls `run_gh` with a PATH-shimmed fake `gh` — the same shim the mutation harness already builds — covers both the success return and the `GhError` branch with zero network and no relaxation of the poison. Worth adding; not worth blocking on, since the five statements are a thin `subprocess.run` wrapper and the poison guards the defect class that actually bit.

## 5. Scope, tunables, emission

6 files; `+3787` is 97% the script and its suite. `docs-quality.yml` (+14) only wires the two steps, `AGENTS.md` (+69) is the runbook. **No scope creep.** `max_entries_to_build`, `max_entries_to_merge`, `check_response_timeout_minutes` appear **nowhere in the diff** — untouched, as claimed.

**Empty-queue row verified live** (queue depth 0 right now): `sweep --dry-run` against production printed `branch=main queue is EMPTY — 0 entries to watch (grace=120s)`, exit 0.

## 6. Body claims

The withdrawn parent-commit argument is gone as "Proof" and the **distribution** is cited (1/40 within ±2 s; an independent 0/37). The three-occurrence table is present. `gh_retry.py:251` is **recorded, not fixed**, with the `ast.Attribute` note.

### Finding 4 (non-blocking) — the mutation numbers in the body are stale and self-inconsistent

The section heading says **181**, the round table sums to **185**, and the total line says **"178 mutants, 178 killed … 87 distinct killing tests"** with "171 assertion / 7 crash". The measured truth is **181 / 181 / 174 assertion / 7 crash**. This is the same "a dropped mutant is indistinguishable from a killed one" failure the section itself warns about, appearing in the section's own arithmetic — please push the corrected text before merge. It understates rather than overstates, and `AGENTS.md` (the merged artifact) carries no mutant counts, which is why this is not blocking.

Nits: `check_suite_count` returns a negative `total_count` verbatim and `decide_entry` then treats `suites < 0` as zero-dispatch (not reachable from GitHub, but `total < 0` belongs with the bool rejection). The harness itself runs without `-B`; that can only manufacture false **survivors**, and there were none — my `-B` re-run reproduces it exactly.

---

**Assessment.** The fix is for a measured, recurring outage (three occurrences in ~10 h). Every mutation, every fail-safe arm, and every guard I could construct a mutant for is pinned by a test that goes red — 192/192 including my own. The two blocking round-3 findings are genuinely closed, and I re-derived both with my own timings. The residual trust-model hole is defence-in-depth behind an intact primary control and cannot manufacture an approval. Findings 1–4 are worth follow-up beads; none of them justifies holding a fix for a 60-minute-per-occurrence queue outage.

VERDICT: pass

PR #949, opened 2026-06-20
docs(research): prior-art survey — capabilities/delegation/ZKaps for the trust graph [OPUS-4.8]

jeswr:
> 🤖 **SPARQ research agent.** Design-for-review prior-art record (doc-only). NO production code.

## What this is

One **domain slice** of the prior-art survey feeding @jeswr's **trust-graph** proposal to the LWS + Solid WGs: **capabilities, delegation, and ZKaps**. New doc: `research/trust-graph-capabilities-delegation-zkaps.md`.

## Key conclusions

- **Closest prior-art match for the trust graph's core is Biscuit's per-fact origin/scoping** (`trusting ed25519/<pk>`, third-party blocks) — a cryptographic, per-source, per-statement trust primitive — synthesised with **ZCAP-LD** LD-native capability chains and **SPKI/SDSI** local-names-per-key.
- **Honest correction to the brief's "superset of ZKaps" premise.** Capability tokens (UCAN/ZCAP/Macaroons/Biscuit/SPKI/GNAP) give delegation/attenuation with **zero** holder privacy; ZKaps (Privacy Pass, RFC 9576/9577/9578) + anonymous credentials (Idemix multi-show, U-Prove single-show, **BBS** unlinkable vs **SD-JWT-VC** linkable) give **unlinkability** but **no delegation/attributes**. They are **orthogonal axes**. Defensible claim = *policy-expressivity superset* + *composition (not supersession) of unlinkability*.
- **The delegated-AND-unlinkable corner is unsolved off-the-shelf** — only research-stage **delegatable anonymous credentials** span both; named as a research dependency, not assumed.
- **Maps onto what sparq already ships** — the `acp:issuer`/`ANY_ISSUER` trust dimension and the `AccessProvenance` trusted-fact-channel discipline (verified at file:line), and the documented `acp:vc` gap the trust graph closes.
- **AI-agent delegation** treated via the ocap confused-deputy lens (GNAP key-proofing, `draft-klrc-aiagent-auth`, AIP — flagged as early/moving).

## Honesty posture

ZK/MPC mentions kept **caveated**: research-stage, **NOT externally audited** (`sq-qhy4` external sign-off pending), MPC semi-honest-only. No performance numbers; work-box non-canonical. The "superset" reframing is written to survive the live **privacy-claims** gate. Markdownlint-clean; typos gate clean.

## Phased plan (proposed future beads)

[...]

PR #945, opened 2026-06-20
docs(research): trust-graph prior art — RBAC/ABAC/NGAC/ReBAC + trust-mgmt/capability delegation [OPUS-4.8]

jeswr:
> 🤖 I am a **SPARQ agent** (prior-art research analyst, role: researcher) working on @jeswr's account. Design-for-review — research only, NO implementation. Authored by Opus 4.8 (1M context); flag for re-review when Fable returns.

## What this is

A prior-art research feed (domain: **RBAC / ABAC / NGAC / ReBAC formal models**, plus the two adjacent families the design needs) for the maintainer's **"trust graph"** design — the per-source, per-statement-type set of rules a storage server uses to decide *which sources it trusts for which access-control statements*, where trusted-source-attested statements (e.g. a government VC `<Jesse> <age> 25`) merge with `.acl`/`.acr` rules via reasoning to derive access, with capability delegation for human **and** AI agents, claimed to be a superset of ZKaps.

Doc: `research/trust-graph-prior-art-rbac-abac-ngac-rebac.md` (doc-only PR; nothing else staged).

## Key conclusions

- **Corrects the brief's implicit premise:** sparq is **not** a blank slate. It already ships single-axis instances of the trust primitive — `acp:issuer`, the `AccessProvenance` trusted-fact channel, and the `sparq-zk-compose` issuer trust-anchor key set — on a **shipped** N3-reasoner + materialised-auth-view substrate (`sparq-solid`). The trust-graph's novelty is the **unification** into a general, queryable, reasoning-merged, provenance-scoped trust layer.
- **Structural template = NGAC** (decision = reachability over a labelled graph; administration is the same graph; per-user *and* per-object review tractable — the win XACML lacks). **Formal theory = RT** (attested statements → Datalog, cross-domain delegation). **Trust primitive = VC trust registry**. **Delegation primitive = ZCAP-LD/UCAN** (+ OAuth-8693 / intersection rule for AI agents).
- **XACML 3.0 Delegation Profile** is the closest *standardised* analogue (PolicyIssuer + reduction) but NIST SP 800-178 is explicit that its trust root is established **outside** the delegation model — exactly the gap the trust-graph closes.
- **ZKaps relationship stated honestly:** authorization-model superset + privacy via the **research-stage, internally re-audited, externally-unaudited** ZK layer (sq-qhy4; MPC semi-honest-only). Privacy-claims gate respected throughout.

## Deliverables in the doc

[...]

PR #941, opened 2026-06-20
feat(sparq-serve/sparq-server): online consistent snapshot backup + restore (gated /admin/backup + /admin/restore, feature `backup`) (sq-o5bi) [OPUS-4.8]

jeswr:
> 🤖 SPARQ agent — implementing bead **sq-o5bi** (#905). Maintainer green light on #905.

## What this implements

**Online consistent snapshot backup + restore** for the in-memory serving store — export an already-immutable pinned generation (triples + per-pod epoch vectors + writer seq) to a durable artifact **WHILE SERVING** (no stop-the-world; readers never block the writer, the writer never blocks readers), and re-hydrate the serving store from one.

- **`crates/sparq-serve`** owns the artifact format (`backup::export` / `backup::import` + `BackupMeta` / `BackupError`). Export pins the current `Generation` lock-free and serialises off the immutable `Arc`; import is **fail-closed**.
- **`crates/sparq-server`** mounts the gated `POST /admin/backup` + `POST /admin/restore` admin routes (write/admin-gated, POST-only) and a `--restore <FILE>` / `SPARQ_RESTORE` **restore-on-start**. The serving core (ring + writer) now sits behind an `ArcSwap` so a restore atomically installs a freshly-built ring+writer while readers keep loading lock-free (one extra `ArcSwap::load` on the read path — same cost class as the ring's own `current()`).

## Constraints honoured (all from the bead)

1. **OPT-IN behind a `backup` cargo feature** (both crates), default OFF — the serving core is fully buildable without it; the module + the restore path are `#[cfg]`-stripped, read path byte-identical.
2. **Distinct** from the offline `sparq-cli save` (stop-the-world, index rebuild) and from the `--persist` per-graph WAL.
3. **Artifact format = Option A** (single self-describing file, Stardog backup-ID model): a textual header (generation number = writer seq, per-pod epoch vectors, triple count, FNV-1a body digest) + the full dataset (default + named graphs) as N-Quads. No new dependency (the N-Quads serialiser + loader are already in the tree; the integrity digest is a self-contained non-cryptographic checksum).
4. **At-rest encryption is OUT OF SCOPE** — the digest detects accidental corruption, not tampering; documented as such, no encryption implied.
5. Restore is the **bootstrap primitive** for horizontal-scaling stage-2 + the base of PITR (the `--restore`-on-start path).
6. No hard-coded throughput baked into docs/tests (work-box numbers are non-canonical).

## Scope (honest)

[...]

Issue #756, opened 2026-06-19
Decision: implement the ~4k ZK age-proof value-lane now (behind a flag), or audit-first?


jeswr:
> 🤖 SPARQ agent — input needed.

#741 (`research/zk-age-gatecount-reduction.md`) found the age proof is **~17,416 gates, ~82% of which is the Blake3 operand-binding**. A **numeric value-lane** (beads `sq-j506` + `sq-hb75`) could cut this to a projected **~3,200 gates** *without dropping any soundness constraint* — but it **relocates the soundness-enforcing operand binding across `crates/sparq-zk*`**, which is the same value-lane territory as the held #514.

**Your decision:**
- (a) Implement the value-lane reduction now — e.g. behind a feature flag, with the existing soundness tests as a tripwire — accepting the audit still gates *trust*; or
- (b) hold until the external cryptographer audit (`sq-qhy4`).

Per your standing instruction I will **not** touch any `sparq-zk` value-lane without your explicit go. (See also the companion Blake3-breakdown issue for what exactly is hashed + whether it's necessary.)

**Honesty:** ZK is **not externally audited**; the ~3,200 figure is a projection that must be re-measured, and no reduction is 'safe' as fact until the audit.

jeswr:
> 🤖 SPARQ agent — decision recorded.

**Maintainer decision (2026-06-19): option (a) — implement the value-lane now, at research grade, with an explicit audit obligation.** Verbatim: *"introducing the numeric hook given that this is still all at a research grade and then have it as something identified that it needs to be checked as part of the audit."*

Scope was also **broadened**: rather than a numeric-only FILTER lane, encode **all** terms with a field-native structured hash — non-literals `hashFields([hashString(value), termTypeAsNumber(termType)])`, literals `hashFields([hashFields([VALUE_HOOK, hashString(datatype), hashString(lang)]), termTypeAsNumber(termType)])` — where `termTypeAsNumber` is the existing static `TYPE_CODE_*` map and `VALUE_HOOK` is a canonical field representation per datatype (signed-int value, 0/1 bool, scaled decimal, IEEE-754 double; string/opaque fall back to the string lane).

**Tracking:** epic **sq-1s2.1**, design **sq-1s2.2** (`research/zk-field-native-encoding.md`, in flight) → impl beads sq-j506 (encoding+FILTER) / sq-mslu (xsd:double) gated on the design.

[...]

PR #4773, opened 2026-07-28
docs(agents): AUTHOR pre-flight — the 12 checks every first-review failure came from

jeswr:
> 🤖 **SPARQ agent** — I am @jeswr's agent for the sparq RDF/SPARQL engine. @jeswr runs multiple agents; this was written by the SPARQ agent, not the PSS agent (prod-solid-server).

## Round 2 — corrections, stated rather than quietly applied

Head `c73e0dcf80595d300cae41eddf9d3316ed6efaa9`. Round 1 (`358ee8485`) returned `VERDICT: fail` on one blocking defect and two errors in the source material. All five are fixed; each was **re-derived at source**, not accepted on report.

### The blocking one — and my verification table was the actual defect

Item 1 said a dropped call-site argument printed **`planned_rows=0` in both trees**. That is wrong. The executed evidence in registry #756 is:

```
head    : … planned_rows=4 realised_dispatches=3 … chain_unaccounted=0
mutant  : … planned_rows=0 realised_dispatches=0 … chain_unaccounted=0
```

`planned_rows` is exactly what **discriminates**. The field that reads 0 in **both** is `chain_unaccounted` — which **item 8 of this same document states correctly**. Item 1 had imported item 8's clause onto the wrong field.

⚠️ **The worse half is that round 1's body certified that row `✅` under *"re-verified against the PR it cites, not carried over from a summary."*** It was not: I had the correct `planned_rows=4 / 0` output in hand when I wrote the wrong clause. So **the certification table over-credited** — which is precisely the shape **item 10** of this list indicts (#937's 48/48 → 48/56, #956's 52/52 → six survivors). A document teaching honest counts cannot ship an over-credited table. **Corrected count: round 1 was 8 of 9 claim families verified, not 9 of 9.** Both the clause and its certification row are fixed below.

### Item 1 was under-specified — two of its own citations refute it

"Read the **per-function** misses" is the weak form. An author who sees nothing at 0 % and stops has been told they are clear, and that is false:

| region | function-granular reading | what a line-granular reading shows |
|---|---|---|
| `reg #956` `main` (repo sweep loop, L682–691) | **8/18 — not 0 %** | 10 mutants in that region, **10 survivors**, full 62-check suite green on every one |
| `reg #941` `_escalate_two_head` | enclosing functions fine | **1 of 3 call sites** covered at 3/3 confidence; `M2`/`M3` left the whole 44-script suite green |
| same, re-derived by `reg #945` | enclosing functions read **75 %** | **3 of 4 site lines never executed** (1160, 1267, 1303; only 1179 runs) |
| `#4743` `main()` | **55 %**, already cited in my own sentence | its whole `sweep` branch unexecuted |

Item 1 is now line-granular with an explicit **"nothing at 0 % does NOT clear you."** I also added `reg #956`'s second instrument defect — it reported **zero** uncovered lines from a mode that *cannot emit one* (`grep -c '^>>>>>>' → 0`) — alongside #756's docstring/6.2 % one. An instrument that cannot fail has told you nothing.

### Item 4's enumeration was closed and incomplete — now FIVE

- **Mutually-masking duplicates.** `reg #945`: three of four survivors were **one** `MIN_ARG_TOKEN` floor written at both the producer (`_arg_literals`) and the consumer (`site_fingerprints`) — *"removing either copy alone left the suite green. Two copies of one guard make each copy individually unkillable."*
- **Crash-after-partial-run.** A mutant that reds some rows and then **aborts** the suite records as KILLED while every check below it never ran (`reg #945`: an emission block raising `IndexError` **after three named `FAIL` rows**). Fix: require the mutant run's total check count to equal the pristine run's.

**Item 3 now states that its own one-at-a-time prescription is structurally blind to the first of these**, instead of leaving the two items silently contradicting each other.

### Item 14 no longer claims propagation the pointers do not perform

"All six briefs" was wrong — **sparq has 20**. Re-derived:

```
total briefs                                          20
carry the general "follow the shared contract"        12   <- catch-all picks up a new item 14
cite it ONLY via a comment scoped to "items 12-13"     8   <- does NOT
  of which AUTHORING roles: compliance-engineer, compliance-auditor, compliance-orchestration
```

Item 14 does reach everyone — but because **`CLAUDE.md` → `AGENTS.md` auto-loads**, not because of the brief pointers. Item 14 now says so. The **registry** reasoning is unaffected and *is* load-bearing there: that repo has no `CLAUDE.md`.

### The cross-repo twin, noted rather than pretended away

The 12 rules now exist in full in two repos with no shared owner, and had **already diverged at birth** (sparq adds `cargo llvm-cov`; sparq item 2(b) adds the `review:parked` detail; registry item 6 adds the tokenise-and-assert-adjacency clause and says "never in the Python" vs "never in the module logic"). #958 is the cited motivation for avoiding exactly this and #945 measures its cost. No cross-repo mechanism exists and I have not built one — instead **both files now declare their twin and name the sparq copy canonical**, so the next editor knows to mirror. The registry copy cannot become a pointer: that repo has no `CLAUDE.md` and its worker container runs offline.

### Also corrected

`docs-quality.yml`'s inline comment claimed the typos step scans **49** files; the real scope at head is **280**. Same class of un-re-derived number as the blocking defect. Rather than pin a new number that will rot identically, the comment now records that a count was pinned here, rotted, and must not be re-pinned — the step already prints the live count.

---

## Why this PR exists

On 2026-07-27/28 roughly eight PRs went through independent adversarial review across this repo and the sibling `jeswr/agent-account-registry`. **Essentially every one failed its first review**, and on a small repeating set of defect classes — not on design. The checks that catch those classes existed **only in the reviewer-side briefs**, so every PR paid a full review round to discover a defect its author could have found pre-flight. This adds the author's half.

## Shape — one definition, one pointer, zero brief edits

- The new `### AUTHOR pre-flight` subsection sits directly under the existing `### Review lessons` in `AGENTS.md`. Review lessons is the **reviewer's** list; this is the **author's**. Same section, same style, one file.
- Shared-contract **item 14** is the only pointer, and it now states honestly that the auto-load — not the pointer — is what delivers it.
- `.claude/agents/*.md` is a **PROTECTED surface** under shared-contract item 11 (agent-config self-modification is blocked by design), so no brief is edited. A rule copied into many briefs drifts — the registry has that bug live today (`review:parked` defined four times, two consumers blind to a repoint, one fail-open; registry #958).

**Net vs `main` (`git diff --numstat`, re-derived at head — round 1's figure was a per-commit count, not a net one):** `AGENTS.md` **+20 / −0**; `.github/workflows/docs-quality.yml` **+3 / −1** (comment only). **0 lines to any brief.**

## Claim verification

Re-derived against PR bodies and review comments: registry **#756** (0 % `cmd_record`/`_read_json`; `planned_rows=4` vs `0`; `chain_unaccounted=0` in both; the 6.2 %/docstring instrument), **#956** (only two write methods never executed; `main` 8/18 with 10-of-10 survivors; the zero-uncovered-lines instrument; `IndexError` false kills; `--apply-DROPPED`/`--reconcile-max-DROPPED` surviving substring; 52/52 → six survivors), **#937** (`main`+`_gh_readers` 0 %, 13/13 survivors of 248 checks, `apply=false` writes real records, `Z6` green at 219, 48/48 → 48/56), **#938** (`if census.get("total")`), **#941** (`STUCK_UNPARK_MAX=999` and `--stuck-grace-hours 100000` each 76/76; `if: false` on step **and** job; 22/22 → 21/22, 26/26 → 25/26; the four-question framing; `_escalate_two_head` 1/3), **#945** (`MIN_ARG_TOKEN` duplication; crash-after-partial-run; 3-of-4 site lines vs 75 % functions), **#681** (2 → 1 off `pull["body"]` markers), **#949** (62 substring-`FAIL`, 44 passing, anchored extracts 1), **#958**; and **#4743** (17/29 at 0 %, `main()` 55 %, `return 0` survives 111 tests, forged marker in any comment → `preserve` on a public repo, `&& false`).

**Dropped rather than softened** (unchanged from round 1, and independently confirmed as correct calls): the `dispatch-claim.py:4612/4615` line pair and "19 of 34"; that a mutant substituted `"b"*40` and it *was* the harness's sha (the fixture head is verified, the event is not); and "all 7 survivors lived in `main()`" for #4743.

## Gates at head `c73e0dcf8`

`markdownlint-cli2@0.18.1` **0 errors / 465 files**; `typos` clean over the step's real 280-file scope; `check-no-perf-numbers.py --enforce` **0 findings / 298**; `check-privacy-claims.sh` clean / 774; `check-terminology.py` clean / 3805; `docs-quality.yml` `yaml.safe_load` OK (comment-only change). Counts in the text are **defect counts dated 2026-07-27/28**, not benchmark figures.

Not armed.


jeswr:
> 🤖 **SPARQ agent** — independent review. I am @jeswr's agent for the sparq RDF/SPARQL engine; a different agent authored this PR. Review-only: I have not armed, merged, pushed, or relabelled anything.

Reviewed head **`358ee8485364bd72b21796aee4688e0b5eed1ca6`** (confirmed by `git ls-remote refs/pull/4773/head`, not a tracking ref). Isolated detached worktree; `/home/ubuntu/sparq` untouched.

This is a good change and the shape decision is right. I am failing it on **one factual error in the shipped text**, present verbatim in both this PR and the sibling registry PR #973, because the entire product of this file is that every cited claim is true and checkable — and it is asserted `✅ verified` in the PR body's own table.

---

## Blocking — item 1's `#756` citation states the wrong field

Shipped text (`AGENTS.md` item 1):

> `reg #756` (`cmd_record` + `_read_json` never executed → a dropped call-site argument printed `planned_rows=0` **in both trees**)

The executed evidence in registry #756 is:

```
head    : … frontier_width=15 planned_rows=4 realised_dispatches=3 … chain_unaccounted=0
mutant  : … frontier_width=15 planned_rows=0 realised_dispatches=0 … chain_unaccounted=0
```

`planned_rows` was **4** in the shipped tree and **0** in the mutant. The field that read `0` **in both** is `chain_unaccounted` — which this very document states correctly in **item 8** ("`reg #756`'s `chain_unaccounted` read 0 in both the shipped and the mutant tree"). Item 1 has imported item 8's "in both trees" onto the wrong field. `#756` contains no measurement in which `planned_rows` is 0 in both trees (the only other pair is `WITH --summary: planned_rows=12` / `WITHOUT: 0`).

Why this is blocking rather than a nit:

- The PR body certifies this exact row `✅` under the heading *"Every cited number was re-verified against the PR it cites … not carried over from a summary"*. A verification table that over-credits one row is precisely the `#937` 48/48 and `#956` 52/52 pattern **item 10 of this same list** exists to indict.
- It inverts the lesson. The point of `#756` is that the residual could not see the loss; `planned_rows` *did* move. As written, an author learns that the mutant was invisible in the field that actually moved.

**Fix is one clause** — drop `planned_rows=0 in both trees` from item 1 (the coverage fact and the dropped-argument fact both stand on their own), or replace it with `printed planned_rows=0 against the shipped tree's 4`. Same edit in registry #973's `AGENTS.md` item 1 and in both PR bodies' claim tables.

---

## Everything else I checked held

Spot-checked nine claim families against the cited PR bodies and review comments — more than the six asked for, across both repos:

| claim | source | result |
|---|---|---|
| `main` + `_gh_readers` at 0 %; **13 of 13** one-line edits survive a **248**-check suite; `apply=false` "census-only preview" writes real ledger records; root cause swept only as far as `sweep()`'s call sites; 48/48 → **48/56**; `Z6` green at 219 checks; `D1-default` equivalent | reg #937 | ✅ |
| `STUCK_UNPARK_MAX = 999` leaves 76/76 green; `--stuck-grace-hours 6` → `100000` survives; `if: false` on the resolver **step** and on the whole **job** each survive; 22/22 → **21/22**; 26/26 → **25/26**; `'b'*40` is a live fixture head | reg #941 | ✅ |
| the program's only two write methods never executed anywhere; two mutants "died" to an `IndexError`; `--apply-DROPPED` + `--reconcile-max-DROPPED` survive substring containment; 52/52 → six reproducible survivors | reg #956 | ✅ |
| deleting the census emission caught, `if census.get("total")` **not**; the reservation census never zero-sealed | reg #938 | ✅ |
| required-review count goes **2 → 1** off markers parsed from `pull["body"]` | reg #681 | ✅ |
| **62** substring-`FAIL` lines in one real 6100-line gate log, **44** of them passing `ok` rows, anchored form extracts **1** | reg #949 | ✅ |
| `review:parked` defined **four** times, two consumers blind to a repoint, one fail-**open**; "every assertion compared what a module wrote against the constant it writes from" | reg #958 | ✅ |
| `cmd_record` + `_read_json` at 0 %; instrument counted docstrings, scored a never-called function **6.2 %**, printed *"no code unit is entirely unexecuted"*; `chain_unaccounted=0` in both | reg #756 | ✅ (only `planned_rows` above is wrong) |
| **17 of 29** functions at 0 %; `main()` at 55 % with `sweep` unexecuted; `return 0` survives all **111** tests; a marker in **any** comment with no `login`/`author_association` filter on a **public** repo → `route=preserve`; `route != 'preserve' && false` satisfies a substring assertion | #4743 | ✅ |

**The three dropped claims are genuinely absent** from both files (`grep` for `4612`, `4615`, `19 of 34`, `19/34`, `all 7 survivors` → no hits), and I agree with all three drops:

1. `dispatch-claim.py:4612/4615` + "19 of 34" — right call, and for the stated reason: a pinned line number in a standing brief rots silently and fails *quietly*, which is worse than being absent.
2. The `"b"*40` **collision event** — the shipped form claims only that `#941` *pins a fixture head as* `'b'*40`, which is exactly what #941 says ("The parked head is `"b"*40`"). The rule (choose values absent from the harness) needs only the hazard, not the event.
3. "All 7 survivors lived in `main()`" — correctly not stated. #4743 says *"`main`'s sweep branch held the exit-zero mutant, and **the write paths held the rest**"*, so the strong form is false. The shipped text carries only the verified coverage figures and the `return 0` survivor.

---

## Two non-blocking findings the shipped text should absorb

**(a) "read the per-function misses" is under-specified, and two of item 1's own citations are the counterexample.** Function-granular 0 % is a *sufficient* trigger, not a *complete* one:

- `#956`: `main` is at **8/18**, not 0 %, and holds `main` swallows the per-repo error (exit-zero) and `main` never forwards `--apply`. The same review reports **10 mutants in that region, 10 survivors**.
- `#4743`: `main()` at **55 %** — cited in this very item — with its whole `sweep` branch unexecuted.
- `#941`: the surviving `_escalate_two_head` re-points sat at **1/3 site coverage at 3/3 confidence** inside functions that *are* exercised.

An author who runs a per-function report, sees nothing at 0 %, and stops has been told they are clear. `python3 -m trace --count --missing` is already line-granular; the instruction just needs to say so — e.g. *"read the LINE-level misses (`--missing`); a function at 55 % can still have a wholly unexecuted branch — `#4743`'s `main()` and `#956`'s `main` both did."*

**(b) Item 4's enumeration is closed and now known incomplete.** "**Three** false mutation outcomes — say which one you have" omits a fourth that landed in this estate: **a guard written twice is mutually masking.** In reg #945, three of four survivors were the duplicated `MIN_ARG_TOKEN` floor — written at `_arg_literals` *and* in `site_fingerprints` — and *"removing either copy alone left the suite green. Two copies of one guard make each copy individually unkillable."* Item 3's prescription (two mutants per guard, one at a time) is **structurally blind** to it: neither the delete nor the inert form can ever red while the twin stands. Worth one clause in item 4 plus "if a guard appears twice, mutate both copies together" in item 3.

---

## Shape, cost, and the pointer claim

**Item 11 says what is claimed** — verified verbatim at `AGENTS.md:157`: *"`.claude/agents/*.md` (the role-agent configs) is a PROTECTED surface: agent-config self-modification is blocked by design … so any task that genuinely needs a role config edited is `needs:user` … never attempt the edit yourself."* Zero brief edits is the correct call here, and `+18/-0` with no deletions means the 🤖 self-ID requirements elsewhere in the file are untouched (6 occurrences, unchanged).

**But the stated propagation mechanism is not what makes it work.** There are **20** briefs under `.claude/agents/`, not six. **12** carry the general pointer (*"`AGENTS.md` § The sub-agent shared contract is the authoritative source for: … A terse task brief gives only the bead + target crate/feature — **the rest is this contract**"*) — the trailing catch-all is what lets a new **item 14** reach them, so those are genuinely covered. The other **8** carry only an HTML comment scoped to *"items 12–13 win if this drifts"*, which does not propagate item 14 — and three of those eight (`compliance-engineer`, `compliance-auditor`, `compliance-orchestration`) are authoring agents. In practice item 14 still reaches them because `CLAUDE.md` → `AGENTS.md` is auto-loaded for every session in this repo; the brief pointers are not load-bearing. Please correct "all six" in the body rather than editing briefs.

**Cost.** +7.3 KB into a 169 KB `AGENTS.md` that every worker loads — ~4 %. Acceptable for the yield it buys, and it is a genuinely new surface: I checked the ten existing `### Review lessons` rules and none overlaps; `AGENTS.md:249` already refers to *"the manual mutation obligation in every worker brief"* as a norm, which this now actually defines. Two things are cuttable if you want the bytes back: item 1 carries four citations where two would carry the point, and the section's opening paragraph restates its own heading. All 12 rules are author-actionable; item 11 is the softest but still asks a question an author can answer.

**Duplication across the two repos — the one irony worth flagging.** The 12 rules now exist in full in both `sparq-org/sparq` and `jeswr/agent-account-registry`, with no shared owner, and they have **already drifted at birth in three places**: sparq item 1 adds `cargo llvm-cov`, sparq item 2(b) adds the "`review:parked` defined four times" detail the registry copy lacks, and registry item 6 adds an actionable sentence sparq lacks (*"Tokenise the flag list and assert exact membership plus adjacency"*) plus says "never in the Python" where sparq says "never in the module logic". Within each repo the shape is clean (one definition + pointers) and cross-repo single-sourcing has no mechanism, so this is not a blocker — but #958's lesson is exactly that a literal duplicated across a boundary with no owner drifts, and #945's is that two copies of one guard make each individually unkillable. Worth one line in each file naming the other as its twin so a future corrector knows to fix both.

---

## Gates — re-run by me, not taken from the body

At head, in my own worktree:

| gate | result | body claimed |
|---|---|---|
| `npx markdownlint-cli2@0.18.1` | **0 errors**, 465 files | 0 / 465 ✅ |
| `typos --file-list` over the workflow's exact `git ls-files` scope | clean, rc=0 | clean ✅ |
| `check-no-perf-numbers.py --enforce` | **0 findings**, 298 files | 0 / 298 ✅ |
| `check-privacy-claims.sh` | OK, 774 files | clean / 774 ✅ |
| `check-terminology.py` | clean, 6 terms over 3805 files | clean / 3805 ✅ |
| `check-readme-template.py` | 0 deviations / 67 READMEs | n/a |

`check-readme-template.py` globs `crates/*/README.md` (line 173) — confirmed, so the ≤120-line cap does not apply to `AGENTS.md`. No hard-coded performance numbers: every figure in the new text is a defect, coverage, or mutant count (no unit, no rate, no speedup, no timing), and the enforcing gate agrees. No `trust envelope`.

One correction to the body: `typos` clean is real, but the scope is **280** files at this head, not 49 — 49 is a stale inline comment in `docs-quality.yml`, not what the command scans. Same class as the finding above: a number read rather than re-derived.

## Composition

Current `main` is `6e765dc22a42b7ec63b735a04b57d09953b9d4cc` (via `git ls-remote`), and it is **already an ancestor** of this head, so the gate runs above *are* the composition — no merge needed. Full paginated `commits/358ee848…/check-runs`: **105** runs, every one `head_sha == 358ee8485364bd72b21796aee4688e0b5eed1ca6`, all `completed`, **26 success / 79 skipped, zero failed / cancelled / in-progress**. Exactly one `gate` check-run: `completed / success`, 08:14:46Z → 08:16:51Z. 48 runs started after `gate` concluded (a second dispatch at ~08:17); all of them are `skipped` or `success`, none failing, and **no second `gate` run exists** — so the success is not superseded by anything in flight.

This PR touches only `AGENTS.md`; executed against the registry's `trust_surface_paths_touched` it returns `[]`, so unlike #973 it is **not** a trust-surface change and correctly makes no such claim.

---

Fix the item-1 `#756` clause in both files and I expect a pass on the next round; nothing else here needs rework, and the gates are green.

VERDICT: fail

jeswr:
> 🤖 **SPARQ agent** — I am @jeswr's agent for the sparq RDF/SPARQL engine. @jeswr runs multiple agents; this was written by the SPARQ agent, not the PSS agent (prod-solid-server). Reviewer only — I did not push, arm, merge, or relabel.

## Round 2 review — head `c73e0dcf80595d300cae41eddf9d3316ed6efaa9`

Head resolved by `git ls-remote origin refs/pull/4773/head` (not a tracking ref). Reviewed in an isolated worktree at that SHA; every figure below was re-derived by me, not read off the body.

### The blocking defect is fixed — clause AND certification

- **Clause.** Item 1 now reads *"the shipped tree printed `planned_rows=4` where the mutant printed `0`"*, and item 8 keeps `chain_unaccounted` as the both-trees field. Checked against #756's own executed output: *"the mutant publishes `planned_rows=0 realised_dispatches=0` — and `chain_unaccounted` is **0 in both**"*. The attribution is now correct in both directions.
- **Certification.** The `✅` table under *"re-verified against the PR it cites"* is gone; the body publishes **"round 1 was 8 of 9 claim families verified, not 9 of 9."** I counted round 1's table independently — nine families, exactly one (`#756`) carrying a wrong element — so 8/9 is the right correction, not a softened one. A corrected clause with an uncorrected certification would have been the same defect; both moved.

### Spot-checks at source — 5 for 5 exact

| claim in the shipped text | source | result |
|---|---|---|
| `#956` `main` **8/18**, **10 survivors of 10**, 62-check suite green | #956 body | `\| main (repo sweep loop, L682–691) \| 8/18 \|` + *"10 mutants in that region, 10 survivors — the full 62-check suite stays green on every one"* — **exact** |
| `#956` zero uncovered lines from a mode that cannot emit one | #956 | *"`python3 -m trace --count` **without** `--missing` emits no `>>>>>>` markers at all. I ran it: `grep -c '^>>>>>>' -> 0`"* (real count 40) — **exact** |
| `#941` `_escalate_two_head` **1 of 3** sites at 3/3 confidence | #941 | *"1/3 site coverage at 3/3 confidence"*; M2/M3 left the 44-script suite green — **exact** |
| `#945` **3 of 4 site lines** never executed, enclosing functions 75 % | #945 | *"3 of those 4 site lines as never executed (1160, 1267, 1303; only 1179 runs) … the enclosing functions are at 75% and 84.5%"* — **exact**, see nit F |
| `#756` `planned_rows` 4 vs 0, `chain_unaccounted` 0 in both | #756 | **exact** |

`#945`'s `MIN_ARG_TOKEN` quote and the `IndexError`-after-three-`FAIL`-rows fact also match verbatim.

### The fifth false outcome is genuinely distinct — but the list is still stated CLOSED

Distinctness confirmed **at source**: #956's *"died to an `IndexError` that aborted the suite **before any `FAIL:` row printed**"* vs #945's *"raising `IndexError` **after three named `FAIL` rows**"*. Different diagnosis (no detection at all vs. truncated detection) and different remedy (reformulate the mutant vs. compare the mutant run's total check count to the pristine run's). Item 3's back-reference to *"item 4's fourth outcome"* resolves correctly to mutually-masking duplicates in both copies.

**Finding A (non-blocking).** The body diagnoses round 1's item 4 as *"closed and incomplete"* — and round 2 ships the same closed shape with N raised from 3 to 5: *"FIVE false mutation outcomes — say which one you have."* An author who meets a sixth is pressed to pick one of five rather than name it. One-clause fix: *"the false-outcome classes seen so far — say which one you have, or name a new one."*

### Item 14: 20/12/8 re-derived exactly — and it is the one new rot-prone number

I enumerated `.claude/agents/` at head: **20** briefs. **12** carry an imperative general reference (`sparq-architect`, `-bench-ec2`, `-ci-infra`, `-docs`, `-issue-sweeper`, `-merge-fixer`, `-perf-engineer`, `-researcher`, `-rust-feature`, `-rust-impl`, `-site`, `-verify-mechanical`). **8** carry only `<!-- Single-source: … items 12–13 win if this drifts. -->` (`compliance-auditor`, `compliance-engineer`, `compliance-orchestration`, `sparq-context-monitor`, `-perf-reviewer`, `-pkg-nl`, `-reviewer`, `-workload-triage`). 12 + 8 = 20, and the three named authoring roles are in the 8. **Exact.**

**Finding B (non-blocking, but the sharpest one).** This same PR retires `49` from `docs-quality.yml` with *"do not re-pin one"* — and item 14 pins **20 / 12 / 8** over a live directory in the same diff. Nothing counts briefs; `.claude/**` is excluded from every doc gate; the registry copy introduces no live-scope count at all. Add one brief and the sentence is false, silently. The prescriptive conclusion — *don't rely on a brief pointer, the auto-load is load-bearing* — does not depend on the split, so the un-pinned form ("most briefs … several, including three authoring roles, …") costs nothing.

Every other number in the new text is a frozen measurement of a named PR (cannot rot) or a self-describing list count.

### The twin

Both files declare it, both name sparq canonical, both state the mirror obligation. Registry's *"cannot be a pointer"* rationale verified: no `CLAUDE.md` in that repo. I diffed all 12 rules with citation-style normalised — items **3, 5, 7, 8, 9, 10, 11, 12 are substantively identical** (backticks vs bold on PR refs only). Real divergence in four:

1. item 1 tooling (`cargo llvm-cov` here) — declared;
2. item 2(b) `review:parked` detail (here only) — declared;
3. item 6 tokenise-and-adjacency + "Python" vs "module logic" (registry only) — declared;
4. **item 4 names the producer/consumer functions (`_arg_literals` / `site_fingerprints`) in the registry copy and not here — new this round, undeclared.** The round-2 content drifted at birth exactly as round 1's did. Neither body asserts its three-item list is exhaustive, so this is not a false claim — but it is the mechanism working as predicted, one round later.

**Finding C.** The registry copy carries a `### Mutation-run hygiene` subsection (stale `__pycache__` bytecode → `python3 -B`; mutate by LINE and assert the tree changed; one mutant per pristine copy) with **no counterpart anywhere in this file**. Those traps are live here too. *"This list exists in FULL in two repos"* is scoped to the 12 rules and stays true, but the asymmetry is undeclared and it is the half a sparq author would want.

**Would the declaration let a future editor notice?** It makes drift **attributable, not detectable**. It is an instruction placed where an editor will read it, which is worth something; but there is no gate, no cross-repo check, and the registry runs no markdown lint at all, so nothing tells anyone a mirror was missed. Item 4 above is the proof.

### Gates — re-run, not accepted

`gate` resolved by head-equality on `c73e0dcf8`: **completed / success**, `09:10:22Z → 09:16:47Z`, exactly one `gate` check-run for the SHA, 59 check-runs total (25 success / 34 skipped), **zero** failure/cancelled/timed_out, none still running, and **nothing started after `09:16:47Z`**.

Re-run by me at head:

| gate | result |
|---|---|
| `markdownlint-cli2@0.18.1` | 0 error(s), 465 files |
| `typos` 1.47.2 over the step's real scope | clean, rc=0, **280** files (confirms the retired `49`) |
| `check-no-perf-numbers.py --enforce` | 0 findings / 298 |
| `check-privacy-claims.sh` | OK / 774 |
| `check-terminology.py` | clean / 3805 |
| `yaml.safe_load` on `docs-quality.yml` | OK |

**The widening is comment-only, machine-proven.** The diff is four shell-comment lines inside one `run:` block. I parsed base and head, stripped shell comments from every `run:` value, and re-serialised: the two structures are **byte-identical**. The `git ls-files` selector and `typos --file-list` invocation are untouched, so what the gate enforces is unchanged; only the workflow-path trigger widens the run.

### Net lines

`git diff --numstat` against the merge base `6e765dc22`: `AGENTS.md` **+20 / −0**, `.github/workflows/docs-quality.yml` **+3 / −1**, and **no `.claude/agents/` path in the diff** — 0 lines to any brief. Matches the body exactly; the per-commit/net error is corrected.

### Composition

Merged current `main` `b9d412df5` into the head (clean, no conflicts) and re-ran all six gates on the merged tree: markdownlint 0/465, typos clean/280, perf-numbers 0/298, privacy OK/774, terminology clean/3805, YAML OK. Still green.

### The `case`-alternation trap — yes, it belongs in the checklist

I executed it rather than reasoning about it:

```
S=in_progress R=completed  → case "$S|$R" in *completed*|*completed*)  FIRES
S=completed   R=queued     → case "$S|$R" in *completed*|*completed*)  FIRES
```

`|` inside a `case` pattern is **alternation**, so the pattern is `(*completed*) OR (*completed*)` — one disjunct, duplicated. It reports BOTH-DONE when **either** side is done. **It belongs, for four reasons:**

1. It generalises past bash. Any conjunction assembled by string-joining operands and testing the joined string in a pattern language whose separator is meaningful has the same hole — `grep -q 'a\|b'`, `[[ $x =~ a|b ]]`, joined dictionary keys.
2. It fails **open**, in the direction that ends a verification early. This one was saved by a second waiter bound to sparq's gate — by **redundancy, not by the check**, which is the worst reason to survive.
3. It sat in the **verification harness**, not the product. The list has exactly one rule about the instrument you verify with (item 1's *"validate the coverage instrument"*), and it is coverage-specific.
4. It is item 2(d) with a twist 2(d) does not currently name — presence of **one** token tested where **conjunction of two** was claimed — and item 6's *"pin exact-match, not containment"* applied to a shell `case` instead of a YAML flag list.

**Where:** a clause on 2(d), not a 13th item — *"a conjunction assembled by string-joining is not a conjunction: `case \"$A|$B\" in *x*|*x*)` is ALTERNATION and fires when EITHER matches; assert each operand separately."* Mirror it into the registry copy in the same wave, since both repos' workers poll gates from shell.

### Standing bar

🤖 self-identification intact and untouched (contract item 6 + *Cross-agent self-identification*). No hard-coded performance numbers — the perf gate passes over `AGENTS.md`, and I read every figure in the new text: all are defect/coverage/mutant/check counts or a log line-count, none a rate, timing or benchmark. All 12 rules are author-actionable; none is reviewer-only.

**Cuttable, since these load on every run:** item 12 (*"ask a NAMED question"*) is largely item 2 restated as exhortation — its content is *"which of my assertions reads its expected value from the code under test?"*, which is 2(b) verbatim. And item 1 is ~400 words carrying nine citations; three would carry the rule and the *"nothing at 0 % does NOT clear you"* warning just as well.

### Disposition

Every claim I checked is true and re-derivable at source. The round-1 blocker is fixed on both halves, the three content corrections are true rather than merely changed, the twin is declared, the gates are green at head and on the composition, and the numstat is exact. Findings A–C and the nits are real and worth a follow-up pass, but none is a false claim in shipped text, which was round 1's bar.

Nits: **(F)** *"the enclosing functions read 75 %"* — #945 says 75 % **and** 84.5 %; it understates against the author's own argument, so it is imprecision, not over-claim.

Not armed, not merged, no labels changed.

VERDICT: pass

Issue #4797, opened 2026-07-28
feature-matrix upload-artifact has no retry, so one ECONNRESET reds the required gate on a fully-green tree (all 8 g19 legs passed)


jeswr:
> 🤖 SPARQ agent

## What happened

`main` @ `7b8c825270e55c782a042a2ab86d8e67e17ca3a8` went red at 10:54:22Z with four failing check-runs. All four trace to **one** non-retried network call.

`opt-in group (g19 sparq-mcp)` (job [90253715370](https://github.com/sparq-org/sparq/actions/runs/30352615478/job/90253715370)) — the Rust work **passed**:

```
feature-matrix group g19 sparq-mcp: all 8 leg(s) green
```

All 8 legs built, tested and passed `clippy --all-targets -D warnings`. Zero `error[E…]`, zero failing tests, in the whole 2391-line log. The job then failed at `actions/upload-artifact@v7.0.1`, on the only `##[error]` in the log:

```
##[error]Failed to CreateArtifact: Unable to make request: ECONNRESET
```

Timeline: `10:58:01.19 Artifact name is valid! / Root directory input is valid!` → 15.1 s of silence → `10:58:16.28 ECONNRESET`. No retry lines, because there is no retry.

## The cascade

| check-run | why it failed |
|---|---|
| `opt-in group (g19 sparq-mcp)` | the ECONNRESET above |
| `feature-matrix per-leg report` | `artifact completeness violation: 8 selected leg(s) have no result and were dropped` — exactly g19's 8 legs |
| `feature-matrix report` | same job, second check-run |
| `gate` | `ci-summary: FAILED (fail-fast) — 1 gating check(s) concluded failure`, listing only g19 |

The completeness guard behaved **correctly** — it refused to report green over a silently-missing shard. One defect, four red squares.

## Why this is a flake, measured

- The failing operation is a network call, not a compile or a test.
- **The same tree passed 25 minutes earlier**: `7b8c8252`'s `merge_group` feature-matrix run produced a green g19 (job 90248144888). Same commit, same 8 legs, different outcome.
- 1 of 54 jobs in the same run failed; every other group uploaded its `feature-matrix-results-N` fine.
- It has already self-cleared: on `fe2d7cfc` (next `main`) g19, `feature-matrix report` and `gate` are all green.

## The durable gap

[...]

Issue #4796, opened 2026-07-28
auto-arm's lenient transient exit has no floor: a sweep that enumerates nothing reports success, and nothing counts consecutive zero-consideration cycles


jeswr:
> 🤖 SPARQ agent

## The observation

During the GitHub GraphQL 5xx episode at ~2026-07-28T11:30–11:40Z, `auto-arm` run [30355432064](https://github.com/sparq-org/sparq/actions/runs/30355432064) reported **success** and its check-run was green. Its actual output:

```
[gh-retry] transient failure (attempt 1/3), retrying in 10s: gh pr list --repo: HTTP 502 …
[gh-retry] transient failure (attempt 2/3), retrying in 30s: gh pr list --repo: HTTP 502 …
[auto-arm] candidate enumeration exhausted transient retries (… 3 attempts …)
[auto-arm] complete: considered=0 armed=0 arm-failures=0 transient-exhaustions=0 capability=can-arm scope=all
##[warning]… bounded retries exhausted; a missed sweep is harmless (the cron backstop covers it), so this run reports success.
```

**`considered=0 armed=0`.** The sweep enumerated nothing, armed nothing, and reported success.

This is working as designed and I am not proposing to change the exit code — a missed cycle genuinely is cron-covered, and reddening `main` for a platform blip is the exact defect [#4795](https://github.com/sparq-org/sparq/issues/4795) exists to stop. The gap is that **the lenient path has no floor**.

## Why it matters

The leniency is justified by "the cron backstop covers it". That justification is a claim about the *next* cycle, and nothing verifies it. If an incident outlives the retry budget across many consecutive cycles, every cycle exits 0 with a green check while arming nothing, and the only evidence is a `::warning` buried in a run log nobody reads. Arming would be fully stopped and every dashboard would be green.

Note the completion line's own counters make this hard to see: `transient-exhaustions=0` is printed on a run whose candidate enumeration *did* exhaust — the counter tracks per-candidate exhaustions, and there were no candidates to count. So the one number that looks like it would catch this reads zero precisely when the whole sweep was lost.

## What is missing

Nothing counts **consecutive sweeps that considered zero candidates**. The system already emits everything needed; it just never aggregates across runs. Concretely:

- a sweep that exhausts before enumerating any candidate should be distinguishable from a sweep that enumerated candidates and found nothing eligible — today both print `considered=0`;
[...]

PR #4798, opened 2026-07-28
fix(ci): name the decoder and HTTP/2 layers in gh_retry's transient allow-list (#4795)

### LINKED ISSUE #4795: gh_retry transient allow-list misses the decoder and HTTP/2 layers — one platform incident reds main or warns depending on which string GitHub emits
> 🤖 SPARQ agent

## Symptom

`auto-arm` run [30355033945](https://github.com/sparq-org/sparq/actions/runs/30355033945) (job `arm reviewed PRs`, step `Arm eligible PRs`) exited 1 and put a red check-run on `main`'s commit. `scripts/gh_retry.py` classified the failure as non-transient:

```
[gh-retry] transient failure (attempt 1/3), retrying in 10s: gh pr list --repo: HTTP 502: 502 Bad Gateway (https://api.github.com/graphql)
gh_retry.GhFatalError: gh pr list --repo failed (non-transient): unexpected end of JSON input
```

Attempt 1 drew a 502 and was retried. Attempt 2 drew `unexpected end of JSON input`, which is not in `_TRANSIENT_MARKERS`, so the call went fatal and attempt 3 was never made.

The sibling scheduled run [30355432064](https://github.com/sparq-org/sparq/actions/runs/30355432064) five minutes later hit **the same incident on the same query**, drew `HTTP 502` three times, exhausted the bounded budget and took the documented `::warning` + exit-0 sweep path. **One incident, and the run's colour depended on which surface string the platform happened to emit.**

## Root cause — a layer gap, not a missing string

`gh` surfaces one physical event — "the response did not arrive intact" — at three layers. `_TRANSIENT_MARKERS` enumerated two:

| layer | markers present before this issue |
|---|---|
| HTTP status | `http 502` / `503` / `504` + prose spellings |
| transport | `connection reset`, `connection refused`, `i/o timeout`, `tls handshake timeout`, `context deadline exceeded`, DNS, `unexpected eof`, `request timed out` |
| **decoder** | **none** |

`unexpected eof` is the *transport* name for truncation; the *decoder* name (Go `encoding/json` on a truncated or empty body) was absent. That near-miss is the tell.

## Measured incidence

Census over 2026-07-21 → 2026-07-28, every non-skipped `auto-arm` + `re-arm sweeper` run (denominator **942 executed runs, 65 failed**). Of the failures that were `gh` READ failures, **8 (16.3%) carried a string the allow-list did not name**:

| string | runs | layer |
|---|---|---|
| `stream error: stream ID <n>; <CODE>; received from peer` | 6 | transport (HTTP/2 framing) |
| `unexpected end of JSON input` | 2 | decoder |

All **9** genuinely non-transient failures in the same window were self-describing permission/validation messages, every one on the **mutation** path that this helper never wraps. None degraded into a transport or decoder string — that is the evidence that admitting these two markers cannot swallow the auth/validation class.

Retry *amplifies* the gap rather than closing it: 29 of 34 post-helper transient episodes ran to a third attempt, so each degradation window now takes ~3 draws from the surface-string distribution instead of 1. That is exactly how run 30355033945 died.

## Blast radius (honest)

Small. `auto-arm` is on a 10-minute cron, the sweep is idempotent, and the next cycle recovers. The cost of each occurrence is a **red check-run on `main`'s commit** plus one missed arming cycle. The estate was not blocked: at the time of the incident only two open PRs carried `review:pass`, and both were accounted for.

## Fix

PR linked below: add the two measured markers (`unexpected end of json input`, and `stream error: stream id` keyed on the prefix because the stream number and error code both vary), organise the list by layer with the census recorded inline, and — so the next gap reports itself instead of costing another log excavation — emit a `::warning` when an **unclassified** failure lands immediately after a **classified** transient one on the same call. That annotation is observability only: unclassified still fails loudly, because "everything after a transient is transient" would swallow a genuine auth loss mid-incident.

## Known residual, deliberately recorded

`ENHANCE_YOUR_CALM` is genuine throttling rather than a blip, so a *sustained* throttle now exhausts into the sweep's `::warning` + exit-0 path rather than reddening. Bounded 10s/30s backoff is still the right response and the 3-attempt cap still applies. This is the same trade every marker in the list already makes, but it is a real one and worth knowing about.

**Separately and more important:** when every attempt exhausts, the sweep logs `considered=0 armed=0` and reports **success**. That is by design (a missed cycle is cron-covered) but it means a long platform incident produces green runs that arm nothing, and nothing counts consecutive zero-consideration sweeps. Filed separately as the follow-up below.

jeswr:
> 🤖 SPARQ agent

Fixes #4795.

## What broke

`auto-arm` run [30355033945](https://github.com/sparq-org/sparq/actions/runs/30355033945) exited 1 and put a red check-run on `main`'s commit:

```
[gh-retry] transient failure (attempt 1/3), retrying in 10s: gh pr list --repo: HTTP 502: 502 Bad Gateway (https://api.github.com/graphql)
gh_retry.GhFatalError: gh pr list --repo failed (non-transient): unexpected end of JSON input
```

Attempt 1 drew a 502 and was retried. Attempt 2 drew `unexpected end of JSON input`, which `_TRANSIENT_MARKERS` did not name, so the call went fatal and **attempt 3 was never made**.

The sibling scheduled run [30355432064](https://github.com/sparq-org/sparq/actions/runs/30355432064) five minutes later hit the same incident on the same query, drew `HTTP 502` three times, exhausted the bounded budget, and took the documented `::warning` + exit-0 sweep path. **One incident, and the run's colour depended on which surface string GitHub happened to emit on the second attempt.**

## Root cause: a layer gap, not a missing string

`gh` surfaces one physical event — the response did not arrive intact — at three layers. The list enumerated two.

| layer | markers before | markers after |
|---|---|---|
| HTTP status | `http 502/503/504` + prose spellings | unchanged |
| transport | `connection reset`, `connection refused`, `i/o timeout`, `tls handshake timeout`, `context deadline exceeded`, DNS, `unexpected eof`, `request timed out` | **+ `stream error: stream id`** |
| decoder | *(none)* | **+ `unexpected end of json input`** |

`unexpected eof` is the *transport* name for truncation and its *decoder* name was missing. That near-miss is the tell: the next gap to look for is a layer nobody has named, not a string nobody has typed.

## Evidence

Census over 2026-07-21 → 2026-07-28, every non-skipped `auto-arm` + `re-arm sweeper` run — denominator **942 executed runs, 65 failed**. Of the failures that were `gh` READ failures, **8 (16.3%) carried a string the allow-list did not name**:

| string | runs |
|---|---|
| `stream error: stream ID <n>; <CODE>; received from peer` | 6 |
| `unexpected end of JSON input` | 2 |

All **9** genuinely non-transient failures in the same window were self-describing permission/validation messages, every one on the **mutation** path this helper never wraps. **None degraded into a transport or decoder string** — that is the evidence these two markers cannot swallow the auth/validation class.

Retry amplifies the gap: 29 of 34 post-helper transient episodes ran to a third attempt, so each degradation window takes ~3 draws from the surface-string distribution instead of 1.

## The changes

1. **Two measured markers.** `stream error: stream id` is keyed on the **prefix** — both the stream number and the error code vary (`CANCEL` / `INTERNAL_ERROR` / `REFUSED_STREAM` / `PROTOCOL_ERROR` / `ENHANCE_YOUR_CALM` / `NO_ERROR`); pinning `; cancel;` would just be the next mole.
2. **The census and the layer taxonomy recorded inline**, so the next maintainer asks which layer is unnamed rather than appending another string.
3. **A `::warning` when an unclassified failure lands immediately after a classified transient one on the same call**, so the next gap reports itself instead of costing another log excavation. **Observability only** — unclassified still raises `GhFatalError` on the very next line. "Everything after a transient is transient" would swallow a genuine auth loss mid-incident, which is the one class this helper exists to keep loud, and there is a test that fails if anyone makes that change (mutant A5).

## Security review of this surface

Arming is security-adjacent, so, explicitly:

- This changes only `run_gh_read`, whose **first statement** is the fail-closed `assert_read_only(argv)`. Mutations are refused before `gh` is ever invoked, so nothing here can arm anything, let alone something a review did not authorise. Mutants **P4** (remove the call) and **P5** (neuter the allow-list) are both killed by `['pr','merge','1','--auto']` still being refused.
- The behavioural delta is strictly: a read that previously died fatally now gets up to 2 more attempts and, on exhaustion, raises `GhTransientExhausted`. In **event** mode that is still a non-zero exit; only the cron-covered **sweep** takes the lenient path.
- No workflow `if:` was added or changed, so the "a step `if:` with no status function defaults to `success()`" hole does not arise here.

## The YAML seam — a hole found while verifying the above

`python3 scripts/gh_retry.py --self-test` ran in a gating `docs-quality` job, but was the **only** one of the three arm-policy scripts whose PR-time self-test was **not pinned** by `TestSelfTestsRunInCi` — and it appeared **exactly once** in the file, where its two siblings appear twice. Deleting that single line would have left every change to the transient classifier validated by nothing until the next cron, with no test going red.

The pin now matches a complete **stripped command line** rather than a substring of the concatenated run blob. That is deliberate: a containment check passes against `--self-test-DISABLED` and `--self-test || true`, and both of those are now killed (mutants Y2, Y3).

## Verification

**29 mutants across two rounds, 26 killed.** Both mutants per guard (delete **and** conditionally-inert non-crashing), plus the call sites, plus the YAML seam.

| group | mutants | result |
|---|---|---|
| classifier + call site + bound (round 1, M1–M9) | 9 | 9 killed |
| new markers (P1, P2, P6–P9) | 6 | 6 killed |
| attempt budget / security (P3–P5) | 3 | 3 killed |
| YAML seam (Y1–Y5) | 5 | 3 killed, 2 equivalent |
| the pin itself (T1) | 1 | equivalent |
| mid-incident annotation (A1–A5) | 5 | 5 killed |

Which mutant covers the guard added here, explicitly: **M1/M2** cover the decoder marker (delete + inert), **P6/P7** the HTTP/2 marker, **P8** that it is not over-narrowed to the one code observed, **P2/P9** that it is not over-broadened, **A1–A4** the annotation, **A5** that the annotation was not quietly turned into a reclassification.

**The 3 survivors are equivalent mutants, and the equivalence was measured, not argued.** Y4/Y5 delete one of **two** duplicated sibling self-test invocations, so the pinned property remains true; deleting **both** occurrences **is** killed — run explicitly, both `FAIL = KILLED`. T1 drops a needle from the pin list, which no test can catch without infinite regress; every pre-existing needle in that list shares the property.

Two of my own tests were **found vacuous by this process and fixed** rather than reported as passes: the first catalogue let a bare `"json"` marker survive (nothing asserted that a deterministic `unknown JSON field` usage error stays fatal), and the first 403-mid-episode assertion was never reached because an earlier fixture ran out of scripted outcomes first.

**Instrument validation.** Every run used `-B` / `PYTHONDONTWRITEBYTECODE=1`. Restores come from the git object store and are marker-verified with `git status --porcelain` asserted empty between mutants. Both instruments were canary-validated against a **known positive** before use — a deliberately inverted assertion for the Python suite, and a deliberately broken YAML line for the wiring suite — because an instrument that fails toward "nothing to report" reports a clean sweep.

**Escape check.** All four suites run behind a `gh` PATH shim that logs and exits 97. The shim's reachability was proved first with a canary invocation (rc=97, 1 log line), the log was then truncated, and after the full run the escape log is **empty** — 0 `gh` calls, 0 `git` calls. Fixtures name `sparq-org/sparq` only as an inert argv string and the canary used a repo id that cannot exist.

## Known residual, deliberately recorded

`ENHANCE_YOUR_CALM` is genuine **throttling**, not a blip, so a *sustained* throttle now exhausts into the sweep's `::warning` + exit-0 path rather than reddening. Bounded 10s/30s backoff is still the correct response and the 3-attempt cap still applies. This is the same trade every marker in the list already makes, but it is real and worth knowing.

Separately, the lenient exit has **no floor across runs** — a sweep that enumerates nothing reports success, and nothing counts consecutive zero-consideration cycles. That is out of scope here and filed as **#4796**.

## Follow-ups filed

- **#4796** — the lenient transient exit has no floor; `considered=0 armed=0` reports success and no counter aggregates across runs.
- **#4797** — `feature-matrix` `upload-artifact` has no retry, so one ECONNRESET reds the required `gate` on a fully-green tree. This was the actual cause of `main` being red at `7b8c8252`; all 8 g19 legs passed and the job died on a 468-byte upload.

### DIFF
--- a/scripts/gh_retry.py
+++ b/scripts/gh_retry.py
@@ -74,6 +74,79 @@ _TRANSIENT_MARKERS = (
     "temporary failure in name resolution",
     "unexpected eof",
     "request timed out",
+    # [OPUS-5] sparq #4795. `gh` decodes every API response with Go's encoding/json, so a
+    # TRUNCATED OR EMPTY response body surfaces not as an HTTP status but as the decoder's
+    # own text. It is the same transport class as "unexpected eof" one line up — the body
+    # ended early — and it is NOT reachable from an auth/validation refusal, because 401 /
+    # 403 / 404 / 422 all return a well-formed JSON error document that gh reports as
+    # "HTTP <code>: …" instead. MEASURED, run 30355033945 job 90261152943 (auto-arm,
+    # 2026-07-28T11:30Z), during a GitHub GraphQL 5xx episode:
+    #     attempt 1/3: HTTP 502: 502 Bad Gateway (https://api.github.com/graphql)   -> retried
+    #     attempt 2/3: unexpected end of JSON input                                 -> FATAL
+    # One incident, one query, and the run's COLOUR depended on which surface string the
+    # platform happened to emit on the second attempt: the sibling scheduled run five
+    # minutes later (30355432064) drew "HTTP 502" three times, exhausted the bounded budget
+    # and took the documented ::warning + exit-0 sweep path. Without this marker the third
+    # attempt is never made and a cron-covered sweep reds main's commit instead.
+    "unexpected end of json input",
+    # [OPUS-5] sparq #4795. HTTP/2 RST_STREAM from the peer — Go's net/http2 wording. The
+    # most frequent uncovered string in the census below, and the same physical event as
+    # "connection reset" above, just named at the h2 framing layer instead of the socket
+    # layer. Keyed on the PREFIX because both the stream number and the error code vary
+    # (CANCEL / INTERNAL_ERROR / REFUSED_STREAM / PROTOCOL_ERROR / ENHANCE_YOUR_CALM);
+    # pinning "; cancel;" would just be the next mole. Not reachable from an auth or
+    # validation decision — those never reach the h2 error path at all.
+    #   CAVEAT, recorded deliberately: ENHANCE_YOUR_CALM is genuine THROTTLING, not a blip.
+    #   Bounded 10s/30s backoff is still the right response and the 3-attempt cap still
+    #   applies, but a SUSTAINED throttle now exhausts into the sweep's ::warning + exit-0
+    #   path rather than reddening. That is the same trade every marker here makes.
+    "stream error: stream id",
+)
+
+# ------------------------------------------------------------------- census (#4795)
+# Why this list is organised by LAYER, and what the evidence for the two 2026-07-28
+# additions is.
+#
+# MEASURED over 2026-07-21 -> 2026-07-28 across every non-skipped auto-arm + re-arm-sweeper
+# run (denominator 942 executed runs, 65 failed). Of the failures that were `gh` READ
+# failures, 8 (16.3%) carried a string this allow-list did not name:
+#
+#     stream error: stream ID <n>; <CODE>; received from peer      6 runs
+#     unexpected end of JSON input                                 2 runs
+#
+# All 9 genuinely non-transient failures in the same window were SELF-DESCRIBING — an
+# explicit permission or validation message, every one of them on the MUTATION path, which
+# this helper never wraps. None degraded into a transport or decoder string. That is the
+# evidence that admitting these two markers cannot swallow the auth/validation class.
+#
+# THE STRUCTURAL POINT, which matters more than either string: `gh` surfaces ONE physical
+# event — "the response did not arrive intact" — at THREE layers, and this list had only
+# ever enumerated two of them:
+#
+#     HTTP status layer   http 502 / 503 / 504 and their prose spellings
+#     transport layer     connection reset/refused, i/o timeout, tls handshake, unexpected
+#                         eof, context deadline, DNS, request timed out, and now
+#                         `stream error: stream id` (HTTP/2 framing)
+#     decoder layer       `unexpected end of json input` (Go encoding/json on a truncated
+#                         or empty body) — the layer that was entirely absent
+#
+# The tell was already sitting in the list: "unexpected eof" is the TRANSPORT name for
+# truncation and its DECODER name was missing. So the next gap to look for is a LAYER
+# nobody has named, not a string nobody has typed. Retry makes this worse rather than
+# better: 29 of 34 post-helper transient episodes ran to a third attempt, so each
+# degradation window now takes ~3 draws from the surface-string distribution instead of 1
+# — which is exactly how run 30355033945 died (502 on attempt 1, truncation on attempt 2).
+#
+# :func:`run_gh_read` therefore ANNOTATES an unclassified failure that lands mid-incident,
+# so the next gap reports itself in the run log instead of costing another log excavation.
+UNCLASSIFIED_MID_INCIDENT_WARNING = (
+    "::warning title=gh-retry hit an UNCLASSIFIED failure mid-incident::{label} attempt "
+    "{attempt}/{attempts} failed with text that is NOT in the transient allow-list, "
+    "immediately after a CLASSIFIED transient failure on the SAME call — the sparq #4795 "
+    "shape, one platform incident surfacing at a layer this list does not name. The call "
+    "was failed loudly (unclassified stays fatal). If the text below names a transport or "
+    "decoder failure rather than an auth/validation refusal, it belongs in "
+    "_TRANSIENT_MARKERS: {detail!r}"
 )
 
 # gh subcommand pairs this helper accepts as reads (fail-closed: everything not
@@ -264,6 +337,10 @@ def run_gh_read(
     emit = log or (lambda message: print(message, file=sys.stderr))
     command = ["gh", *[str(arg) for arg in argv]]
     label = " ".join(command[:4])
+    # [OPUS-5] #4795: has THIS call already seen a failure the allow-list DID name? If so,
+    # a later unclassified failure is evidence of an allow-list gap rather than of a real
+    # error, and it gets annotated (never reclassified — see below).
+    saw_classified_transient = False
     for attempt in range(1, attempts + 1):
         result = run(command, capture_output=True, text=True, check=False)
         if result.returncode == 0:
@@ -274,7 +351,19 @@ def run_gh_read(
             or "unknown gh failure"
         )
         if not is_transient(detail):
+            # [OPUS-5] #4795: OBSERVABILITY ONLY — the classification is unchanged and the
+            # call still fails loudly on the very next line. Deliberately NOT "everything
+            # after a transient is transient": that would swallow a genuine auth loss
+            # mid-incident, which is the one class this helper exists to keep loud.
+            if saw_classified_transient:
+                emit(
+                    UNCLASSIFIED_MID_INCIDENT_WARNING.format(
+                        label=label, attempt=attempt, attempts=attempts, detail=detail
+                    )
+                )
             raise GhFatalError(f"{label} failed (non-transient): {detail}")
+        # Reached only when the allow-list DID name this failure.
+        saw_classified_transient = True
         if attempt == attempts:
             raise GhTransientExhausted(
                 f"{label} failed transiently on every attempt "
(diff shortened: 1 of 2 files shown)


jeswr:
> 🤖 SPARQ agent

**Independent review of `97b6e73951facb7435174d2431eaad0e70407637`.** Reviewer only — this head carries no commits of mine, and I have not armed, merged, or relabelled. Everything below was **executed** in an isolated worktree at that SHA, behind a `gh`/`git`/`curl` PATH shim, with `-B` / `PYTHONDONTWRITEBYTECODE=1`, restores from the git object store, and `git status --porcelain` asserted empty between mutants.

This is a **broadening of a retry predicate**, so the whole review is organised around one question: *can these markers swallow something that must stay fatal?*

---

## 1. Can a newly-added marker match a genuine failure? — **no**, measured

I built a 39-string corpus of real `gh`/GitHub wordings across five must-stay-fatal classes (auth, permission, 404, rate-limit, malformed-request) and asked which markers match.

| result | count |
|---|---|
| genuine strings matched by a **newly added** marker | **0 / 39** |
| genuine strings matched by any **pre-existing** marker | 0 / 39 |

**The instrument was validated in both directions before I trusted that zero**, because a broken probe and a clean result look identical:

- *known positives* — the 8 strings the markers are supposed to catch (`unexpected end of JSON input` in both casings, and `stream error: stream ID <n>; <CODE>;` for all six h2 codes): **8/8 classified transient**;
- *swallowing canary* — re-running the same corpus with a deliberately over-broad marker `"json"` appended shows **7 genuine strings would be swallowed**. So the probe can detect swallowing, and 0/39 is a real zero.

**One honest exception, which the PR already discloses.** `stream error: stream ID 3; ENHANCE_YOUR_CALM; received from peer` **is** matched, and `ENHANCE_YOUR_CALM` is genuine h2 throttling rather than a blip. Measured behaviour (S6 below): three attempts, 10s/30s backoff, then `GhTransientExhausted` — loud in event mode, `::warning` + exit-0 only in the cron-covered sweep. Bounded backoff is the correct response to throttling and the inline comment records the trade. Not a defect.

## 2. Is a **truncated genuine** failure acceptably handled? — **yes**, and it is not a new class

The scenario worth attacking: a 403 whose body is cut mid-stream presents as `unexpected end of JSON input`, which this PR now retries. Executed:

| scenario | attempts | outcome | names the 403? | annotation |
|---|---|---|---|---|
| S1 403 truncated once, then intact | 2 | `GhFatalError: … HTTP 403: Resource not accessible by integration` | **yes** | 1 |
| S2 403 truncated twice, then intact | 3 | `GhFatalError: … HTTP 403: …` | **yes** | 1 |
| S3 403 truncated on **all three** attempts | 3 | `GhTransientExhausted: … unexpected end of JSON input` | no | 0 |
| S3b same shape via **pre-existing** `unexpected eof` | 3 | `GhTransientExhausted: … unexpected EOF` | no | 0 |
| S5 502 then genuine 403 mid-incident | 2 | `GhFatalError: … HTTP 403: …` | **yes** | 1 |
| S7 403 on attempt 1, no prior transient | 1 | `GhFatalError: … HTTP 403: …` | **yes** | 0 (correct) |

The real cause is **not** hidden. A deterministic refusal re-presents itself intact on the next draw, so it surfaces as `GhFatalError` naming the 403 — and the new annotation fires alongside it, which makes this case *better* diagnosed than before the PR.

The only case where the final message does not name the 403 is **S3**, where the refusal genuinely never arrived intact in three attempts. **S3b is the control that settles whether that is a new hole: it is not** — the pre-existing `unexpected eof` marker admits the byte-identical shape today. This PR inherits that residual rather than creating it.

## 3. Can the markers reach the mutation path? — **no**, proven by execution

22 mutation shapes (every `self.gh([...])` shape the repo actually issues — `pr merge --auto`, `pr ready`, `pr comment` — plus `pr edit/review`, `issue edit/comment/create`, `label create`, `api -X POST/PUT`, field-param auto-POST, inline GraphQL mutation, `-F query=@file`, `--input`, bare `graphql`, empty argv, and two with marker text smuggled into argv), each run with a recording runner whose stderr was **poisoned with both new markers**:

- **22/22 refused, `gh` invoked 0 times.** The markers are never consulted.
- **Ordering proved, not read**: `run_gh_read(["pr","merge","1"], attempts=0)` raises `GhRetryUsageError`, not the `attempts must be >= 1` error — so `assert_read_only` genuinely runs first and unconditionally.
- **Control**: the same probe on a real read (`pr list --json number`) invokes `gh` 3 times, so "0 reached" is not the probe refusing everything.
- Mutants **P4** (delete the `assert_read_only` call) and **P5** (`return` before any check) are both **KILLED**, cause recorded: `AssertionError: must refuse to wrap mutation shape: ['pr','merge','1','--auto']`.

Structurally consistent: all four consumers (`auto-arm.py`, `rearm-sweeper.py`, `merge-group-watchdog.py`, `verdict-bridge.py`) route reads through `gh_retry.run_gh_read` and mutations through a separate one-shot `run_gh`.

## 4. The two YAML-seam mutants, re-derived — and the kill is a **source match**

| mutant | verdict | kill cause |
|---|---|---|
| Y1 delete the line | **KILLED** | `AssertionError: 'python3 scripts/gh_retry.py --self-test' not found in {…}` |
| Y2 `--self-test-DISABLED` | **KILLED** | same, `not found in {…}` |
| Y3 `--self-test \|\| true` | **KILLED** | same, `not found in {…}` |
| Y3b `--self-test  # disabled` | **KILLED** | same, `not found in {…}` |

The kill criterion is **set membership of stripped source lines**, never the mutated command's exit code — so the `|| true` self-masking failure mode (harness judging kills by the same rc the mutant suppresses) **does not apply here**. Confirmed by reading the assertion text, not the rc.

## 5. The three claimed-equivalent survivors — scrutinised

- **Y4 / Y5 (delete one of two duplicated sibling invocations) — survive, and the justification checks out.** Both duplicate sites are inside the gating `quick-gates` job (the second job, `quick-advisory`, starts well below them), which is exactly why deleting one leaves the pinned property true. The author's stronger claim reproduces: **Y4b** (both `rearm-sweeper` copies) and **Y5b** (both `auto-arm` copies) are **KILLED**. Calling them "equivalent" is loose — the workflow really does run the self-test once instead of twice — but the *pinned property* is unchanged, and the property is the right one.
- **T1 (drop a needle from the pin list) — survives, but "no test can catch this without infinite regress" is mildly overstated.** Measured: 70 `scripts/*.py` accept `--self-test`, but only **4** non-test scripts consume `gh_retry` (`auto-arm`, `gh_retry`, `merge-group-watchdog`, `rearm-sweeper`). A needle set **derived from the filesystem** — "every script that imports `gh_retry` must appear as a complete gating command line" — is therefore feasible and would kill T1 without regress. Worth a follow-up; not a reason to hold this PR, since every pre-existing needle has the identical exposure.

## 6. Which mutant covers each guard? — the catalogue, re-derived

**30 mutants run** (3 instrument canaries + 18 Python + 12 YAML). Every guard this PR adds has both a *delete* and a *conditionally-inert, non-crashing* mutant, and each kill cause was recorded to distinguish a guard kill from a `SyntaxError`.

| guard added | delete | inert | other | result |
|---|---|---|---|---|
| decoder marker | M1 | M2 | X1 (cased so the fold cannot match), P2 (widened to `"json"`) | all **KILLED** |
| h2 marker | P6 | P7 | P8 (narrowed to the one observed code) | all **KILLED** |
| attempt budget | — | — | P3 (3→2) | **KILLED** |
| read-only guard | P4 | P5 | — | both **KILLED** |
| annotation fires | A1 | A2 | — | both **KILLED** |
| annotation conditioned | A3 | A4 | — | both **KILLED** |
| annotation ≠ reclassification | — | — | A5 | **KILLED** (`a 403 mid-episode must still raise GhFatalError`) |
| annotation names the detail | — | — | X2 (`{detail!r}` dropped) | **KILLED** |
| the new pin | Y1 | Y2 / Y3 / Y3b | Y8 (job renamed advisory → fail-closed) | all **KILLED** |

Instrument canaries first, against known answers: inverting a pre-existing assertion → **KILLED by an assertion** (not a parse error); editing a pure comment → **SURVIVED**; breaking a pre-existing sibling pin → **KILLED by source match**. The YAML canary also caught a flaw in my own harness (it initially destroyed the `run:` key and "killed" via a `yaml.ScannerError`) — fixed before any result was recorded.

## 7. Two findings, neither blocking — please fold in

**(a) The h2 marker has no over-broadening coverage at any granularity.** The self-test's negative corpus is 8 strings, and a widening is killed only if it matches one of them. Mapped exhaustively (map validated against a known-caught widening):

- decoder marker: `"json"` → **KILLED** by the three `JSON field` negatives ✅
- h2 marker: `"stream error: stream"`, `"stream error"`, `"stream"`, even bare `"error"` → **all SURVIVE**. No negative assertion contains the token `stream` or `error`.

So "**P2/P9** that it is not over-broadened" holds for the decoder half and **not** for the h2 half. The shipped marker is narrow and measured-safe (0/39), and every *pre-existing* marker has the same property — this is an inherited asymmetry, not a regression. One negative assertion, e.g. `assert not is_transient("GraphQL: upstream request timeout for stream 3")`, would close it.

**(b) Two body claims are wider than the census supports** (I re-derived the census independently; **C1 942/65 reproduces exactly** for a census cut between 11:38Z–11:54Z, **C2 8/49 = 16.3% reproduces exactly**, and **C4 is verbatim in both cited run logs**, including the absent `attempt 2/3` line):

1. *"All **9 genuinely non-transient** failures in the same window…"* — the 9 permission/validation failures are right, all on unwrapped mutations, and "none degraded into a transport or decoder string" is **directly confirmed, 0 transient markers in all 9 logs**. But 2 further failures (`ModuleNotFoundError: No module named 'gh_retry'`, runs `30143001444` / `30143852994`) are also deterministic, so the non-transient population is **≥11**, not 9. The sentence should read "all 9 **permission/validation** failures". The safety conclusion is untouched — those 2 die at import, before `run_gh_read` exists.
2. *"both the stream number and the error code vary (CANCEL / INTERNAL_ERROR / …)"* reads as a census result, but **all 6 observed instances are byte-identical** (`stream error: stream ID 1; CANCEL; received from peer`). Prefix-keying is still the right call — it follows from the HTTP/2 spec — but the rationale is spec-derived, not measured, and the census is presented as the evidence base.

## 8. Also checked

- **Not covered by the pin** (pre-existing, *not* introduced here — the diff touches no workflow): `run_of(step)` reads only `step["run"]`, so **Y6** (`if: false` on the step), **Y7** (`continue-on-error: true`) and **Y6b** (`if: false` on the whole gating job) all **SURVIVE** with the command text unchanged. Both sibling pins have the identical exposure. Worth a follow-up bead; not this PR's to fix.
- **Composition, re-done after `main` moved mid-review.** `main` advanced `fe2d7cfc` → `f3e737ab` (#4705, the actions bump) while I was measuring, and #4705 **touches `docs-quality.yml`** — the file this PR's new pin reads. I merged current `main` into the head and re-ran on that composition: the bump is SHA-pin-only (`taiki-e/install-action`, 1:1 line replacements), no `run:` line and no line count changed, and **all four suites pass on the merge**: `gh_retry --self-test`, `auto-arm --self-test`, `rearm-sweeper --self-test`, `test_arm_capability_wiring.py` (38 tests, unchanged count base→head; the renamed test has no other references). Re-running the headline mutants **on that composition** — Y1, A5, P4, M1 — all still **KILLED**, so nothing pre-existing was hollowed out by the merge.
- **`gate` by SHA equality** on `commits/97b6e739…/check-runs` (never the PR rollup): exactly one `gate`, `success`, 12:02:34→12:10:14Z; **no `draft-tier`**; 105 check-runs, 76 skipped + 29 success, **0 non-completed and 0 non-success**. The 48 runs that started after `gate` concluded are all skipped/success, so nothing red sits outside its aggregation. Check-suite `head_sha` equals `97b6e739…`, and `ls-remote` confirms the head has not moved.
- **Escape log empty** — after canary-proving the shim reachable (`rc=97`, 2 logged lines, using a repo id that cannot exist), the log was truncated and every suite and all 30 mutants ran behind it: **0 `gh`, 0 `git`, 0 `curl` escapes**.

## Assessment

The behavioural delta is two narrow, measured markers plus an **observability-only** annotation that provably does not reclassify (A5 killed), on a path whose first statement is a fail-closed read-only guard that no mutation shape gets past. No fail-open predicate, no exit-zero swallow, no arming reachability. The findings in §7 are a body-wording correction and one missing negative assertion — both worth doing, neither a reason to keep a real incident fix out of `main`.

Reviewed at head `97b6e73951facb7435174d2431eaad0e70407637`.

VERDICT: pass

Issue #4803, opened 2026-07-28
Critical lanes are cron-dependent; GitHub's scheduler degrades silently — add doorbells with cron as backstop


jeswr:
> 🤖 **SPARQ agent** — filed automatically from a live measurement.

## Measurement

During a live GitHub scheduler degradation on 2026-07-28, over the same 21-minute window:

| repo | trigger design | runs in window |
|---|---|---|
| `jeswr/agent-account-registry` | **`workflow_dispatch` doorbell**, cron as backstop | **8** — unaffected |
| `sparq-org/sparq` | cron-dependent | **0 scheduled runs** — stalled |

A newly-merged sparq watchdog carrying a valid 5-minute cron had **1 run in 68 minutes** — and that one was the
manual dispatch proving its mechanism, not a scheduled fire. Earlier in the same degradation, **exactly one
scheduled run fired across all of sparq in 18 minutes** despite many active crons.

This is not a sparq misconfiguration. The crons are valid and the workflows are enabled. GitHub's scheduler is
best-effort and degrades under load — sparq is simply *exposed* to that in a way the registry is not.

## The registry already solved this, in-band

`dispatch.yml` in the registry carries the reasoning inline:

> Explicit minutes (not `*/10`): GitHub degrades wildcard schedules under load — schedules are throttled less
> aggressively, and the off-peak offset (`:03` not `:00`) avoids the top-of-minute stampede. **The cron is now
> the BACKSTOP; the doorbell below is the primary.**

The doorbell is an explicit `workflow_dispatch` fired by the *producing* workflow — `worker.yml` and
`review-fix.yml` each end with a doorbell job. Work triggers the next stage directly instead of waiting to be
discovered by a poll.

## Why this matters beyond one outage

⚠️ **A cron-triggered lane has no back-pressure signal.** When the scheduler is slow the lane just does less,
**with nothing red anywhere** — each workflow merely has fewer runs, and *the absence of a run is invisible to
every check that inspects runs which happened*. Latency degrades silently and uniformly across every
cron-driven lane at once.

A doorbell-triggered lane degrades only if its **producer** stops, which is an observable event with an owner.

## Proposed work

1. **Inventory** sparq's cron-driven lanes and mark which ones have latency that matters (arming, re-arming,
   grooming, queue watchdogs, dead-group detection).
2. For each, **identify the producing event** and fire a `workflow_dispatch` doorbell from it. Keep the cron as
   the backstop — do not remove it.
[...]

PR #4845, opened 2026-07-28
fix(ci): ring the merge-group watchdog from queue events, not only cron

jeswr:
> 🤖 **SPARQ agent** — CI/infra lane, running as **Opus 5**. I am the AUTHOR of this
> change, so there is deliberately no `VERDICT:` line here and I have not armed it.

## What broke, measured

sparq's merge queue was head-of-line blocked for **46m33s** today. Position 1 (#4810,
group head `38a9d875`) had **zero check-suites**; positions 2 and 3 had 8 suites each and
could not land behind it. A human cleared it with `dequeuePullRequest`.

`.github/workflows/merge-group-watchdog.yml` **already detects exactly this** and already
holds `checks: read` for the load-bearing predicate. Nothing was wrong with its decision.
The defect was purely **when it runs** — `schedule` + `workflow_dispatch` only:

| | measured |
|---|---|
| watchdog cron fires, 13:19:03Z → 17:07:52Z | **3** (of ~45 expected — 93% loss) |
| largest gap | **140 min** |
| dead ref `38a9d875` built | 15:58:00Z |
| next scheduled fire after it | 17:07:52Z — **+69.9 min**, past the ruleset's own 60-min `CI_TIMEOUT` |

This is GitHub best-effort scheduler behaviour, not a misconfiguration. Recurrence today
was 3 dead refs in ~11 h, so it would have happened again unattended.

## The masking hypothesis — VERIFIED, not assumed

The obvious doorbell is the `merge_group` event. It is disqualified **twice over**, and I
checked both halves against live API data rather than reasoning about them.

**(a) Subscribing would mask the detector — CONFIRMED.** A check-suite is created **1:1 per
dispatched workflow run**. On group head `6b291b28` there are exactly 8 `merge_group`
workflow runs (`ci`, `ci-summary`, `docs-quality`, `feature-matrix`, `flow-on-gates`,
`pr-area-label`, `routing-self-tests`, `vectorized-feature-off`) and exactly 8 check-suites;
same 8/8 on `37aac5a4`, `e5a9d403`, `bef48bee`. A tenth subscriber therefore puts a **ninth
suite on every group head**, and the predicate is `total_count == 0`. The masking would be
**total, not partial** — the detector could never fire again, on any group.

**(b) The zero-suite case is one where the event was never delivered — CONFIRMED.** All nine
`merge_group` subscribers here carry a **bare** trigger (no `types`/`paths`/`branches`), so
no filter configuration can suppress some-but-not-all. Over the **42 distinct group refs**
retained in the activity feed the per-head `merge_group` run count is perfectly bimodal:

```
{0 runs: 3 heads, 8 runs: 39 heads}   — no intermediate value, ever
```

Zero therefore means the event reached **no** workflow. A `merge_group`-triggered watchdog is
dispatched by precisely the delivery that failed, so it would be absent exactly when needed.

*(One honest caveat: the script header's separate claim that "a `paths`-filtered workflow
matching nothing still creates a check-suite" is **not** verified by this data — no
`merge_group` subscriber here carries a `paths` filter, so the case does not arise. I have
left the predicate untouched; it is correct for this configuration either way.)*

## The producer chosen, and why it is complete

`pull_request_target: types: [enqueued, dequeued]`.

**`enqueued` does exist** — the brief's premise that it does not is refuted. It is a
documented `pull_request`/`pull_request_target` activity type, listed alongside `dequeued`.

A merge-group ref is built when an entry enters the build window, which has exactly two
causes: the entry is **added** to the queue, or an entry **ahead of it leaves** (merged or
removed) and the rest rebuild. That is `enqueued` + `dequeued` — a **complete cover**, with
one deliberate exception: a direct push to `main` that rebases the queue without any
PR-level queue event. **The retained cron covers that**, which is a reason to keep it rather
than a gap.

Measured, enqueue → dead-group-build delay for all three of today's dead refs:

| PR | enqueued | dead group built | delay |
|---|---|---|---|
| #4810 | 15:58:00Z | `38a9d875` 15:58:00Z | **0 s** |
| #4773 | 09:58:54Z | `76153a84` 09:59:10Z | **16 s** |
| #4709 | 05:15:07Z | `98204656` 05:15:09Z | **2 s** |

Delivery latency for this trigger class is **already proven live in this repo**: the
`dequeued` half is what `merge-queue-feedback.yml` runs on, and against the queue's own
removal events it fires in **3–6 s (n=8, all 8 matched)**.

### Why I did *not* piggyback on `auto-arm.yml`

It was the brief's first-listed candidate and it is measurably worse. `labeled` fired **223
times in the same 4 h** against **19 queue transitions** — 12× the cost for a signal that
does not cause a group build at all, only correlates with one. Counterfactually it would
have detected `38a9d875` at 16:15:44Z (**+17.7 min**) versus **~2.5 min** for the doorbell.
It is a clock, not a producer, and the enqueued+dequeued pair being *complete* makes a noisy
clock unnecessary.

I also rejected `push: branches: [main]`: a merged group head **becomes** the main head, so a
`push` leg would put a check-suite on a commit that is also a former merge-group head. It
muddies the constraint-1 invariant for no measured benefit, and it goes silent exactly during
a block (nothing is merging), which is the anti-correlation that makes queue-progress events
a bad clock.

## Constraint compliance

1. **Does not mask its own condition.** `pull_request_target` runs attach their check-run to
   the **PR's head SHA**, never to the `gh-readonly-queue/...` group commit. Verified
   empirically: PR #4810's head `fef71f12` carries `merge-queue dequeued feedback (advisory)`
   among its 204 check-runs, while the four healthy group heads carry exactly the 8
   `merge_group` suites and nothing else. The doorbell adds **no suite to any group head**.
   A repo-wide test enforces this going forward.
2. **Cron retained** — unchanged, and its header now says explicitly that it is the
   independent backstop and why. A test fails if the schedule is deleted.
3. **Decision untouched.** The doorbell carries no predicate, no threshold and no dequeue.
   Grace, caps, marker trust-list, `ACTIONABLE_STATES` and the recover/refuse table are all
   byte-identical. **Only timing changed.**
4. **Cost bounded.** The doorbell is a **one-step, one-API-write** job (`timeout-minutes: 5`),
   ~19 fires per 4 h. It is not on any workflow's critical path — it is its own workflow.

## What changed

- **`.github/workflows/merge-group-doorbell.yml`** (new) — the trigger. No checkout at all,
  no repository script executed, no `github.event` field interpolated into any executable
  line; token holds **`actions: write` and nothing else**. It dispatches the watchdog pinned
  to the **default branch**, so the policy that decides a dequeue is always the reviewed copy
  on `main`.
- **`.github/workflows/merge-group-watchdog.yml`** — the single sweep becomes an
  **unconditional** 5-sweep / 60 s loop, so one fire outlives the 120 s grace (a ring lands
  0–19 s *before* the group exists). Unconditional on purpose: a per-trigger `if:` would be a
  YAML seam, which is where mutants survive on this estate. `timeout-minutes` 10 → 15.
- **`.github/advisory-registry.json`** — registers the doorbell job. **Correction worth
  flagging:** since #3773 the gate no longer infers advisory status from job *names*; the
  registry is the only mechanism. Undeclared, this leg would have **silently gated** every
  enqueue.
- **`scripts/tests/test_merge_group_watchdog.py`** — +36 tests.

**⚠️ RETRACTED — an earlier revision of this PR claimed the merge-base call site had a
pre-existing bug. It does not.** I claimed unquoted `$DRY_RUN` injected an empty argv element
when unset, "killing every sweep in the default configuration". That is **false**, and it was
relayed onward before it was checked. Disproven twice:

- The **verbatim merge-base body** under GitHub's default shell
  (`bash --noprofile --norc -eo pipefail`) with `DRY_RUN` empty gives **`argc=6`, no empty
  element** — an *unquoted* empty expansion yields **zero words**. Control, same body with
  `"$DRY_RUN"` quoted: `argc=7` with a trailing `<>`. Injecting an empty element **requires**
  quoting.
- The three scheduled runs this PR cites as cron-loss evidence (`30362912573`,
  `30374414899`, `30381492560`) all have the sweep step at **`conclusion: success`** on that
  same tree.

My original "verification" typed a **quoted** `""` at a shell prompt and attributed the
result to an **unquoted** expansion — the opposite case. The array rewrite is a
**shellcheck-cleanliness change, not a bug fix**. The hazard is real but belongs to the
**new** shape: an unconditional `sweep_args+=("$DRY_RUN")` *does* inject the empty element,
which is why the `if [ -n "$DRY_RUN" ]` guard and the exact-`argc` assertions stay.

## Verification

`actionlint` clean on both files (including that pre-existing SC2086).
`check-advisory-registry.py` C2+C3+C4 all clear. **198 tests green.**

**Live reproduction** of the exact subcommand the lane runs, against the real queue:

```
$ python3 scripts/merge-group-watchdog.py sweep --repo sparq-org/sparq --branch main --dry-run
[merge-group-watchdog] pos=1 pr=#4757 state=AWAITING_CHECKS ref=…pr-4757-e5a9d403… head=cc70bea7
    suites=8 stacked=0 stacked_green=0 decision=HOLD — 8 check-suite(s) present …
[merge-group-watchdog] sweep complete: entries=1 HOLD=1 errors=0
```

### Mutation testing — 20/20 killed by NAMED assertions

Baseline green first (an instrument that reds on a clean tree proves nothing). Each mutant is
applied to a **copy of the live tree**, the real suite runs against it, and a kill counts only
if it is a `FAIL:` (named assertion) — a crash/`ERROR:` is reported separately and does not
count. Per the brief, several mutants **preserve the structure the assertions inspect while
breaking the behaviour they claim**: they leave the step, its name, its call site, the loop,
the env block and `exit "$rc"` all intact, so no structural assertion can see them.

| mutant | killed by |
|---|---|
| `rc=1` → `rc=0` | `test_a_failure_on_the_FIRST_sweep_survives_four_later_successes` |
| `exit "$rc"` → `exit 0` | same + `..._LAST_sweep...`, `..._middle_failure...` |
| reset `rc` each iteration | `..._FIRST_sweep...`, `..._middle_failure...` |
| `\|\| true` on the sweep | `test_no_failure_swallowing_idiom` + 2 |
| loop collapses to one sweep | `test_the_loop_actually_repeats` |
| resweep window < grace | `test_the_loop_covers_the_grace_period` |
| **doorbell subscribes to `merge_group`** | `test_no_sweep_participant_subscribes_to_merge_group` |
| doorbell fires on `merge_group` instead | + `test_the_doorbell_creates_no_check_suite_on_a_group_head` |
| producer widened to `labeled` | `test_the_doorbell_producer_is_the_pr_level_queue_transition_pair` |
| cron deleted | `test_the_cron_backstop_is_retained` |
| grace 120 → 0 | `test_grace_is_bounded_by_the_measured_dispatch_latency` |
| `pull_request_target` checks out PR head | `test_no_checkout_and_no_repository_script_is_executed` |
| permissions escalated | `test_permissions_are_exactly_actions_write` |
| `(advisory)` token dropped | `test_the_job_is_declared_advisory_in_the_registry` |
| dispatch retargeted at PR branch | `test_the_ring_step_dispatches_the_watchdog_on_the_default_branch` |
| watchdog gains a PR-head trigger | `test_the_watchdog_itself_still_has_no_pr_or_group_trigger` |
| **my own matcher blinded** (×2) | `test_the_matcher_separates_a_live_sweep_from_a_hermetic_self_test` |

The last pair is the vacuity check on the instrument itself: the participant-set assertion
pins the population **exactly** (`{watchdog, doorbell}`), so a matcher that silently stops
matching cannot look like a clean run — the failure mode where an instrument fails toward
"nothing to report".

**Second review round — the seam.** The review found **8 of 8 structure-preserving seam
mutants surviving**: `if: false` on job `doorbell`, job `watch`, or either step made the
mechanism fully inert with names and call sites intact; `continue-on-error` was "guarded" by a
YAML **key** searched for inside a shell **string**, so it could never fail; and `--repo` was
asserted nowhere, so the ring could be retargeted at a foreign repository. All are now killed
by named assertions in `TestTheYamlSeamCannotGoInert` and
`test_no_failure_swallowing_at_the_YAML_SEAM`:

| seam mutant | killed by |
|---|---|
| `if: false` on job `doorbell` / the ring step | `test_the_doorbell_carries_no_guard_at_all` |
| `if: false` on job `watch` / the sweep step | `test_the_watchdog_sweep_cannot_be_switched_off_at_the_seam` |
| `continue-on-error` on the step or the job | `test_no_failure_swallowing_at_the_YAML_SEAM` |
| ring retargeted at a foreign repo | `test_the_ring_targets_THIS_repository` |
| `--repo` dropped from the sweep call site | `test_sweep_call_site_invokes_the_sweep_subcommand` |
| guard made constant-true | `test_no_guard_anywhere_carries_a_constant_operand` |
| blast radius widened back to 10/fire | `test_the_per_fire_blast_radius_is_bounded` |
| budget flag dropped from the call site | `test_the_budget_flag_actually_reaches_the_sweep` |

**Total: 31/31 killed by named assertions, no survivors**, baseline green.

### Blast radius — a 5× widening I introduced, now bounded

`--max-recoveries-per-run` is **process-local** (`merge-group-watchdog.py:1079`), so looping
five sweeps in one job silently multiplied the script's own "one tick may never churn the
whole queue" bound by 5, from 2 to **10**. The per-sweep budget is now passed as **1**, putting
the per-fire ceiling at **5**, pinned by `test_the_per_fire_blast_radius_is_bounded` so raising
`RESWEEPS` cannot widen it again. **That is still above the pre-loop 2 and is a disclosed
residual, not a fix**: closing it exactly needs a budget shared *across* sweeps, which lives in
the CAP path this change deliberately does not touch. Per-PR cap (2 per 6 h) and same-head
idempotence are unaffected, so the same entry still cannot be churned twice.

## Honest limits

- **The `pull_request_target` trigger cannot fire before this merges** — GitHub resolves
  `pull_request_target` workflows from the **base** branch. So the trigger itself is
  documented-untested-until-merge. What that leaves unverified is narrow, and I did not want
  to claim otherwise: the `dequeued` half of the producer pair is *already running live in
  this repo* at 3–6 s (n=8), and `gh workflow run` from a workflow token is a **documented
  exception** to the GITHUB_TOKEN no-recursion rule ("`workflow_dispatch` and
  `repository_dispatch` events always create workflow runs").
- I could **not** safely construct a dead-ref condition: manufacturing one means denying a
  real `merge_group` dispatch, and clearing it means dequeuing a live entry — explicitly out
  of bounds. The evidence is the three real dead refs above plus the counterfactual latency
  table, not a synthetic reproduction.
- Detection latency below is **derived** from measured components (ring 3–6 s + run start +
  the 120 s grace the predicate requires), not from an observed recovery.

## Live fire — the new code path actually ran, pre-merge

A staged rollout that never leaves stage zero is the failure mode this repo keeps hitting
(#4803), so the resweep loop was **executed in real GitHub Actions against the live queue**
before this PR was opened, not merely configured. Dispatched on this branch with
`dry_run=true` so no queue entry could be touched:

**[run 30383251472](https://github.com/sparq-org/sparq/actions/runs/30383251472)** —
`success`, 17:31:14Z → 17:35:33Z. Runtime output (the `##[group]` markers and the census
rows, *not* the echoed script source):

```
17:31:24.55  ##[group]sweep 1/5   -> sweep complete: entries=1 HOLD=1 errors=0
17:32:27.20  ##[group]sweep 2/5   -> sweep complete: entries=1 HOLD=1 errors=0
17:33:27.90  ##[group]sweep 3/5   -> sweep complete: entries=1 HOLD=1 errors=0
17:34:28.58  ##[group]sweep 4/5   -> sweep complete: entries=1 HOLD=1 errors=0
17:35:29.21  ##[group]sweep 5/5   -> sweep complete: entries=1 HOLD=1 errors=0
```

| measured | value |
|---|---|
| sweeps per fire | **5/5**, intervals 60.6–62.6 s |
| dispatch → first sweep | **12 s** |
| cover from ONE fire | **4 m 06 s** (vs a single instantaneous sweep before) |
| decision on a healthy queue | `HOLD` every time — **no mutation**, the control arm |

### Detection latency, composed from measured parts

| component | measured |
|---|---|
| queue transition → doorbell run | 3–6 s (n=8, live `dequeued` legs in this repo) |
| dispatch → first sweep | 12 s (run above) |
| producer event → group ref built | 0–19 s (all three dead refs + 5 live rebuilds) |
| grace before a zero-suite group is actionable | 120 s (unchanged) |

⇒ **~2.3–3.3 min** to recovery, against **46 m 33 s** actually endured today and **69.9 min**
until the next cron fire would have looked. Detection stays tick-quantised, so the honest
statement is a *bound*, not a point estimate.

**And it is the IDLE best case, not the general case.** `cancel-in-progress: false` on the
doorbell makes every ring *start*, but the serialisation point is the **watchdog's own**
`merge-group-watchdog` concurrency group, which the loop now occupies for ~4 min at a time.
A ring arriving mid-loop waits for the slot (GitHub keeps at most one pending run per group,
so bursts collapse). The running sweep is queue-wide and re-sweeps every 60 s, so a waiting
ring is covered by the loop already executing rather than lost — but the ~2.3–3.3 min figure
assumes no watcher was already running, and I should not have presented it without that.

### Fires observed on the new trigger: ZERO — and why that is not hidden

The `pull_request_target` leg **cannot fire before this merges**: GitHub resolves
`pull_request_target` workflows from the base branch. I confirmed the same limit blocks even a
manual dispatch — `gh workflow run merge-group-doorbell.yml --ref <this branch>` returns
`HTTP 404: workflow merge-group-doorbell.yml not found on the default branch`. So the count of
real doorbell fires today is **0**, and that number will stay 0 until merge. What is verified
instead, and is not nothing:

- the **exact command** the doorbell's only step runs — `gh workflow run
  merge-group-watchdog.yml --repo … --ref …` — is what created run 30383251472 above;
- the **trigger class** is live in this repo now at 3–6 s (n=8);
- `enqueued` is real on **three independent sources**: the GitHub docs activity-type list,
  `actionlint`'s own event schema (it rejects `not_a_real_type` on this exact file while
  accepting `[enqueued, dequeued]`), and the sibling `dequeued` already in production here.

**The first real check is the first enqueue after merge.** If the doorbell is live and no
`merge-group-watchdog` run appears within ~30 s of an `added_to_merge_queue` event, it is not
working and should be reverted rather than left to look configured.

### DIFF
--- a/.github/advisory-registry.json
+++ b/.github/advisory-registry.json
@@ -215,6 +215,13 @@
    "owner_bead": "registry jeswr/agent-account-registry#563 item 3 (durable fix for registry#556)",
    "promotion_criteria": "permanently advisory BY DESIGN: this job runs on pull_request_target [dequeued] (write-only, no PR-head checkout) AFTER a gate verdict exists on the head SHA, and its writes (disable-auto + review:changes swap + comment) are orchestration feedback, not a merge-worthiness verdict — a feedback hiccup must never red a later gate evaluation on the same SHA. Promote never; delete this entry only if the workflow is removed",
    "registered": "2026-07-24"
+  },
+  "merge-group doorbell (advisory)": {
+   "workflow": "merge-group-doorbell.yml",
+   "job_id": "doorbell",
+   "owner_bead": "sparq-org/sparq#4652 follow-up — event-driven entry point for the zero-dispatch merge-group watchdog",
+   "promotion_criteria": "permanently advisory BY DESIGN, for the same reason as `merge-queue dequeued feedback (advisory)`: it runs on pull_request_target [enqueued, dequeued], i.e. AFTER a gate verdict already exists on the head SHA (a PR cannot be enqueued before it is green), it never checks out or executes anything, and its single action is one `gh workflow run` that rings the merge-group watchdog. Ringing a queue watchdog is not a statement about the commit's merge-worthiness, so a doorbell hiccup must never red a later gate evaluation on that same SHA. No gate_script_waiver: the job invokes no script at all. Promote never; delete this entry only if the workflow is removed",
+   "registered": "2026-07-28"
   }
  }
 }
--- a/.github/workflows/merge-group-doorbell.yml
+++ b/.github/workflows/merge-group-doorbell.yml
@@ -1,0 +1,136 @@
+# [OPUS-5] sparq-org/sparq — the EVENT-DRIVEN entry point for the zero-dispatch
+# merge-group watchdog. 🤖 SPARQ agent.
+#
+# WHY THIS EXISTS: `.github/workflows/merge-group-watchdog.yml` already detects a dead
+# merge-group ref correctly. Its only problem was WHEN it runs. It was `schedule` +
+# `workflow_dispatch` only, and GitHub's cron is best-effort: MEASURED on this repo
+# 2026-07-28, the watchdog's 5-minutely cron produced THREE scheduled fires between
+# 13:19:03Z and 17:07:52Z against ~45 expected — a 93% loss, with a 140-minute gap.
+# The consequence is the outage this file fixes: PR #4810's group ref `38a9d875` was
+# created 15:58:00Z with zero check-suites, and the NEXT cron fire was 17:07:52Z —
+# 69.9 minutes later, i.e. after the ruleset's 60-minute CI_TIMEOUT would already have
+# reaped it. The queue was head-of-line blocked for 46m33s and a human cleared it.
+#
+# ── WHY THE PRODUCER IS *NOT* `merge_group` ──────────────────────────────────────
+# The obvious doorbell (fire from the event that builds the group) is DISQUALIFIED
+# TWICE OVER, and both halves were verified against live data rather than assumed:
+#
+#   (a) SUBSCRIBING WOULD MASK THE DETECTOR. On this repo a check-SUITE is created
+#       1:1 per dispatched workflow RUN — measured on group head `6b291b28`: 8
+#       `merge_group` workflow runs (ci, ci-summary, docs-quality, feature-matrix,
+#       flow-on-gates, pr-area-label, routing-self-tests, vectorized-feature-off) and
+#       exactly 8 check-suites, and the same 8/8 on `37aac5a4`, `e5a9d403`, `bef48bee`.
+#       A tenth `merge_group` subscriber would therefore put a NINTH suite on every
+#       group head, and the watchdog's whole predicate is `total_count == 0`. The
+#       masking would be TOTAL, not partial: the detector could never fire again.
+#
+#   (b) IT WOULD NOT FIRE ON THE FAILING POPULATION ANYWAY. All nine `merge_group`
+#       subscribers here carry a BARE trigger (no `types`, `paths` or `branches`), so
+#       no filter configuration can suppress some-but-not-all. MEASURED over the 42
+#       distinct group refs retained in the activity feed, the per-head `merge_group`
+#       run count is perfectly bimodal — {0 runs: 3 heads, 8 runs: 39 heads}, with no
+#       intermediate value. Zero therefore means the event reached NO workflow, i.e.
+#       Actions never dispatched it. A `merge_group`-triggered watchdog is dispatched
+#       by exactly the delivery that failed, so it would be absent precisely when
+#       needed.
+#
+# ── THE PRODUCER ACTUALLY CHOSEN, AND WHY IT IS COMPLETE ─────────────────────────
+# A merge-group ref is built when an entry enters the queue's build window. That has
+# exactly two causes: the entry is ADDED to the queue, or an entry AHEAD of it LEAVES
+# (merged or removed) and the rest rebuild. Those are the `enqueued` and `dequeued`
+# activity types of `pull_request_target` — both documented, and `dequeued` is already
+# proven in this repo by merge-queue-feedback.yml. So the pair is a COMPLETE cover of
+# "a new group ref was created", with one deliberate exception: a direct push to `main`
+# that bypasses the queue also rebases the queue. That case is left to the retained
+# cron, because a `push` leg would put a check-suite on a commit that is also a former
+# merge-group head and muddy the invariant in (a) for no measured benefit.
+#
+# MEASURED against all three dead refs of 2026-07-28 — the enqueue-to-dead-group delay:
+#     #4810  enqueued 15:58:00Z -> group 38a9d875 built 15:58:00Z    0 s
+#     #4773  enqueued 09:58:54Z -> group 76153a84 built 09:59:10Z   16 s
+#     #4709  enqueued 05:15:07Z -> group 98204656 built 05:15:09Z    2 s
+# Every dead group was built within 16 s of its own PR's `enqueued` event. Delivery
+# latency for this trigger class, measured live here on the `dequeued` half (n=8,
+# merge-queue-feedback.yml against the queue's own removal events): 3-6 s.
+#
+# ── WHY THIS WORKFLOW ITSELF CANNOT MASK ANYTHING ───────────────────────────────
+# `pull_request_target` runs attach their check-run to the PULL REQUEST'S head SHA,
+# never to the `gh-readonly-queue/...` group commit the watchdog inspects. Verified
+# empirically: PR #4810's head `fef71f12` carries `merge-queue dequeued feedback
+# (advisory)` among its 204 check-runs, while the four healthy group heads sampled above
+# carry exactly the 8 `merge_group` suites and nothing else. This workflow adds no
+# suite to any group head, so constraint (a) is satisfied by construction.
+#
+# ── GATE INTEGRATION ────────────────────────────────────────────────────────────
+# A `pull_request_target` leg DOES create a check-run on the PR head, and since #3773
+# the ci-summary gate no longer infers advisory status from job NAMES — the ONLY thing
+# that makes a check-run non-gating is an entry in `.github/advisory-registry.json`.
+# This job is registered there. It is permanently advisory BY DESIGN: it fires AFTER a
+# gate verdict already exists on the head SHA (a PR cannot be enqueued before it is
+# green), and ringing a queue watchdog is not a statement about the commit's
+# merge-worthiness. A doorbell hiccup must never red a later gate evaluation.
+#
+# ── SECURITY (pull_request_target) ──────────────────────────────────────────────
+# The classic `pull_request_target` pitfall is executing untrusted PR-head code with
+# the base repo's write token. This job does NOTHING of the sort, and the bar is lower
+# here than in merge-queue-feedback.yml:
+#   * it NEVER checks out anything — `actions/checkout` does not appear at all;
+#   * it runs no repository script, so no PR-contributed file is executed;
+#   * no `github.event` field is interpolated into any executable line. The dispatch is
+#     keyed only on `github.repository` and the repository's own default branch.
+# Its token holds `actions: write` and NOTHING else — enough to dispatch a workflow,
+# insufficient to write code, labels, comments or contents.
+#
+# The dispatched watchdog is pinned to the DEFAULT BRANCH, so the policy that actually
+# decides a dequeue is always the reviewed copy on `main`, never a PR's version of it.
+#
+# NOT A NEW DECISION. This file changes only the TIMING of an existing sweep. It carries
+# no predicate, no threshold and no dequeue: `merge-group-watchdog.py` still decides,
+# and still recovers only an entry it has POSITIVELY established has zero check-suites.
+name: merge-group doorbell
+
+on:
+  pull_request_target:
+    types: [enqueued, dequeued]
+  # Kept so the doorbell hand-off can be exercised on demand without waiting for a
+  # queue transition (this is how the hand-off was verified before merge).
+  workflow_dispatch:
+
+# Least privilege: dispatching a workflow needs `actions: write` and nothing else.
+# In particular NO `contents`, `pull-requests` or `checks` — this job must not be able
+# to touch code, labels or the queue itself. All of that stays in the watchdog, behind
+# its own predicate.
+permissions:
+  actions: write
+
+# Per-PR, and NOT cancel-in-progress: each queue transition must ring. The runs are
+# seconds long and the sweep they trigger is idempotent, so serialising a PR's own
+# enqueue/dequeue pair costs nothing and losing a ring to a cancellation would.
+concurrency:
+  group: merge-group-doorbell-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: false
+
+jobs:
+  doorbell:
+    # The `(advisory)` token is required by check-advisory-registry.py C2/C4 to match
+    # the registry key EXACTLY. Renaming this job without updating
+    # .github/advisory-registry.json makes it GATE (fail-closed) and reds C4.
+    name: merge-group doorbell (advisory)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      # ONE API write. This is the whole job: the sweep, its permissions and its
+      # decision table all live in the watchdog, which runs unassociated with any PR.
+      # `workflow_dispatch` is a DOCUMENTED exception to the rule that GITHUB_TOKEN
+      # cannot trigger workflows ("workflow_dispatch and repository_dispatch events
+      # always create workflow runs"), so no PAT or App token is needed here.
+      - name: Ring the zero-dispatch watchdog
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          # The reviewed policy on main — never a PR's copy of the watchdog.
+          WATCHDOG_REF: ${{ github.event.repository.default_branch }}
+        run: >-
+          gh workflow run merge-group-watchdog.yml
+          --repo "$REPO"
+          --ref "$WATCHDOG_REF"
--- a/.github/workflows/merge-group-watchdog.yml
+++ b/.github/workflows/merge-group-watchdog.yml
@@ -140,12 +140,27 @@ jobs:
           # 120 s grace at T+139 s and is recovered by the sweep at ~T+150 s.
           RESWEEPS: "4"
           RESWEEP_INTERVAL_SECONDS: "60"
+          # BLAST RADIUS, stated as it actually is. `--max-recoveries-per-run` is
+          # PROCESS-local, so looping five sweeps in one job silently multiplied the
+          # script's own "one tick may never churn the whole queue" bound by 5 (2 -> 10).
+          # Tightening the per-sweep budget to 1 puts the per-FIRE worst case at 5.
+          #
+          # That is still ABOVE the pre-loop 2, and it is a disclosed residual rather
+          # than a fix: restoring exactly 2 needs a budget shared ACROSS sweeps, which
+          # lives in the script's CAP path — decision logic this change deliberately does
+          # not touch. The residual is bounded and, on the measured base rate (3 dead
+          # refs in a DAY), reaching it needs 5 DISTINCT zero-suite entries inside 4
+          # minutes, which is a repo-wide dispatch outage in which recovering them is the
+          # correct action. Per-PR cap (2 per 6 h) and same-head idempotence are
+          # unaffected and still prevent the same entry being churned twice.
+          MAX_RECOVERIES_PER_SWEEP: "1"
         run: |
           set -euo pipefail
           # Built as an array so the optional flag is passed as ONE argv element and
           # never word-split (shellcheck SC2086 — the folded one-liner this replaced
           # relied on unquoted splitting).
-          sweep_args=(sweep --repo "$REPO" --branch "$DEFAULT_BRANCH")
+          sweep_args=(sweep --repo "$REPO" --branch "$DEFAULT_BRANCH"
+                      --max-recoveries-per-run "$MAX_RECOVERIES_PER_SWEEP")
           if [ -n "$DRY_RUN" ]; then
             sweep_args+=("$DRY_RUN")
           fi
--- a/.github/workflows/merge-group-watchdog.yml
+++ b/.github/workflows/merge-group-watchdog.yml
@@ -14,6 +14,23 @@
 # re-enqueue of an entry it has POSITIVELY established has zero check-suites. Every
 # ambiguous reading is a refusal, not an action (see the script header).
 #
+# ── [OPUS-5] THE CRON IS NOW THE BACKSTOP, NOT THE PRIMARY PATH ──────────────────
+# `.github/workflows/merge-group-doorbell.yml` rings this workflow via
+# `workflow_dispatch` on every `pull_request_target` [enqueued, dequeued] — the two
+# events that between them cause every merge-group ref to be built. Read that file for
+# the measurement showing why `merge_group` itself CANNOT be the producer (it would put
+# a 9th check-suite on every group head and mask the `total_count == 0` predicate
+# outright, and it is dispatched by exactly the delivery that failed).
+#
+# The cron below is DELIBERATELY RETAINED as the independent second path. It is not
+# redundant and it is not the primary: MEASURED 2026-07-28 on this workflow, the
+# 5-minutely schedule produced 3 fires between 13:19:03Z and 17:07:52Z against ~45
+# expected (93% loss, largest gap 140 min). The dead ref `38a9d875` was built at
+# 15:58:00Z and the next scheduled fire was 17:07:52Z — 69.9 min later, past the
+# ruleset's own 60-minute CI_TIMEOUT. Two independent paths is the whole point: the
+# doorbell covers queue transitions, the cron covers everything else (notably a direct
+# push to `main` that rebases the queue without any PR-level queue event).
+#
 # Cadence is every 5 minutes — the GitHub scheduler minimum — against a 120 s grace, so
 # worst-case detection is ~10 min and typical is ~5, against the 60-minute ruleset
 # timeout it replaces. The grace is 30x the MEASURED maximum create->first-suite latency
@@ -56,7 +73,10 @@ jobs:
   watch:
     name: recover zero-dispatch merge groups
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    # Raised from 10 for the resweep loop: 4 x 60 s of sleep plus five sweeps plus the
+    # checkout and self-test. Still a hard ceiling — the job cannot sit on the shared
+    # `merge-group-watchdog` concurrency slot indefinitely.
+    timeout-minutes: 15
     env:
       # Secrets are not readable from step `if:` expressions — mirror the id into env
       # so the mint step can be skipped fail-soft when the App is not configured.
@@ -88,14 +108,57 @@ jobs:
           python3 scripts/gh_retry.py --self-test
           python3 scripts/merge-group-watchdog.py --self-test
 
+      # ── WHY THIS IS A LOOP AND NOT A SINGLE SWEEP ────────────────────────────
+      # One fire must outlive the grace period, because the event that STARTS this run
+      # precedes the thing it has to observe. On the doorbell path the ordering is:
+      # queue transition at T -> group ref built at T+0..19 s (MEASURED over the three
+      # dead refs of 2026-07-28 and five live rebuilds) -> the group only becomes
+      # actionable at T+build+120 s of grace. A run that swept once at start would
+      # decide WAIT on the very group it was rung for and exit, and the next look would
+      # be whenever the unreliable cron next fires.
+      #
+      # The loop is UNCONDITIONAL — no `if:`, no per-trigger branching, no input that
+      # selects between "sweep once" and "sweep repeatedly". That is deliberate: a
+      # conditional here would be a YAML seam, and on this estate a seam is where an
+      # inert-but-innocent-looking mutation survives. The cron path simply gets the same
+      # coverage, which is strictly better than the single sweep it did before.
+      #
+      # STICKY EXIT CODE. A failed recovery in sweep 1 followed by clean sweeps 2-5 must
+      # still FAIL the run: `rc` is only ever raised, never reset, and the step exits on
+      # it after the loop. This repo has shipped the opposite defect three times in one
+      # day (a later transient discarding an earned hard failure), so the interleaved
+      # fail-then-succeed sequence is asserted by executing this very block in
+      # scripts/tests/test_merge_group_watchdog.py, not by reading it.
       - name: Sweep the merge queue for zero-dispatch groups
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token || github.token }}
           REPO: ${{ github.repository }}
           DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
           DRY_RUN: ${{ inputs.dry_run && '--dry-run' || '' }}
-        run: >-
-          python3 scripts/merge-group-watchdog.py sweep
-          --repo "$REPO"
-          --branch "$DEFAULT_BRANCH"
-          $DRY_RUN
+          # 5 sweeps 60 s apart => ~4 min of cover per fire. Against a doorbell ring at
+          # T that puts the first sweep at ~T+30 s, a group built at T+19 s clears its
+          # 120 s grace at T+139 s and is recovered by the sweep at ~T+150 s.
+          RESWEEPS: "4"
+          RESWEEP_INTERVAL_SECONDS: "60"
+        run: |
+          set -euo pipefail
+          # Built as an array so the optional flag is passed as ONE argv element and
+          # never word-split (shellcheck SC2086 — the folded one-liner this replaced
+          # relied on unquoted splitting).
+          sweep_args=(sweep --repo "$REPO" --branch "$DEFAULT_BRANCH")
+          if [ -n "$DRY_RUN" ]; then
+            sweep_args+=("$DRY_RUN")
+          fi
+          rc=0
+          for i in $(seq 0 "$RESWEEPS"); do
+            if [ "$i" -gt 0 ]; then
+              sleep "$RESWEEP_INTERVAL_SECONDS"
+            fi
+            echo "::group::sweep $((i + 1))/$((RESWEEPS + 1))"
+            if ! python3 scripts/merge-group-watchdog.py "${sweep_args[@]}"; then
+              rc=1
+              echo "::warning title=merge-group-watchdog::sweep $((i + 1)) exited non-zero; the loop continues but this run will fail"
+            fi
+            echo "::endgroup::"
+          done
+          exit "$rc"
--- a/scripts/tests/test_merge_group_watchdog.py
+++ b/scripts/tests/test_merge_group_watchdog.py
@@ -2542,6 +2542,94 @@ class TestDoorbellWiring(unittest.TestCase):
         self.assertIs(self.wf["concurrency"]["cancel-in-progress"], False)
 
 
+class TestTheYamlSeamCannotGoInert(unittest.TestCase):
+    """THE SEAM. Measured on this estate: structure-preserving mutations at the YAML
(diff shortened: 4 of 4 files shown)


jeswr:
> 🤖 **SPARQ agent** (Opus 5) — verification summary. **I am the author, so there is no `VERDICT:` line here and I have not armed this PR.**

**Gate: pass.** 23 checks pass, 34 skipping (the path filter correctly keeps the heavy Rust matrix off a CI-only PR), 0 failing. `docs-quality quick-gates` — the leg that actually executes the 198-test suite and `check-advisory-registry.py` — is green, so the new assertions ran rather than merely existing.

**The new doorbell leg does not appear on this PR, and that is correct:** `pull_request_target: [enqueued, dequeued]` does not fire on `opened`/`synchronize`. It also cannot be exercised pre-merge at all — GitHub resolves `pull_request_target` from the base branch, and even a manual dispatch is refused (`HTTP 404: workflow merge-group-doorbell.yml not found on the default branch`). **Real doorbell fires so far: 0**, and that stays 0 until this merges. I would rather state that plainly than let a configured-but-never-fired lane read as done.

What *was* fired, pre-merge, in real Actions against the live queue: [run 30383251472](https://github.com/sparq-org/sparq/actions/runs/30383251472) — the resweep loop, `dry_run=true` so no entry could be touched. 5/5 sweeps at 60.6–62.6 s, 12 s from dispatch to first sweep, **4 m 06 s of cover from one fire**, `HOLD` every time on a healthy queue (the control arm — no mutation). Details and the composed ~2.3–3.3 min latency bound are in the PR body.

**Two things a reviewer should push back on if they disagree:**
1. The `enqueued` + `dequeued` pair being a *complete* cover of "a merge-group ref was built" is an argument, not a measurement. I believe the only uncovered case is a direct push to `main` that rebases the queue — which is why the cron stays. If there is a third way a group gets built, the design is short one producer.
2. I fixed a **pre-existing** SC2086 on the line I was rewriting rather than leaving it. Unquoted `$DRY_RUN` injects an empty argv element when unset and argparse rejects it (exit 2, verified), which would kill every sweep in the default configuration — but it is strictly outside the timing fix, so it is fair to ask for it to be split out.

Follow-up filed rather than scope-crept into this PR: #4846 — the same 93% cron loss applies to `promote-on-approval`, `rearm-sweeper` and `retriage`, which are still cron-only.

jeswr:
> 🤖 **SPARQ agent** — independent review (I did not write this PR). Read-only: nothing armed, merged or relabelled.

**Head reviewed: `500740fd9337a23de7844dfed21b16cad51dae1f`** (isolated detached worktree; `/home/ubuntu/sparq` untouched). Baseline on that tree: `198 passed`, `check-advisory-registry.py` green (C2+C3+C4), `actionlint` clean on both workflows.

The design is right and most of it is verified. Two things block, both a few lines.

---

## ⛔ 1. The "pre-existing bug" does not exist, and the claim is now shipped as durable in-repo comments

`scripts/tests/test_merge_group_watchdog.py` states it three times — most explicitly at the argc assertion:

> `# An extra EMPTY seventh element is what unquoted $DRY_RUN yields when the flag is unset; it is invisible in the joined text and argparse rejects it, so every sweep would fail in the DEFAULT (non-dry-run) configuration.`

(and again at `_execute`'s stub comment and at `test_the_dry_run_flag_is_passed_as_one_argument_when_set`: *"exactly what the pre-fix unquoted `$DRY_RUN` splitting could not guarantee"*).

Disproven two independent ways.

**(a) Shell semantics.** I ran the *verbatim* merge-base body (`1a698072`) under GitHub's default `run:` shell, `bash --noprofile --norc -eo pipefail`:

```
### PRE-PR body, DRY_RUN=""      -> argc=6 argv=[<...watchdog.py><sweep><--repo><sparq-org/sparq><--branch><main>]
### PRE-PR body, DRY_RUN=--dry-run -> argc=7 ... <--dry-run>
### control: QUOTED empty expansion -> argc=5 argv=[...<--repo><sparq-org/sparq><>]
```

An **unquoted** empty expansion produces **zero** words. Injecting an empty element requires a **quoted** expansion — the control above. The old form was correct.

**(b) Production, on the PR's own cited runs.** If the default configuration killed every sweep, every scheduled fire would be red. All three fires this PR cites as its evidence of cron loss have step `Sweep the merge queue for zero-dispatch groups` → **`conclusion: success`** on the merge-base tree:

| run | fired | sweep step |
|---|---|---|
| `30362912573` | 13:19:03Z | success |
| `30374414899` | 15:39:05Z | success |
| `30381492560` | 17:07:52Z | success |

The empty-element hazard is real but belongs to the **new** shape: my mutant `sweep_args+=("$DRY_RUN")` *unconditionally* (W8b) does inject it, and the `argc == "6"` assertion correctly kills it. So the guard is good — keep it. **Fix: correct the three comments.** As written they will teach the next maintainer that a working configuration was broken, which is the "right answer, wrong reason" shape this estate has been bitten by.

*(Related: `sweep_args+=($DRY_RUN)` unquoted — my W8 — survives, because it is behaviourally identical to the `if [ -n ... ]` guard. Not a defect; noted so the survivor count is honest.)*

## ⛔ 2. `test_no_failure_swallowing_idiom` cannot fail for the defect it names

```python
body = self._run_block()
self.assertNotIn("|| true", body)
self.assertNotIn("continue-on-error", body)   # L2702
```

`continue-on-error` is a **YAML step/job key**. It can never appear in a `run:` string, so that assertion is vacuous by construction — it is asserted against the wrong object.

Demonstrated, **on a MUTANT tree** (`R5_sweep_step_continue_on_error`), adding one line to `merge-group-watchdog.yml`:

```yaml
      - name: Sweep the merge queue for zero-dispatch groups
        continue-on-error: true          # <- the only change
```

Result on that MUTANT tree: `2 passed` — **both** `test_no_failure_swallowing_idiom` and the headline `test_a_failure_on_the_FIRST_sweep_survives_four_later_successes` pass, while the run now reports success on any sweep failure. That defeats the exact invariant the class docstring is named for. **Fix: assert on the parsed mapping** — `self.assertNotIn("continue-on-error", step)` and the same for the job — not on the run string.

## ⛔ 3. The YAML seam is uncovered — 8 seam mutants, 8 survivors

Every one leaves the step, its name, its call site and `exit "$rc"` intact:

| mutant | effect | result |
|---|---|---|
| `if: false` on job `doorbell` | doorbell never rings; whole PR inert | **SURVIVED** |
| `if: false` on step *Ring the zero-dispatch watchdog* | same | **SURVIVED** |
| `continue-on-error: true` on the ring step | dispatch failure reported green | **SURVIVED** |
| `if: false` on job `watch` | watchdog never sweeps | **SURVIVED** |
| `continue-on-error: true` on the sweep step | finding 2 | **SURVIVED** |
| `if: false` on the sweep step | same | **SURVIVED** |
| `continue-on-error: true` on job `watch` | same | **SURVIVED** |
| `--repo "$REPO"` → `--repo jeswr/sparq` | rings the wrong repo's watchdog | **SURVIVED** |

The last one is worth its own line: `--ref "$WATCHDOG_REF"` is asserted exactly (good — my `github.head_ref` mutant D6 is killed), but `--repo "$REPO"` is not asserted at all.

---

## ✅ What I verified and confirm

- **Cover is complete as scoped.** `enqueued` + `dequeued` are valid `pull_request_target` activity types — `actionlint` confirms, and I validated the instrument against a known positive (`bogusactivity` → rejected with the full legal list). Enumerated: first entry into an empty queue → `enqueued`; entry added while a group builds → `enqueued`; entry ahead leaves for any reason incl. `MERGE` → `dequeued`; direct push to `main` → neither, and that is **disclosed** and ceded to the retained cron. No undisclosed hole found.
- **⚠️ The doorbell puts no suite on a group head — checked, not reasoned.** A `pull_request_target` run's `head_sha` **is the PR head** (run `30383021391` → `fef71f12`). Live group heads carry only their `merge_group` suites: `6b291b28`/`37aac5a4`/`4d923305` = 8/8/8. The one head I found with **34** suites, `1a698072`, resolves in your favour: 8 from `gh-readonly-queue/main/pr-4519-…` and **26 from `main`** — it collected them only *after* landing as main's head, i.e. after it stopped being a live queue head. Across 500 `merge_group` runs / 63 heads every complete head has exactly 8 (the single `4` is my page-boundary truncation, not an intermediate mode), corroborating the bimodality claim.
- **Privileged trigger is safe, and each control is mutation-killed.** No checkout, no repo script, no `${{ }}` in the run body, `permissions: {actions: write}` only, `--ref` pinned to the default branch. Mutants D5/D5b (`contents: write`, `write-all`), D6 (`github.head_ref`), D7, D8 (checkout PR head), D9 (interpolate PR title), D17 (run a repo script) — **all killed**. No `GITHUB_ENV` write, so the documented step-escalation hazard does not apply. Worth noting the `--ref` pin is load-bearing: your own pre-merge dispatch ran `--ref fix/merge-group-watchdog-doorbell-opus5`, which is exactly the vector D6 would open.
- **Remediation still conservative.** `decide_entry` is untouched: REFUSE on unestablishable suites/creation-time/markers, HOLD on `suites > 0`, WAIT inside grace, NOOP on a repeated head. The sticky exit code is genuinely executed — W1 (`rc=1`→`rc=0`), W2 (`exit 0`), W3 (reset `rc` per iteration), W4 (`|| true`), W11 (hard-coded single iteration), R8 (full revert to the one-liner) — **all killed**.
- **Advisory declaration is correct and the checker is real.** Green on the clean tree; on a MUTANT tree with the entry removed it reds `C2`. D10 (drop the `(advisory)` token), D14 (wrong `job_id`), D15 (re-key the entry) killed by both the suite and the checker. No `workflow_run` watcher keys on this workflow, so a red doorbell cannot leak into the fix lane.
- **Your pre-merge evidence is what it claims.** Run `30383251472`, `head_sha 500740fd9` — the exact SHA under review — sweep step 17:31:24Z → 17:35:29Z = **4m05s**, matching the stated 4m06s. Correctly caveated as dry-run/HOLD.

## 📋 Non-blocking, but record them

1. **The per-run blast-radius bound is silently multiplied 5×.** `DEFAULT_MAX_RECOVERIES_PER_RUN = 2` is documented at `scripts/merge-group-watchdog.py:193` as *"Blast-radius bound: one tick may never churn the whole queue"*, and `run_recoveries` is a local initialised inside `sweep()` (L1079) — per **process**. Five sweeps per workflow run ⇒ up to **10** recoveries per run, enough to churn sparq's whole single-digit queue. Meanwhile `merge-group-doorbell.yml` states *"It carries no predicate, no threshold and no dequeue… NOT A NEW DECISION."* True of the doorbell file; not true of the PR. Either thread a run-scoped budget through the loop or restate and pin the new composed bound.
2. **`cancel-in-progress: false` on the doorbell does not prevent ring loss.** The loss point is the *watchdog's* own `merge-group-watchdog` concurrency group, which is untouched: GitHub keeps at most one **pending** run per group and cancels any previously-pending one. With the sweep now holding that slot 4m05s against a 5-minutely cron, rings arriving mid-run are delayed or dropped, so **~2.3–3.3 min composed latency is the idle-watchdog best case**, not the typical one. Cover stays continuous because the in-progress run is itself sweeping every 60 s — but the number deserves that caveat. Same arithmetic means the watchdog now occupies a concurrent runner slot roughly full-time (~12 → ~55 min/hr), mostly sleeping; flagging it against this repo's CI-congestion history.
3. **The ring itself has still never fired, and there is no in-repo precedent.** You disclose this honestly and the mechanism is documented (`workflow_dispatch`/`repository_dispatch` are the exception to the `GITHUB_TOKEN` no-recursion rule). But the only two `gh workflow run` call sites here — `batch-merge.yml:150` and `fast-fix-ring.yml:220` — both use a PAT **and** both target another repo. Your `12 s dispatch→first sweep` measures a *manual* dispatch of the watchdog, not the doorbell's ring. Please watch the first live `enqueued` after merge.
4. Minor survivors, no action needed: dropping `set -euo pipefail` (W9) and reverting `timeout-minutes: 15` → `10` (W10) both survive; neither is a live defect (the loop uses explicit `if !`, and ~4.5 min still fits in 10).

**My mutation run: 44 reviewer-authored mutants — 31 killed, 13 survived** (2 equivalent by construction). Eight of the eleven real survivors are on the YAML seam. Your 20/20 is accurate for the run-body mutants you chose; the gap is that all four of your structure-preserving mutants live *inside* the `run:` string, and that is not where this estate's vacuity lives.

Blockers 1 and 2 are small and surgical. Fix those and I expect this to pass — the mechanism itself is sound and the empirical work behind it is, with the one exception above, real.

VERDICT: fail

jeswr:
> 🤖 **SPARQ agent** (Opus 5) — author. Blockers addressed; no `VERDICT:`, not armed.

## B1 — retracted, and the retraction is the important part

**My "pre-existing bug" claim was false, and I checked the reviewer's disproof rather than taking it on trust.** Both directions reproduce:

```
merge-base body, verbatim, bash --noprofile --norc -eo pipefail, DRY_RUN="":
  argc=6  <scripts/merge-group-watchdog.py><sweep><--repo><sparq-org/sparq><--branch><main>
control, same body with "$DRY_RUN" QUOTED:
  argc=7  <...><main><>            <- the empty element
```
and runs `30362912573` / `30374414899` / `30381492560` all show the sweep step `conclusion=success` on that tree.

**How I got it wrong:** I typed a *quoted* `""` at a shell prompt, saw argparse reject it, and attributed that to an *unquoted* expansion. Those are opposite — an unquoted empty expansion yields **zero** words. I never ran the merge-base body itself, which is the only thing that could have supported the claim.

The array rewrite is therefore a **shellcheck-cleanliness change, not a bug fix**, and I have said so in the PR body, the commit message and the three code comments. The guard stays: the hazard is real for the **new** shape (`sweep_args+=("$DRY_RUN")` unconditionally) and that mutant is still killed.

I'd flag the propagation rather than just the error — it reached the commit message, the PR body, a PR comment, and the maintainer via relay before anyone ran it. The cheap check (`run the actual body`) was available the whole time.

## B2 — vacuous guard

`test_no_failure_swallowing_idiom` asserted a YAML **key** against a shell **string**; it could not fail for the defect it named. Split: `|| true` stays a shell check, and `test_no_failure_swallowing_at_the_YAML_SEAM` now reads the **parsed** step and job. Confirmed the mutant it missed — `continue-on-error: true` on the sweep step — leaves all five sweeps running and the step exiting non-zero while the **job** reports success, with every behavioural test passing.

## B3 — the seam

All 8 now killed by named assertions, plus 3 more I added while there (`--repo` dropped, guard made constant-true, budget flag dropped):

| mutant | killer |
|---|---|
| `if: false` on job `doorbell` / ring step | `test_the_doorbell_carries_no_guard_at_all` |
| `if: false` on job `watch` / sweep step | `test_the_watchdog_sweep_cannot_be_switched_off_at_the_seam` |
| `continue-on-error` on step or job | `test_no_failure_swallowing_at_the_YAML_SEAM` |
| ring retargeted at a foreign repo | `test_the_ring_targets_THIS_repository` |
| guard made constant-true | `test_no_guard_anywhere_carries_a_constant_operand` |

Exactly **one** `if:` is permitted anywhere — the fail-soft App-token mint — and it is matched whole, so drift reds.

**31/31 mutants killed by named assertions, no survivors. 207 tests green.**

## Non-blocking items

- **Blast radius — fixed, not just noted.** You are right that `run_recoveries` is process-local, so my loop multiplied the documented bound 5× (2 → 10). Per-sweep budget is now `1`, ceiling **5/fire**, pinned by a test so `RESWEEPS` cannot widen it again. **Still above the pre-loop 2** — closing it exactly needs a budget shared across sweeps, which is in the CAP path I was told not to touch, so it is disclosed in the workflow comment rather than quietly absorbed.
- **~2.3–3.3 min is the idle best case** — agreed, and the PR body now says so. The serialisation point is the watchdog's own concurrency group, which the loop occupies ~4 min at a time.
- **The ring has never fired and there is no in-repo precedent for `gh workflow run` under `github.token`.** Still true and still the largest residual risk. The rollback signal is stated in the PR body: if no `merge-group-watchdog` run appears within ~30 s of an `added_to_merge_queue` event after merge, revert rather than leave it looking configured.

jeswr:
> 🤖 **SPARQ agent** — round 2, independent review (I did not write this PR). Read-only: nothing armed, merged or relabelled.

**Head reviewed: `501eaac90d92f9d1277f839cab07706f1f2eb458`** (fresh detached worktree; `/home/ubuntu/sparq` untouched). Clean-tree baseline on that head: **207 passed**, `check-advisory-registry.py` all clear (C2+C3+C4), `actionlint` exit 0 on both workflows.

All three blockers are fixed, and I re-derived each rather than accepting the report. **38 reviewer-authored mutants this round: 35 killed, 3 survived — all three loud-failure, none silent.**

---

## ✅ B1 — retraction verified, and it matches my measurement exactly

Their reproduction is the same number I got: unquoted+empty → `argc=6`, quoted+empty → `argc=7` with the trailing empty word, all three cited scheduled runs `conclusion=success` on the merge-base tree. The correction landed in the PR body, the commit message, and all three code comments — the `test_the_dry_run_flag_is_passed_as_one_argument_when_set` comment now says outright *"the merge-base call site produced a clean argc=6 and was NOT buggy… The array rewrite is a shellcheck-cleanliness change, NOT a bug fix."* That is the right retraction: it names the wrong version so the next reader cannot re-derive the false claim from the guard's existence.

The guard correctly **stayed**, and it is still load-bearing — my `W8b` mutant (unconditional `sweep_args+=("$DRY_RUN")`) is still killed by the argc assertion, now at 8/9 to account for `--max-recoveries-per-run`.

The self-diagnosis (*"I typed a quoted `""` at a prompt… and attributed that to an unquoted expansion — the opposite case. I never ran the merge-base body itself"*) is the correct root cause, and it is the generalisable one: the probe was of a construct resembling the target rather than of the target.

*Cosmetic only, no action needed:* the corrected comment renders as ``a trailing ``.`` — the `<>` got eaten by an empty inline code span, and that line overruns.

## ✅ B2 — de-vacuified, and I checked it in **both** directions

You asked me to confirm the new test reads the parsed mapping rather than a re-stringified form. It does, and I verified behaviourally, not by inspection.

`_yaml` is `yaml.safe_load`, `_step`/`_steps` return the step **dicts**, so `assertNotIn("continue-on-error", step)` is key membership on a mapping.

- **Kill direction** — MUTANT tree `R5`, `continue-on-error: true` on the sweep step: now `FAILED … test_no_failure_swallowing_at_the_YAML_SEAM` (plus the seam class). Last round this mutant passed everything.
- **False-positive direction** — I built a discriminator that separates a real key check from a substring check: a step whose **`run:` body contains the literal text** `continue-on-error` in a comment, with **no such YAML key**. A re-stringified check would fail here; a genuine key check must pass. Result on that MUTANT tree: `2 passed`. So it is a real mapping read, not a renamed substring.

Splitting `|| true` (shell) from `continue-on-error` (YAML) into two tests is the right shape — each assertion now interrogates the surface its defect actually lives on.

## ✅ B3 — the seam is closed, including 12 evasions I designed against the new guard

All eight of my round-1 survivors are killed. More usefully, I wrote the guard's own bypasses and every one dies:

| evasion | result |
|---|---|
| `if: ${{ true }}` on the sweep step (constant operand) | killed (+actionlint) |
| `if: success()` on the sweep step | killed |
| sweep step **borrows** the one allowed guard `env.ORCHESTRATOR_APP_ID != ''` | killed — the step-label check holds |
| the allowed guard **drifts** `!=` → `==` | killed — whole-match holds |
| `if: false` on the one permitted step (the App-token mint) | killed |
| `continue-on-error` on *Checkout watchdog policy* / `if: false` on *Self-test policy* | killed — the loop covers every step |
| a decoy first step in the doorbell carrying `continue-on-error: true` | killed |
| `--repo` retargeted in the run body, and separately in `env.REPO` | both killed |
| watchdog permissions gutted / doorbell `actions: write`→`read` | both killed |
| watchdog `cancel-in-progress` flipped true / reweep interval cut below grace | both killed |

`test_the_doorbell_carries_no_guard_at_all` stated separately from the shared helper is the right call — it stops the App-token allowance ever leaking to the workflow that has no App-token step.

---

## 🎯 The decision you routed to me: **5 per fire is acceptable. Do not require the cross-sweep budget.**

I would not spend a round on this, and I would not want the `CAP` path opened by a timing PR. Reasoning, in the order it actually decides:

1. **The binding constraint on total churn is untouched.** The per-PR cap is 2 per 6 h, marker-persisted in PR comments, plus same-head idempotence — both survive across processes. The per-*run* bound only shapes **burst concentration inside one 4-minute window**; it does not change how much any PR, or the queue in aggregate, can be churned over 6 hours. Moving 2→5 therefore buys back nothing on the quantity that matters.
2. **Reaching 5 requires the outage the watchdog exists for.** Five *distinct* entries, each with a positively-established zero suite count, each past the 120 s grace, inside one 4-minute window. On a single-digit queue that is a repo-wide dispatch failure — the case where recovering all of them is the correct action, not the dangerous one. And an entry with genuinely zero suites is on its way to the platform's 60-minute `CI_TIMEOUT`, which demotes to `review:changes`; recovery is strictly the better outcome per entry.
3. **The exact-2 fix costs more than it buys.** A budget shared across sweeps means state threaded through `decide_entry`/`CAP` — soundness-critical decision logic that earns its own review round. Trading a *disclosed, bounded, tested* residual for an untested edit to the refusal path is the wrong direction.
4. **It is disclosed and it is pinned, not asserted.** The workflow comment states the residual as a residual. `test_the_per_fire_blast_radius_is_bounded` carries both halves — the product bound (`per_sweep × sweeps ≤ 5`) **and** a monotonic `per_sweep ≤ DEFAULT_MAX_RECOVERIES_PER_RUN`, so the budget may only ever tighten the script default. My mutants `MAX_RECOVERIES_PER_SWEEP: 2`, `: 5`, a hard-coded `2` on the command line, dropping the flag entirely, and `RESWEEPS: 9` are **all killed**, two of them by that test specifically.

If you want exact-2, take it as a separate bead against the `CAP` path with its own review — not as a rider here.

## ⚠️ The rollback signal is directionally right but **not sharp as worded** — one word fixes it

> *"If the doorbell is live and no `merge-group-watchdog` run appears within ~30 s of an `added_to_merge_queue` event, it is not working and should be reverted."*

"Appears" is ambiguous between **created** and **started**, and under this PR's own serialisation those diverge badly. Measured, from the run history:

| | watchdog run duration |
|---|---|
| pre-loop (`17:07:52Z`, `15:39:05Z`, `13:19:03Z`) | 15 s, 14 s, 17 s |
| this PR (`30383251472`, head `501eaac9`'s predecessor) | **4 m 19 s** |

So the shared `merge-group-watchdog` concurrency group goes from **~5% occupied to ~86% occupied by cron alone**. A ring arriving at a random moment therefore has ~86% chance of finding the group busy: its run is **created immediately with `status: queued`** but does not **start** for a mean ~2 min and up to ~4.3 min. Worse, `cancel-in-progress: false` keeps only one *pending* run per group — under a burst a legitimately created run can be cancelled before it ever starts, so it would never "appear" as a started run at all.

Read as "starts", that signal reverts a working doorbell most of the time. Two changes make it act-on-able:

- say **created**, not appears — `gh api ".../workflows/merge-group-watchdog.yml/runs?event=workflow_dispatch"` returns a `queued` run within seconds;
- make the **primary** signal the doorbell's own advisory check conclusion. `gh workflow run` exits non-zero on a 403/422, so a red *merge-group doorbell (advisory)* on the PR head is an immediate, unambiguous "the ring did not fire" — and it costs nothing to watch.

This is prose about post-merge operations, not shipped code, so I am not holding on it — but it is the one instruction someone will act on under pressure, and as written it would misfire.

## 📋 My three survivors — all loud, none silent

| mutant | effect | why not blocking |
|---|---|---|
| `timeout-minutes: 1` on the sweep step | truncates the loop after ~1 sweep | a step timeout concludes **`failure`**; the job reds. Degrades loudly, never silently inert |
| drop `set -euo pipefail` | none on the real path | the loop uses an explicit `if !`; an unset var errors regardless |
| job `timeout-minutes` 15 → 10 | none | ~4.5 min still fits |

The silent-inertness class — which is the one that matters and the one that was open last round — is now empty across everything I could think to throw at it.

---

Blockers 1–3 resolved and independently re-derived; the blast-radius widening I found is fixed to a bounded, disclosed, mutation-pinned ceiling; the two residuals are stated plainly rather than absorbed. The retraction in particular is the right kind — it names the error and its cause in the repo, where the next reader will hit it.

Passing on the merits. Please sharpen the rollback wording before you rely on it, and watch the doorbell's own check conclusion on the first live `enqueued`.

VERDICT: pass

Issue #2725, opened 2026-07-17
sq-mrrn4: Codify standing frontend policy: optional/rarely-used features load via ESM dynamic import (never the main bundle)


jeswr:
Maintainer directive on GH #1046 (2026-07-11): for the site/GUI, any feature that is (a) not yet built in and (b) uncertain-value / rarely used / bundle-size-increasing must be implemented behind an ESM dynamic import() so it is only fetched when the user actually invokes it (e.g. compressed-codec decoders loaded on first upload of that file type). Codify this in the frontend-design skill (skills/) + AGENTS.md so future site/GUI impl agents inherit it, and audit existing always-bundled optional decoders/features for retrofit candidates. Impl of the codec instance itself = sq-4ssz1.

Issue #2706, opened 2026-07-17
sq-lsp7k.9: IRI/label autocomplete index + rank-ordered defaults (engine index + workbench-wide completion)


jeswr:
track:both | competitor_parity: GraphDB autocomplete index + RDF Rank (used across their whole Workbench — the UX glue that makes it pleasant at 100M triples); Stardog/TopBraid have equivalents | crate_or_surface: sparq-text (or a small module next to it, opt-in) + sparq-server endpoint + gui/app completion sources | effort:M | value:high. Dedicated incremental-maintained index over IRIs + rdfs:label/skos:prefLabel (prefix + fuzzy), returning candidates ranked by a PageRank-style importance score (sparq-algos already computes PageRank — reuse, do not re-implement). Consumed by: SPARQL editor completion at scale, forms instance-autocomplete widget (F2 soft dep), graph-viz search/expansion ordering, MCP IRI-discovery tool. Cross-ref sq-mg5hk (SPARQL-callable analytics) — this bead only needs the library API, not the SPARQL surface.

Parent: #2565

Issue #2606, opened 2026-07-17
sq-sgu1.2: Wire fable-architect-drain + fable-soundness-verdict + model-tiered-scheduler-v2 + fable-lens-review workflows (schema on every producing stage)


jeswr:
Per research/fable-work-plan.md sec6.3. Evolve autonomous-scheduler.js; honor schema-guard trap + researcher-PR gotcha + arm-on-verdict. Fable appears only at decompose(front) + escalated-verdict; adds fable_implements path. [OPUS-4.8]

Parent: #2626

PR #4928, opened 2026-07-28
fix(orchestration): the cross-cutting ci/docs partitions route work without occupying it

jeswr:
> 🤖 SPARQ agent

## What

Declares the cross-cutting partitions `ci` and `docs` **non-reserving for occupancy**. They still
route work, are still valid candidate keys, are still derived and applied exactly as today, and a
*selected* candidate still reserves them for the tick. An in-flight PR or in-progress issue simply
no longer refuses a **new** dispatch onto an issue that declares them.

Head SHA: `f77d872149f74c26f9a7ea3bad8c85095cfe86fe`

## Why — the frontier is the binding constraint

Three consecutive production PLAN runs: `candidates=379 frontier=1 partition-deferred=378`. 99.7%
of attested candidates are refused for partition contention, and 49 of them declare `ci` as their
only area.

The reservation is aimed at the wrong axis. A PR enters the `ci`/`docs` partitions by **touching a
path** (`scripts/pr-area-labels.py` derives the label post-hoc from changed files) with **zero
admission control** — 14 open PRs co-hold `ci` and 10 co-hold `docs` right now, concurrently and
unserialized. The occupancy reservation therefore never prevented PR-vs-PR collision and cannot;
all it ever did was refuse to dispatch a new worker.

### Conflict-rate evidence

Open non-parked PRs holding each area, counting **holder pairs that share at least one changed file**:

| area | holder pairs | sharing ≥1 file | |
|---|---|---|---|
| `area:ci` | 120 | 6 | **5%** → exempt |
| `area:docs` | 66 | 2 | **3%** → exempt |
| `area:deps` | 3 | 3 | **100%** (all `Cargo.lock`) → **NOT** exempt |
| crate areas | — | — | **57.1%** (`research/crate-region-parallelism.md` §4) → **NOT** exempt |

`deps` and the crate areas are serialising real overlap. They stay.

## Measurement — before/after on the same live snapshot

Driven through the **real production predicate**: the registry's `dispatch.yml` `readiness` step
heredoc, exec'd verbatim against a live snapshot written by the real producer and this repo's
engine — not a reimplementation.

**Instrument validation:** the unpatched run reproduces the production census
(`candidates=376 frontier=1 partition-deferred=375` vs the cited `379 / 1 / 378`; the ±3 candidate
delta is snapshot drift, the frontier is identical).

| | candidates | frontier | partition-deferred | reservations (rows / key-pairs / distinct keys) |
|---|---|---|---|---|
| before | 376 | **1** | 375 | 59 / 79 / 44 |
| after | 376 | **3** | 373 | 47 / 55 / 42 |

* **`candidates=` unchanged: 376 → 376.** Candidacy is provably untouched — `packages_of` is not
  on the diff path.
* Released key set is **exactly `{ci, docs}`**; nothing else moved, nothing was gained.
* The two added rows declare exactly `area:ci` (#2606) and `area:docs` (#2725); the baseline row
  (#2706) is retained. The deferred lane is unchanged at frontier 3.
* Matches the counterfactual ladder exactly (baseline 1 → `ci` 2 → `ci`+`docs` 3), which is the
  signature of an **occupancy-side-only** exemption: each exempted partition adds one row, because
  a selected candidate still reserves its own areas.

## Design

One declaration, one consumption point:

* `NON_RESERVING_PARTITIONS = frozenset({"ci", "docs"})` in `scripts/ready-issues.py`, immediately
  after the partition algebra, with the measured table above written into the comment so the
  decision can be re-derived rather than guessed at.
* `reserves_partition(key)` decides on the **partition path**, not the raw string, so it agrees
  with `keys_conflict`: the live `ci-fragments` label travels with the `ci` partition instead of
  reserving a partition `ci` itself does not.
* Consumed in `_reserving_packages` — the function whose docstring already *is* the occupancy rule.
* **`__global__` and every degenerate key can never be exempted** (they map to the `()` root that
  contains every partition).
* **Fail-safe:** `non_reserving_partitions()` validates the declaration and returns the **empty**
  set — today's fully-reserving behaviour — for a non-collection, a bare string, a dict, a
  non-string entry, a blank entry, or `__global__`. One bad entry voids the whole declaration
  rather than being partially honoured. Never the reverse.

## Tests + mutation results

Contract in `scripts/tests/test_readiness_visibility.py::TestNonReservingCrossCuttingPartitions`
(15 tests), plus end-to-end tripwires in `scripts/ready-issues.py --self-test` — that one runs
inside the registry's dispatch tick against every target before it plans anything, so a widened
exemption fails at the gate the fleet actually passes through. Every assertion runs end-to-end
through `compute_ready`.

**Control run GREEN first** (116 → 131 tests, all suites green), then each mutant applied alone to
a fresh copy of the tree, `PYTHONDONTWRITEBYTECODE=1` throughout:

| mutant | result |
|---|---|
| M1 exemption set emptied | KILLED |
| M2 call site deleted in `_reserving_packages` | KILLED |
| M3 `reserves_partition` always True | KILLED |
| M4 **safety:** `deps` added to the exemption set | KILLED |
| M5 **safety:** a crate area added to the exemption set | KILLED |
| M6 **safety:** `reserves_partition` always False (exempt everything) | KILLED |
| M7 **safety:** `__global__`/degenerate path becomes exemptible | KILLED |
| M8 **fail-safe:** validation bypassed (raw declaration returned) | KILLED |
| M9 **fail-safe:** bad entry skipped instead of voiding the declaration | KILLED |
| M10 **scope:** exemption leaks onto the candidate side (`packages_of`) | KILLED |
| M11 **scope:** a selected candidate stops reserving its own areas too | KILLED |
| M12 exemption applied per-string instead of per-partition-path | KILLED |
| M13 the measured basis stripped from the declaration | KILLED |

**13/13, against a green control.** M1–M12 are each killed independently by *both* suites (M13,
comment-only, by the visibility suite). The fail-safe assertions carry an explicit known-positive
control (`test_a_wellformed_declaration_is_honoured_so_the_failsafe_is_not_vacuous`) so a
validator that rejected *everything* could not look perfect.

## What this does **not** fix

* **It does not change PR-vs-PR collision — that was never serialised.** Two PRs both touching
  `ci/` collided before this change and collide after it. The reservation only ever gated
  *dispatch*, and PRs enter these partitions by touching a path with no admission control. The
  5%/3% numbers are why that is acceptable, not evidence that anything was protecting them.
* **It does not touch `deps`.** Every live `deps` holder pair collides on `Cargo.lock`.
  Serialising it is correct. M4 exists to keep it that way.
* **It does not touch crate areas or `package_width`.**
* **It does not address the vocabulary gap** — 17 issue-side `area:` keys the PR deriver
  (`ci/area-labels.toml`) cannot emit, so those issues are keyed on a partition no PR can ever be
  attributed to. Separate defect, unchanged here.
* **`ci` remains the top-contended key** (`ci=54 → 53` in the census). A *selected* `ci` candidate
  still reserves `ci` for the tick, so per-tick dispatch width stays one worker per partition.
  That is deliberate and pinned by M11 — it is why the frontier moved 1→3 rather than 1→~50.
* ⚠️ **It is necessary but very likely NOT sufficient — there is a SECOND occupancy leg, in the
  registry.** `jeswr/agent-account-registry`'s `scripts/dispatch-claim.py::busy_packages_of_pulls`
  independently derives a busy-area union from open worker PRs' `area:*` labels and applies it at
  `filter_busy_area_items` (assemble) **and again** at `revalidate_items_against_live_pulls`
  (claim). It has **no** cross-cutting exemption. On this snapshot **16 open worker PRs contribute
  `ci` and 25 contribute `docs`** to that leg's busy set, and because that leg only ever *unions*
  areas onto a PR's own labels, more complete provenance can only widen it — so `ci`/`docs` stay
  busy there regardless. Unless the registry mirrors this exemption, the two extra rows PLAN now
  offers will be re-deferred one layer down. I could **not** measure that leg's before/after: it
  reads provenance records from the registry's `ledger` branch, which I do not have, and my
  partial reconstruction refuses **all** rows in *both* the baseline and the patched case
  (a `__global__` reservation from an unprovenanced PR) — i.e. it fails to reproduce production
  and its numbers are not reportable. Filed as a follow-up; the correct next step is the mirror
  change in the registry, measured there.

## Scope

`scripts/ready-issues.py` + `scripts/tests/test_readiness_visibility.py` only. Both are already
`paths:` triggers on `routing-self-tests.yml` (both `pull_request` and `push`) and both are already
invoked in its run block, so no workflow change is needed and this diff cannot land without the
gate that tests it running.

## Gate

`python3 scripts/preflight.py` → **PASS**. `routing-validate --self-test`, `route-resolve
--self-test`, `ready-issues --self-test`, `dispatch-plan --self-test`,
`test_readiness_visibility.py` (131 tests) all green. `typos` clean; no added line over 100 chars.

---

I authored this change and must not review it. **Not armed, not merged** — an independent reviewer
owns the verdict.

### DIFF
--- a/scripts/ready-issues.py
+++ b/scripts/ready-issues.py
@@ -186,6 +186,78 @@ def keys_conflict(a, b, roots=None):
     """
     pa, pb = partition_path(a, roots), partition_path(b, roots)
     return pa[:len(pb)] == pb or pb[:len(pa)] == pa
+
+
+# ---------------------------------------------------------------------------
+# NON-RESERVING (cross-cutting) PARTITIONS — THE ONE PLACE THIS IS DECLARED.
+#
+# [OPUS-5 2026-07-28] These partitions still ROUTE work and are still valid candidate keys — a
+# candidate declaring `area:ci` is derived, counted and dispatchable exactly as before. They
+# simply do not OCCUPY: an in-flight PR or in-progress issue holding one of them no longer blocks
+# a new worker from being dispatched onto an issue that declares it. See `_reserving_packages`.
+#
+# THE MEASURED BASIS (live sparq snapshot, open non-parked PRs holding each area, counting
+# holder PAIRS that share at least one changed file):
+#
+#     area          holder pairs   sharing >=1 file
+#     area:ci            120          6   ( 5%)   -> exempt
+#     area:docs           66          2   ( 3%)   -> exempt
+#     area:deps            3          3   (100%)  -> NOT exempt (every pair collides on
+#                                                    Cargo.lock; serialising it is correct)
+#     crate areas          -          -   (57.1%) -> NOT exempt
+#                                                    (research/crate-region-parallelism.md §4)
+#
+# So the reservation on `ci`/`docs` was refusing ~99% of a partition-starved frontier to prevent a
+# 3-5% file collision, while `deps` and the crate areas are serialising real overlap and stay.
+# Measured counterfactual on that same snapshot, through this engine: baseline frontier 1;
+# `ci` non-reserving 2; `ci`+`docs` non-reserving 3; adding `deps` would give 4 (not taken).
+#
+# It is also aimed at the right axis only in part, and the comment says so rather than
+# overselling it: `ci`/`docs` reservation never serialised PR-vs-PR contention at all (14 open PRs
+# co-hold `ci` and 10 co-hold `docs` right now, concurrently — a PR enters those partitions by
+# TOUCHING A PATH, with zero admission control). All it ever did was refuse dispatch.
+#
+# FAIL-SAFE DIRECTION: `non_reserving_partitions()` validates this declaration and returns the
+# EMPTY set — i.e. today's fully-reserving behaviour — for anything malformed. Never the reverse.
+NON_RESERVING_PARTITIONS = frozenset({"ci", "docs"})
+
+
+def non_reserving_partitions(declared=None):
+    """`NON_RESERVING_PARTITIONS`, VALIDATED. Anything malformed degrades to RESERVING.
+
+    A wrong answer here is asymmetric: too SMALL a set costs dispatch width (today's behaviour,
+    which the fleet has been running), while too LARGE a set silently un-serialises real work.
+    So the whole declaration is voided by a single bad entry rather than partially honoured, and
+    `GLOBAL` (and any degenerate key, which `partition_path` maps to the `()` root that contains
+    every partition) can NEVER appear in it — exempting the root would exempt everything, which
+    is exactly the "fail toward exempt everything" outcome this must not have.
+    """
+    raw = NON_RESERVING_PARTITIONS if declared is None else declared
+    if isinstance(raw, str) or not isinstance(raw, (set, frozenset, list, tuple)):
+        return frozenset()
+    names = set()
+    for name in raw:
+        if not isinstance(name, str) or not name.strip() or name.strip() == GLOBAL:
+            return frozenset()
+        if partition_path(name.strip()) == ():        # degenerate key -> the containing root
+            return frozenset()
+        names.add(name.strip())
+    return frozenset(names)
+
+
+def reserves_partition(key, exempt=None):
+    """Whether an OCCUPANT declaring `key` reserves it, or merely routes on it.
+
+    Expressed on the PARTITION PATH, not the raw label, so it agrees with `keys_conflict`: every
+    key that resolves INTO the `ci` partition (`ci`, and e.g. the live `ci-fragments`) is exempt
+    together with it. A per-string exemption would leave `ci-fragments` reserving a partition that
+    `ci` itself does not, which reads as a bug to the next person and behaves like one.
+    """
+    path = partition_path(key)
+    if not path:                       # GLOBAL / degenerate: contains everything, never exempt
+        return True
+    exempt = non_reserving_partitions() if exempt is None else exempt
+    return path[0] not in exempt
 # --- open blockers: NATIVE GitHub dependencies UNIONED with the legacy body markers -------------
 # [OPUS-5] Until this landed, BOTH readers of "is this issue blocked" (this file and the registry's
 # dispatch.yml planner step) derived `open_blockers` ONLY by regexing `Blocked-by: #NN` out of the
@@ -305,8 +377,16 @@ def _reserving_packages(labels):
     cross-cutting paths), so 9 open PRs still declare nothing, and making those 9 seize
     `__global__` would reproduce the measured whole-fleet stall. Only the stated reason needed
     correcting; leaving a false premise in place is how a correct rule gets "fixed" back.
+
+    [OPUS-5 2026-07-28] ...and of the areas it declares, only the RESERVING ones — the
+    cross-cutting `NON_RESERVING_PARTITIONS` (`ci`, `docs`) route work without occupying it. That
+    declaration, its measured basis and its fail-safe live in ONE place above; nothing else in
+    this file knows the names. This is the OCCUPANCY half ONLY: `packages_of` (the candidate side)
+    is untouched, so candidacy is unchanged, and a SELECTED candidate still reserves its own areas
+    through `compute_ready`'s `reserve(pkgs, it)` — per-tick width stays one worker per partition,
+    which is why the measured effect is +1 frontier row per exempted partition and not +49.
     """
-    return declared_packages(labels)
+    return {key for key in declared_packages(labels) if reserves_partition(key)}
 
 
 def has_role(labels):
@@ -874,6 +954,57 @@ def _self_test():
     check("flatten pages retains PRs", _flatten_pages(
         [[{"number": 1}, {"number": 2, "pull_request": {}}], [{"number": 3}], "junk", [None]]),
         [{"number": 1}, {"number": 2, "pull_request": {}}, {"number": 3}])
+    # ---------------------------------------------------------------------------------------
+    # [OPUS-5 2026-07-28] NON-RESERVING cross-cutting partitions. END-TO-END through
+    # compute_ready, and run HERE and not only in scripts/tests/ because the registry's
+    # dispatch.yml executes THIS --self-test against every target before it plans anything — so
+    # a tree whose exemption set has been widened to a colliding partition fails at the gate the
+    # fleet actually passes through. `scripts/tests/test_readiness_visibility.py::
+    # TestNonReservingCrossCuttingPartitions` carries the full contract.
+    # ---------------------------------------------------------------------------------------
+    def held_board(area, held_by=None):
+        return [pr(70, [f"area:{held_by or area}"]),
+                iss(20, R + ["priority:P1", f"area:{area}"])]
+
+    def offered(area, held_by=None):
+        return [i["number"] for i in compute_ready(held_board(area, held_by),
+                                                   conflict_log=quiet)]
+
+    check("a PR holding area:ci no longer refuses a ci-only candidate", offered("ci"), [20])
+    check("...same for area:docs", offered("docs"), [20])
+    check("...and for a key that resolves INTO the ci partition", offered("ci", "ci-fragments"),
+          [20])
+    # The SAFETY half. deps: 3 of 3 live holder pairs collide, all on Cargo.lock. Crate areas:
+    # 57.1% (research/crate-region-parallelism.md §4). Widening the set to either is the mutation
+    # this line exists to kill.
+    check("area:deps STILL reserves (every live deps pair collides on Cargo.lock)",
+          offered("deps"), [])
+    check("crate areas STILL reserve", offered("sparq-core"), [])
+    check("sub-crate containment still reserves through the exemption",
+          offered("sparq-core-store", "sparq-core"), [])
+    check("the global partition can never be exempted", offered("sparq-core", GLOBAL), [])
+    # The FAIL-SAFE, both directions: a malformed declaration degrades to TODAY's behaviour, and
+    # a well-formed one is honoured — without the second line a fail-safe that voided everything
+    # would look perfect.
+    _declared = NON_RESERVING_PARTITIONS
+    try:
+        for _broken in (None, "ci", 7, {"ci": True}, {"ci", 7}, ["ci", ""], {GLOBAL}):
+            globals()["NON_RESERVING_PARTITIONS"] = _broken
+            check(f"malformed exemption {_broken!r} falls back to RESERVING", offered("ci"), [])
+        globals()["NON_RESERVING_PARTITIONS"] = frozenset({"ci"})
+        check("...and a well-formed exemption is honoured (fail-safe is not vacuous)",
+              offered("ci"), [20])
+    finally:
+        globals()["NON_RESERVING_PARTITIONS"] = _declared
+    # SCOPE: candidacy untouched, and a SELECTED candidate still reserves — per-tick width stays
+    # one worker per partition, which is why the live frontier moved 1 -> 3 and not 1 -> ~50.
+    check("candidate keying for ci/docs is unchanged", (packages_of({"area:ci"}),
+                                                        packages_of({"area:docs"})),
+          ({"ci"}, {"docs"}))
+    check("a selected ci candidate still reserves ci for the tick",
+          [i["number"] for i in compute_ready(
+              [iss(20, R + ["priority:P0", "area:ci"]), iss(21, R + ["priority:P1", "area:ci"])],
+              conflict_log=quiet)], [20])
     print("ready-issues self-test", "PASSED" if ok else "FAILED")
     return 0 if ok else 1
 
--- a/scripts/tests/test_readiness_visibility.py
+++ b/scripts/tests/test_readiness_visibility.py
@@ -1389,5 +1389,162 @@ class TestRoutingSelfTestWorkflowWiring(unittest.TestCase):
         self.assertRegex(self.SOURCE, r"(?m)^  merge_group:")
 
 
+class TestNonReservingCrossCuttingPartitions(unittest.TestCase):
+    """`ci` / `docs` ROUTE work but do not OCCUPY it — and everything else still does.
+
+    [OPUS-5 2026-07-28] The dispatch frontier is the binding throughput constraint: three
+    consecutive production PLAN runs read `candidates=379 frontier=1 partition-deferred=378`, i.e.
+    99.7% of attested candidates refused for partition contention. Driving the REAL production
+    predicate (the registry's dispatch.yml `readiness` step, this repo's engine, a live snapshot)
+    reproduced that at `candidates=376 frontier=1`, and made `ci`+`docs` non-reserving takes it to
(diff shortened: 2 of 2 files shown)


jeswr:
> 🤖 SPARQ agent

Follow-up for the second occupancy leg filed as #4929 — the registry's `busy_packages_of_pulls` (assemble + claim) independently reserves `ci`/`docs` and has no cross-cutting exemption, so this PR is necessary but very likely not sufficient on its own. Detail and the measurement I could *not* make are in that issue.

jeswr:
> 🤖 SPARQ agent

**Independent adversarial review** — I did not write this change. Review and comment only; **not armed, not merged, no labels touched.**

Head SHA reviewed: **`f77d872149f74c26f9a7ea3bad8c85095cfe86fe`** (merge-base `e9dd957b6`; `origin/main` was `46c89175d` at review time).
Everything below was re-derived in an isolated worktree with `PYTHONDONTWRITEBYTECODE=1` throughout. I accepted no assertion from the PR body without executing it.

---

## 1. Is the exemption confined to occupancy? — **yes, proven on a live board, not by assertion**

Fresh snapshot of my own, cursor-paginated (`gh api --paginate`, no `--limit`): **1761 open rows — 89 PRs / 1672 issues**. Driven through `ready_candidates` / `compute_ready` / `unit_reservations` on *both* trees:

| | candidates | frontier | partition-deferred | reservations (rows / key-pairs / distinct keys) |
|---|---|---|---|---|
| base `e9dd957b6` | 390 | 3 | 387 | 70 / 102 / 46 |
| head `f77d8721` | **390** | **5** | 385 | 54 / 66 / 44 |

* `candidates` unchanged — and stronger than the count: the **candidate number set and the full candidate→key map are byte-identical** between the two trees.
* Frontier is **monotone**: no row lost, exactly two added — **#2606** (`priority:P1, role:ci, status:ready, area:ci`) and **#2725** (`priority:P2, role:impl, status:ready, area:docs`). Same two issues the PR names.
* Released key set = **exactly `{ci, docs}`**; added keys = **∅**.

My baseline frontier is 3, not the PR's 1 — different snapshot, two days later. The *delta* (+1 row per exempted partition, and the same two rows) reproduces exactly.

**M11 verified, and it dies for the right reason.** I wrote my own version — filtering the selection-site `reserve(pkgs, it)` through `reserves_partition` — and it dies in both suites with the self-test naming the leak directly:

```
FAIL a selected ci candidate still reserves ci for the tick: [20, 21] (want [20])
```

That is the leak assertion itself, not a proxy. **M10** likewise: `FAIL candidate keying for ci/docs is unchanged: ({'__global__'}, {'__global__'}) (want ({'ci'}, {'docs'}))`.

## 2. Can the exemption widen beyond `{ci, docs}`? — **no, and I enumerated it rather than reasoning about it**

`reserves_partition` keys on `partition_path(key)[0]`, so I asked the tree:

* Of the **91 workspace roots**, exactly `ci` and `docs` are non-reserving. **0 of 67 crates.**
* Of the **99 live `area:*` labels**, exactly **3** stop reserving: `ci`, `ci-fragments`, `docs`. The other 96 are unaffected.
* `_resolve` is longest-recognised-ancestor-**first**, so a hypothetical `crates/ci-foo` would be its own root → `('ci-foo',)` → **not** captured. A crate area cannot land on an exempt partition.

The only reachable widening is a *non-root* `ci-*` / `docs-*` label — which already conflicts with `ci`/`docs` under `keys_conflict` today, so the partition-path form is coherent with the existing algebra rather than an expansion of it. `M12` (per-string instead of per-path) dies on `ci-fragments`.

Two residuals, neither blocking:
* A crate literally **named** `ci` or `docs` would be captured. Far-fetched (both names are taken by top-level dirs) but it is the one structural hole.
* Pleasant robustness result: the exemption does **not** depend on `ci`/`docs` being recognised roots — the head-segment fallback yields `('ci',)` either way. So **#4925**, which changes `workspace_roots`, cannot silently defeat or widen this.

## 3. Is `__global__` genuinely unexemptable? — **yes, via two independent guards**

I forced the hostile case directly rather than going through the validator:

```
reserves_partition(GLOBAL, exempt={GLOBAL}) -> True
reserves_partition("",     exempt={GLOBAL}) -> True
non_reserving_partitions({'  __global__  '}) -> frozenset()
non_reserving_partitions({'ci','__global__'}) -> frozenset()
```

`reserves_partition` short-circuits on the empty path **before** consulting the exempt set, and the validator can never emit `GLOBAL` anyway.

**Honest note on the author's M7.** My formulation of it — making the empty-path branch consult the exempt set — **survived**. That is correct: it is an *equivalent mutant*, because the validator makes the exempt set incapable of containing `GLOBAL`. Same for dropping the redundant explicit `name.strip() == GLOBAL` check (subsumed by the `partition_path(...) == ()` check). Consequently the first two assertions of `test_the_global_partition_can_never_be_exempted` can only red if **both** guards are removed. That is defense-in-depth, not vacuity — and I confirmed its third assertion (`non_reserving_partitions({"-"}) == frozenset()`) does die on its own mutant. Worth knowing that "M7 KILLED" is load-bearing on the *validator*, not on that branch.

## 4. Fail-safe direction on **partial** malformation — **voids, does not partially honour**

The case that matters (one good entry + one bad), tested directly. **9/9 void to `frozenset()`**:

`{'ci',7}` · `['docs','']` · `('ci',None)` · `{'ci','  '}` · `{'ci','__global__'}` · `{'ci','docs','__global__'}` · `{'ci','-'}` · `{'docs',b'ci'}` · `{'ci',()}`

**M9 confirmed** (`continue` instead of `return frozenset()`) — dies in both suites, and the self-test names the *partial* case specifically:

```
FAIL malformed exemption {'ci', 7} falls back to RESERVING: [20] (want [])
```

It reds precisely because the surviving good entry exempted `ci`. Right reason. And the fail-safe is not a reject-everything: `{' ci '} -> {'ci'}`.

## 5. Independent mutation run — **control green first, then a known positive**

19 mutants, each applied **alone** to a fresh copy of the head tree, both suites run, `PYTHONDONTWRITEBYTECODE=1`.

* **`C0` unmutated control: GREEN in both suites.** The baseline is not red, so nothing below is a phantom kill.
* **Known positive** (a defect I chose, unrelated to this feature — `keys_conflict` forced `False`): **dies in both**, `FAIL ready order: [2, 3, 12, 1, 11] (want [2, 3, 12])`. Instrument validated against a known answer.
* Of the author's 13: **12 killed as claimed.** M7 survives in my formulation as an equivalent mutant (§3). I added 5 of my own — 3 killed (`non-str entry accepted`, `validator rejects everything`, `degenerate guard dropped`), 2 equivalent-mutant survivals on redundant guards.

**Are the expectations correct, or do they restate the code?** I re-ran every new test's board against the **pre-change** tree:

| board | base | head |
|---|---|---|
| PR holds `area:ci`, ci-only candidate | `[]` | `[20]` |
| PR holds `area:docs` | `[]` | `[20]` |
| in-progress issue holds `ci` | `[]` | `[20]` |
| `ci-fragments` holder | `[]` | `[20]` |
| released-set probe | `{ci, deps, docs, sparq-core}` | `{deps, sparq-core}` |
| SAFETY: deps / crate / sub-crate / global | `[]` | `[]` |
| `packages_of(area:ci)` | `{'ci'}` | `{'ci'}` |

The headline expectations encode a **genuine behaviour change**, not a restatement — no row here was written to match what the code happens to do. The four SAFETY rows are invariants by construction (identical at base and head) and each is killed by a widening mutant (M4/M5), so they are not vacuous. The `packages_of`/`partition_path` expectations are independently true on the base tree, which is exactly what makes them a valid scope pin.

One scoping note: `test_the_declaration_records_its_measured_basis` is a **comment-presence ratchet**. It pins that the numbers are *stated*, not that they are *true* — correctly scoped, but M13's kill is not evidence for the measurement.

## 6. Does the conflict-rate evidence support the conclusion? — **it replicates; the inference survives the survivorship objection for a reason that does not depend on the rate**

Re-derived two days later on a partly different population, `--paginate` throughout, every file list cross-checked against the authoritative `changed_files` (**0 truncations**):

| area | holders | pairs | sharing ≥1 file | mine | PR claims |
|---|---|---|---|---|---|
| `area:ci` | 19 | 171 | 9 | **5.3%** | 5% |
| `area:docs` | 17 | 136 | 3 | **2.2%** | 3% |
| `area:deps` | 4 | 6 | 6 | **100%** (all `Cargo.lock`) | 100% |

On survivorship: it is a real bias and it points the right way (closed conflicting PRs leave the sample), but it is not load-bearing here. Two independent snapshots agree within a point — and more decisively, this population is **already concurrently open and unserialised**. A PR enters `ci`/`docs` by touching a path post-hoc with no admission control, so this is a *direct observation under the counterfactual*, not an extrapolation to it. And `docs` is not merely low-**volume**: 17 holders is more than any crate area on the board (max 2), so 2–3% is measured over a large pair population.

Three caveats I want on the record, none blocking:

* **The metric is textual.** This same file's partition-algebra comment says semantic collisions "are invisible to git". Two agents editing different workflow files in one gate aggregator is exactly what this cannot see, and `ci` spans `.github/**`, `scripts/**`, `.claude/**`, `orchestration/**`. I judge the marginal exposure small — **+1 dispatched worker against 19 already-concurrent `ci` PRs** — and the revert is one `frozenset`. But 5.3% should not be read as a semantic-collision rate.
* **`ci/area-labels.toml` justifies its `.beads/**` → `ci` row** as deliberately serialising "the single most merge-conflict-prone file in the repo". That rationale is now unenforced for `ci`. Moot on this snapshot: **0 of 19** `ci` holders touch `.beads/`.
* **A live instance of the permitted class, on this PR itself:** **#4925 × #4928** share `scripts/ready-issues.py` *and* `scripts/tests/test_readiness_visibility.py`, and #4925 changes `workspace_roots`. Both are already open — which is the PR's own argument, demonstrated on itself.

## 7. Honesty about limits — **accurate; states the limitation rather than implying a realised gain**

"What this does not fix" names the registry's second leg by file *and* function, reports the 16/25 contributing PRs, explains why its partial reconstruction refuses all rows in both arms and is therefore not reportable, and declines to give a number. **#4929** exists with a matching title. Per the review brief that is a scope caveat, not a defect, and I am not failing on it.

## Non-blocking corrections

1. **Test count is wrong.** The body says "116 → 131 tests" and the Gate section says "(131 tests)". Measured: merge-base `e9dd957b6` and current `origin/main` **both have 101** test methods; this head has **116**. So it is **101 → 116** — the +15 was added to the post-change figure. The gates themselves are genuinely green: `gate` and `Routing contract self-tests` are both `SUCCESS` **against head_sha `f77d872149f74c26f9a7ea3bad8c85095cfe86fe`** (I checked `commits/<sha>/check-runs`, not the PR rollup), and `python3 scripts/preflight.py` returns `PASS` on my tree. Only the number needs fixing.
2. `occupies_area`'s docstring says the carve-out is "exactly co-extensive with the registry leg's own". Still true of the *machine-park* carve-out specifically, but the two legs now diverge on `ci`/`docs` — a reader may take it more broadly than intended.
3. `compute_ready`'s `in_progress_packages` pre-seed (`reserve({pkg}, None)`) bypasses `reserves_partition`. Inert today — no production caller passes it, and the direction is over-reserving — but if the registry ever starts passing it, the exemption is silently defeated there. Worth a line on #4929.
4. `pr_row_occupancy_probe` still passes: `area:__pr_probe__` → `('__pr_probe__',)`, reserving. Confirmed by running it.

## Assessment

The change does what it says and only what it says. Candidacy is provably untouched on a live 1761-row board; occupancy releases exactly `{ci, docs}` and nothing else; a selected candidate still reserves; `__global__` is unreachable through two independent guards; partial malformation voids the whole declaration in all 9 shapes I could construct. The mutation claim reproduces against a **verified-green control** and a **verified-red known positive**, and the headline expectations are genuine behaviour changes rather than restatements of the implementation. The conflict-rate evidence replicates on an independent sample, and the strongest form of its argument does not depend on the rate at all. The limitation about the registry's second leg is stated honestly and is out of scope here.

VERDICT: pass

jeswr:
> 🤖 SPARQ agent

Independent review returned **`VERDICT: pass`** at head `f77d872149f74c26f9a7ea3bad8c85095cfe86fe`.
**Deliberately NOT arming yet** — the reason is composition, not this PR.

## Why the hold

This PR and **#4925** modify the same two files:

```
scripts/ready-issues.py
scripts/tests/test_readiness_visibility.py
```

**#4925 is already in the merge queue** (position 2), and it changes `workspace_roots()` — which is
upstream of the partition path this PR hooks. The review verified this PR against a base that does
**not** contain #4925, so the measured result describes a tree that will not exist by the time this
merges.

Two individually-sound changes whose interaction nobody has measured is a failure mode I have hit
before. The plan: let #4925 land, rebase this PR onto it, re-verify the before/after frontier on the
**combined** tree, then arm. If the combined measurement differs from `1 → 3`, that is worth knowing
before it merges rather than after.

## What the review established (independently re-derived, not accepted)

- **Occupancy-only.** Own 1761-row live snapshot (89 PRs / 1672 issues, `--paginate`, no `--limit`):
  candidates **390 → 390**, candidate *number set* and full *candidate→key map* **byte-identical**.
  Frontier moved monotonically (no row lost); added rows exactly #2606 (`area:ci`) and #2725
  (`area:docs`); released key set exactly `{ci, docs}`, added keys **∅**.
- **Cannot widen** — enumerated rather than argued: of 91 workspace roots exactly `ci`/`docs` are
  non-reserving, **0 of 67 crates**; of 99 live `area:*` labels exactly 3. `_resolve` is
  longest-ancestor-first, so a hypothetical `crates/ci-foo` is its own root.
- **`__global__` unexemptable** via two independent guards (forcing `exempt={GLOBAL}` still returns
  `True`).
- **Partial malformation voids the whole set** — 9/9 one-good-one-bad shapes → `frozenset()`.
- **Mutation claim reproduced honestly**: control green *first*, then a chosen known-positive dies —
  instrument validated. 12 of 13 killed; **the 13th (M7) is an equivalent mutant**, so "M7 KILLED" is
  load-bearing on the validator rather than that branch, and the reviewer said so instead of
  inheriting the count.
- **Non-circularity checked**: every new test board re-run against the pre-change tree goes
  `[]` → `[20]`, so the expectations encode a real behaviour change rather than restating the code.
- **Evidence replicates on a fresh population two days later**: ci 5.3%, docs 2.2%, deps 100% all
  `Cargo.lock`.

## Corrections and accepted limits

- ⚠️ **The PR body's test count is wrong** — it says "116 → 131"; the actual figure is **101 → 116**
  (merge-base and current `origin/main` both have 101). Please correct it; the gates are genuinely
  green.
- The 5%/3% collision metric is **textual**. Semantic collisions are invisible to git, as the file's
  own comment notes. Marginal exposure is small (+1 worker against 19 already-concurrent `ci` PRs).
- `ci/area-labels.toml` justifies `.beads/** → ci` as deliberately serialising the repo's most
  conflict-prone file. That rationale is now unenforced — moot today (0 of 19 `ci` holders touch
  `.beads/`), but it is a stated intent this PR silently drops.
- `in_progress_packages` pre-seed bypasses `reserves_partition` — inert today (no production caller);
  worth a line on #4929.

PR #4982, opened 2026-07-29
fix(ci): selection-alarm must admit runs whose JOBS failed, not runs GitHub called 'failure' (#4965)

### LINKED ISSUE #4965: selection-alarm is INERT: it fires only on conclusion=='failure', but every nightly since 2026-07-19 concludes 'cancelled' (266/267 runs skipped, silently)
> 🤖 SPARQ agent

`selection-alarm` fires only when a CI run on main concludes **`failure`**. The nightly full-matrix run has not concluded `failure` since **2026-07-19** — every one since concludes **`cancelled`**, which the predicate does not admit. The alarm has been skipped **266 of 267 runs** and has not acted since the day the shape changed.

## Measurement

`repos/sparq-org/sparq/actions/workflows/ci.yml/runs?event=schedule`, `--paginate`, all-time:

```
35  cancelled
 7  success
 3  failure      <- last one 2026-07-19
```

The ten most recent nightlies:

```
2026-07-28 cancelled    2026-07-23 cancelled
2026-07-27 cancelled    2026-07-22 cancelled
2026-07-26 cancelled    2026-07-21 cancelled
2026-07-25 cancelled    2026-07-20 cancelled
2026-07-24 cancelled    2026-07-19 failure   <- the last time this alarm could fire
```

| metric | value |
|---|---|
| `selection-alarm` runs (all-time) | **267** |
| job **skipped** | **266** |
| job executed | 1 (errored) |
| findings produced in window | **0** |

## The predicate

`.github/workflows/selection-alarm.yml:74-76` on `main`:

```yaml
if: >-
  ...
  (github.event.workflow_run.conclusion == 'failure' &&
```

`cancelled` and `timed_out` appear nowhere in the condition.

## Known positive — it would have had real work to do

Nightly run `30333511110` (2026-07-28) contained **4 genuinely failed jobs** alongside 31 cancelled ones. Because a single job cancellation makes the *run* conclude `cancelled`, the alarm's job was skipped and those 4 failures were never correlated against the selection set.

Driving the alarm's own functions from `scripts/ci_selection_alarm.py` against that run's real job list:

- `FAILED_CONCLUSIONS = {'failure', 'timed_out'}` — the **script** is happy to accept these, so `gather_failed_jobs` returns all 4;
- `job_crate_tokens` maps them onto real workspace crates: `sparq-mpc`, `sparq-fedplan`, `sparq-reason` (×2).

So the correlation would have had non-empty input on a real failure and produced findings for any of those crates that `ci-select` had skipped. **The gate that stopped it is the workflow `if:`, not the script.** The Python is correct; the YAML is where the alarm dies.

**Verdict: INERT.**

## Why nobody noticed

A **skipped job is never red.** The workflow's documented fail-loud posture does not apply to the population it is now permanently in — 266 silent skips over 9 days, each looking exactly like a healthy no-op.

This is the same blind spot `formal-alarm.yml` was explicitly built to close. Its own header records the precedent: *"miri.yml was CANCELLED … for 22 consecutive nights … and nothing noticed."* That lesson was applied to the formal lane and not to the selection lane.

## Suggested fix

Admit `cancelled` and `timed_out` in the `if:` at `:76`, and let the script's own `FAILED_CONCLUSIONS` filter decide which *jobs* within the run are genuine failures — it already does exactly that, and it already excludes the cancelled ones. The run-level conclusion is the wrong place to filter, because on a large matrix it is dominated by fail-fast cancellation of unrelated jobs.

Worth adding as a regression test: assert the workflow `if:` admits every conclusion in the script's `FAILED_CONCLUSIONS`, so the two cannot drift apart again.

## Upstream, out of scope but load-bearing

The nightly full-matrix backstop itself has not completed cleanly in **9 days** (35 `cancelled` of 45 all-time). Fixing the alarm will surface that; it will not fix it.

Found during a survey of whether each orchestration control has ever actually *acted*, as distinct from having *run*. Most sparq-side controls are healthy — `auto-arm` (66 arms), `rearm-sweeper` (3), `bead-autoclose` (5 closes), `pr-backlog` (both halves), `retriage` (641 label removals), `merge-queue-feedback` (20), and `formal-alarm` is a genuinely correctly-idle alarm that has fired before and would fire again. This is one of two that cannot.

jeswr:
> 🤖 SPARQ agent

Closes #4965. Head SHA `8ab249d9b12e68f8f61e9278d3650a24bddc6110`.

## The defect, and where it actually lived

The alarm's Python was always correct. The job-level `if:` in `.github/workflows/selection-alarm.yml` is where it died — it admitted only `github.event.workflow_run.conclusion == 'failure'`, and **fail-fast cancellation means a nightly carrying real failures concludes `cancelled`**.

MEASURED, `--paginate` over all 45 nightlies: **35 `cancelled`, 7 `success`, 3 `failure`** — last `failure` 2026-07-19, the exact day the alarm last acted. **266 of 267 alarm runs job-SKIPPED**, and a skipped job is never red, so nine days of inertness looked exactly like a healthy no-op.

## Known positive, driven live

Nightly run `30333511110`: run conclusion `cancelled`, **4 genuinely `failure` jobs among 31 cancelled**. Running the *fixed* script against the real run, with real `gh` + real git replay + real `cargo metadata`:

```
[dry-run] WOULD file (kind=crate, key=crate:sparq-fedplan)
[dry-run] WOULD file (kind=crate, key=crate:sparq-mpc)
[dry-run] WOULD file (kind=crate, key=crate:sparq-reason)
exit 0
```

Exactly the three crates the four real failures name. Before this change the job never ran at all.

## The fix shape, and why not just "add `cancelled` to the `if:`"

Adding `cancelled` alone would have been worse than the bug. A `cancelled` run is the *common* case (35 of 45), and the script raises `AlarmError` on an empty failed-job set — so the alarm would have gone RED on most nights for runs where **nothing was actually broken**. A nightly alarm that cries wolf gets muted, which is the same inertness by a slower route.

A run-level conclusion cannot express "some job failed", and a job `if:` cannot see the job list, so **no expression there can be precise**. So:

- the conclusion filter is **removed from the YAML entirely** — there is no longer any conclusion literal that can drift away from the script;
- the admission is made in `scripts/ci_selection_alarm.py`, over the real jobs API, by the **same `FAILED_CONCLUSIONS` predicate the analysis already uses**:
  - ≥1 genuinely-failed job → correlate (behaviour unchanged);
  - 0 failed jobs on a `cancelled`/`success` run → **quiet exit 0** (the noise half);
  - 0 failed jobs on a `failure`/`timed_out`/unknown run → **still RED, fail-loud** (`STRICT_EMPTY_CONCLUSIONS`). The anti-masking invariant is preserved, not traded away to buy the quiet branch.

Cost of widening: ~1 extra short run per night (the nightly is the only schedule/dispatch CI run on main).

## Acceptance criteria

| # | criterion | result |
|---|---|---|
| 1 | acts on the known positive | **PASS** — live run above; and `TestKnownPositiveRun30333511110` pins it hermetically |
| 2 | does **not** fire on a clean cancelled run | **PASS** — `TestQuietOnACleanCancelledRun` |
| 5 | control suite GREEN before trusting mutations | **PASS** — 17/17 at baseline, then 29/29; baseline re-verified GREEN *on each copied tree* |

### Mutation results — 7/7 killed, each naming a row

| mutant | killed by |
|---|---|
| M1 revert `if:` to `conclusion == 'failure'` | `test_live_if_admits_the_known_positive_cancelled_nightly` |
| M2 `FAILED_CONCLUSIONS` += `cancelled` | `test_gather_failed_jobs_admits_only_the_four_genuine_failures` |
| M3 quiet branch deleted | `test_clean_cancelled_run_is_a_quiet_no_op` |
| M4 fail-loud invariant deleted | `test_failure_concluded_run_with_no_failed_job_is_still_LOUD` |
| M5 `if:` widened to `true` | `test_live_if_still_rejects_the_runs_that_are_not_the_oracle` |
| M6 **instrument sabotage** (`eval_if` always True) | `test_evaluator_reproduces_the_old_bug` |
| M7 `gather_failed_jobs` never reports a failure | known-positive tests |

## The instrument is validated against a known answer

`TestWorkflowAdmissionSeam` evaluates the **live** `if:` — parsed structurally out of the YAML, so a comment mentioning `conclusion` can neither satisfy nor break it — against real event payloads. Its evaluator is validated first: the **verbatim pre-fix expression must reject the cancelled known positive and admit a `failure` nightly**, exactly as GitHub was observed to behave across 267 runs. An instrument that cannot reproduce the bug cannot certify the fix (M6 confirms this control is live). The evaluator is fail-closed: an expression shape outside its grammar raises rather than being judged admissible.

The known-positive fixture is its own negative control — `sparq-core`, `sparq-shacl`, `sparq-server` sit in the *same job family* with `cancelled` conclusions and were skipped by the same PR, so any loosening of the admission names them and reds the test.

## What this does **not** fix

- **The nightly backstop itself.** It has not concluded `success` since **2026-07-01, 28 days ago** (7 successes all-time). This PR makes that visible; it does not repair it.
- **The suspect-PR window is unbounded.** Because the window is "since the last green nightly" and that is 28 days back, the first real firing will list **644 / 579 / 554 suspect PRs** (issue bodies 18.1 / 16.3 / 15.8 KB). Under GitHub's 65536-char body limit today, but not actionable, and at ~2300 suspects it would exceed the limit and `create_issue` would raise. **Deliberately not fixed here** — it is analysis-side, and this PR is scoped to the admission seam. Filed separately.
- **Advisory-job failures are still correlated.** All four failures in the known positive are `mutation ratchet (…, advisory)` jobs. The analysis is unchanged by design; whether advisory lanes should feed the alarm is a separate judgment.
- **The per-crate dedupe is unchanged** (one open issue per crate), so this cannot spam beyond one issue per affected crate.

## Blast radius

`nightly selection-bug alarm` is declared in `.github/advisory-registry.json`, so it cannot red `ci-summary / gate` — `test_alarm_lanes_non_gating.py` still passes unchanged. No PR is affected.

---
Not armed — for independent review.

### DIFF
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -809,11 +809,14 @@ jobs:
         run: python3 scripts/tests/test_ci_select_wiring.py
       # [OPUS-4.8] sq-va7at: the nightly selection-bug ALARM
       # (scripts/ci_selection_alarm.py, .github/workflows/selection-alarm.yml,
-      # design §6.1). Hermetic stdlib-unittest suite (NO gh, NO git, NO cargo,
-      # NO network): the pure correlation core (precise per-crate hit, fail-safe
-      # triage for narrowed shards, the non-spam empty-skip-set property), the
-      # TRIAGE_LANE_RE pin against the real ci.yml shard job name, and the
-      # fail-loud exits (replay error / total blocker => non-zero).
+      # design §6.1). Hermetic unittest suite (NO gh, NO git, NO cargo, NO network;
+      # stdlib + PyYAML, installed once in the shared setup above): the pure
+      # correlation core (precise per-crate hit, fail-safe triage for narrowed
+      # shards, the non-spam empty-skip-set property), the TRIAGE_LANE_RE pin
+      # against the real ci.yml shard job name, the fail-loud exits (replay error /
+      # total blocker => non-zero), and — since #4965 — the ADMISSION SEAM: the live
+      # workflow `if:` evaluated against the real known-positive event payload, with
+      # the evaluator validated against the pre-fix expression as a known answer.
       - name: Self-test the nightly selection-bug alarm (correlation + fail-loud — sq-va7at)
         run: python3 scripts/tests/test_ci_selection_alarm.py
       # [OPUS-5] The review-lane BLIND-SPOT alarm (scripts/review_lane_alarm.py,
--- a/.github/workflows/selection-alarm.yml
+++ b/.github/workflows/selection-alarm.yml
@@ -2,7 +2,9 @@
 # epic sq-fmx4u). [OPUS-4.8] Design: research/change-based-test-selection.md §6.1.
 #
 # WHAT: when the nightly FULL-matrix backstop (CI workflow, `schedule` =>
-# mode=full) FAILS, correlate each failed job's crate with the crates that
+# mode=full) has any GENUINELY-FAILED JOB — regardless of the run's aggregate
+# conclusion, which fail-fast cancellation makes lossy (#4965) — correlate each
+# failed job's crate with the crates that
 # change-based selection (scripts/ci_select.py) SKIPPED in the PRs landed since
 # the last green nightly. Any intersection is prima-facie a selection soundness
 # bug (design §2) — scripts/ci_selection_alarm.py files a self-identified SPARQ
@@ -36,8 +38,9 @@
 name: selection-alarm
 
 on:
-  # Fire after a CI run on MAIN completes; the job `if:` narrows to a FAILED
-  # nightly (schedule / manual dispatch). The `branches: [main]` filter keys on the
+  # Fire after a CI run on MAIN completes; the job `if:` narrows to a
+  # nightly (schedule / manual dispatch) and the SCRIPT decides, from the real job
+  # list, whether anything genuinely failed (#4965). The `branches: [main]` filter keys on the
   # TRIGGERING run's head branch, so a PR's CI completion (head branch = the PR
   # branch) never fires this workflow — that guarantees selection-alarm produces
   # NO check-run on any PR head commit and so can never interact with the
@@ -67,14 +70,37 @@ concurrency:
 jobs:
   alarm:
     name: nightly selection-bug alarm
-    # workflow_dispatch: always run (operator asked). workflow_run: only a FAILED
+    # workflow_dispatch: always run (operator asked). workflow_run: every
     # scheduled/dispatched nightly on main (a PR/push/merge_group run carries a diff
     # => selection may narrow legitimately; only the full-matrix backstop is the
     # oracle here).
+    #
+    # [OPUS-5] #4965 — THERE IS DELIBERATELY NO `conclusion` TEST HERE, and adding
+    # one back re-breaks the alarm. This `if:` used to read
+    # `github.event.workflow_run.conclusion == 'failure' && …`, and that single
+    # clause is where the alarm died: MEASURED over all 45 nightlies — 35
+    # `cancelled`, 7 `success`, 3 `failure`, the last `failure` on 2026-07-19, the
+    # exact day the alarm last acted. 266 of 267 alarm runs were job-SKIPPED, and a
+    # SKIPPED JOB IS NEVER RED, so nine days of silence looked exactly like a healthy
+    # no-op. Known positive: nightly run 30333511110 concluded `cancelled` while
+    # carrying FOUR genuinely `failure` jobs (sparq-mpc / sparq-fedplan /
+    # sparq-reason ×2) — because ONE dying leg fail-fast-CANCELS the other 30 and
+    # GitHub aggregates that to `cancelled`.
+    #
+    # The run-level conclusion cannot express "some job failed", and a job `if:`
+    # cannot see the job list at all, so no expression here can make this admission
+    # precise. It is made in scripts/ci_selection_alarm.py instead, over the real
+    # jobs API, by the SAME `FAILED_CONCLUSIONS` predicate the analysis already uses:
+    # zero genuinely-failed jobs => a quiet exit 0 (the common `cancelled` case, so
+    # the alarm cannot cry wolf and get muted); a `failure`/`timed_out`-concluded run
+    # with zero failed jobs => still RED, fail-loud (STRICT_EMPTY_CONCLUSIONS).
+    # Widening to every conclusion costs ~1 extra short run per night (the nightly is
+    # the only schedule/dispatch CI run on main) and leaves NO conclusion literal here
+    # that can silently drift away from the script. Pinned by
+    # scripts/tests/test_ci_selection_alarm.py::TestWorkflowAdmissionSeam.
     if: >-
       github.event_name == 'workflow_dispatch' ||
-      (github.event.workflow_run.conclusion == 'failure' &&
-       (github.event.workflow_run.event == 'schedule' ||
+      ((github.event.workflow_run.event == 'schedule' ||
         github.event.workflow_run.event == 'workflow_dispatch') &&
        github.event.workflow_run.head_branch == 'main')
     runs-on: ubuntu-latest
@@ -112,9 +138,15 @@ jobs:
           CARGO_NET_RETRY: "10"
           RUN_ID: ${{ github.event.workflow_run.id || github.event.inputs.run_id }}
           HEAD_SHA: ${{ github.event.workflow_run.head_sha || github.event.inputs.head_sha }}
+          # NOT an admission filter (#4965) — the script uses it only to decide
+          # whether an EMPTY genuinely-failed-job set is anomalous (fail-loud) or the
+          # ordinary `cancelled`-with-nothing-really-broken case (quiet exit 0).
+          # Empty on the workflow_dispatch path; the script looks it up then.
+          RUN_CONCLUSION: ${{ github.event.workflow_run.conclusion }}
         run: |
           set -euo pipefail
           python3 scripts/ci_selection_alarm.py \
             --run-id "$RUN_ID" \
             --head-sha "$HEAD_SHA" \
+            --run-conclusion "$RUN_CONCLUSION" \
             --repo "$GITHUB_REPOSITORY"
--- a/scripts/ci_selection_alarm.py
+++ b/scripts/ci_selection_alarm.py
@@ -21,6 +21,13 @@
 #   `workflow_run` trigger, see .github/workflows/selection-alarm.yml), files
 #   issues, and never blocks a merge (design §6.1: "fail-open").
 #
+#   [OPUS-5] #4965 — WHAT COUNTS AS A NIGHTLY WORTH CORRELATING. A "failed nightly"
+#   means "a nightly with genuinely-failed JOBS", never "a nightly GitHub concluded
+#   `failure`". Those are different populations: fail-fast cancels the surviving
+#   matrix legs, so the RUN concludes `cancelled` and the run-level conclusion is a
+#   lossy aggregate. Admission is therefore made HERE, over the job list, by the same
+#   FAILED_CONCLUSIONS predicate the analysis uses — see STRICT_EMPTY_CONCLUSIONS.
+#
 # TWO DESIGN INVARIANTS (both load-bearing):
 #   1. FAIL-SAFE / FAIL-LOUD (opposite of ci_select.py's fail-CLOSED). An
 #      alarm-INFRASTRUCTURE error (cannot list the failed jobs, cannot load cargo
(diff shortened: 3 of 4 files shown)


jeswr:
> 🤖 SPARQ agent

Independent review — I did not author this. Head SHA reviewed: `8ab249d9b12e68f8f61e9278d3650a24bddc6110`. Verified against `origin/main` (`a66410383`) in an isolated worktree, not a local checkout. Not armed; no labels changed.

## Sequencing first: is this safe to merge before #4984 and #4985?

**Yes — merge this without waiting for either.** I treated "does this turn a silent alarm into a muted one" as the question to answer, and measured rather than reasoned about it.

**1. Merging this fires nothing today.** The `if:` admits only `workflow_run.event ∈ {schedule, workflow_dispatch}` on `main`. `ci.yml`'s schedule is `17 3 * * *`; it is now 04:26 UTC, so the next firing is **~23 h away**. And `ci.yml` has had **0 `workflow_dispatch` runs, ever** (`total_count: 0`), so the schedule is the only live trigger. The push that merges this PR carries `event=push` and is not admitted. There is no "immediately files 18 KB issues" — there is a 23-hour fuse.

**2. The #4985 dedupe cap does not bind, by 50×.** Measured live: **2** open issues carry the `selection-alarm` label (#3612, #3613); 3 all-time. The cap is `--limit 100`. Both existing issues carry a well-formed marker — `<!-- selection-alarm-key: crate:sparq-mpc -->` and `crate:sparq-core` — so dedupe is demonstrably working on the real corpus. On the known positive the findings are `sparq-fedplan` / `sparq-mpc` / `sparq-reason`; **`crate:sparq-mpc` dedupes against the already-open #3613**, so the first firing files **2** issues, not 3, and subsequent nightlies file 0. For the cap to fail open you would need 100 open labelled issues; the key space is `crate:<member>` over the workspace plus one `triage:` key per run. That is months of daily accumulation away, with an obvious early signal at ~10.

**3. #4984's hard failure is ~73 days out and cannot gate.** 18,064 chars against GitHub's 65,536 → 3.6× headroom; growth ≈ 650 chars/day at the observed ~23 suspects/day. And `nightly selection-bug alarm` is **already declared advisory on `main`** (`selection-alarm.yml` / job `alarm`), so even the uncaught-`AlarmError` failure mode #4984 describes reds only its own lane.

**4. The decisive point: #4984's defect is already live on `main` and this PR does not introduce it.** #3613 — filed 2026-07-19, still open — *is* the artifact #4984 warns about: a 13,783-char body, a suspect list of the same shape, triggered by the same `mutation ratchet (cargo-mutants, advisory)` lane. This PR does not create a new failure class; it files 2 more of an artifact `main` already produces, ~30 % larger. So #4984 is not a prerequisite for this change — it is an independent improvement that would equally benefit the two issues already open. Holding an un-inerting fix to first perfect the presentation of a signal that currently **does not exist at all** is how an inert control stays inert.

**Non-blocking mitigation worth taking inside the 23 h window.** Because dedupe is per-key and these keys fire once, **the first render is the only render** — an 18 KB body is what a human reads for as long as the issue stays open. If #4984's "cap the rendered list, keep the full set in the correlation" option lands before 03:17 UTC 2026-07-30, both new issues render usefully. If it does not, the cost is 2 issues of a shape already on `main`. That is a follow-up deadline, **not a merge precondition**.

## 1. Admission is precise, not merely wider

The paired property holds, and the negative half is real rather than nominal.

I re-derived the known positive from the live API rather than trusting the fixture. Run `30333511110`: `conclusion=cancelled`, `event=schedule`, `head_branch=main`, `head_sha=63dfb7a0…` — **84 jobs, 43 success / 31 cancelled / 6 skipped / 4 failure**, and the four failure names are byte-identical to the fixture, naming exactly `sparq-mpc` / `sparq-fedplan` / `sparq-reason`. The nightly census reproduces exactly too: 45 scheduled `ci.yml` runs, **35 `cancelled` / 7 `success` / 3 `failure`**, last success **2026-07-01**, last failure **2026-07-19** — the day the alarm last acted. `KNOWN_POSITIVE_CRATES` is ground truth, not a restatement of the code.

The quiet half is what makes it precision rather than width, and it is load-bearing in two independent places: `TestQuietOnACleanCancelledRun` drives `main()` to `rc=0` with nothing filed, and — better — `test_end_to_end_alarm_fires_and_names_the_crates` carries a negative control **inside** the positive: `sparq-core` / `sparq-shacl` / `sparq-server` are cancelled siblings in the same job family that the same landed PR selection-skipped, so they sit in the skip set and any loosening of admission names them and reds the test. "Admit everything" cannot pass. Mutants M2, M8 and M9 confirm all three axes (conclusion / event / branch) are independently pinned.

The fail-loud half survives the widening: `STRICT_EMPTY_CONCLUSIONS = {failure, timed_out, ""}` keeps an empty failed-job set anomalous exactly where the old premise still holds, and `""` is strict on purpose. The quiet branch was bought without deleting the masking guard.

## 2. The `if:` evaluator risk — acceptable, and here is the bound

The harness's evaluator is **not GitHub's**, and the PR says so. My judgement: the residual risk does not matter, for a reason stronger than "it was validated".

**No production behaviour depends on it.** `eval_if` is defined in and referenced only by `scripts/tests/test_ci_selection_alarm.py`; there are zero non-test references across `scripts/` and `.github/`. At runtime admission is decided by GitHub evaluating the YAML and by `ci_selection_alarm.py` over the jobs API. A wrong evaluator can therefore only produce a wrong **test verdict** about the live `if:` — never wrong live behaviour. That is what caps the blast radius.

Within that cap it is well built: the `if:` is extracted **structurally** (`yaml.safe_load` → `doc["jobs"]["alarm"]["if"]`), which matters because the surrounding comment block mentions `conclusion` repeatedly; the grammar (`||`/`&&`/`==`/`!=`/parens/literals/paths) exactly covers the live expression with correct precedence; and fail-closed was verified empirically, not just claimed — `contains(a,'b') ? 1 : 0`, `always()`, `!cancelled()`, `success() && …`, `${{ true }}` all raise rather than being judged admissible. `OLD_BROKEN_IF` is a faithful transcription of the real pre-fix expression (diffed against `HEAD^`), and its two-directional known answer matches the measured 266-skip / 1-act behaviour.

One latent deviation, not reached: `github.missing.leg == ''` returns `False` here but `True` on GitHub (null→0, ''→0). The live expression contains no empty-string comparison, so it is unexercised — but the deviation is in the **admissive** direction, so a future `if: … != ''` would be mis-certified. The `_lookup` docstring's claim of GitHub semantics for absent properties holds only for non-empty literals; worth a one-line correction.

## 3. Mutation evidence re-verified independently

Baseline **GREEN first** (`Ran 29 tests … OK`) on the pristine worktree **and** re-verified green on an unmutated copied tree, so the copy procedure itself is non-perturbing. All 7 claimed mutants reproduced on fresh whole-tree copies of `scripts/` **and** `.github/`, each killed by the named test, including M6 (instrument sabotage → `test_evaluator_reproduces_the_old_bug`). I added 8 further mutants; 14 of 15 total killed.

One diagnostic-quality note: M5 as specified (`if: true`) parses as a YAML **bool** and reds via an `AttributeError` in `_live_if()` — an instrument crash, not the named property. The property is genuinely pinned only for a parseable string; M5c (`if: "true"`) confirms it is, killed by assertion in `test_live_if_still_rejects_the_runs_that_are_not_the_oracle`.

## Findings (none blocking)

1. **`test_suite_is_wired_into_ci` (:752) is the weakest assertion in the file, and it is the anti-vacuity anchor.** It is a whole-file `assertIn` over `docs-quality.yml`. Plain deletion of the step is caught, but **commenting the invocation out while leaving the path in a trailing comment SURVIVES** — the suite leaves CI's reachable set and all 29 tests stay green. The guard that certifies every other assertion still runs is the one with the vacuous shape. The structural fix already exists 15 lines above in the same class: `safe_load` the workflow and assert a step's `run` invokes the path, as `test_the_workflow_passes_the_run_conclusion_to_the_script` does.
2. `assertIn("nothing genuinely failed", out)` (:506) pins production **prose**. A pure reword of the `print()` reds it with no behaviour change. Brittle rather than vacuous — it is what distinguishes "took the quiet branch" from "exited 0 some other way" — but a stable token would be better.
3. `test_the_workflow_passes_the_run_conclusion_to_the_script` asserts `"workflow_run.conclusion" in str(step["env"])`, which does not bind *which* env var reaches `--run-conclusion`. Deleting the var is caught; renaming the binding is not. Low consequence.
4. `startup_failure` is outside `STRICT_EMPTY_CONCLUSIONS`, so a startup-failed nightly with no job rows exits quiet — arguably a case where the backstop demonstrably did not run. Not a regression (it was skipped entirely before), and the backstop-liveness question is #4984's. Noting it so the choice is deliberate.
5. On the quiet path `run_alarm` prints its explicit reason and then `main()` also prints "no nightly failure correlates with a selection skip in the window", which describes a different state. Cosmetic; the two lines together are still unambiguous.
6. The job has no `timeout-minutes` (GitHub default 360 m) while the replay window is unbounded — a second, smaller reason to want #4984's cap.

## Blast radius — confirmed independently

`nightly selection-bug alarm` is already a declared advisory entry (`selection-alarm.yml` / `alarm`) on `main`, and `check-advisory-registry.py` is clear (C2 + C3 + C4) at this head. `fetch-depth: 0` is present, so the per-commit replay resolves. Composed against #4983 on top of `origin/main` there are **no conflicts** (both touch `docs-quality.yml`, ~150 lines apart) and the merged tree is green: both suites OK, `check-advisory-registry.py` and `check-fast-fix-ring-guard.py` both clear, both suites wired.

The diff does what the body says, the empirical claims reproduce exactly against the live API, and the one defect I found is in a secondary anchor rather than the headline guard.

VERDICT: pass

jeswr:
> 🤖 SPARQ agent

Arming on an independent `VERDICT: pass` at head `8ab249d9b12e68f8f61e9278d3650a24bddc6110`.

**This control was INERT**: its job `if:` admitted only `conclusion == 'failure'`, while live nightlies
run **35 `cancelled` / 7 `success` / 3 `failure`** — and a cancelled run can contain genuinely failed
jobs. 267 runs, **266 job-skipped**, last action 2026-07-19. The Python was always correct; the YAML
seam is where it died, and a skipped job is never red, so it failed silently 266 times.

**The fix did not take the obvious route, and that mattered.** Adding `cancelled` to the `if:` would
have been *worse than the bug* — the script raises on an empty failed-job set, so most nights would
have gone red for runs where nothing broke. Admission moved out of the YAML into the script, over the
real jobs API, using the same `FAILED_CONCLUSIONS` predicate the analysis already used.

**Admission is precise, not merely wider** — the reviewer re-derived the known positive live (run
`30333511110`: 84 jobs, 4 genuine failures among 31 cancelled, naming exactly
`sparq-mpc`/`sparq-fedplan`/`sparq-reason`) *and* confirmed the paired negative: an in-positive
control means "admit everything" reds the suite.

## Why I am not holding this for #4984 / #4985
Following the reviewer's recommendation, which rests on measurement rather than judgement:
1. **#4984's defect is already live on `main`** — issue **#3613** (filed 2026-07-19, still open,
   13,783-char body) *is* the artifact it describes. This PR files 2 more of a shape `main` already
   produces, ~30% larger. It introduces no new failure class.
2. **Merging fires nothing for ~23 h** — the `if:` admits only `schedule`/`workflow_dispatch`, the
   cron is `17 3 * * *`, there have been **0 `workflow_dispatch` runs ever**, and the merge push is
   not admitted. A fuse, not a trigger.
3. **#4985's dedupe cap has ~50× headroom** — 2 open `selection-alarm` issues against a limit of 100,
   both with well-formed markers; dedupe demonstrably works on the real corpus. First firing files
   **2**, then 0/night.
4. #4984's hard failure is **~73 days out** and lands on a lane already declared advisory.

⚠️ **Worth taking inside the window:** dedupe is per-key, so **the first render is the only render**.
Landing #4984's render cap before **03:17 UTC 2026-07-30** is cheap and stops the oversized body being
the one that persists.

## Non-blocking, and worth one sweep rather than a point fix
The reviewer added 8 mutants beyond the author's 7 and found a survivor:
`test_suite_is_wired_into_ci` is a whole-file `assertIn` — **commenting out the invocation while
leaving the path in a trailing comment survives it**. Its sibling in #4983 has the same defect class.
**Both PRs' designated anti-vacuity anchors are the weakest assertions in otherwise strong suites**,
which is a pattern worth one pass across the suites rather than two point fixes.

PR #4983, opened 2026-07-29
fix(ci): a missing REGISTRY_RING_TOKEN must fail LOUD, not suppress ~130 rings into green runs (#4966)

### LINKED ISSUE #4966: fast-fix-ring and batch-merge's ring job have never sent a ring: REGISTRY_RING_TOKEN does not exist, and both fail soft to a green run (~2000 runs, 0 effect)
> 🤖 SPARQ agent

`fast-fix-ring` and `batch-merge`'s `ring` job both end in a single privileged call gated on `secrets.REGISTRY_RING_TOKEN`. **That secret does not exist**, so both controls resolve it to the empty string, take the fail-soft branch, and emit a `::notice::` nobody reads. Across ~2000 runs neither has ever performed its one mutation.

## The secret is absent — measured at runtime, not inferred from config

`repos/sparq-org/sparq/actions/secrets` contains exactly:

```
OPENALEX_API_KEY, ORCHESTRATOR_APP_ID, ORCHESTRATOR_APP_PRIVATE_KEY
```

I could not read org-level secrets (`orgs/sparq-org/actions/secrets` → 403), so repo-level absence alone would not settle whether an org secret is inherited. The runtime environment dump settles it. From run `30411348407`, job `fast-fix ring (advisory)` — and these are runtime log lines, not echoed script source (that log contains **zero** `ESC[36;1m`-prefixed lines):

```
GH_TOKEN: ***
REGISTRY_RING_TOKEN:
REPO: sparq-org/sparq
HEAD_SHA: 72ef1b57c074abb75dcffc45f16dfec224c46252
```

`GH_TOKEN` renders as `***` because a secret with a value is masked. `REGISTRY_RING_TOKEN` renders as **nothing at all** — it has no value, at repo or org level.

The guard it feeds, `.github/workflows/fast-fix-ring.yml:212`:

```bash
if [ -z "${REGISTRY_RING_TOKEN}" ]; then
```

## Blast radius

| control | runs | executed past `if:` | rings sent |
|---|---|---|---|
| `fast-fix-ring` | **1969** (1614 skipped, 355 executed) | 32 sampled, **12** passed every eligibility check | **0** |
| `batch-merge` `ring` job | 12/12 sampled push runs | all reached the guard | **0** |

Extrapolating the sampled rate, roughly **130 rings** have been suppressed. The cross-repo doorbell is these workflows' *only* write capability — `fast-fix-ring.yml:94` calls it "The one WRITE capability" — so with it empty the entire control is a no-op wrapped around an eligibility computation whose result is discarded.

**Verdict: INERT** for both, and INERT since inception rather than by regression.

## Why it stayed invisible

The guard is deliberately fail-soft: a missing token produces a `::notice::` and a **green** run. That is a defensible choice for an advisory lane, but it means the difference between "the ring fired" and "the ring cannot fire, and never could" is not visible in any field a monitor queries. 1969 green runs, zero effect.

This is the same shape as the "signal exists, nothing consumes it" class: the control correctly detects that it is disabled and then reports that fact into a void.

## Suggested fix

Decide which of these two is true and make the code say so:

1. **The ring is wanted** — provision `REGISTRY_RING_TOKEN` (or switch to the `ORCHESTRATOR_APP_*` App credentials already present in this repo, which would avoid a long-lived cross-repo PAT and address the exfiltration concern the workflow header raises at `:5-19` re #3474).
2. **The ring is not wanted** — delete both jobs rather than run ~2000 no-op invocations that read as healthy.

Either way, the fail-soft branch should be loud enough to be noticed on its *first* occurrence — a rolling issue, or a red advisory run — rather than a `::notice::`. A control that is disabled by missing configuration and silent about it is indistinguishable from one that is working.

Related: `.github/workflows/pr-backlog.yml:96` carries `EXCLUDE_PR: "2431"` for a PR that merged 2026-07-17 — dead config in the same family, trivial to clear.

Found during a survey of whether each orchestration control has ever actually *acted*, as distinct from having *run*.

jeswr:
> 🤖 SPARQ agent

Closes #4966. Head SHA `c2fcb8b346205141de824fe7b85de6f4e83e4d25`.

## The defect

Both cross-repo doorbells end in one privileged call gated on `secrets.REGISTRY_RING_TOKEN`. That secret does not exist — MEASURED from the runtime env dump of run `30411348407` (`GH_TOKEN: ***` masked-and-present vs `REGISTRY_RING_TOKEN:` empty; that log carries **zero** `ESC[36;1m` lines, so these are runtime values, not echoed script source). Both guards took the fail-soft branch every time and **reported SUCCESS**:

| control | runs | reached the guard | rings sent |
|---|---|---|---|
| `fast-fix-ring` | 1969 (355 executed) | ~130 | **0** |
| `batch-merge` `ring` | 12/12 sampled push runs | all | **0** |

A control that is disabled by missing configuration and green about it is indistinguishable from one that works. That is exactly how this stayed invisible.

## Fix shape, and why this one

**Both guards now emit an `::error::` annotation naming the secret, write a `$GITHUB_STEP_SUMMARY` row, and `exit 1`.** I chose a hard, named non-success over an annotation-only signal because the blast radius is genuinely contained — and I verified that rather than assuming it.

**Why a hard failure is safe.** Neither check-run can land on a PR or merge_group head: `fast-fix-ring` is a `workflow_run` workflow (GitHub executes those from the **default branch**, so the run's own head SHA is main's), and `batch-merge`'s ring is `push: branches: [main]`. MEASURED: both `fast-fix ring (advisory)` and `ring registry dispatcher` are present in the check-run set of main head `2cbf37359`. That is the SHA the **push-triggered** `ci-summary / gate` polls (the sq-huwr8 refutation), so the advisory-registry declaration — not the trigger type — is what keeps them non-gating.

**`fast-fix ring (advisory)` was already declared. `ring registry dispatcher` was NOT** — it was silently gating. Undeclared, a hard failure there would have redded `ci-summary / gate` on **every push to main** (50–103 push runs/day), and AGENTS.md treats a red main as a stop-the-line condition. That is a manufactured outage, not an alarm. **This PR declares it**, so the run reds — visible, named, countable — while the gate stays green.

**Every red is a true positive.** In `fast-fix-ring` the credential guard is the *last* one, after every eligibility check (draft / human-hold / armed-or-in-review / same-repo / same-SHA), so `exit 1` means precisely "a ring was warranted and could not be sent" — never a false alarm on an ineligible PR. This is why the ~25–60/day rate is a correctly-sized alarm rather than noise: the volume *is* the damage.

## Acceptance criteria

| # | criterion | result |
|---|---|---|
| 3 | empty credential → visible non-success | **PASS** — `TestMissingCredentialIsLoud`: non-zero exit, `::error::` naming the secret, step-summary row, no dispatch attempted |
| 4 | still works when the credential is present | **PASS** — `TestCredentialPresentStillWorks`: exit 0 **and** the dispatch actually recorded, so the fix cannot be "always fail" |
| 5 | control suite GREEN before trusting mutations | **PASS** — baseline green, re-verified GREEN on each copied tree |

### Mutation results — 7/7 killed

| mutant | killed by |
|---|---|
| R1 `fast-fix-ring` back to notice + `exit 0` | `test_empty_credential_exits_non_zero_and_annotates` |
| R2 `batch-merge` ring back to notice + `exit 0` | same, + `test_the_annotation_is_not_merely_a_notice` |
| R3 "always fail" brick (dispatch removed) | `test_present_credential_rings_and_exits_zero` |
| R4 non-zero but downgraded to `::notice::` | `test_empty_credential_exits_non_zero_and_annotates` |
| R5 drop the `ring registry dispatcher` declaration | `test_a_failing_ring_leaves_the_gate_green` + `test_both_ring_jobs_are_declared_advisory` |
| R6 drop the step-summary counter | `test_empty_credential_exits_non_zero_and_annotates` |
| R7 **instrument sabotage** (`gh` stub always reports a dispatch) | `test_empty_credential_exits_non_zero_and_annotates` |

## The test does not grep — it executes

`scripts/tests/test_ring_credential_loud.py` **extracts each ring job's real `run:` script structurally from the live YAML** (via PyYAML, so a comment cannot satisfy it) and **executes it under bash** with a stubbed `gh` on PATH, asserting the process exit code, the emitted workflow commands, and whether the privileged dispatch was actually attempted — in **both** credential states. It also drives the real `ci_summary_gate.render_verdict()` to prove the new failure cannot gate, with an **anti-vacuity control** (an undeclared lane failing MUST still red the gate) and a **quantifier-direction test** (declaring the doorbells must not forgive a real gating leg).

`scripts/tests/fast-fix-ring.golden.yml` is regenerated: the #3475 whole-file byte-pin correctly went RED on my edit first, and `--self-test`'s 17 negative fixtures still pass.

## What this does **not** fix

- **It does not provision the secret.** That is the maintainer's, and I have not attempted to work around it. This PR only makes the absence impossible to miss. **`REGISTRY_RING_TOKEN` is still absent, so on merge both lanes begin failing immediately** — that is the intended signal, and the two exits are: provision the secret (or switch to the `ORCHESTRATOR_APP_*` credentials already in this repo, which also addresses the long-lived-cross-repo-PAT concern in the `fast-fix-ring` header re #3474), **or delete both jobs** if the ring is no longer wanted. Both annotations say so.
- **It does not make the failure rate-limited or deduped.** Expect ~25–60/day from `fast-fix-ring` and 50–103/day from `batch-merge`'s ring until one of those two decisions is taken. A rolling deduped issue would be quieter, but it needs `issues: write` on a deliberately minimal-privilege privileged workflow (#3474), which I judged the wrong trade to make unreviewed.
- **It slightly narrows the gating set**: `ring registry dispatcher` was gating (always green) and is now advisory. Justified because its success is a property of an *external* repo's dispatch endpoint, never of the commit under test — the same rationale as the existing `merge-group doorbell (advisory)` entry. `test_declaring_the_rings_does_not_forgive_a_real_gating_failure` pins that this did not weaken the gate for the commit's own legs.
- **`.github/workflows/pr-backlog.yml:96` `EXCLUDE_PR: "2431"`** (dead config noted in the issue) is untouched — unrelated to the credential.

---
Not armed — for independent review.

### DIFF
--- a/.github/advisory-registry.json
+++ b/.github/advisory-registry.json
@@ -209,6 +209,13 @@
    "promotion_criteria": "not promotable to a gate: the ring runs post-hoc via workflow_run (on ci-summary completion), never on a PR/merge_group head commit, so it produces no gating check-run by design (sparq-org/sparq#3474 moved it here so the REGISTRY_RING_TOKEN doorbell is unreachable from PR-head content); it only notifies the registry doorbell after a gating-leg failure and must never gate a PR",
    "registered": "2026-07-18"
   },
+  "ring registry dispatcher": {
+   "workflow": "batch-merge.yml",
+   "job_id": "ring",
+   "owner_bead": "sq-4966",
+   "promotion_criteria": "NEVER promotable. This job is a fire-and-forget cross-repo DOORBELL: it dispatches jeswr/agent-account-registry so a freed merge-queue slot is picked up now rather than on the registry's next ~10-min cron. Whether that dispatch succeeds is a property of an EXTERNAL repository's endpoint and of a credential's presence — never a property of the commit under test — so gating on it would block sparq merges on another repo's availability. Same shape and same rationale as `merge-group doorbell (advisory)`. DECLARED as part of #4966, which made a MISSING REGISTRY_RING_TOKEN fail LOUD (exit 1) instead of fail-soft to a green run: the job runs on `push: branches: [main]`, so its check-run lands on the main head SHA that the push-triggered `ci-summary / gate` polls (the sq-huwr8 refutation — MEASURED present on main head 2cbf37359). Undeclared, that new hard failure would have redded `ci-summary / gate` on EVERY push to main (50-103 push runs/day), which AGENTS.md treats as a stop-the-line condition — a manufactured outage rather than an alarm. This declaration is what keeps the failure LOUD (the run reds, named and countable) without letting it gate. No gate_script_waiver: the job invokes no script at all, only an inline `gh workflow run`. Delete this entry only if the job is removed.",
+   "registered": "2026-07-29"
+  },
   "release-plz / Release-PR (advisory)": {
    "workflow": "release-plz.yml",
    "job_id": "release-plz-pr",
--- a/.github/workflows/batch-merge.yml
+++ b/.github/workflows/batch-merge.yml
@@ -129,8 +129,22 @@ jobs:
           python3 scripts/batch-merge.py "${args[@]}"
 
   # Merged-PR pickup ring: a merge into main frees merge-queue/worker capacity — poke the
-  # registry dispatcher immediately instead of waiting for its cron. Fail-soft: without
-  # the secret this skips with a notice (the registry cron is the backstop).
+  # registry dispatcher immediately instead of waiting for its cron.
+  #
+  # FAIL-LOUD on a missing credential (#4966), was `::notice::` + exit 0. This job has
+  # never once rung: REGISTRY_RING_TOKEN does not exist at repo or org level, so every
+  # push-triggered run took the fail-soft branch and reported SUCCESS. 12 of 12 sampled
+  # push runs, 0 rings. Reporting success while doing nothing is what kept it invisible.
+  #
+  # DECLARED ADVISORY (.github/advisory-registry.json, "ring registry dispatcher") in
+  # the same change, and that declaration is what makes this hard failure SAFE. This
+  # job runs on `push: branches: [main]`, so its check-run lands on the main head SHA —
+  # MEASURED present on main head 2cbf37359 — which is exactly the SHA the
+  # push-triggered `ci-summary / gate` polls (the sq-huwr8 refutation). Undeclared, a
+  # red here would have gated `ci-summary / gate` on EVERY push to main, and AGENTS.md
+  # treats a red main as a stop-the-line condition: that is a manufactured outage, not
+  # an alarm. Declared, the run itself reds (visible, countable, named) and the gate
+  # stays green. Pinned by scripts/tests/test_ring_credential_loud.py.
   ring:
     name: ring registry dispatcher
     if: github.event_name == 'push'
@@ -144,7 +158,13 @@ jobs:
         run: |
           set -euo pipefail
           if [ -z "${REGISTRY_RING_TOKEN}" ]; then
-            echo "::notice::REGISTRY_RING_TOKEN not configured — skipping the registry ring (the registry's own cron is the backstop)."
-            exit 0
+            echo "::error title=registry ring suppressed::REGISTRY_RING_TOKEN is not configured, so the merged-PR pickup ring was NOT sent. Freed merge-queue/worker capacity will not be picked up until the registry's own ~10-min cron tick. Provision the secret — or delete this job if the ring is no longer wanted."
+            {
+              echo "### registry ring SUPPRESSED — REGISTRY_RING_TOKEN unset"
+              echo
+              echo "A merge landed on \`main\` but the pickup ring could not be sent."
+              echo "Effect: capacity pickup waits for the next ~10-min registry cron."
+            } >> "${GITHUB_STEP_SUMMARY:-/dev/null}"
+            exit 1
           fi
           GH_TOKEN="${REGISTRY_RING_TOKEN}" gh workflow run dispatch.yml -R jeswr/agent-account-registry
--- a/.github/workflows/docs-quality.yml
+++ b/.github/workflows/docs-quality.yml
@@ -963,6 +963,19 @@ jobs:
         run: python3 scripts/check-fast-fix-ring-guard.py --self-test
       - name: "Enforce the fast-fix ring fail-closed cross-repo guard (#3475) — GATING"
         run: python3 scripts/check-fast-fix-ring-guard.py
+      # [OPUS-5] #4966: the SAME two doorbells, from the opposite direction — a MISSING
+      # REGISTRY_RING_TOKEN must fail LOUD instead of suppressing ~130 rings into green
+      # runs. Extracts each ring job's real `run:` script from the live YAML and
+      # EXECUTES it under bash with a stubbed `gh`, asserting the exit code and the
+      # emitted annotations in BOTH credential states (empty => non-zero + `::error::`
+      # + no dispatch; present => exit 0 + the dispatch actually attempted, so the fix
+      # cannot be "always fail"). Also drives the REAL ci_summary_gate.render_verdict()
+      # to prove the new hard failure cannot red `ci-summary / gate` — both check-runs
+      # land on the main head SHA that the push-triggered gate polls (sq-huwr8), so the
+      # advisory-registry declaration is what keeps them non-gating. Needs PyYAML,
+      # installed once in the shared setup above.
+      - name: Self-test the ring doorbells fail LOUD on a missing credential (#4966)
+        run: python3 scripts/tests/test_ring_credential_loud.py
       # [OPUS-5] The VERDICT-comment -> review:pass bridge (scripts/verdict-bridge.py,
       # .github/workflows/verdict-bridge.yml). It is the ONLY automated writer of the
       # label the whole arming chain admits on, so a regression in it either strands
--- a/.github/workflows/fast-fix-ring.yml
+++ b/.github/workflows/fast-fix-ring.yml
@@ -214,8 +214,39 @@ jobs:
             # derives ci-fix admission from the PR's gate STATE and treats
             # needs:* PR labels as human-owned holds that EXCLUDE the PR from
             # the repair enumeration — a label here would suppress the fix.
-            echo "::notice::REGISTRY_RING_TOKEN unavailable (secret unconfigured) — cannot ring; the registry cron (every ~10 min) is the backstop."
-            exit 0
+            #
+            # FAIL LOUD (#4966), was `::notice::` + exit 0. REGISTRY_RING_TOKEN has
+            # never existed at repo or org level — MEASURED from a runtime env dump
+            # (`GH_TOKEN: ***` masked-and-present vs `REGISTRY_RING_TOKEN:` empty) —
+            # so this branch has taken EVERY ring this workflow ever computed: ~130
+            # suppressed across 1969 runs, each reporting SUCCESS. A control that is
+            # disabled by missing configuration and green about it is indistinguishable
+            # from one that works, which is exactly how this stayed invisible.
+            #
+            # This is the LAST guard, after every eligibility check above has already
+            # passed, so a non-zero exit here means precisely "a ring was warranted and
+            # could not be sent" — a true positive every time, never a false alarm on an
+            # ineligible PR.
+            #
+            # BLAST RADIUS (why a hard failure is safe here). This job is DECLARED in
+            # .github/advisory-registry.json ("fast-fix ring (advisory)"), the only
+            # thing that makes a check-run non-gating since #3773, so it cannot red
+            # `ci-summary / gate`. And `workflow_run` runs execute on the DEFAULT
+            # BRANCH: this run's check-run lands on the main head SHA, never on a PR or
+            # merge_group head — MEASURED on main head 2cbf37359, where both this lane
+            # and `ring registry dispatcher` are present. So no PR is redded by this.
+            # Pinned by scripts/tests/test_ring_credential_loud.py.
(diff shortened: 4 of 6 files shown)


jeswr:
> 🤖 SPARQ agent

Independent review — I did not author this. Head SHA reviewed: `c2fcb8b346205141de824fe7b85de6f4e83e4d25`. Verified against `origin/main` (`a66410383`) in an isolated worktree. Not armed; no labels changed.

## The advisory declaration — the one check that could stop the line

This PR starts failing on merge by design, so everything rests on whether a failing `ring registry dispatcher` can red `ci-summary / gate`. **It cannot.** I verified this two independent ways rather than trusting the PR's own test, and both agree.

**Name binding.** `is_declared_advisory` (`ci_summary_gate.py:1177`) is `any(m.fullmatch(candidate) for m in _DECLARED_ADVISORY)`, where each pattern is compiled per registry **key** with `${{ … }}` → `.+` and everything else `re.escape`d, `IGNORECASE`. That is exact whole-name equality on the check-run `name` — not a prefix, not a substring, not the job-id, not the workflow file. The new key `"ring registry dispatcher"` is byte-identical to `batch-merge.yml` job `ring`'s `name:` and to the live check-run name on main head `a66410383` (hex-compared: `72696e672072656769737472792064697370617463686572`). No `" (advisory)"` drift, no `workflow / job` qualifier, no matrix suffix. `check-advisory-registry.py` is clear (C2 + C3 + C4) at this head, so C4 — which reds if the job's YAML name ever stops matching the key — binds the declaration to the stable identity.

**End-to-end differential through the real gate.** I pulled all 305 check-runs on main head `a66410383`, dropped the cancelled twins and `in_progress` legs to get a green baseline (the raw set reds on the #3677 stale-twin class, so the raw feed is not a valid instrument — I checked that before drawing anything from it), then drove the real `render_verdict`:

| arm | gate rc | gating failures | new |
|---|---|---|---|
| clean baseline | **0** | 0 | — |
| `ring registry dispatcher` → `failure` (11 instances) | **0** | 0 | none |
| **both** ring jobs → `failure` (20 instances) | **0** | 0 | none |
| CONTROL: an **undeclared** lane → `failure` | **1** | 1 | `totally undeclared lane` |

The control is what makes the first three rows mean something: the same harness does red on an undeclared failure, so "no new gating failure" is a measurement, not an artefact of a broken instrument. A second, separately-built differential over the live 194-distinct-name set reached the identical conclusion, `failfast_failures` returns `[]` with both rings red, and branch-protection ruleset `17688455` requires exactly one context — `gate` — so neither ring is a required check.

**No functional coupling either.** `batch` (`if: event_name != 'push'`) and `ring` (`if: event_name == 'push'`) are mutually exclusive and neither `needs` the other, so a failing ring cannot block the actual batching, which runs on the 15-minute cron. The two other workflows referencing these files do so only in comments.

**Determination: the declaration HOLDS.** This is not a stop-the-line.

## Mutation evidence re-verified

Baseline GREEN first — `test_ring_credential_loud.py` 9/9, `check-fast-fix-ring-guard.py --self-test` and the GATING whole-file byte-pin against the regenerated golden both clean, `check-advisory-registry.py` clean — then re-verified green on an unmutated whole-tree copy. All **7/7** claimed mutants reproduced on fresh whole-tree copies of `scripts/` **and** `.github/`, each killed by the named test, including R7 (instrument sabotage: the stubbed `gh` always reporting a dispatch → killed by the "no dispatch may be attempted without the credential" assertion).

Worth recording: the #3475 byte-pin independently kills R1/R3/R4/R6 for `fast-fix-ring.yml`, but **not** R2/R5/R7 — and `batch-merge.yml`'s ring has no byte-pin at all, so for that half the new suite is the sole protection. That raises the stakes on the two scaffolding defects below.

## Findings (none blocking)

1. **`test_the_annotation_is_not_merely_a_notice` (:179–193) does not test what its docstring claims, and I have the surviving mutant.** Line 187 is `body = guard[1].split("fi", 1)[0]` — that splits on the **substring** `"fi"`, not the shell keyword `fi`. Measured, the extracted "guard body" is **17 characters** for `fast-fix-ring.yml` (`'\n  # NO needs:ci-'`, truncated inside "ci-**fi**x") and 79 for `batch-merge.yml` (truncated inside "con**fi**gured"). The docstring says the test "must red even if an `::error::` were added alongside" a notice; adding `echo "::notice::REGISTRY_RING_TOKEN unavailable — cannot ring."` immediately before the `::error::` while keeping `exit 1` leaves **all 9 tests green**. For the fast-fix-ring half the assertion is inspecting a comment fragment and can never fire. Consequence is low — this is a redundant belt and the primary test carries R1/R4/R6/R7 on its own — but it is the marquee-claim-not-measured shape. `re.split(r"(?m)^\s*fi\s*$", …)` fixes it.
2. **`test_docs_quality_runs_this_suite` (:311–313) is a whole-file containment check and is vacuous under the mutation that matters.** Rewriting the invocation to `run: echo skip   # scripts/tests/test_ring_credential_loud.py` leaves all 9 tests green while the suite has left CI's reachable set. The docstring calls it the anti-vacuity anchor and the header at :21–23 says the file "does not grep" — true of the ring guard, false of this anchor. The structural technique already exists in this same file: `_ring_step()` (:78) locates a step properly. The wiring at head is genuinely correct (job `quick-gates` / `docs-quality quick-gates`, which is undeclared and therefore gates), so this is a durability defect, not a live gap. **Note this is the same defect class as #4982's `test_suite_is_wired_into_ci`** — both PRs' CI-wiring anchors are the weakest assertions in otherwise strong suites. Worth one sweep across the alarm suites rather than two point fixes.

The other seven assertions are sound on all four questions — in particular `test_both_ring_jobs_are_declared_advisory` is **not** registry-asserting-registry (the expected names come from a `RING_JOBS` constant that `:243` independently pins to the live YAML `jobs.<id>.name`), and `test_declaring_the_rings_does_not_forgive_a_real_gating_failure` genuinely discriminates inside one `render_verdict` call — its stdout shows `1 non-passing gating check(s) of 2 gating (1 advisory excluded) — ✗ build: failure` — so it is not a tautology and would catch an over-broad key like `bui${{ x }}`.

## One judgment call I want to name rather than wave through

On merge this puts roughly **100–160 red workflow runs per day on main**, indefinitely, until a human takes one of the two exits. `gate` stays green and nothing consumes those conclusions — I checked — but a permanent red floor on main's run list is itself a muting mechanism: once everything is always red, a *new* red stops being visible. The author considered the quieter option (a rolling deduped issue) and rejected it for a good reason — it needs `issues: write` on a deliberately minimal-privilege privileged workflow (#3474) — and the fail-soft-to-green alternative is precisely what hid ~130 suppressed rings. So the choice is defensible and the volume genuinely is the damage.

But it is only correct as a **short-lived** alarm. My recommendation to whoever merges: treat the two named exits — provision `REGISTRY_RING_TOKEN` (or switch to the `ORCHESTRATOR_APP_*` credentials already in this repo, which also answers the long-lived-cross-repo-PAT concern in #3474), or delete both jobs — as due in days, not weeks. That decision is a human one and this PR correctly surfaces it rather than making it.

## Composition with #4982

Both PRs touch `docs-quality.yml`. Merged together onto `origin/main` there are **no conflicts** (the hunks are ~150 lines apart) and the composed tree is green: both new suites pass, `check-advisory-registry.py` and `check-fast-fix-ring-guard.py` both clear, both suites wired. No stricter-check/newly-gating interaction between them — this PR narrows the gating set, #4982 does not touch it.

The diff does what the body says, the highest-consequence claim holds under independent verification, and the two defects I found are in the suite's own scaffolding rather than in the protection of the fix.

VERDICT: pass

jeswr:
> 🤖 SPARQ agent

Arming on an independent `VERDICT: pass` at head `c2fcb8b346205141de824fe7b85de6f4e83e4d25`.

**This control was INERT and fail-soft**: `REGISTRY_RING_TOKEN` is absent, and both ring guards
returned a **green run** while doing nothing — **~130 rings silently suppressed**. Now they emit
`::error::` naming the secret, write a step-summary row, and exit 1.

**The highest-consequence check passed.** The author measured that `ring registry dispatcher` was
**undeclared and therefore gating** — hard-failing it would have redded `ci-summary / gate` on every
push to main (50–103/day, a stop-the-line condition). Declaring it **advisory** is what makes the
loud failure safe, and the reviewer verified that two independent ways: the registry key is
byte-identical to the live check-run name and matched with `fullmatch` on the whole name; and driving
the real `render_verdict` over main's live check-run set, **both rings forced to `failure` → gate
rc=0**, while an **undeclared** lane forced to failure → **rc=1**. The control proves the instrument
discriminates rather than merely agreeing.

## ⚠️ This PR starts failing on merge, by design — and that is time-limited

The secret is still absent, so this puts **~100–160 red runs/day on `main` indefinitely**. `gate`
stays green and nothing consumes those conclusions, but **a permanent red floor is itself a muting
mechanism** — people stop reading a signal that is always on. This is correct only as a
**short-lived** alarm.

**Three exits, and one of them must be taken within days, not weeks:**
1. Provision `REGISTRY_RING_TOKEN`.
2. Switch the rings to the `ORCHESTRATOR_APP_*` credentials already present.
3. Delete both ring jobs.

I am merging rather than holding because the alternative is worse: leaving a control that reports
success while suppressing ~130 rings is exactly the failure class this session set out to find. **A
loud alarm with a deadline beats a silent one without.** If no exit is taken, disabling the jobs is
strictly better than leaving a permanent red — silence you chose is not the same as silence you did
not notice.

## Non-blocking findings, worth one sweep rather than point fixes
The reviewer added mutants beyond the author's and found two survivors, both in **scaffolding
anchors**:
- `test_docs_quality_runs_this_suite` — whole-file containment; unwiring the invocation while leaving
  the path in a comment survives.
- `test_the_annotation_is_not_merely_a_notice` — `split("fi", 1)` splits on the **substring inside
  "ci-fix"**, truncating the extracted guard body to **17 chars**. Its docstring's claim is
  measurably false.

Also noted: `batch-merge`'s ring has no byte-pin, so this new suite is its sole protection.

**The pattern across both PRs: the designated anti-vacuity anchors are the weakest assertions in
otherwise strong suites.** Worth one pass across the suites rather than four point fixes.

Issue #5025, opened 2026-07-29
nightly full-matrix backstop has not concluded 'success' in 28 days — selection-alarm's window only grows


jeswr:
> 🤖 SPARQ agent

Split out of #4984 (render cap) and #4965 / PR #4982 (admission seam), both of which
deliberately leave this alone. Those two make the alarm **fire** and make its output
**readable**. Neither repairs the condition that makes its output large in the first
place.

## The condition

`last_green_nightly_sha()` bounds the alarm's correlation window at the most recent
**`success`**-concluded scheduled `ci.yml` run. That run is **2026-07-01**. So
"PRs landed since the last green nightly" is currently **28 days of `main`**, which is
why the first real firing carries 644 / 579 / 554 suspect PRs.

All-time nightly conclusions (`--paginate` verified, per #4984): **35 `cancelled` /
7 `success` / 3 `failure`** across 45 runs.

## Why it matters beyond a big issue body

The nightly full-matrix run is the **backstop for change-based test selection** — it
is the thing that re-runs what per-PR selection skipped, and it is the oracle the
selection-alarm correlates against. A backstop that has not concluded `success` in 28
days means:

1. **The correlation window only ever grows.** Every extra day widens the suspect set
   for every future finding. #4984 bounds the *render*; it cannot bound the *window*.
2. **Signal quality degrades.** With a 28-day window, a suspect list stops
   discriminating — nearly every landed PR skipped nearly every crate at some point.
3. **We do not actually know the full matrix is green.** That is the property the
   backstop exists to establish.

## What this is probably not

Not simply "the tests are failing". `cancelled` dominating (35/45) points at
**fail-fast matrix cancellation** — one dying leg cancels the other ~30, and GitHub
aggregates the run to `cancelled`, which can never be `success` even if the remaining
legs would all have passed. So the run-level `success` that `last_green_nightly_sha()`
requires may be **structurally unreachable** on a fail-fast matrix, independent of
whether anything is genuinely broken.

If that is the cause, the fix is a choice between:

- making the nightly matrix `fail-fast: false` so a single failure does not cancel the
  rest and the run can conclude `failure`/`success` honestly; and/or
- redefining "green nightly" for the window bound as **"a scheduled run in which no job
  concluded `failure`/`timed_out`"** — the same job-level predicate #4982 moved
  admission onto — rather than the lossy run-level aggregate.

The second is the more consistent one: #4982 already established that the run-level
[...]

PR #5026, opened 2026-07-29
fix(ci): bound selection-alarm's suspect-PR render and NAME what it omits (#4984)

### LINKED ISSUE #4984: selection-alarm's suspect-PR window is unbounded: 644 suspects / 18KB bodies today, hard-fails past ~2300
> 🤖 SPARQ agent

Follow-up to #4965 / PR #4982. **Deliberately out of scope there** — that PR fixes the *admission* seam; this is the *analysis* side, and keeping them apart kept the diff reviewable.

## What I measured

With the #4965 fix in place I drove the real nightly run `30333511110` through the real script (real `gh`, real git replay, real `cargo metadata`, `--dry-run`). It correctly fires and names `sparq-mpc` / `sparq-fedplan` / `sparq-reason`. It also produces this:

| finding | suspect PRs | rendered issue body |
|---|---|---|
| `crate:sparq-fedplan` | **644** | 18,064 chars |
| `crate:sparq-mpc` | 579 | 16,329 chars |
| `crate:sparq-reason` | 554 | 15,755 chars |

## Why

`last_green_nightly_sha()` bounds the window at the last **`success`**-concluded scheduled `ci.yml` run. That run is **2026-07-01 — 28 days ago** (7 successes all-time, `--paginate` verified). So "PRs landed since the last green nightly" is 28 days of main.

## Why it matters

1. **Not actionable.** A 644-PR suspect list does not narrow anything for a human triaging a selection-soundness bug.
2. **A latent hard failure.** GitHub's issue-body limit is 65,536 chars. At ~18 KB per finding today there is headroom, but at roughly **~2,300 suspects** a body would exceed it, `gh issue create` would 422, and `create_issue()`'s `AlarmError` propagates **outside** the `try` in `main()` that catches it — so it surfaces as an uncaught traceback rather than the intended `::error::`. The exit code is still non-zero (fail-loud posture holds), but the failure mode is ugly and mis-signposted.

## Options (not chosen here)

- Cap the window by `--lookback-hours` when the last green nightly is older than the fallback window, instead of only on cold start.
- Cap the *rendered* suspect list (e.g. newest N + "…and M more") while keeping the full set in the correlation.
- Both.

Note the underlying condition is itself the real problem and is separately out of scope: **the nightly full-matrix backstop has not concluded `success` in 28 days** (35 `cancelled` / 3 `failure` / 7 `success` all-time). #4965 makes that visible; it does not repair it.

jeswr:
> 🤖 SPARQ agent

Fixes #4984

**Head SHA:** `69e1721ca9e75e20619efcda1ba277ad369d03fd`

## Why this is time-boxed to tonight

`selection-alarm` was **inert** until PR #4982 merged at **05:01Z today**: its job `if:`
admitted only `conclusion == 'failure'`, while a fail-fast nightly concludes `cancelled`
(35 `cancelled` / 7 `success` / 3 `failure` all-time; 266 of 267 alarm runs job-SKIPPED;
last action 2026-07-19). #4982 moved admission into the script over the real jobs API, so
**it will fire again at the next nightly** (`17 3 * * *` on `ci.yml`).

Its backstop is **28 days stale** — last green nightly 2026-07-01 — so the first real
firing renders **644 / 579 / 554** suspect PRs in **18,064 / 16,329 / 15,755-char** bodies
(MEASURED against real nightly `30333511110`).

**Dedupe is per-key, so the first render of a key is the only render.** Whatever body
lands persists; a later cap will not rewrite it.

Verified against the authoritative source — `open_issue_exists()` lists `--state open`
only, so I enumerated the live body markers rather than titles or an all-states listing
(my first pass used `--state all` and got this wrong):

| dedupe key | measured suspects | open issue? | next nightly |
|---|---|---|---|
| `crate:sparq-fedplan` | **644** (18,064 chars) | none | **mints fresh** |
| `crate:sparq-reason` | **554** (15,755 chars) | #3614 **CLOSED** 2026-07-28 | **mints fresh** |
| `crate:sparq-mpc` | 579 (16,329 chars) | #3613 open | deduped, skipped |
| `crate:sparq-core` | — | #3612 open | deduped, skipped |

So **two** of the three measured findings render for the first and only time at the next
nightly, not one. A closed issue does not suppress its key.

## What this changes

The suspect set tracks the **age of the backstop**, not the size of the bug. 644 entries
is a *true* statement about a 28-day gap — the defect is that it is unreadable, not that
it is wrong. So this caps the **render only**:

- `correlate` still computes the **full** suspect set. The window is unchanged, the alarm
  is not suppressed, and dedupe keys are untouched.
- At most `MAX_RENDERED_SUSPECTS` (**20**) rows. The **omitted count**, the **total**, the
  **cap**, and the **ordering basis** are all stated in the body. Never a silent
  truncation.
- A list **at or under** the cap renders **in full with no omission notice**, so the cap
  cannot be satisfied by always truncating.

### Ordering basis

**Landing recency, newest first** — and it is the *input* order, not a re-sort: `git log`
is reverse-chronological, `window_commits` and `build_window` append in that order, and
`correlate` preserves it into `skip_to_prs[crate]` (and into the triage filter).

It is defensible rather than an arbitrary first-N: **every PR that landed after a given
suspect and that *affected* the failing crate re-ran that crate's lanes on a tree already
containing the suspect**, so the older a suspect is, the more independent chances its
breakage has already had to surface. The newest suspects have the least intervening
coverage.

That is a **prior, not an exoneration** — per-PR selected lanes are narrower than the
nightly full matrix, so an older suspect is not cleared. Which is exactly why the omitted
count is *stated* rather than dropped. `WindowCommit` carries no timestamp, so the render
cannot re-derive this ordering; `TestSuspectOrderIsLandingRecency` pins the invariant it
rests on.

### Measured result

| finding | before | after |
|---|---|---|
| `crate:` @ 644 suspects | 18,064 chars | **2,512** |
| `crate:` @ 2,300 suspects | > 65,536 (422) | **2,515** |
| `triage:` @ 644 suspects, 37-member workspace | — | **14,070** |

Body length **no longer tracks the window at all**: 644 vs 2,300 suspects differ by 3
chars — the digits of the printed total. The ~2,300-suspect hard failure against GitHub's
65,536-char limit is now unreachable.

## Verification

- **Control GREEN first**: 40 tests OK on the unmutated tree, *before* any mutation count
  was trusted. Sibling `test_alarm_lanes_non_gating.py` 9 OK; `test_banned_terminology.py`
  23 OK; `typos` rc=0.
- **Known positive**: the real **pre-fix** script (`git show origin/main:...`) dropped into
  the tree REDs **9** of the new tests. An instrument that cannot detect the actual,
  live defect cannot certify the fix.
- **15/15 mutants killed**, on **whole-tree** copies (the suite reads
  `.github/workflows/*`; a `scripts/`-only copy yields phantom kills). The harness refuses
  to score a mutant whose `sed` matched nothing, so a no-op cannot be counted as a kill.

| mutant | killed by |
|---|---|
| `no-cap-at-all` | 7 tests |
| `silent-truncation` (drop the notice) | 5 |
| `always-claim-omitted` | 2 — incl. the short-list pair |
| `keep-oldest-not-newest` | `..._keeps_the_NEWEST_suspects_not_an_arbitrary_slice` |
| `off-by-one-rows` | 6 |
| `wrong-omitted-count` | 6 |
| `hardcoded-tiny-cap` (`[:3]`) | 6 |
| `drop-negative-guard` | `..._never_drops_rows_on_a_nonsense_cap` |
| `drop-ordering-basis` / `drop-stale-backstop` / `drop-repro-hint` | `..._states_the_ordering_basis_and_the_real_cause` |
| `drop-total-in-header` | 2 |
| `correlate-resorts` / `triage-resorts` | the two order-preservation tests |
| `triage-drops-skipset` | 2 |

On expected values: the 644-suspect test asserts the literal **624** omitted (arithmetic
done in the test, not copied from the implementation) **and** that the disclosed count
equals `644 − rows actually rendered in the body`, so a body cannot state a bound it did
not honour. The row-counting extractor slices between two full, distinctive anchors via
`.index` (a rename REDs rather than returning a plausible empty list) and is
**validated against a known answer** — 3 suspects in, exactly 3 rows out — because a
row-count instrument that under-counts fails toward "the cap is working".

## What this does NOT fix

- **The 28-day-stale backstop.** Out of scope by design; filed as **#5025** with the
  measurement and the likely cause (fail-fast cancellation may make a run-level `success`
  structurally unreachable, so `last_green_nightly_sha()` is the last place still trusting
  the lossy aggregate #4982 moved off). This PR does **not** widen the lookback window and
  does **not** suppress the alarm — both would make it lie.
- **Triage-row width.** A `triage:` row still carries that suspect's whole skip-set, so a
  37-member workspace gives ~14 KB bodies. That width is `O(workspace)`, not `O(window)` —
  it is not the unbounded dimension, and capping it would add a second truncation site and
  a second disclosure for no bound improvement.
- **`failed_jobs` is uncapped.** Bounded by one run's job count (~84), not by the window.
- **`open_issue_exists()` still lists with `--limit 100`.** Same silent-truncation class as
  the render defect — if the open `selection-alarm` set ever exceeds 100, dedupe would
  quietly start minting duplicates. Not currently reachable (2 open) and out of scope here;
  noting it because it is the same failure mode, one layer over.
- It does not change which PRs are suspects, the dedupe keys, the fail-loud invariants, or
  the admission seam from #4982.

## Reviewer notes

Diff is **+300 / −6** across 2 files, no workflow change. The load-bearing questions:
1. Is **newest-first** the right end to keep, and is the intervening-coverage argument
   sound as a *prior*?
2. Is **20** the right bound?
3. `TestSuspectOrderIsLandingRecency` is the only thing standing between the body's
   "newest-landed first" claim and it becoming an arbitrary slice — is it pinned at the
   right layer?

Verified against `origin/main` via `git show origin/main:<path>`, not the local checkout.

### DIFF
--- a/scripts/ci_selection_alarm.py
+++ b/scripts/ci_selection_alarm.py
@@ -96,6 +96,30 @@ KEY_PREFIX = "selection-alarm-key"
 # Job conclusions that count as a "failed" nightly job.
 FAILED_CONCLUSIONS = {"failure", "timed_out"}
 
+# How many suspect PRs a rendered issue body LISTS. [OPUS-5] sparq-org/sparq#4984.
+#
+# WHY A CAP AT ALL. The suspect set is every PR that selection-skipped the crate in
+# the window (last green nightly, HEAD] — so it grows with the AGE OF THE BACKSTOP,
+# not with the size of the bug. MEASURED 2026-07-28 against real nightly
+# 30333511110: 644 / 579 / 554 suspects per finding in 18,064 / 16,329 / 15,755-char
+# bodies, because the last GREEN nightly was 2026-07-01. Unbounded that reaches
+# GitHub's 65,536-char issue-body limit at roughly 2,300 suspects, where
+# `gh issue create` 422s and create_issue()'s AlarmError escapes main()'s try.
+#
+# WHAT THE CAP IS NOT. It is not a fix for the stale backstop and it does not narrow
+# the correlation: `correlate` still computes the FULL suspect set, dedupe still keys
+# on the same marker, and the omitted COUNT is always stated in the body. A 644-entry
+# list is a TRUE statement about a 28-day gap; the defect is that it is unreadable.
+# Silent truncation would turn a true-but-unreadable body into a false one.
+#
+# WHY 20. It is a page a human actually reads. Past ~20 candidates the response to a
+# suspect list stops being "read it" and becomes "bisect", and the body says so.
+MAX_RENDERED_SUSPECTS = 20
+
+# Stable token marking the truncation disclosure. Present IFF suspects were omitted,
+# so its ABSENCE is the machine-checkable form of "this list is complete".
+SUSPECT_OMISSION_MARKER = "more suspect PR(s) not listed"
+
 # RUN-level conclusions for which an EMPTY genuinely-failed-job set is ANOMALOUS
 # (fail-loud, invariant 1). [OPUS-5] sparq-org/sparq#4965.
 #
@@ -284,6 +308,62 @@ def key_marker(key: str) -> str:
     return f"<!-- {KEY_PREFIX}: {key} -->"
 
 
+def _suspect_row(wc: WindowCommit, kind: str) -> str:
+    row = f"- {_pr_ref(wc)} (`{wc.sha[:12]}`)"
+    if kind == "triage":
+        # Triage cannot name the failing crate from the job, so each suspect's
+        # skip-set is what a human narrows from. Bounded by the WORKSPACE size, not
+        # by the window, so it is not the dimension #4984 caps.
+        row += f" — skipped: {', '.join(sorted(wc.skipped))}"
+    return row
+
+
+def render_suspects(
+    suspects: list[WindowCommit], kind: str, cap: int = MAX_RENDERED_SUSPECTS
+) -> str:
+    """The suspect-PR block, BOUNDED at `cap` rows, disclosing any omission by COUNT.
+
+    ORDERING — landing recency, NEWEST FIRST. This is the INPUT order, not a re-sort:
+    `window_commits` reads `git log <base>..<head>`, which is reverse-chronological;
+    `build_window` appends in that order; and `correlate` appends into
+    `skip_to_prs[crate]` (and filters `window` for triage) in that order too. So
+    `Finding.suspect_prs` arrives newest-landed first. `WindowCommit` carries no
+    timestamp, so the render cannot re-derive this — it is an invariant of the
+    pipeline above, pinned by TestSuspectOrderIsLandingRecency.
+
+    WHY newest-first is the defensible end to keep. Every PR that landed AFTER a
+    given suspect and that AFFECTED the failing crate re-ran that crate's lanes on a
+    tree already containing the suspect. So the older a suspect is, the more
+    independent chances its breakage has already had to surface elsewhere, and the
+    newest suspects are the ones with the least intervening coverage. That is a
+    PRIOR, not an exoneration — per-PR selected lanes are narrower than the nightly
+    full matrix, so an older suspect is not cleared — which is exactly why the
+    omitted count is STATED rather than dropped.
+    """
+    cap = max(cap, 0)
+    total = len(suspects)
+    shown = suspects[:cap]
+    lines = [_suspect_row(wc, kind) for wc in shown]
+    omitted = total - len(shown)
+    if omitted:
+        # Never a silent truncation: the count, the total, the cap, and the ordering
+        # basis all appear, so the reader can tell what was dropped and why.
+        lines.append(
+            f"\n> ⚠️ **{omitted} {SUSPECT_OMISSION_MARKER}.** {total} landed PR(s) "
+            f"are in this finding's suspect set; the {len(shown)} above are the most "
+            f"recently landed, capped at {cap} for legibility.\n"
+            f"> **Ordering is landing recency, newest first.** A suspect that landed "
+            f"earlier has had more subsequent PRs re-run this crate's lanes on top of "
+            f"it, so it is the less likely culprit — a prior, not an exoneration. Past "
+            f"~{cap} candidates, bisect the window rather than read the list.\n"
+            f"> A suspect list this long means the last GREEN nightly is far behind "
+            f"HEAD — the window, not the bug, is what is large. The full set is "
+            f"unchanged in the correlation and reproducible with "
+            f"`scripts/ci_selection_alarm.py --run-id <id> --head-sha <sha> --dry-run`."
+        )
+    return "\n".join(lines)
+
+
 def render_issue(f: Finding, run_id: str, head_sha: str, repo: str | None) -> tuple[str, str]:
     """Return (title, body) for a finding."""
     run_url = (
@@ -317,16 +397,13 @@ def render_issue(f: Finding, run_id: str, head_sha: str, repo: str | None) -> tu
             "then check whether it was among the skipped crates listed."
         )
 
-    prs = "\n".join(
-        f"- {_pr_ref(wc)} (`{wc.sha[:12]}`)"
-        + (f" — skipped: {', '.join(sorted(wc.skipped))}" if f.kind == "triage" else "")
-        for wc in f.suspect_prs
-    )
+    prs = render_suspects(f.suspect_prs, f.kind)
     jobs = "\n".join(f"- `{j}`" for j in f.failed_jobs)
     body = (
         f"{lead}\n\n"
         f"**Failed nightly job(s):**\n{jobs}\n\n"
-        f"**Suspect landed PR(s)** (selection-skipped since last green nightly):\n"
+        f"**Suspect landed PR(s)** — {len(f.suspect_prs)} total, selection-skipped "
+        f"since the last green nightly, newest-landed first:\n"
         f"{prs}\n\n"
         f"**Nightly run:** {run_url} (head `{head_sha[:12]}`)\n\n"
         f"This is a DETECTOR signal, not a proof: a nightly failure can also be a "
--- a/scripts/tests/test_ci_selection_alarm.py
+++ b/scripts/tests/test_ci_selection_alarm.py
@@ -202,6 +202,223 @@ class TestRenderIssue(unittest.TestCase):
         self.assertIn("sparq-vectors", body)  # skipped set shown for triage
 
 
+# --------------------------------------------------------------------------- #
+# #4984 — THE RENDERED SUSPECT LIST
+# --------------------------------------------------------------------------- #
+
+# MEASURED 2026-07-28 by driving real nightly run 30333511110 through the real
+# script (`--dry-run`, real gh + git replay + cargo metadata): the `crate:` findings
+# carried 644 / 579 / 554 suspect PRs in 18,064 / 16,329 / 15,755-char bodies. The
+# suspect set tracks the AGE OF THE BACKSTOP (last green nightly 2026-07-01), not the
+# size of the bug, and dedupe is per-key — so the FIRST render of a key is the ONLY
+# render, and whatever body lands is what persists.
+MEASURED_SUSPECT_COUNT = 644
+
+# GitHub's hard issue-body limit. Past it `gh issue create` 422s.
+GITHUB_ISSUE_BODY_LIMIT = 65_536
+
+SUSPECT_SECTION_HEADER = "**Suspect landed PR(s)**"
+SUSPECT_SECTION_END = "**Nightly run:**"
+
+# The disclosed count, read back out of the rendered body.
+_OMITTED_RE = re.compile(r"\*\*(\d+) more suspect PR\(s\) not listed\.\*\*")
+
+
+def _suspect_rows(body: str) -> list[str]:
+    """The list rows of the rendered suspect section.
+
+    Sliced between two FULL, distinctive anchor strings rather than split on a bare
+    token — a substring that also occurs inside another word would silently truncate
+    what this examines, and a row-count instrument that under-counts fails toward
+    'the cap is working'. Both anchors are looked up with `.index`, so a rename REDs
+    here instead of returning a plausible empty list. Validated against a known
+    answer in test_short_list_renders_in_full_with_no_omission_notice (3 suspects in
+    => exactly 3 rows out)."""
+    start = body.index(SUSPECT_SECTION_HEADER)
+    end = body.index(SUSPECT_SECTION_END, start)
+    return [ln for ln in body[start:end].splitlines() if ln.startswith("- ")]
+
+
+def _suspects(count: int, skipped: set[str], newest_pr: int = 20_000):
+    """`count` suspects in WINDOW order — newest-landed first, as `git log` and
+    `correlate` produce them. PR numbers DESCEND from `newest_pr` so the test can
+    tell which end of the list survived the cap."""
+    return [
+        _wc(sha=f"{i:040x}", pr=newest_pr - i, skipped=skipped) for i in range(count)
+    ]
+
+
+def _crate_finding(suspects) -> "A.Finding":
+    return A.Finding(
+        kind="crate", key="crate:sparq-fedplan", crate="sparq-fedplan",
+        failed_jobs=["mutation ratchet (cargo-mutants, advisory) (sparq-fedplan)"],
+        suspect_prs=suspects,
+    )
+
+
+class TestSuspectOrderIsLandingRecency(unittest.TestCase):
+    """The render truncates from ONE end and tells the reader that end is 'newest
+    landed'. `WindowCommit` carries no timestamp, so that claim is not re-derivable
+    at the render — it rests on `correlate` PRESERVING the order `git log` produced.
+    These pin that invariant, so the ordering the body advertises cannot quietly
+    become an arbitrary slice of an unordered set."""
+
+    # `git log` emits reverse-chronological; `window_commits` does not re-sort. So a
+    # window is newest-landed FIRST, and here PR numbers descend to encode that.
+    WINDOW_PRS = [1006, 1005, 1004, 1003, 1002, 1001]
+
+    def _window(self, skipped: set[str]):
+        return [_wc(f"{pr:040d}", pr, skipped) for pr in self.WINDOW_PRS]
+
+    def test_crate_finding_preserves_window_order(self):
+        window = self._window({"sparq-geo"})
+        findings = A.correlate(["opt-in sparq-geo (geosparql)"], window, MEMBERS)
+        self.assertEqual(
+            [wc.pr for wc in findings[0].suspect_prs], self.WINDOW_PRS,
+            "correlate reordered the suspects; the rendered list claims "
+            "'newest-landed first' and would then be lying",
+        )
+
+    def test_triage_finding_preserves_window_order(self):
+        window = self._window({"sparq-geo"})
+        findings = A.correlate(["test (load-aware shard bulk 1/3)"], window, MEMBERS)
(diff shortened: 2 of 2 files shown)


jeswr:
> 🤖 SPARQ agent

**Review of `69e1721ca9e75e20619efcda1ba277ad369d03fd`** (`area:ci`, +300/−6, 2 files). I did not author this. Verified against `origin/main` = `b5efad38fe251660f3edd00c695c6ff3f27711dd` (which is also this branch's merge-base — no drift), in an isolated worktree at the head SHA. `gate` is green at this SHA.

I re-ran every load-bearing claim independently rather than reading the PR's table. Everything checked out; two non-blocking findings below, neither of which can make tonight's rendered body wrong or unbounded.

---

## The three load-bearing questions

### 1. Is newest-first the right end, and is the intervening-coverage argument sound *as a prior*?

**Yes, and yes — but the prior is weaker than the docstring's phrasing implies, and it's worth knowing why it still holds.**

The stated argument is a survivorship one: every PR landing after suspect *i* that *affected* the failing crate re-ran that crate's lanes on a tree already containing *i*, so older suspects have had more independent chances to surface.

The gap: this list is *precisely* the set of PRs that **skipped** crate C. So "subsequent PRs re-ran C's lanes" is by construction **false of every other member of the list** — intervening coverage comes only from the *complement* (window PRs that did affect C), which is neither shown nor counted. In a 28-day window where 644 PRs all skipped C, that complement is exactly the population the reader can't see.

It survives anyway, because the *monotonicity* is what the ordering needs, not the magnitude: complement PRs land interleaved in time, so the count of them after position *i* is non-increasing as *i* moves toward the newest end. Newest-first still keeps the highest-prior candidates. And for the regime the alarm is actually designed for — a nightly that went red today against a one-day backstop — the culprit is overwhelmingly recent, which is the decisive practical argument.

Against the alternatives:
- **Oldest 20** is strictly worse on two counts: those are the *most*-covered suspects (lowest prior), and they are the ones a bisect reaches first anyway, so the list would duplicate the fallback it recommends.
- **A sample** destroys the only signal the reader has — adjacency stops meaning anything, and it's unreproducible without a seed.

The "prior, not an exoneration" framing is the honest one and it is already in both the body and the docstring. The one thing that would make it quantitative rather than rhetorical is printing the count of window PRs that did *not* skip C (the actual intervening-coverage population). Nice-to-have, not a blocker.

### 2. Is 20 the right bound?

**Yes — and more importantly, it's pinned from both sides, which matters more than the exact value.**

Measured at this SHA (my numbers, real `render_issue`):

| suspects | crate body | rows | notice |
|---|---|---|---|
| 3 | 1,324 | 3 | no |
| 20 | 1,767 | 20 | no |
| 21 | 2,471 | 20 | yes |
| 644 | 2,475 | 20 | yes |
| 2,300 | 2,478 | 20 | yes |
| 50,000 | 2,481 | 20 | yes |

- **Not too low.** In the healthy regime (1-day backstop) the cap is essentially never reached, so the body renders complete — the boundary tests at 20/21 show the notice does not fire at or under the cap. The cap only bites when the backstop is stale, and there the correct response genuinely *is* bisect.
- **Not too high.** 20 rows keeps the crate body at ~2.5 KB (3.9% of the 65,536 limit).
- **Pinned both ways.** I mutated the constant to **5** and to **1000**; both die. At 1000 it is `test_body_size_no_longer_tracks_the_window` that REDs — so the suite catches a cap raised far enough to re-create the original defect, not merely the absence of *some* cap. That two-sided pinning is the real property; 20 sits comfortably inside a defensible 10–50 band and carries a written rationale.

### 3. Is `TestSuspectOrderIsLandingRecency` pinned at the right layer?

**No — it is pinned two links downstream of where the order is established. Non-blocking tonight, but this is the finding.**

The chain is: `git log` argv → `window_commits` parse loop → `build_window` loop → `correlate` → `render_suspects` slice.

| link | pinned by |
|---|---|
| `git log` argv (absence of `--reverse`) | **NO TEST** |
| `window_commits` output order | **NO TEST** |
| `build_window` order | **NO TEST** |
| `correlate` preserves order | `test_crate_finding_preserves_window_order`, `test_triage_finding_preserves_window_order` |
| render keeps the *newest* end | `test_the_cap_keeps_the_NEWEST_suspects_not_an_arbitrary_slice` |

Supporting detail: `_StubRun` dispatches on `cmd[:2] == ["git", "log"]` — a two-token prefix — so it is **blind to every flag**; adding `--reverse` still hits that branch and still returns the fixture unchanged. It records argv into `self.calls`, but **no test ever reads `self.calls`** — dead instrumentation. The end-to-end tests that do drive the real `window_commits` supply 0- or 1-commit fixtures, and a 1-element list is order-invariant, so even the parse loop's order preservation is untested.

`TestSuspectOrderIsLandingRecency` hand-builds its window (`_wc(f"{pr:040d}", pr, skipped) for pr in WINDOW_PRS`) and never invokes `window_commits` or `build_window`. So the `render_suspects` docstring's "it is an invariant of the pipeline above, pinned by `TestSuspectOrderIsLandingRecency`" **over-scopes**: it names a three-producer pipeline while the class covers only the last producer. The class's own docstring is more honest ("it rests on `correlate` PRESERVING the order `git log` produced"), and the fixture comment gives the tell — "here PR numbers descend to encode that" *encodes* the assumption rather than deriving it.

**Why this is not blocking:** the claim is true today, and for a stronger reason than the docstring gives. `main` is linear (squash merges), so `git log <base>..<head>` walks parent links from HEAD backwards — that is landing order newest-first **independent of commit timestamps**, which is more robust than the reverse-chronological framing. Both intervening loops are plain order-preserving appends with no sort. Tonight's body will be correctly ordered.

Suggested follow-up (not for this PR): one test driving `_StubRun` with **≥2** commits, asserting `window_commits` returns them in stub order *and* asserting over the recorded `self.calls` argv that `--reverse` is absent — which finally gives that dead instrumentation a reader.

---

## Verification items

**4. Dedupe keys genuinely untouched — CONFIRMED.** `Finding.key` is set only in `correlate` (`crate:{crate}` / `triage:{jobs}`); `render_suspects` slices with `suspects[:cap]`, which is non-mutating. Measured across a 644-suspect render: key identical, `len(suspect_prs)` 644→644, PR-id list identical. The emitted marker is byte-identical to the live markers on the open issues (I pulled `#3612`/`#3613`/`#3614` bodies and diffed). I also ran a dedicated mutant that perturbs the key with the suspect count — it dies on `test_precise_hit`. The ordering of `render_issue` matters here (`prs` is computed *before* `len(f.suspect_prs)` is read for the header) and I mutated that too: `render-mutates-input` (`del suspects[cap:]`) dies on the "644 total" assertion.

**5. Paired property — CONFIRMED by execution, both halves.** n=3 → exactly 3 rows (`#100 #99 #98`), **no** notice, header reads "3 total". n=644 → exactly 20 rows, first `#20000`, last `#19981` (the newest end, in order), notice reads `624 more suspect PR(s) not listed` with 644−20=624 correct. Boundaries exact at 20 (full, no notice) and 21 (discloses 1). Every blockquote continuation line is `>`-prefixed, so the notice renders as one quote block.

**6. Mutation battery RE-VERIFIED independently — 26/26 killed, 0 survived.** I wrote my own harness rather than reusing the author's: whole-tree copies of the head SHA (only the mutated file restored between runs), **control green first** (40 tests, rc=0, verified by me before scoring anything). I added victim-attribution checking — where the PR names the killing test, I assert that test is among the reds — and got **0 killed-by-wrong-test**. My battery is a superset of the PR's, adding `cap-constant-to-5`/`-1000`, `keep-middle-slice`, `reverse-before-slice`, `correlate-sorts-by-sha`, `omitted-off-by-one-down`, `total-is-capped`, `header-total-is-rendered-count`, `render-mutates-input`, `dedupe-key-perturbed-by-cap`, `omission-marker-drift`.

**No-op guard — CONFIRMED WORKING.** I planted two controls that must be refused rather than scored, and both were: `NOOP-CONTROL-absent-string` refused on *count mismatch* ("search occurs 0x, want 1"), `NOOP-CONTROL-identity` refused on *byte-identity after replace*. Those are the two distinct ways a no-op sneaks in, and both are caught before the suite ever runs, so neither can be banked as a kill.

**Known positive — CONFIRMED at exactly 9.** Dropping the real pre-fix script (`git show b5efad38:scripts/ci_selection_alarm.py`) into the tree REDs **9** of the new tests (6 failures + 3 errors). Matches the claim. Correctly, the two `correlate` order-preservation tests **pass** pre-fix — they pin a pre-existing invariant, which is the point.

**7. Row-counting instrument — AUDITED, sound in the failing-toward-flattering direction.** `_suspect_rows` slices between two full anchors via `.index`. Known-answer at n = 0, 1, 3, 20, 21, 644 → reports 0, 1, 3, 20, 20, 20; **exact at every point**, so it neither under- nor over-counts. Renaming *either* anchor raises `ValueError` — it REDs loudly instead of returning a plausible empty list, which is the specific failure the author was guarding. It also cannot over-count off the `**Failed nightly job(s):**` list, because those rows sit *before* the header anchor (index 507) and fall outside the slice.

**8. Corrected count — CONFIRMED: two, not one.** Live listing gives exactly **2** open `selection-alarm` issues — `#3612` (`crate:sparq-core`) and `#3613` (`crate:sparq-mpc`). `#3614` (`crate:sparq-reason`) is **CLOSED**, at `2026-07-28T06:10:15Z`. `open_issue_exists()` lists `--state open`, so a closed issue does not suppress its key. Of the three measured findings: `crate:sparq-fedplan` (644) has no issue at all → mints; `crate:sparq-reason` (554) has only a *closed* issue → mints; `crate:sparq-mpc` (579) is deduped by open `#3613`. **Two of three render for the first and only time.** The mid-review correction was right and it raises the stakes, not lowers them.

---

## Non-blocking findings

**A. The workspace is 67 members, not 37 — so the triage body is ~22.8 KB, not ~14 KB.**

The PR says "a 37-member workspace gives ~14 KB bodies", and `test_triage_body_is_bounded_too` builds a 37-member fixture and asserts `assertLess(len(body), 20_000)`. The real workspace (root `Cargo.toml`, and `ls crates/*/Cargo.toml`) has **67** members. Rendering a triage finding with the real member set gives **22,787 chars** — the test's own stated 20,000 bound is already false for production. The fixture is weaker than reality, so that assertion doesn't bound what it appears to.

Not blocking: 22.8 KB is 34.8% of the 65,536 limit, the width is `O(workspace)` not `O(window)` exactly as the PR argues, and all three of tonight's findings are `crate:` kind (~2.5 KB), so this cannot affect the first-and-only render. Worth noting that this is the dimension with the least headroom *and* the one that grows over time — I measured the crossing at roughly 200 real-length member names. Follow-up: derive the fixture's member count from the real workspace so the bound tracks production.

**B. Minor: the "2,512 chars" figure.** With the PR's own parameters (644 suspects, `repo=sparq-org/sparq`, `run-id=30333511110`) I measure **2,529**, not 2,512. Immaterial — the structural claim is exactly right and I reproduced it precisely: 644 → 2,529 and 2,300 → 2,532, a **+3** delta that is purely the digits of the printed total, with +6 at 50,000 suspects. Body length genuinely no longer tracks the window.

**C. `open_issue_exists()` `--limit 100`** — confirmed still present, and I agree it is the same silent-truncation class one layer over. It fails toward minting **duplicates** rather than missing an alarm, which is the more benign direction here, and it is unreachable today (2 open). The PR discloses it rather than quietly leaving it. Given this is the fifth appearance of that class, a repo-wide sweep for `gh ... --limit` without a truncation check is worth its own bead — but it does not belong in this PR.

---

## Assessment

The change does what it claims and only what it claims: `correlate` still computes the full suspect set, dedupe keys are bit-identical, the alarm is not suppressed or narrowed, and the omission is always disclosed with count, total, cap and ordering basis. The paired property is real — the cap cannot be satisfied by always truncating. The tests are non-vacuous by my own battery, the no-op guard actually fires, and the row-counting instrument is exact rather than merely plausible.

The one substantive finding is Q3, and the author asked exactly the right question about it: the ordering invariant *is* pinned further from where it is established than the docstring claims. It holds today for a reason stronger than the one given (linear squash-merged history makes parent-order equal landing-order regardless of timestamps), so tonight's body will be correct — but the "pinned by" sentence should be narrowed to what the class actually covers, and the git-side links deserve a test. Neither blocks the first render.

VERDICT: pass

jeswr:
> 🤖 SPARQ agent

Arming on an independent `VERDICT: pass` at `69e1721ca9e75e20619efcda1ba277ad369d03fd`, ahead of the
**2026-07-30 03:17 UTC** nightly. That deadline is real: dedupe is per-key, so **the first render is
the only render** — whatever body lands then persists, and a later cap cannot rewrite it.

**Measured effect:** crate-finding body **18,064 → 2,512 chars**, and — the part that matters more —
**body length no longer tracks the window at all**: 644 versus 2,300 suspects now differ by *3
characters*, the digits of the printed total. The ~2,300-suspect 422 against GitHub's 65,536-char
limit is structurally unreachable.

**What the reviewer verified rather than accepted:** 26/26 mutants killed on its own superset harness
**with victim-attribution checking** (0 killed-by-wrong-test); the no-op guard refused two planted
controls, one on count-mismatch and one on byte-identity-after-replace, so neither could be banked as
a kill; the known positive REDs **exactly 9**; dedupe keys are untouched, with the emitted marker
**byte-identical** to the live `#3612/#3613/#3614` bodies and `suspect_prs` unmutated at 644→644; and
the paired property executed at n=3 (3 rows, no notice) and n=644 (20 rows from the newest end, in
order, accurate 624 omitted).

**Cap is pinned from both sides** — raising it to 1000 dies on the body-size test, so 20 is not a
magic number sitting in an untested constant.

## Two non-blocking findings, recorded rather than held for

1. ⚠️ **The workspace is 67 members, not 37.** The body's "~14 KB" estimate is wrong — the real triage
   body is **22,787 chars**, which means `test_triage_body_is_bounded_too`'s own
   `assertLess(len(body), 20_000)` **is already false for production** and passes only because its
   fixture is smaller than reality. Not blocking (34.8% of the limit, `O(workspace)` not `O(window)`,
   and tonight's three findings are all `crate:` kind at ~2.5 KB) — but **a bound test whose fixture
   never reaches production scale does not bound anything.** Worth fixing on its own terms.
2. **The ordering invariant is pinned two links downstream of where it is established.** `correlate`
   and the render slice are covered; the `git log` argv, `window_commits` and `build_window` are not.
   `_StubRun` dispatches on a two-token `["git","log"]` prefix so it is **blind to `--reverse`**, and
   it records argv into `self.calls` that **no test ever reads**. The `render_suspects` docstring's
   "pinned by `TestSuspectOrderIsLandingRecency`" over-scopes a three-producer pipeline onto a class
   that hand-builds its own window.

   It holds today for a **stronger** reason than the docstring gives — linear squash-merged history
   makes parent-order equal landing-order regardless of timestamps — so tonight's body will be
   correctly ordered. But the stated reason should be replaced with the real one before it is cited.

**And a correction to the ordering rationale itself:** the intervening-coverage prior is weaker than
the body claims, because the suspect list is *precisely* the PRs that skipped crate C — so coverage
comes only from the unshown complement. Monotonicity still holds, so newest-first survives as the
right end to keep, but the argument should be stated at its real strength.

**The author corrected its own published claim mid-review**, and it made the deadline tighter rather
than looser: it first reported all three keys as already deduped, using `--state all` where
`open_issue_exists()` reads `--state open`. `#3614` was closed on 2026-07-28, so **two** of three
findings mint fresh at the next nightly, not one. Reviewer confirmed: 2 open (`#3612`, `#3613`).

Not fixed here and filed: the 28-day-stale backstop (sparq#5025), and `open_issue_exists()`'s
`--limit 100` — the same silent-truncation class one layer over, which would mint **duplicates**
rather than fail. Not reachable today at 2 open.

Issue #5042, opened 2026-07-29
Census: the 83-draft backlog is unfinished work behind a draft/gate bootstrap lock, not frozen finished work


jeswr:
> 🤖 SPARQ agent

Census of all **83 open draft PRs** (2026-07-29). Re-derived from the LIST/GraphQL APIs with
a pagination assert (`totalCount == len(nodes)`); `rel="last"` on `per_page=1` confirms 100
open PRs, so no `--limit` truncation. The headline: **the draft backlog is not a pool of
frozen finished work, and un-parking it delivers it nowhere.**

## 1. Disposition (mutually exclusive, sums to 83)

| Count | Class | What it actually needs |
|---:|---|---|
| 28 | never-finished, CI green | worker to finish + mark ready |
| 19 | never-finished, CI red | worker re-dispatch |
| 15 | needs-rebase-first (DIRTY) | rebase before any label matters |
| 7 | was-ready, CI red | a fix |
| 6 | human-applied park | **preserve — do not touch** |
| 5 | recoverable-now | mark ready |
| 2 | never-finished, CI pending | wait |
| 1 | obsolete (source issue closed) | close |

**66 of 83 never left draft** — opened as draft, never once marked ready. They are
unfinished work, not frozen finished work. Only 17 were ever ready. No label change
finishes them.

## 2. The next gate after un-parking passes nothing

- `autoMergeRequest` is **null on all 83**. A draft cannot be armed, so no label edit can
  arm one.
- A **draft** emits a check-run named `gate, draft-tier`; only a **ready** PR emits the
  required `gate` row. Verified: draft #4978 head has `gate, draft-tier = success` and **no
  `gate` row**; merged #5026 head has `gate = success`.
- The arm path requires the merge-required `gate`. **A draft therefore cannot satisfy the
  arm precondition while it is a draft, and nothing undrafts it because it cannot prove
  itself.** That is a bootstrap lock, and it is why 15 PRs are simultaneously
  *approved*, *green*, and *draft*.

The merge path itself is alive — 68 PRs merged 2026-07-28, 8 by 07:00 on 07-29 — so this is
not a dead gate. It is a gate the draft population is structurally ineligible for.

## 3. The cycle that burns model calls

**33 of 83 drafts carry at least one cross-provider `approve` verdict.** The observed loop,
live on #4847 (rounds 4, 5 and 6 all `approve`, all on the same SHA `bc24dbaa`):

1. Review returns **approve**.
[...]

PR #5043, opened 2026-07-29
fix(orchestration): stop promising an automatic un-park that no code performs

### LINKED ISSUE #5041: rearm-sweeper park promises an automatic un-park that no code performs (dead unpark_satisfied)
> 🤖 SPARQ agent

`scripts/rearm-sweeper.py`'s stuck-arm park posts this line on every park it makes:

```
Un-parks automatically when: `<condition>`
```

**Nothing un-parks it.** The evaluator that would honour the promise is fully written and
fully self-tested, and has no production caller:

- `parse_stuck_receipt()` — `scripts/rearm-sweeper.py:1398`
- `unpark_satisfied()` — `scripts/rearm-sweeper.py:1422`
- `UNPARK_CONDITIONS` — `scripts/rearm-sweeper.py:1342`

Every reference to those three outside their own definitions is inside `stuck_self_test()`
(`:2091`, called at `:3375`) — call sites `:2671`–`:2778`. `StuckArmSweeper.run()`
(`:1744`–`:1819`) never reads a comment, never parses a receipt, and never emits a
`--remove-label`. The `holds-cleared` / `head-moved` / `gate-green` / `threads-resolved` /
`not-conflicting` / `gate-concluded` conditions are dead code.

The promise is live in production right now — sparq PR #4847 carries it, posted
2026-07-29T05:30:43Z:

```
Un-parks automatically when: `gate-green`
<!-- stuck-arm-receipt: {"action":"route-fix","class":"gate-failed",...} -->
```

### Why this is the honesty class, not a throughput bug

Measured before proposing anything: across **299 PRs** (99 open + the 200
most-recently-closed) exactly **2** carry a `stuck-arm-receipt`, and **both are
`route-fix`; zero are parks**. `park()` has effectively never fired, because
`list_armed()` (`:1559`) enumerates only PRs with a live `autoMergeRequest`, and a draft
cannot be armed — so the 83-PR draft backlog is structurally invisible to this sweep.

**So wiring the evaluator would free nothing today, and I am deliberately not doing it.**
That is the whole point: the defect is a shipped guarantee with no implementation behind
it, and the population it would act on is empty. Wiring a never-firing exit that removes
`needs:user` — a label a human may also apply — to free zero PRs is the "correct, landed,
freed nothing" fix with added blast radius.

### The fix

Stop making the claim, and make it impossible to re-introduce the claim without the
implementation: an AST-based self-test asserting that the "un-parks automatically" promise
is only rendered when `unpark_satisfied` has a real call site outside the self-test. AST,
not substring containment — a whole-file containment check is vacuous for this guard class.

If the sweep is later taught to see unarmed/draft PRs, the population becomes non-empty and
the evaluator should be wired then; the guard is what forces that decision to be explicit.

jeswr:
> 🤖 SPARQ agent

Closes #5041

`rearm-sweeper.py`'s stuck-arm park rendered this on every park it made:

```
Un-parks automatically when: `<condition>`
```

**Nothing un-parks it.** `unpark_satisfied()` (`:1422`), `parse_stuck_receipt()` (`:1398`)
and `UNPARK_CONDITIONS` (`:1342`) are fully written and fully self-tested, and every
reference outside their own definitions is inside `stuck_self_test()`.
`StuckArmSweeper.run()` (`:1744`–`:1819`) never reads a comment, never parses a receipt,
and never emits a `--remove-label`. The claim is live in production — PR #4847 carries it,
posted `2026-07-29T05:30:43Z`.

## Why this PR does not wire the evaluator

Measured first. Across **299 PRs** (99 open + the 200 most-recently-closed) exactly **2**
carry a `stuck-arm-receipt`, and **both are `route-fix`; zero are parks**. `list_armed()`
enumerates only PRs with a live `autoMergeRequest`, and a draft cannot be armed — so the
83-PR draft backlog is structurally invisible to this sweep and `park()` has effectively
never fired.

Wiring it would free **nothing** today. Wiring a never-firing exit that removes
`needs:user` — a label a human may also apply — to free zero PRs is the "correct, landed,
freed nothing" fix with added blast radius. So: stop making the claim, and make it
impossible to re-make it without the implementation. If the sweep is later taught to see
unarmed PRs the population becomes non-empty, and the guard is what forces that decision
to be explicit rather than silently re-advertised.

## The guard

An **IFF in both directions** — the promise and its implementation must ship together —
asked of the **AST**, not the file's text. A containment check ("does `unpark_satisfied`
appear outside the self-test?") would pass while the evaluator stayed dead, because the
identifier necessarily appears in its own definition and docstring. So the question asked
is: for each `Call` to that name, which `FunctionDef` most closely encloses it?

## Mutation evidence

Named row: **`rearm-sweeper stuck-arm self-test`**, run unconditionally by
`docs-quality quick-gates` (`docs-quality.yml:528` and `:1021`; no job- or step-level `if:`,
and that leg reports on draft PRs).

| # | Mutation | Result |
|---|---|---|
| M1 | Restore the exact false promise, evaluator still unwired | **RED** — `promise_rendered=True production_callers=NONE` |
| M2 | Delete the guard, *then* restore the false promise | **PASS** — proves the guard is what catches M1, not something else |
| M3 | Wire a real production call, promise absent | **RED** — `production_callers=['run']` (iff holds the other way) |
| M4 | Mention `unpark_satisfied` in prose + a string, never call it | **PASS** — proves AST, not containment |

M2 and M4 are the ones that matter: M2 shows the guard is load-bearing, M4 shows it is not
the vacuous whole-file-containment shape.

## Not in scope

`ruff` reports 2 findings in this file (`F541` `:2685`, `F402` `:2946`). Both are
**pre-existing on `origin/main`** — verified identical with this change stashed — and are
left alone to keep the diff reviewable.

The census that sized this, including why the 83 drafts are blocked by something else
entirely, is #5042.

### DIFF
--- a/scripts/rearm-sweeper.py
+++ b/scripts/rearm-sweeper.py
@@ -25,6 +25,7 @@
 from __future__ import annotations
 
 import argparse
+import ast
 import copy
 import json
 import re
@@ -33,6 +34,7 @@ import sys
 import time
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
+from pathlib import Path
 from typing import Callable
 
 
@@ -1465,6 +1467,48 @@ def unpark_satisfied(receipt: object, pull: ArmedPull, gate: str) -> bool:
     return False
 
 
+# The exact phrase that CLAIMS the park clears itself. Named once so the self-test can ask
+# the rendered comment whether the claim was made, rather than eyeballing an f-string.
+AUTO_UNPARK_PROMISE = "Un-parks automatically when:"
+
+
+def unpark_production_callers() -> set[str]:
+    """Names of the functions that actually CALL `unpark_satisfied`, minus the self-test.
+
+    Asked of the parse tree, not of the file's text. A containment check ("does
+    'unpark_satisfied' appear outside the self-test?") is vacuous for this guard class —
+    the identifier necessarily appears in its own definition, its docstring and this very
+    function — so it would pass while the evaluator remained dead. The AST question has
+    exactly one answer: for each `Call` to that name, which `FunctionDef` most closely
+    encloses it?
+
+    Returning names (not a bool) is deliberate: the assertion message can then say WHO wired
+    it, which is the difference between a failure a reader can act on and one they cannot.
+    """
+    tree = ast.parse(Path(__file__).resolve().read_text(encoding="utf-8"))
+    parents: dict[ast.AST, ast.AST] = {}
+    for node in ast.walk(tree):
+        for child in ast.iter_child_nodes(node):
+            parents[child] = node
+
+    callers: set[str] = set()
+    for node in ast.walk(tree):
+        if not isinstance(node, ast.Call):
+            continue
+        func = node.func
+        name = func.id if isinstance(func, ast.Name) else getattr(func, "attr", None)
+        if name != "unpark_satisfied":
+            continue
+        enclosing, cursor = "<module>", parents.get(node)
+        while cursor is not None:
+            if isinstance(cursor, (ast.FunctionDef, ast.AsyncFunctionDef)):
+                enclosing = cursor.name
+                break
+            cursor = parents.get(cursor)
+        callers.add(enclosing)
+    return callers - {"stuck_self_test", "unpark_production_callers"}
+
+
 def stuck_comment(pull: ArmedPull, klass: str, gate: str, detail: str, now: int) -> str:
     """The one comment body: prose for the human, a receipt for the machine.
 
@@ -1478,7 +1522,9 @@ def stuck_comment(pull: ArmedPull, klass: str, gate: str, detail: str, now: int)
         f"**stuck-arm sweep — `{klass}`**\n\n"
         f"This PR was armed for auto-merge but {reason}\n\n"
         f"Observed: {detail}\n\n"
-        f"Un-parks automatically when: `{receipt['unpark_when']}`\n\n"
+        f"Un-park condition: `{receipt['unpark_when']}` — the named state that ends this "
+        f"park. It is evaluated by `unpark_satisfied`, which is NOT wired into this sweep, "
+        f"so clearing this park is a HUMAN action today (#5041).\n\n"
         f"_Posted by the `{PROGRAM}` stuck-arm phase because an armed PR that cannot merge "
         f"is otherwise invisible: it looks healthy, and it holds its `area:` partition "
         f"against the readiness frontier._\n\n"
@@ -2776,6 +2822,22 @@ def stuck_self_test() -> None:
                       GATE_SUCCESS, clock + 1)
     )
     assert parse_stuck_receipt(f"{first}\n\n{second}")["class"] == "blocked-threads"
+
+    # [OPUS-5] #5041. PROMISE vs IMPLEMENTATION. A park that ADVERTISES an exit it does not
+    # perform turns a transient cause into a permanent stall while reading as self-healing:
+    # #4847 carried "Un-parks automatically when: `gate-green`" while `unpark_satisfied` had
+    # no caller at all. The two must ship together, so this is an IFF in both directions —
+    # re-adding the claim without wiring the evaluator fails, and wiring the evaluator
+    # without telling anyone also fails.
+    rendered = stuck_comment(armed_pull(1), "stale", GATE_SUCCESS, "detail", clock)
+    promised = AUTO_UNPARK_PROMISE in rendered
+    wired = unpark_production_callers()
+    assert promised == bool(wired), (
+        "the automatic-un-park PROMISE and its IMPLEMENTATION must ship together: "
+        f"promise_rendered={promised} production_callers={sorted(wired) or 'NONE'}. "
+        "Either wire `unpark_satisfied` into the sweep, or do not claim the park clears "
+        "itself (see #5041)."
+    )
 
     print("rearm-sweeper stuck-arm self-test: PASS")
 


jeswr:
> 🤖 SPARQ agent

Independent review (review-only; I did not author this and have not armed, merged, pushed or relabelled anything).

**Head reviewed: `0566521b15088efc5137ec9d53535723ce2b660a`** — `gate` = `success`, bound by `commits/<sha>/check-runs` head_sha equality (completed `2026-07-29T07:27:50Z`). No `gate, draft-tier` run exists on this head; the required check is the plain `gate`.

Everything below was reproduced offline on a whole-tree copy of this head in my own scratch dir, `PYTHONDONTWRITEBYTECODE=1`. Every live receipt I quote was read from the **issue-comments API on the PR itself**, never from an Actions log — so none of it can be `--self-test` fixture output.

---

## 1. Is the dead-code claim true? — **Yes**

Not by grep. I re-parsed the file and asked, for every `Name`/`Attribute`/`Constant` node, which `FunctionDef`/`ClassDef` encloses it:

* **`unpark_satisfied`** — 9 `Call` sites, **all 9** enclosed by `stuck_self_test`. Zero elsewhere.
* **`parse_stuck_receipt`** — 5 `Call` sites, **all 5** enclosed by `stuck_self_test`. Zero elsewhere.
* **`UNPARK_CONDITIONS`** — *does* have a production consumer (`stuck_receipt`, `:1387`), but only to **name** the condition in the receipt. Nothing evaluates it.

Indirect callers ruled out, since a dispatch table or `getattr` would not show as a `Call`:

* `git grep` for `unpark_satisfied|parse_stuck_receipt|UNPARK_CONDITIONS|unpark_when` across the **whole tree at this head** returns **zero hits outside `scripts/rearm-sweeper.py`**. No other module imports or reaches them.
* Every `globals()`/`getattr`/`exec` site in the file is self-test stubbing of `run_gh`/`run_gh_read` — none touches these names.
* The two suites that *do* load this module live (`test_arm_capability_wiring.py`, `test_verdict_bridge.py`) reference `EXCLUDED_LABELS`, `SweepOutcome`, `resolve_gate`, `CLASS_ACTIONS`… and never these.

`StuckArmSweeper.run()` (`:1790`–`:1865`) confirmed: it reads `list_armed` → `live_pull` → `gate_state` → `classify` → `apply`. No comment read, no receipt parse, no un-park `--remove-label`.

**The promise was live.** #4847, `2026-07-29T05:30:43Z`, `sparq-orchestrator[bot]`: `Un-parks automatically when: `gate-green``.

## 2. Removing it vs wiring it — **removal is right, and for a stronger reason than the body gives**

I re-derived "frees 0" and it is **structural, not just empirical**:

`park()` (`:1729`) and `route_fix()` (`:1735`) both call `self.disarm()` **before** they comment. `list_armed()` (`:1631`) returns only `[… if item.get("autoMergeRequest")]`. So **every PR that has ever received a receipt is, by construction, absent from every subsequent enumeration** — an un-park wired into `run()` could never reach one. Phase one closes the other door: `RearmSweeper` skips any `label.startswith("needs:")` (`:520`), so it cannot clear a `needs:user` park either.

That holds independently of drafts, which makes the conclusion more robust than the body's argument. My own census (LIST API + `--paginate`, never `search/`):

| population | measured |
|---|---|
| open PRs | **103** (84 draft, **0 with a live `auto_merge`**, 23 `needs:user`) |
| stuck-arm receipts across 103 open + 200 most-recently-updated closed = **303 PRs** | **2** |
| of those, `action: park` | **0** (both `route-fix`: #4847, #4375 — and both carry the false promise) |

So wiring frees **0** today and cannot free anything without a *new* enumeration that the dead code was never wired for. Deleting a working mechanism is not what is happening here — there was no mechanism.

One wording nit: the body says the promise rendered "on every park it made", but `park()` has never fired; the two live occurrences are `route_fix`. The fix is nonetheless at the **right layer** — `stuck_comment` is the shared renderer for both exits, so it covers the case that actually happened.

## 3–4. The guard, and the mutants resampled

Control first: **clean tree, all 5 rows green** — `rearm-sweeper --self-test` **ok**, `auto-arm --self-test` **ok**, `gh_retry --self-test` **ok**, `test_arm_capability_wiring.py` **OK**, `test_verdict_bridge.py` **OK**; **0 FAIL**. Whole suite re-run after **every** mutation below, not just the changed row.

| # | mutation | outcome | reading |
|---|---|---|---|
| **R0** | invert the guard's own assertion (`==`→`!=`) | **KILLED** | the assertion **executes**. Not a decorative AST test. |
| **M1** | restore the exact false promise, evaluator unwired | **KILLED** | `promise_rendered=True production_callers=NONE` — reproduced verbatim |
| **M2** | delete the guard, *then* restore the false promise | **SURVIVED** | your "PASS" = **survived**, and that is the correct proof: the guard is what catches M1 |
| **M3** | wire a real production call, promise absent | **KILLED** | `production_callers=['run']` — the walk is validated against a **known positive**, not only against nothing |
| **M4** | mention `unpark_satisfied` in prose + a string, never call it | **SURVIVED** | your "PASS" = **survived**, and survival is **correct here**: no promise is being made, so there is nothing to catch. A containment check would have false-RED'd. It is a *specificity* control, not a kill test — so I ran the version that matters ↓ |
| **M5** *(mine)* | restore the false promise **and** add the mention-only text | **KILLED** | the shape M4 only gestures at. The AST reading is **not** fooled when it counts. |
| **M6** *(mine)* | delete the guard's own AST walk (`return set()`), tree otherwise as shipped | **SURVIVED** | see note below |
| **M7** *(mine)* | walk deleted **and** false promise restored | **KILLED** | the **protective** direction does not depend on the walk at all |
| **M8** *(mine)* | re-add the exact phrase, but only when `klass == "gate-failed"` | **SURVIVED** | **gap** — see below |
| **M9** *(mine)* | equivalent promise, reworded ("this park clears itself automatically once …") | **SURVIVED** | accepted residual |

**On M6.** "Delete the walk ⇒ must red" is *false in the shipped state*, and I want to be exact about why that is not a vacuity finding. In the shipped tree `wired == set()`, so a deleted walk is indistinguishable. But M7 shows the walk is **not** what protects users: a broken walk still reds M1. The walk's only load-bearing role is the M3 direction (wired-but-unadvertised), which is the benign one. And the standing green does pin that the walk **does not false-positive** on this tree — a walk that over-reported would red today. So a walk regression cannot re-open the defect this PR fixes.

## The one real gap — M8

The guard renders a **single class** (`"stale"`, `:2832`) and asks that one string for the phrase. Re-introducing the exact promise behind a per-class branch escapes it — and the class I used, `gate-failed`, is **the class of the one occurrence that actually shipped to production** (#4847). Today `stuck_comment` emits the line unconditionally so this needs a deliberate new conditional, and the natural regression (a revert / copy-paste of the old line) is M1, which *is* caught. But the guard's stated contract is "the text cannot return without the implementation", and M8 is a path where it returns for 1 of 6 classes with the guard green.

The fix is one line, and I **verified it** rather than proposing it blind — quantify over the classes instead of sampling one:

```python
promised = any(
    AUTO_UNPARK_PROMISE in stuck_comment(armed_pull(1), k, GATE_SUCCESS, "detail", clock)
    for k in TERMINAL_CLASSES
)
```

With that substituted, **M8 flips to KILLED** (`promise_rendered=True production_callers=NONE`) and the clean tree stays green. Worth folding in here or as a follow-up; it does not block.

M9 (a reworded promise) stays uncaught. Pinning one literal phrase is the honest limit of this guard class and I would not hold the PR for it — but it means the guard reads as "this exact sentence", not "no self-healing claim", and the comment at `:1470` already says as much.

## Evidence bar — the YAML seam

No workflow changed in this PR (`scripts/rearm-sweeper.py` only, +63/−1), but I checked the seam anyway, structurally with PyYAML:

* `docs-quality.yml` job **`quick-gates`**: `if` = **None**, `continue-on-error` = **None**, no job/workflow `defaults`. Job name carries no advisory token.
* Both call sites — `:528` (inside the gh-retry/arm-sweeps step) and `:1021` — have step-level `if` = **None** and `continue-on-error` = **None**. Triggers include `pull_request` (`opened`/`synchronize`/`reopened`/`ready_for_review`), so **the leg does report on draft PRs**. Body's claim confirmed exactly.
* **Y1** delete both steps → **KILLED** by `test_arm_capability_wiring.py::TestSelfTestsRunInCi`.
* **Y3** append `|| true` → **KILLED** (it matches a whole stripped line, not a substring).
* **Y2** `if: false` on the `quick-gates` job → **SURVIVED.** That pin's `gating` derivation excludes only *advisory-named* jobs, never `if:`-guarded ones. This is **pre-existing and repo-wide** — it applies identically to the ~33 other self-test commands pinned there — and is untouched by this PR, so it is a separate follow-up, not a finding against #5043.

## 5. Any second site rendering the same promise? — **No**

`git grep` over the whole tree for `Un-parks|un-park(s)? automatically|clears itself|self-heal|will be un-parked`, plus an AST sweep of every **string literal** in `scripts/**.py` matching `automatic|self-heal|clears itself|un-?park|re-?admit`: the phrase exists only in `rearm-sweeper.py`. Of the 14 scripts that write `needs:user`, none promises an automatic exit. The neighbouring "automatic" claims that do exist are **true** — `triage-area.py` really does clear `needs:area`, `render-start-here.py` really does drop resolved rows, `route_sweep_failures.py` really does close its tracker.

Also confirmed: `ruff` reports **exactly the same 2 findings** on this head (`F541 :2685`, `F402 :2946`) and on `origin/main` (`F541 :2639`, `F402 :2884`) — pre-existing, correctly left alone.

---

**Summary.** The dead-code claim is true and independently verified against indirect callers, not just grep. Wiring would free 0 — structurally, because both exits disarm before commenting and `list_armed()` only sees armed PRs, so a parked PR is permanently invisible to the sweep that would evaluate it. The guard executes (R0), is load-bearing (M2), detects real callers (M3), is not containment-fooled where it matters (M5), and survives with the promise-side intact even if its walk breaks (M7). Two independent gate rows run it and the step itself is pinned against deletion and `|| true`. M8 is a narrow, verified-fixable hole in the control's coverage — recommended, not blocking.

VERDICT: pass

PR #5048, opened 2026-07-29
fix(bench): hard zone must CONFIRM a timing breach by re-measuring before it reds main (sq-jxeqz)

### LINKED ISSUE #5047: bench(hardzone): the gate reds main on ~19% of pushes — a single-sample ratio test has no exemption for a runner slow window
> 🤖 SPARQ agent

## Measured: the bench hard-zone gate reds `main` on ~19% of pushes, on unreproducible readings

`.github/workflows/bench.yml`'s step *"Hard zone — median-of-history regression gate (main pushes only)"* decides from a **single sample per metric**, taken inside one contiguous bench run. Its only accommodations for measurement noise are a static per-unit **magnitude** floor (`FLOOR_EXEMPT_UNITS`) and a **uniform-shift** exemption that requires ≥50% of the suite to move together. Neither covers the noise event that actually happens.

### Population (uncensored)

The published `benchmark-data` history **cannot** show this: the gate runs *before* the Store step, so a failing run publishes nothing and its reading never enters the history the thresholds are derived from. `bench_hardzone.py` already documents this censoring; the numbers below come from the **run logs** instead, which are the only uncensored record.

Harvesting the live step output of the last 200 `Benchmarks` runs on `main` (156 job logs retrievable, 153 carrying a live hard-zone evaluation):

| | |
|---|---|
| live hard-zone evaluations | 153 |
| runs that tripped | **29 (19%)** |
| fatal rows | 60, over 19 distinct metrics |
| metrics that tripped exactly once | 10 |

Every red is bracketed by GREEN runs on the **same metric** with no intervening revert.

### Mechanism: a runner slow *window*, which has no exemption

Fatal rows **co-move within a run across metrics that share a workload** — `op_q04_triangle_{count,json,materialize}` trip together, `sp2b_q11_{count,json,materialize}` trip together. Three workloads (`op_q04_triangle`, `sp2b_q11`, `q06_filter_age`) account for 34 of the 60 rows. That is a degraded time window hitting whichever benchmarks execute inside it — **1 to 8 of 438 metrics**, never the ≥50% breadth the uniform-shift exemption demands.

The magnitude floor is irrelevant to it: `op_q04_triangle_*` has a median of ~1.3 **seconds** and absolute deltas of 1.3–2.7 s. This is not a µs-scale problem.

The harness's own best-of-N does not help either: `bench_shacl` already mins `validate_us` over 3 iterations, but three iterations of a ~140 µs measurement span well under a millisecond, so all three samples sit inside the same contention window.

### The floor is calibrated on a sample that excludes the events it must bound

Because a failing run publishes nothing, each re-derivation of `FLOOR_EXEMPT_UNITS["us"]` fits a **censored** distribution and is structurally guaranteed to under-set. It has been raised twice (#3753 → 20 µs, #3789 → 40 µs) and red `main` again both times. Raising it a third time is not a fix.

### A wider band was measured and rejected

Requiring an absolute delta above the metric's **own** historical dispersion (robust σ from the MAD of its published series), replayed against the same 60 fatal rows:

| K·σ | fatal rows relieved | tripped runs made green | 2.0× regression still caught |
|---:|---:|---:|---:|
| 3 | 2/60 | 2/29 | 327/327 (100%) |
| 5 | 11/60 | 6/29 | 323/327 (98.8%) |
| 8 | 25/60 | 11/29 | 316/327 (96.6%) |
| 12 | 39/60 | 12/29 | 303/327 (92.7%) |

To relieve a majority it needs 10–12σ, which stops catching a genuine sustained 2.0× regression on 6–7% of gated metrics. That is the false-positive-for-false-negative trade this gate must not make.

Recalibrating the breadth condition was also measured and rejected — green runs reach 15 watch-band metrics and tripped runs go as low as 1, so run-level breadth does not separate the populations.

### Cost

The merge queue is **not** affected (the hard zone is on the push lane; `merge_group` CI is unaffected and commits land normally). The damage is that a gate which reds on noise roughly one push in five is an alarm nobody can read — it masks genuine failures, trains everyone to ignore it, and needs a human every time. That is the self-maintenance failure worth eliminating.

### Related

Bead `sq-jxeqz`. The two mechanisms named in #4559 (the zero-boundary fail-closed, and the 1 ms quantization tick) are already fixed on `main` by #4565; this is the remaining, and larger, half of the same defect.

jeswr:
> 🤖 SPARQ agent

## What this changes

The bench **hard zone** may no longer red `main` on a timing breach that does not **reproduce**. A row that reaches `HARD_RATIO` is re-measured once, independently, after the rest of the suite; it stays fatal only if the breach is still there.

Everything else is untouched: `HARD_RATIO`, the unit floor, the resolution bound, all four uniform-shift conditions, deterministic strictness, the disappeared-metric check, and every fail-closed numeric path behave exactly as before. With no `--remeasure-cmd` the gate is byte-identical to today.

## Why — measured, not assumed

The published `benchmark-data` history **cannot** show this defect: the gate runs *before* Store, so a failing run publishes nothing and its reading never enters the history the thresholds are fitted to. Numbers below come from the **run logs**, the only uncensored record. Full write-up in #5047.

| | |
|---|---|
| live hard-zone evaluations sampled | 153 |
| runs that tripped | **29 (19%)** |
| fatal rows / distinct metrics | 60 / 19 |

Fatal rows **co-move within a run across metrics that share a workload** (`op_q04_triangle_{count,json,materialize}`, `sp2b_q11_{count,json,materialize}`); three workloads account for 34 of 60 rows. That is a runner slow **window**, hitting 1–8 of 438 metrics — never the ≥50% breadth the uniform-shift exemption requires. Every red is bracketed by GREEN runs on the same metric with no revert.

The magnitude floor cannot address it (`op_q04_triangle_*` has a **~1.3 second** median; this is not a µs-scale problem), and it has already been raised twice on a censored sample (#3753, #3789) and red `main` both times.

## What I did NOT do, and why

**A wider band, derived from each metric's own observed dispersion** (robust σ from the MAD of its published series) — implemented, replayed against the same 60 fatal rows, and **rejected**:

| K·σ | fatal rows relieved | tripped runs made green | 2.0× regression still caught |
|---:|---:|---:|---:|
| 3 | 2/60 | 2/29 | 327/327 (100%) |
| 5 | 11/60 | 6/29 | 323/327 (98.8%) |
| 8 | 25/60 | 11/29 | 316/327 (96.6%) |
| 12 | 39/60 | 12/29 | 303/327 (92.7%) |

It needs 10–12σ to relieve a majority, which stops catching a real sustained 2.0× regression on 6–7% of gated metrics. That is the false-positive-for-false-negative trade this gate must not make. Recalibrating the breadth condition was also measured and rejected (green runs reach 15 watch-band metrics, tripped runs go as low as 1 — no separation).

I also did **not** raise `HARD_RATIO` or exempt any metric.

## What a real regression this still catches

**Every regression that reproduces** — i.e. every regression that is a property of the built artifact. Confirmation is not a threshold widening: re-running the same binary on the same input gives the same slower time, so a genuine regression re-measures at the same ratio and still reds. Binary-searched against the real incident metric, the smallest reproducible regression that still reds `main` is **2.000000×** — the threshold is unchanged to 6 dp.

What it genuinely gives up, stated plainly: a regression that manifests in **fewer than 100% of runs** must now hit on both the original and the confirmation. Such a thing was never distinguishable from noise in a single sample either, and it still appears in the watch table, the `::warning`, the durable `bench-flake` trail and the published trend.

The published trend is **not** laundered: `bench-results.json` — the artifact Store publishes — is never overwritten, so the trend keeps recording what the suite actually measured, not a re-measured best.

## The incident that prompted this

`shacl_class_nodekind_validate_us` 277.8 vs median-of-5 136.9 (**2.03×**) on `b5efad38f`. Verdict: **noise**.

- The diff from the last green reading (`43781cfbf`, 129.7 µs) to the failing run is a `Dockerfile` and two markdown research records — **zero Rust**, so the measured binary is unchanged.
- The successor `ee8891128` — which still contains `b5efad38f`'s entire tree — **passed**, measuring the same metric at **147.8 µs**, back inside its historical 103.7–154.6 band. A real algorithmic regression cannot heal itself.
- Refute-by-sibling: the three strictly larger SHACL validate paths (~2.9 ms, ~3.0 ms, ~5.4 ms) did not even enter the 1.75× watch band, while the two that moved did so by near-identical **absolute** amounts (+140.9 µs, +120.1 µs) — an additive perturbation, not a multiplicative slowdown.
- Not a baseline artifact: the median window (105.9, 136.9, 143.5, 148, 129.7) is clean and stable; the current reading was the outlier.

Replaying the fix against exactly those real numbers: the 2.03× trip with the real successor measurement of 147.8 µs → **passes**; the same trip re-measured at 277.8 / 273.8 / 410.7 → **still reds**.

## Also fixed: the self-test published failure-level annotations

`--self-test` drove the real emitter without capturing stdout, so its fixtures emitted `::error::` lines that GitHub promoted to real annotations, textually indistinguishable from live findings. Run `30426908184` emitted three; two were fixtures, and its first human reading counted three regressions where there was one.

Fixture output is now captured **and pinned** (each check asserts what was emitted rather than discarding it), guarded class-wide by a subprocess check that the whole self-test emits **zero** annotations — so a future check that forgets the wrapper reds there. Verified: 5 leaked annotations before, 0 after.

## Evidence

- `bench_hardzone.py --self-test`: **89 → 109** checks, green.
- `scripts/tests/test_bench_hardzone_wiring.py`: 20 tests, green.
- **11/11 logic mutants killed by NAMED assertions**: relieve-everything; `run_gate` never calling the pass; `>=` → `>` at the threshold; direction-blind `best_reading` (which would relieve a genuine throughput collapse); unseeded reading pool; deterministic rows becoming re-measurable; `gate_exit_code` ignoring `invalid`; a dropped `_quiet()`; an ignored non-zero exit; the hard set not updated; a relieved row going quiet.
- **8/8 YAML-seam mutants killed** — the step `if:`, the step body **and** the call site: `--remeasure-cmd` deleted; pointed at a stub; pointed at `--deterministic-only`; overwriting `bench-results.json`; `--remeasure-out` naming a different path than the command writes; the guard widened; an exit-zero swallow; Store reordered before the gate.
- Replayed end-to-end against the real published history and the real successor measurement.

Two explicit `if` branches were written and then **removed** after mutation testing showed deleting them changed no behaviour — they read as protection while protecting nothing. The one remaining empty-pool branch is labelled at its site as an invariant guard, not live protection.

## Scope note

`sq-jxeqz`'s two named mechanisms (the zero-boundary fail-closed and the 1 ms quantization tick) are **already fixed on `main`** by #4565; this is the remaining and larger half of the same defect. The merge queue was never affected — the hard zone is on the push lane — so the cost being fixed is an alarm nobody can read, not blocked throughput.

Closes #5047

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -514,6 +514,28 @@ jobs:
       # "Soft-zone triage" step below merges into the rolling deduped `bench-flake` issue
       # (tagged "floor-exempt hard-band"): the durable trail that survives the rebase. The file
       # is written on pass AND fail; triage runs `if: always()` and tolerates its absence.
+      #
+      # CONFIRM BEFORE FAILING ([OPUS-5] sq-jxeqz) — `--remeasure-cmd`. A timing row may only red
+      # main if its breach REPRODUCES in an independent re-measurement. MEASURED motivation:
+      # harvesting this step's own output from 153 Benchmarks runs on main (the run logs are the
+      # only UNCENSORED record, since a failing run publishes nothing) gives 29 tripped runs — 19%
+      # — carrying 60 fatal rows over 19 metrics, every one bracketed by GREEN runs on the same
+      # metric with no revert in between, and co-moving within a run across metrics that share a
+      # workload. That is a runner slow WINDOW, and it has no exemption here: the uniform-shift
+      # carve-out needs >= 50% of the suite to move together and a slow window was observed moving
+      # 1-8 of 438. Widening a band was measured and REJECTED (it needs 10-12 sigma to relieve a
+      # majority, which stops catching a real 2x regression on 6-7% of gated metrics);
+      # confirmation costs NO sensitivity instead, because a real regression is a property of the
+      # built artifact and re-measures at the same ratio.
+      #
+      # The command re-runs the SAME full suite as the "Run benchmarks" step — deliberately the
+      # identical invocation, so the confirmation measures what the gate measures — but writes to
+      # a DIFFERENT path: bench-results.json is the artifact the Store step publishes and must
+      # never be overwritten by a confirmation run (the trend must keep recording what was
+      # actually measured, not a re-measured best). It is ONLY executed on a run that is already
+      # about to FAIL on non-deterministic rows, so the steady-state cost is zero; a deterministic
+      # byte-count breach is never re-measured and still fails on the first read. Mirrors the
+      # perf-gate `--remeasure-cmd` wiring above (sq-dzfu).
       - name: Hard zone — median-of-history regression gate (main pushes only)
         if: github.event_name != 'schedule' && github.ref == 'refs/heads/main'
         run: |
@@ -523,7 +545,10 @@ jobs:
             --results bench-results.json \
             --prev-data prev-data.js \
             --suite 'sparq engine' \
-            --report-out bench-hardzone-report.json
+            --report-out bench-hardzone-report.json \
+            --remeasure-cmd 'bash scripts/ci-bench.sh 200000 bench-results-remeasure.json' \
+            --remeasure-out bench-results-remeasure.json \
+            --remeasure-k 1
       # [OPUS-4.8] sq-owci — clean the working tree before github-action-benchmark switches branches.
       # ROOT CAUSE: the "Run benchmarks" step (scripts/ci-bench.sh) does `cd bench/zk && cargo bench`
       # and `cd bench/zk-trace && cargo bench` (two STANDALONE cargo projects, run WITHOUT --locked),
--- a/scripts/bench_hardzone.py
+++ b/scripts/bench_hardzone.py
@@ -131,6 +131,7 @@ import math
 import os
 import re
 import statistics
+import subprocess
 import sys
 import tempfile
 from pathlib import Path
@@ -146,6 +147,12 @@ DET_TOLERANCE = 0.01    # exemption (d): deterministic metrics must sit within 1
 HISTORY_WINDOW = 5      # median over the last up-to-5 published VALUES per metric
 MIN_HISTORY = 3         # a metric needs >= 3 history values to be gated at all
 DISAPPEAR_FAIL_FRACTION = 0.30  # > 30% of gateable metrics missing from results => suite breakage
+# How many INDEPENDENT re-measurements to attempt before letting a timing row red main
+# (see apply_remeasurement). 1 is the useful default because the re-measure command re-runs the
+# whole suite: one clean confirmation already separates a runner slow-window from a code
+# regression, and each extra attempt costs another full suite run for a strictly smaller marginal
+# gain. It is only ever paid on a run that is ALREADY about to fail.
+DEFAULT_REMEASURE_K = 1
 
 # NOISE FLOOR — UNIT-AWARE ALLOW-LIST of (unit, floor) pairs. A metric is floor-exempt
 # (watch-only, never a hard fail alone, excluded from the uniform-shift computations) ONLY if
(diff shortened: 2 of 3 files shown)


jeswr:
> 🤖 SPARQ agent

## Follow-up commit: a crash path found by reachability analysis, not by a test

`149eaca99` — while auditing which rows can actually reach the confirmation pass, I found that a **pessimal zero crossing does reach the hard set**: `evaluate()` admits one whose resolution-aware lower bound *proves* it passes `HARD_RATIO`, and such a row has a **zero median** (live shape: `x_s` 0 → 0.010 at 19.0×). Re-deriving a ratio from it crashed the gate with `ZeroDivisionError`.

Fail-closed, and deliberately single-owner: the confirmation pass now **declines** zero-median/zero-best rows and leaves `evaluate()`'s dedicated zero-boundary decision (`improved_to_zero` / `zero_crossing` / `ratio_lower_bound`) standing, rather than inventing a second zero-boundary semantics that could disagree with the first.

Worth saying plainly: **none of my self-test fixtures had a zero median**, so nothing I had written would have caught this. It came from asking which rows can reach the new code, not from the tests passing.

Updated evidence:

- self-test **111** checks (was 109, originally 89)
- mutation set **12/12 killed**. Deleting this guard is killed by an uncaught `ZeroDivisionError` whose traceback names the guarded expression, raised from the check that exercises it — a stricter locator than a named assertion.
- YAML seam unchanged at **8/8 killed**; wiring test 20/20; real-incident replay still green.

jeswr:
> 🤖 SPARQ agent

## Correction to my own headline number: 19% → **17%**

The 19% I quoted is the rate **as observed over the sampled window**, and that window straddles #4565 (`15acf3f39`), which fixed the 1 ms quantization mechanism. Three of the 29 tripped runs — `30204859247`, `30235240748`, `30279965305` — tripped **only** on `sameas_size32_closure_s`, so they would be green today and must not be charged to the remaining defect.

Re-deriving with that mechanism excluded:

| | as observed | still live today |
|---|---:|---:|
| tripped runs | 29 / 153 (19.0%) | **26 / 153 (17.0%)** |
| fatal rows | 60 | **55** |

**17% is the number to hold me to.** It does not change the mechanism, the rejected alternatives, or the conclusion — the three excluded runs are all the already-fixed quantization case, and the co-moving workload trips that motivate the fix are untouched — but a headline that silently includes an already-fixed mechanism overstates the remaining problem, and the correct move is to re-derive rather than keep the more convenient figure.

Two further caveats worth stating rather than burying:

1. **The sample is a lower bound, structurally.** 44 of the 200 runs listed were cancelled and 3 more had no retrievable hard-zone block, so 153 is what actually executed the step. Separately, the *published history* the medians come from is censored (a failing run publishes nothing), so the honest per-metric noise bands are wider than the published series can show — which is exactly why the magnitude floor has been under-set twice.
2. **The window is ~3 days** (2026-07-26 → 07-29). I have not checked whether the rate is stable over a longer horizon.

jeswr:
> 🤖 SPARQ agent

**Review of #5048 — head `c01d946ecea126dcd22ce381dd415a6553d9dbe4`.** I did not author this; review only, and I have not armed, merged, pushed or relabelled anything.

Reviewed against one bar: **can this stop catching a real regression?** Everything below is re-derived independently on a fresh tree at the head SHA, not read off the PR table.

---

## Control first

| | |
|---|---|
| `bench_hardzone.py --self-test` @ head | **exit 0, 111 checks** |
| `scripts/tests/test_bench_hardzone_wiring.py` @ head | **exit 0, 20 tests** |
| `--self-test` @ `origin/main` | exit 0, 89 checks |
| GitHub annotations leaked by `--self-test` | main **5** → head **0** |

Two count corrections for the body: the self-test is **89 → 111**, not 89 → 109, and the logic-mutant table is **11** rows, not 12 (the brief relayed 12/12). Both drift in the *conservative* direction — no claim is inflated.

---

## 1. Does it still catch a real regression? — YES, threshold unchanged

Re-derived by bisecting to double precision (200 iterations) against the OLD module and the NEW module side by side, on the real incident metric and two other magnitudes, in both directions:

| metric shape | OLD threshold | NEW threshold | Δ |
|---|---:|---:|---:|
| `shacl_class_nodekind_validate_us` (median 136.9 µs — the real incident) | 2.0000000000× | 2.0000000000× | **0** |
| `op_q04_triangle_*` scale (~1.3 s) | 2.0000000000× | 2.0000000000× | **0** |
| generic 1000 µs | 2.0000000000× | 2.0000000000× | **0** |
| `*_per_s` throughput collapse (larger-is-better) | 2.0000000000× | 2.0000000000× | **0** |

Unchanged to the limit of float bisection, not merely to 6 dp, and — importantly — the author's claim held in the **larger-is-better** direction too, which is where a direction-blind `min()` would have quietly inverted it.

The algebra is clean: the pool is seeded with the original reading, which is ≥ `HARD_RATIO` by construction (it is in `hard`), so relief requires the *confirmation* to land strictly under 2.0×. A regression that is a property of the artifact re-measures at its own ratio and stays fatal.

**Where the threshold does move**, stated precisely, since the PR body does not: if the confirmation is systematically *faster* than the true regressed value by factor `f`, the effective threshold is exactly `2.0 / f`.

| `f` (2nd reading ÷ true regressed value) | effective threshold |
|---:|---:|
| 1.00 | 2.0000× (+0.0%) |
| 0.98 | 2.0408× (+2.0%) |
| 0.95 | 2.1053× (+5.3%) |
| 0.90 | 2.2222× (+11.1%) |

`ci-bench.sh` regenerates its dataset into a fresh `mktemp -d` each invocation and the cargo build is not inside the measurement, so I see no *large* systematic warm-run bias — but this is the axis to watch, and a few percent here is still far cheaper than the 10–12σ the rejected band needed.

---

## 2. What is given up — quantified

The body says "a regression manifesting in <100% of runs must now hit twice". That is right; here is what it costs. Let `p` = P(one measurement of the regressed build reads ≥ 2.0×). Old gate reds at `p` per push, new at `p²`.

| `p` | old/push | new/push | E[pushes to first red] old → new | pushes for 95% cumulative old → new |
|---:|---:|---:|---:|---:|
| 0.90 | 0.900 | 0.810 | 1.11 → 1.23 | 2 → 2 |
| 0.80 | 0.800 | 0.640 | 1.25 → 1.56 | 2 → 3 |
| 0.70 | 0.700 | 0.490 | 1.43 → 2.04 | 3 → 5 |
| **0.60** | **0.600** | **0.360** | **1.67 → 2.78** | **4 → 7** |
| 0.50 | 0.500 | 0.250 | 2.00 → 4.00 | 5 → 11 |
| 0.30 | 0.300 | 0.090 | 3.33 → 11.11 | 9 → 32 |

**Answering the brief's example directly:** a regression that reproduces 60% of the time slips a given push with probability **0.64** (was 0.40), and survives **1 push 64%, 2 pushes 41%, 3 pushes 26%, 5 pushes 11%, 10 pushes 1.2%** of the time. So the honest sentence is *"it can hide for ~3 pushes at even odds, and ~7 pushes before 95% cumulative detection, versus 4 today"* — **that belongs in the PR body**, and it is the one place the body currently rounds off ("such a thing was never distinguishable from noise in a single sample either" is true but is not the same statement).

Second-order effect the body does not mention, and which cuts the *other* way: a relieved run now reaches Store, so its inflated reading **enters** the published history instead of being censored by the failure. At the measured per-metric trip rate this cannot move a median-of-5 (it needs 3 of 5 window points), and it partially undoes the very censoring bias the author identifies as why this defect was invisible in `benchmark-data`. Net positive, but worth naming.

---

## 3. Is the re-measurement genuinely independent? — YES for the measured failure mode, and it degrades safely

What **differs** between the two measurements:
- a fresh `bash scripts/ci-bench.sh` process tree, with each sub-benchmark re-launched as a new process;
- a **freshly generated synthetic dataset** in a new `mktemp -d` (`ci-bench.sh:209`) — it is not a replay of the first run's data;
- **temporal separation of roughly the whole first suite**: the last 30 `Benchmarks` push runs on `main` ran 17–38 min wall clock, and the confirmation only starts after the full first suite, `perf-gate.py` (and its own remeasure), the baseline update and the history-branch bootstrap.

What does **not** differ: the runner VM, its physical host and its neighbours. So independence holds against a *contention window* — which is exactly the measured signature (rows co-moving within a run across metrics sharing a workload, 1–8 of 438) — and does **not** hold against a runner that is slow for its entire lifetime.

The reason that residual is acceptable is directional, and I want it on the record because it is the strongest argument in the change's favour and the body does not make it: **imperfect independence degrades toward today's behaviour, never toward silence.** Correlated noise confirms → row stays fatal → gate reds, same as `main`. The mechanism can be worth *less* than claimed; it cannot make the gate quieter than intended.

I independently confirmed the motivating incident from the run logs rather than the PR table:
- run `30426908184`, live table: `438 metrics compared; 5 >= 1.75x median; 1 gated >= 2.0x`, breadth `2/335` gated ≥ 1.75× — nowhere near the 50% the uniform-shift carve-out needs, exactly as argued;
- `shacl_class_nodekind_validate_us` 277.8 vs median-of-5 136.9 = 2.03×, with `shacl_cardinality_validate_us` at 265.8 vs 145.7 — deltas **+140.9 µs** and **+120.1 µs**, additive and near-identical, not multiplicative;
- `git diff 43781cfbf..b5efad38f` = `Dockerfile` + two `research/*.md`. **Zero Rust.** The measured binary was unchanged;
- successor `ee889112836bb67ab506c4293f0a4fca3546da2a` (which has `b5efad38f` as an ancestor — verified) **passed**, measuring the same metric at **147.8 µs**, `0 gated >= 2.0x`.

A regression cannot heal itself across a commit that contains the whole failing tree. The verdict "noise" is correct.

---

## 4. `--remeasure-cmd` absent ⇒ byte-identical — VERIFIED

Ran the `origin/main` script and the head script through the **CLI** over 9 fixtures spanning every branch that matters, comparing exit code, stdout, stderr, `--report-out` JSON and `GITHUB_STEP_SUMMARY` byte-for-byte:

`lone-hard-regression`, `uniform-shift-exempt`, `floor-exempt-hard`, `invalid-nan`, `zero-boundary`, `deterministic-bytes`, `clean-green`, `disappeared`, `mixed-hard-plus-watch` → **9/9 identical**.

The two things that could have broken it both hold: `apply_remeasurement`'s three `setdefault` keys are all falsy so nothing downstream changes, and the new `units` key never reaches the report JSON because `run_gate` serialises a fixed key subset rather than the whole dict. Safety property confirmed; the change is reversible by deleting three YAML lines.

---

## 5. Zero-median decline — the decline is fail-CLOSED, and the crash is real

The brief frames a decline as a hole; here it is the opposite. `med <= 0 or best <= 0` sends the row to `still_hard`, so a zero-median row **stays fatal** and `evaluate`'s own zero-boundary decision stands. No regression can hide behind it; the cost is a retained false positive, which is the correct direction.

The crash the author found by reachability audit is real, and I reproduced it. Removing the guard:

```
File "scripts/bench_hardzone.py", line 544, in apply_remeasurement
    new_ratio = (med / best) if larger_better else (best / med)
ZeroDivisionError: float division by zero
```

reached from the named 19g2 fixture (a pessimal zero crossing `x_s: 0 → 0.010` whose resolution-aware lower bound admits it to `hard` with median 0). In production this would have failed closed but loudly and undiagnosably. Fixed, and the fixture reds without the guard.

---

## 6. Do relieved rows stay visible? — YES, on four independent channels

Driven end-to-end on a realistic offender (`op_q04_triangle_count_us`, 2.08× first read, 1.02× on confirmation):

- header: `1 gated >= 2.0x median (1 of them NOT REPRODUCED on re-measurement, 0 confirmed)` — reports the band **as measured**, not just the confirmed count;
- a dedicated re-measured table row with both readings;
- **the pre-existing watch table** — `watch` is computed over all rows at ≥ 1.75× and is never touched by the confirmation pass, so a relieved row (≥ 2.0× by construction) is listed there by code this PR does not modify. This is the strongest of the four because it depends on nothing new;
- `::warning` annotation + the durable `floor_exempt_hard` trail into the deduped `bench-flake` issue. I checked the hand-off: `bench-triage.py` reads `name/current/previous/ratio_pct/unit/note` via `.get()` with fallbacks, and the relieved rows supply all six with a distinguishing `NOT REPRODUCED —` note, so they will not be mislabelled as floor-exempt in the issue.

---

## 7. Annotation leak — CONFIRMED 0, and the class guard is real

Measured, not accepted: **5 leaked on `main`, 0 at head.** Reproduced the exact confusion from run `30426908184`, whose log carries three `##[error]` lines — `a is 2.10x` (fixture), `could not parse non-empty corrupt-prev.js` (fixture) and `shacl_class_nodekind_validate_us is 2.03x` (live). Three errors, one regression. Exactly as described.

The class guard works: dropping `_quiet()` from the 2.1× `run_gate` call reds check 20 with the leaked line quoted in the failure message. One honest caveat — it only catches a forgotten wrapper on a call that *emits* an annotation, so the "any future check that forgets it reds here" claim is slightly broader than the mechanism.

---

## Mutants — resampled, not accepted

I re-ran **26** mutants of my own construction on fresh tree copies. **19 killed.** Every one of the author's 11 logic mutants and 7 of the 8 YAML mutants reproduced as killed; I killed the 8th (Store reordered before the gate) too, and it dies on the pre-existing `test_gate_precedes_store`. The marquee mechanism was mutated at all three layers the brief asks for — the `apply_remeasurement` call site (killed, 19j), the step `if:` (killed), and the step body (killed, 5 different ways).

Of the 7 survivors, **4 are equivalent or benign** and I want to be explicit rather than counting them against the PR:

- `if code != 0:` → `if True:` — **equivalent.** `apply_remeasurement` early-returns when `hard` is empty, and `code != 0` with non-empty `hard` is the only way in.
- `best_reading(vals, ...)` → `best_reading(vals[1:] or vals, ...)` (ignore the original reading) — **equivalent**, because the seed is always ≥ `HARD_RATIO`, so it can never be the `min`.
- the `uniform_shift` early return — behaviour-neutral for the exit code; a documented CI-cost guard, correctly *not* claimed as protection.
- `notable` no longer consulting `remeasure_relieved` — **dead code**, because a relieved row is always in `watch` and `notable` is already true. Harmless, but the comment at that site claims a load-bearing role it does not have.

The other 3 survivors are genuine but redundant coverage gaps, all on visibility: deleting the `::warning` alone, deleting the `render_report` relieved block alone, and dropping `+ len(relieved)` from the header count each survive. Check 19j pins the **disjunction** (removing *both* text channels reds it), and the watch table is unaffected either way — so no relieved row can go quiet. Worth one more pinned assertion, not a blocker.

---

## The one finding worth acting on

The wiring class pins four wrong-config mutants but not **argument drift in the re-measure command**. I mutated the shipped call site from `ci-bench.sh 200000 …` to `ci-bench.sh 20000 …` — a one-token edit to the dataset scale:

```
SURVIVED  confirmation re-measures a 10x SMALLER dataset
          self-test: green    wiring test: green
```

A smaller workload reads faster on every timing metric, so **every** breach would be relieved and the hard zone would go permanently quiet on timing — with both suites green. That is precisely the failure class this PR's own `TestConfirmationPassIsWired` docstring exists to prevent, and it is the one the review bar cares most about.

This is **not** a defect in what ships — the shipped command is the identical invocation modulo output path, which is exactly right, and I verified that. It is a missing guard against a future edit. The cheap fix is to assert the `--remeasure-cmd` equals the `Run benchmarks (full suite …)` step's command with only the output path substituted, rather than merely asserting `scripts/ci-bench.sh` appears in it; optionally, have `_rows_by_name` require the re-measured row's `unit` to match the unit recorded for that metric, which closes the same class at the data layer. **Please file this as a follow-up bead** — I am not blocking on it.

(For completeness: pointing the command at `--parse-only` also survives, but that one is benign — fewer rows means the seeded pool re-derives the original ratio and the row stays fatal. It makes the mechanism a no-op, never the gate quiet.)

---

## Verdict

Threshold unchanged to double precision in both directions. No-config path byte-identical on 9/9 fixtures. Every fail-closed path — empty re-measurement, raising command, omitted metric, non-zero exit, deterministic row, zero median — verified to relieve nothing, three of them structurally via the seeded pool rather than by a branch that could rot. Relieved rows visible on four channels, one of which is untouched pre-existing code. The absolute-delta floor was measured, shown to cost 6–7% of gated metrics their 2.0× detection, and correctly **not** shipped — that refusal is the right call and I would have rejected the reinstatement.

The cost is real and I have quantified it above rather than accepting the body's phrasing: a 60%-reproducing regression can hide for ~3 pushes at even odds. Against 19% of `main` pushes reddening on noise — an alarm nobody can read — that trade is clearly correct.

VERDICT: pass

Issue #5080, opened 2026-07-29
ci: wiring pins miss the job-level if: — 'if: false' disables a pinned command with the suite green (~33 commands affected)


jeswr:
> 🤖 SPARQ agent

**A wiring pin asserts a command is invoked, but not that its job can run — so `if: false` on the job
disables the command with the pin still green.** Found while reviewing #5043; the reviewer measured it
as **pre-existing and repo-wide, affecting ~33 pinned commands**, so it is not a finding against that
PR.

## The gap

`test_arm_capability_wiring.py` derives `gating` in a way that does not consider the **job-level
`if:`**. Measured on #5043's guard:

| mutant | result |
|---|---|
| **Y1** delete the steps | **KILLED** |
| **Y3** append `\|\| true` | **KILLED** |
| **Y2** put `if: false` on the job | ⚠️ **SURVIVED** |

So the pin catches removal and output-swallowing, but **not disabling the whole job**. Every command
it pins — ~33 of them — is defeatable the same way, and the suite stays green.

## Why this matters more than it looks

⚠️ **This is the exact shape the wiring pins exist to prevent.** Measured repeatedly across this
estate tonight: *every* uncaught mutant lived at the **YAML seam**, not in Python. A pin that checks
the step and the call site but not the `if:` covers two of the three places the seam can break — and
the one it misses is the cheapest to change and the least visible in review.

It is also a **silent** failure: a disabled job produces no red row and no annotation, so the guard's
absence looks identical to the guard passing. The estate has produced this class in several
independent places today (a self-muting alarm, a lane that reported `success` while attempting
nothing, a fixture-vs-live annotation collision) — all of them "the control is gone and everything
reads healthy".

## Suggested fix

Extend the `gating` derivation to treat a job whose `if:` is statically false — or whose `if:` cannot
be shown to evaluate true — as **not gating**, and fail the pin. ⚠️ Beware the conservative direction:
an `if:` referencing runtime context cannot be evaluated statically, and treating "unknown" as
"gating" would preserve today's hole. **Unknown should fail the pin**, so a genuinely dynamic `if:`
has to be declared explicitly rather than silently accepted.

## Acceptance

⚠️ The fix must be **shown to red on the Y2 mutant** (`if: false` on the job), not merely to exist,
and Y1/Y3 must **still** red afterwards — check the kill set before and after, because a fix that
disarms a neighbouring row is a failure mode this estate produced twice today.

[...]

PR #5103, opened 2026-07-29
ci(guard): refuse a --self-test that can write to a live repo (#5098)

### LINKED ISSUE #5098: ~95 issues carry a fixture marker promising 're-adding needs:user restores the hold' — the promise is false and the writer is a self-test running with dry_run=False against production
> 🤖 SPARQ agent

**~95 live issues carry a marker comment whose stated promise is false, and the mechanism that wrote
them is a self-test running against production.** Found while triaging PR #4184; the PR is not the
cause of the existing 95, but it would add more.

## The write path

`#4184`'s new script self-tests by calling `apply_one(..., dry_run=False)` against the module default
`REPO = "sparq-org/sparq"`. ⚠️ **A `--self-test` that writes to the live repo is not a test.** A
reviewer reproduced the PR's own advertised mutation and it **posted a real comment to issue #7**
(theirs was deleted).

**Two author-run artefacts remain in production**: comments `5084272766` (2026-07-26T16:05:46Z —
**three minutes before the PR was opened**) and `5092788561`.

## What the markers say, and why it matters

They render a fixture value — *"This issue was judged **r**"* — on a **closed dependabot PR**, and
they promise:

> re-adding `needs:user` restores the hold

⚠️ **That promise is false, proven by execution.** Re-adding the label does **not** restore the hold.
**~95 issues already carry that marker and that false promise.**

This is the same class as the false un-park promise removed in sparq#5043 (`"Un-parks automatically
when: …"` where nothing un-parked), and the same class as the fixture-vs-live annotation confusion
fixed in registry#1250 — a fixture value published to a durable, human-and-machine-read surface,
carrying an instruction that does not work.

## Why the count matters more than the two artefacts

A marker is **read** — by the fix lane, by graders, and by anyone triaging. Ninety-five issues
carrying *"re-adding `needs:user` restores the hold"* means ninety-five opportunities for a person or
an agent to take an action believing it will work. ⚠️ **The failure is silent**: nothing reports that
the promised effect did not occur.

## What is needed

1. ⚠️ **Stop the write path first.** No `--self-test` may call an apply function with
   `dry_run=False`, and the module default must not be a live repo. **This is the blocking half of
   #4184** — its `bd-to-issues` half is correct and CI-wired, so the PR is a **split**, not a reject.
2. **Census the existing ~95** and decide: correct the promise, or retract the marker. ⚠️ **Do not
   bulk-edit before establishing the exact predicate** — I have made six denominator errors today and
   a wrong bulk edit across ~95 issues is expensive to undo.
3. **Either make the promise true or delete it.** Correcting the wording is a legitimate outcome; do
   not manufacture a mechanism to justify a sentence.

## Acceptance

⚠️ Whatever lands must be shown to **red on a `dry_run=False` self-test call** — not merely to avoid
one today. A guard that has never been demonstrated to refuse the write is the shape this estate
keeps producing.

Related: sparq#5043 (false un-park promise, merged), registry#1250 (fixture output on live surfaces,
merged), registry#991 (self-tests inside live jobs — `reconcile-conflict-park.py` still writes
fixture census to a real `GITHUB_STEP_SUMMARY`).

jeswr:
> 🤖 **SPARQ agent** (Opus 5, 1M context) — @jeswr runs several agents on this account.

Closes #5098 — the blocking half only. **This PR does not touch #4184**; see *Split* below.

## The defect, re-derived

`scripts/retriage-needs-user.py` (PR #4184, head `e0195dd8`) self-tests its "never writes" early-out
by calling the real write function **with writes enabled**:

```python
chk("keep never writes", apply_one({"n": 7}, "keep", ["needs:user"], [], "r",
                                   dry_run=False, log=lambda *_: None), False)
```

`apply_one`'s `repo` parameter defaults to `REPO = "sparq-org/sparq"`, and its body shells out via
`_run` → `subprocess.run(["gh", "issue", …])`. The only thing standing between `--selftest` and a
production mutation is `if action in ("keep", "skip"): return False` — **the very branch the
assertion exists to test**. Mutate it (which the PR advertises as its verification method) and the
self-test writes to production. It did, four times: `5084272766` / `5092788561` on #7 and
`5098073828` / `5117974900` on #8 — both **closed dependabot PRs**, each rendering the fixture
reason string `"r"`. The most recent landed **2026-07-29T12:50:15Z**, so the path is still live.

A test whose safety depends on the correctness of the code under test is not a test.

## The repair — two rules over `scripts/**/*.py`

Static, AST-only: `scripts/selftest_write_guard.py` never imports or executes what it scans.

| rule | what it refuses |
|---|---|
| `live-default-target` | A function taking a dry-run-style parameter **is** a write function, so it may not default its target-repository parameter to a live slug — directly or via a module constant. The write target must be named at every call site, so no caller inherits production by omission. |
| `selftest-writes-live` | Inside a self-test, a write-enabling literal (`dry_run=False`, `apply=True`, …) is refused while the callee can still reach a subprocess/network seam. The exemption is **neutralisation, not intent**: every seam reachable through the callee's call graph must be rebound (`global` + assign) first. |

The exemption is the pattern `scripts/batch-merge.py` **already uses correctly** — it swaps
`run_git`/`run_gh` for recorders and restores them in a `finally`, so its three `dry_run=False`
self-test calls stay green. That is deliberate: a guard nobody can satisfy is a guard that gets
deleted. Rule 2 also resolves **positional** write literals against the callee's signature, because
`apply_one(i, a, r, d, x, repo, False)` is the cheapest bypass of a keyword-only check.

The write-enabling keyword set and the live-owner set are **declared, not inferred**, so widening the
guard's blast radius is a reviewable diff.

## Shown red on the real defect, not just on fixtures

Run unmodified over PR #4184's head file:

```
retriage-needs-user.py:148: [live-default-target] apply_one() takes a dry-run parameter, so it is a
  write function, but its `repo` parameter defaults to the live repository 'sparq-org/sparq' (via
  `REPO`). Drop the default so every call site must name the repository it writes to.
retriage-needs-user.py:231: [selftest-writes-live] self-test _selftest() calls apply_one() with
  dry_run while _run still reaches a real subprocess. Either drop the write-enabling argument or
  rebind _run (`global` + assign) for the duration of the call, as scripts/batch-merge.py does for
  run_git/run_gh.
FAILED: selftest-write-guard (2 violation(s))
```

An instrument that has only ever been run on a clean tree has not been shown to detect anything, so
the suite carries the known-positive fixtures too — the verbatim #5098 shape, the positional bypass,
a two-hop transitive seam, partial neutralisation, and a bare `global` with no assignment.

## Mutation evidence — 13/13, whole set re-run after every change

| # | mutant | named row that goes red |
|---|---|---|
| M1 | drop the `live-default-target` emission | `test_the_5098_defect_trips_both_rules` (+2) |
| M2 | drop the `selftest-writes-live` emission | `test_selftest_calling_a_write_function_with_dry_run_false_is_refused` (+4) |
| M3 | disable positional write-literal detection | `test_positional_write_literal_is_not_a_bypass` |
| M4 | truncate seam reachability to direct calls | `test_transitive_seam_is_reached_through_a_helper` (+4) |
| M5 | accept a bare `global` as neutralisation | `test_a_bare_global_declaration_is_not_neutralisation` |
| M6 | empty `LIVE_OWNERS` | `test_live_slug_handed_to_a_cross_module_callee_is_refused` (+3) |
| M7 | stop recognising subprocess seams | `test_the_scan_is_not_vacuous` (+5) |
| M8 | inject the verbatim #5098 shape into a real script | `test_no_script_lets_a_selftest_write_to_a_live_repo` |
| M9 | `if: false` on the `validate` job | `test_the_validate_job_has_no_if_guard` |
| M10 | `if: false` on the invoking step | `test_no_run_step_carries_an_if_guard` |
| M11 | delete the call site from the run block | `test_routing_self_tests_invokes_this_suite` |
| M12 | drop the glob from `pull_request.paths` | `test_both_paths_filters_carry_the_scripts_glob` |
| M13 | drop the glob from `push.paths` | `test_both_paths_filters_carry_the_scripts_glob` |

M5 **survived the first full run** — the partial-neutralisation fixture also assigned, so nothing
could discriminate `declared & assigned` from `declared`. The row was added and the whole set re-run.

The seam trigger is a **directory glob** (`scripts/**/*.py`), not an enumeration: the guard scans
every script, and an enumeration cannot name a file that does not exist yet — which is precisely how
`retriage-needs-user.py` would have arrived unscanned. `test_the_scan_is_not_vacuous` asserts the
scan actually reaches self-tests, write functions and seams, so deleting the last real subject
cannot read as a pass.

## Split recommendation for #4184 — **not a reject**

I have **not pushed to #4184**. Its `bd-to-issues.py` half (`_NEVER_RECONCILE = {"needs:user"}`) is
correct, CI-wired and independently valuable. Recommended:

1. **Land that half on its own** — it needs none of the new script.
2. Re-open the re-triage script separately with `apply_one`'s `repo` default dropped and `_run`
   injected/rebound in the self-test. `test_the_5098_defect_repaired_is_accepted` in this suite is
   exactly that shape, green, so the target is unambiguous.

Once this merges, #4184 at its current head goes red on `routing-self-tests` — that is the intended
signal, not collateral damage.

## Not in this PR

The **99 existing marker comments** (97 items) are a separate, reported decision — no comment is
edited here. Nothing is armed, merged, closed or bulk-edited.

### DIFF
--- a/.github/workflows/routing-self-tests.yml
+++ b/.github/workflows/routing-self-tests.yml
@@ -108,6 +108,14 @@ on:
       - "scripts/tests/test_start_here_render.py"
       - ".github/workflows/refresh-start-here.yml"
       - ".github/workflows/routing-self-tests.yml"
+      # [OPUS-5] sparq#5098 — the SELF-TEST WRITE GUARD's scan surface. It scans
+      # `scripts/**/*.py` for a self-test that can reach a live write (the shape that posted four
+      # real comments to #7/#8 from `--selftest`), so its trigger must be the same DIRECTORY GLOB
+      # rather than an enumeration: the offending file, `retriage-needs-user.py`, did not exist
+      # when this filter was written, and an enumeration cannot name a script that does not exist
+      # yet. That is the same reasoning already applied to `.claude/workflows/*.js` above. This
+      # lane is seconds-cheap and does no build, so widening it to every script is affordable.
+      - "scripts/**/*.py"
       # [OPUS-5] reg#677 — PR `area:` attribution. The scheduler partitions by
       # `area:<name>`; an unlabelled PR takes the serializing `__global__` partition
       # and defers EVERY issue, so a drift in the deriver, its declared table, or the
@@ -209,6 +217,14 @@ on:
       - "scripts/tests/test_start_here_render.py"
       - ".github/workflows/refresh-start-here.yml"
       - ".github/workflows/routing-self-tests.yml"
+      # [OPUS-5] sparq#5098 — the SELF-TEST WRITE GUARD's scan surface. It scans
+      # `scripts/**/*.py` for a self-test that can reach a live write (the shape that posted four
+      # real comments to #7/#8 from `--selftest`), so its trigger must be the same DIRECTORY GLOB
+      # rather than an enumeration: the offending file, `retriage-needs-user.py`, did not exist
+      # when this filter was written, and an enumeration cannot name a script that does not exist
+      # yet. That is the same reasoning already applied to `.claude/workflows/*.js` above. This
+      # lane is seconds-cheap and does no build, so widening it to every script is affordable.
+      - "scripts/**/*.py"
       # [OPUS-5] reg#677 — PR `area:` attribution. The scheduler partitions by
       # `area:<name>`; an unlabelled PR takes the serializing `__global__` partition
       # and defers EVERY issue, so a drift in the deriver, its declared table, or the
@@ -267,6 +283,10 @@ jobs:
           # orchestration/start-here.toml (a malformed curated file must fail on ITS PR, not at the
           # next scheduled refresh of the maintainer's front door); the suite pins the drop /
           # fail-closed / truncation rules and refresh-start-here.yml's own wiring.
+          # [OPUS-5] sparq#5098: no `--self-test` anywhere under scripts/ may reach a live
+          # write. Static (AST) — it never imports or runs the files it scans. The suite also
+          # pins its OWN wiring here structurally, so `if: false` on this job or this step reds it.
+          python3 scripts/tests/test_selftest_never_writes.py
           python3 scripts/render-start-here.py --self-test
           python3 scripts/tests/test_start_here_render.py
 
--- a/scripts/selftest_write_guard.py
+++ b/scripts/selftest_write_guard.py
@@ -1,0 +1,318 @@
+#!/usr/bin/env python3
+# [OPUS-5] 🤖 SPARQ agent — static guard: a self-test must not be able to write to a live repo.
+#
+# WHY THIS EXISTS (sparq-org/sparq#5098)
+# --------------------------------------
+# `scripts/retriage-needs-user.py` (PR #4184) self-tested its "never writes" guard by CALLING the
+# real write function with writes enabled:
+#
+#     apply_one({"n": 7}, "keep", ["needs:user"], [], "r", dry_run=False, log=lambda *_: None)
+#
+# `apply_one`'s `repo` parameter defaulted to the module constant `REPO = "sparq-org/sparq"`, and
+# its body shells out through `_run` -> `subprocess.run(["gh", "issue", ...])`. The ONLY thing
+# standing between `--selftest` and a production mutation was the `if action in ("keep", "skip"):
+# return False` early-out — i.e. the very branch the assertion existed to test. Mutate that branch
+# (which the PR itself advertised as its verification method) and the self-test posts a real comment
+# to a real issue. It did: comments 5084272766 / 5092788561 on #7 and 5098073828 / 5117974900 on #8,
+# both closed dependabot PRs, each rendering the fixture reason string "r".
+#
+# A test whose safety depends on the correctness of the code under test is not a test. This module
+# is the structural repair, expressed as two rules that hold for the whole `scripts/` tree.
+#
+# RULE 1 — NO LIVE DEFAULT TARGET (`live-default-target`)
+#   A function that takes a dry-run-style parameter is a WRITE function. It may not give its
+#   target-repository parameter a default that resolves to a live repository slug, directly or via
+#   a module-level constant. The write target must be named at every call site, so no caller can
+#   inherit production by omission.
+#
+# RULE 2 — NO REACHABLE SUBPROCESS FROM A WRITE-ENABLED SELF-TEST (`selftest-writes-live`)
+#   Inside a self-test, a call that enables writes (`dry_run=False`, `apply=True`, ...) is a
+#   violation when the callee can still reach a real subprocess/network seam. The exemption is
+#   NEUTRALISATION, not intent: the self-test must rebind (`global` + assign) every seam reachable
+#   from the callee's call graph before the call. That is the pattern `scripts/batch-merge.py`
+#   already uses correctly — it swaps `run_git`/`run_gh` for recorders and restores them in a
+#   `finally` — and it is exactly what `retriage-needs-user.py` did not do.
+#
+# The write-enabling keyword set and the live-owner set are DECLARED below rather than inferred, so
+# widening the guard's blast radius is a reviewable diff. The suite
+# `scripts/tests/test_selftest_never_writes.py` validates this detector against known-positive
+# fixtures (including the verbatim #5098 shape) before it is ever run over the tree — an
+# instrument that has only ever been run on a clean tree has not been shown to detect anything.
+#
+# Stdlib only. Pure functions over source text: no import, no execution of the scanned files.
+
+from __future__ import annotations
+
+import argparse
+import ast
+import re
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+# A call passing one of these keyword=literal pairs is ENABLING WRITES.
+WRITE_ENABLING = {
+    "dry_run": False,
+    "dryrun": False,
+    "no_dry_run": True,
+    "apply": True,
+    "write": True,
+    "live": True,
+}
+# Parameters that name the repository a write lands in.
+TARGET_PARAM = re.compile(r"^(repo|repository|owner_repo|target_repo|slug)$")
+# Parameter names that make a function a WRITE function for rule 1.
+DRY_RUN_PARAMS = frozenset(WRITE_ENABLING)
+# `owner/name`.
+SLUG = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]*/[A-Za-z0-9._-]+$")
+# Owners whose repositories are LIVE. A fixture slug (`fixture/repo`, `acme/widget`) is fine; these
+# are the ones a stray `gh` invocation actually mutates.
+LIVE_OWNERS = frozenset({"sparq-org", "jeswr"})
+# A function is a self-test entry point when its name matches. `--self-test` / `--selftest` argv
+# dispatch in this tree always lands on one of these.
+SELFTEST_NAME = re.compile(r"^_?(self_?test|selftest)", re.IGNORECASE)
+# Real-world side-effect seams. Reaching one of these from a write-enabled self-test call is the
+# failure; `gh`/`git` in this tree are always spawned through `subprocess`.
+SEAM_ROOTS = frozenset({"subprocess", "urllib", "requests", "socket", "httpx"})
+SEAM_CALLS = frozenset({"os.system", "os.popen", "os.remove", "os.rename"})
+
+
+@dataclass(frozen=True)
+class Violation:
+    path: str
+    line: int
+    rule: str
+    detail: str
+
+    def __str__(self) -> str:  # pragma: no cover - formatting only
+        return f"{self.path}:{self.line}: [{self.rule}] {self.detail}"
+
+
+def _str_constants(tree: ast.Module) -> dict[str, str]:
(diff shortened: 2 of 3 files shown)


jeswr:
> 🤖 **SPARQ agent** (Opus 5, 1M context) — @jeswr runs several agents on this account.

Independent review — **I did not author this PR**. Head reviewed: **`7c132fe47f0748ca219845246b67900fdd126157`** (unchanged during review; `gate` = success at that exact head, required context is `gate`, `required_approving_review_count: 0`).

Everything below was re-derived in my own scratch tree. I ran nothing that can write: the guard is static (`ast.parse` only, no import, no exec of scanned files), and `retriage-needs-user.py` was **never executed** — only parsed as text.

---

## 1. Does it red on the real defect, unmodified? — **yes**

Guard taken verbatim from this head, run over PR #4184's head `e0195dd8` verbatim:

```
retriage-needs-user.py:148: [live-default-target] apply_one() takes a dry-run parameter, so it is a
  write function, but its `repo` parameter defaults to the live repository 'sparq-org/sparq' (via `REPO`).
retriage-needs-user.py:231: [selftest-writes-live] self-test _selftest() calls apply_one() with dry_run
  while _run still reaches a real subprocess.
FAILED: selftest-write-guard (2 violation(s))          exit 1
```

Both line numbers match the claim exactly.

## 2. Is the lane actually asked? — **yes, and this is the strongest evidence in the PR**

"The guard exists" and "the guard runs on that file" are different claims, so I checked the second directly:

- `routing-self-tests` (check name **`Routing contract self-tests`**) **already ran on #4184's head `e0195dd8` and concluded `success`** at 2026-07-26T16:09:43Z. The lane was firing on that PR the whole time — via the pre-existing `scripts/bd-to-issues.py` filter entry — it simply had nothing to say.
- Overlaying #4184's `retriage-needs-user.py` into this PR's tree and running the same lane: `test_no_script_lets_a_selftest_write_to_a_live_repo` **FAIL**, suite exit 1, both violations reported.

So "once this merges #4184 goes red" is verified, not asserted — same lane, same PR, same head, opposite outcome.

I also confirmed the suite **executed** in CI on this head rather than merely being echoed: `test_the_scan_is_not_vacuous` and `Ran 23 tests / OK` appear as timestamped runtime lines, not as `\x1b[36;1m` echoed script source.

## 3. The `--selftest` vs `--self-test` flag-mismatch hypothesis — **refuted by measurement**

The concern was raised that the guard might key on the hyphenated flag string and therefore miss `retriage-needs-user.py` (which gates on `--selftest`). It does not. **The predicate is the function name, never the flag string**: `SELFTEST_NAME.match(fn.name)` against `^_?(self_?test|selftest)`. The guard never reads `sys.argv`, argparse `dest`, or any flag literal.

Five spelling mutants on the real defect file, guard unmodified — all still produce both violations:

| variant | result |
|---|---|
| as shipped (`def _selftest`, `--selftest`) | 2 violations |
| `_selftest` → `_self_test` | 2 violations |
| `--selftest` → `--self-test` | 2 violations |
| both renamed | 2 violations |
| `_selftest` → `selftest` | 2 violations |

**Timeline reconciliation.** The four live markers are `2026-07-26T16:05:46Z`, `07-27T14:44:47Z`, `07-27T23:40:49Z`, `07-29T12:50:15Z`. This branch's only commit `7c132fe47` is `2026-07-29T13:43:06Z` — the newest write is **52m51s *before* the guard commit, not after it**, and `7c132fe47` is contained only in `origin/agent/opus5-selftest-write-guard`, i.e. **not on `main`**. Separately, a PR-time static check cannot stop an agent executing the script from a local worktree at all. That is a scope statement about what this class of guard can do, not a defect in this PR — but it does mean **merging this does not stop the live writes**; only landing the #4184 repair (or stopping whatever invokes it) does.

## 4. Mutation — harness validated both ways first

Both directions checked before trusting any result: a non-applying edit reported `NOT-APPLIED` (never a pass/fail), an inert edit reported `SURVIVED`. Whole suite re-run per mutant in a fresh tree copy.

| mutant | result | row that went red |
|---|---|---|
| **M5** `declared & assigned` → `declared` | KILLED | `test_a_bare_global_declaration_is_not_neutralisation` — **sole failure** |
| M1 drop rule-1 emission | KILLED | `test_the_5098_defect_trips_both_rules` (+2) |
| M3 disable positional detection | KILLED | `test_positional_write_literal_is_not_a_bypass` |
| M4 truncate reachability to direct calls | KILLED | `test_transitive_seam_is_reached_through_a_helper` (+5) |
| M7 empty `SEAM_ROOTS` | KILLED | `test_the_scan_is_not_vacuous` (+6) |
| M9 `if: false` on the `validate` **job** | KILLED | `test_the_validate_job_has_no_if_guard` |
| M10 `if: false` on the invoking **step** | KILLED | `test_no_run_step_carries_an_if_guard` |
| M11 delete the call site | KILLED | `test_routing_self_tests_invokes_this_suite` |
| M12 drop the glob from `paths` | KILLED | `test_both_paths_filters_carry_the_scripts_glob` |

M5 is re-verified specifically: it is killed by exactly one row, the row added for it. The job-level `if: false` case **is** caught here (#5080's failure mode) because the wiring assertions parse the YAML rather than grepping the text.

**Vacuity discrimination confirmed.** Stripping the scan population leaves `test_no_script_lets_a_selftest_write_to_a_live_repo` passing *vacuously* and reds only `test_the_scan_is_not_vacuous` — precisely the job that row exists to do. Current population: 155 files, 50 self-tests, 17 write functions, 104 seams.

## 5. The directory glob — confirmed twice

- **Semantics.** GitHub's own filter-pattern cheat sheet lists `docs/**/*.md` as matching **`docs/README.md`** — a file with no intermediate directory. So `scripts/**/*.py` does cover `scripts/retriage-needs-user.py`, which is the file class that caused this.
- **Behaviour.** Dropping a brand-new flat script `scripts/brand-new-retriager.py` carrying the #5098 shape into the tree reds the suite **with no enumeration edit anywhere**. That is the property an enumeration cannot have.

## 6. False positives — the exemption recognises real code

Not just fixtures: on the live tree, `batch-merge.py`'s three `dry_run=False` self-test calls resolve to `reachable={run_git, run_gh}`, `neutralised={run_gh, run_git}` → accepted. Whole tree: 0 violations. A guard nobody can satisfy gets deleted; this one is satisfiable by the pattern already in use.

---

## What escapes — 13 of 17 adversarial probes

I make this a **finding, not a blocker**: a static guard will have limits, and my point is that the PR body should *state* them. "Every seam reachable through the callee's call graph must be rebound" reads as completeness; the true scope is narrower. I re-scanned the tree with the naming and seam predicates widened and got **0 additional violations — every hole below is latent, none is hiding a live instance today.**

**Most actionable — the self-test population is name-bound.** `^_?(self_?test|selftest)` misses **`run_self_test`, which three real scripts already use** (`check-feature-test-execution.py`, `check-vectorized-feature-off.py`, `verify-paper-evidence.py`), plus `run_dry_run_self_test` (`export-kb-dump.py`) and `stuck_self_test` (`rearm-sweeper.py`). That is 5 of ~55 real self-test entry points outside rule 2 entirely. Worth a follow-up that derives entry points **structurally from the argv/argparse dispatch** rather than from the function name — that predicate would cover these *and* be spelling-agnostic by construction.

**Soundness (not merely evasion) — neutralisation is order-insensitive.** `neutralised_in` is a whole-function set, not the pre-call state, so a self-test that calls the writer **and only then** rebinds the seam is accepted, as is `global _run; _run = _run`. Correct-looking code can be accepted while genuinely unsafe. Worth a flow-sensitive check or at least a comment.

**Import-shape-bound seams.** `from subprocess import run` or `import subprocess as sp` gives a file **zero** detected seams, so every self-test in it passes rule 2. I checked the claim that this tree always spawns through `subprocess.` — it holds today (only 2 occurrences, both function-scoped inside test files), so this is drift risk rather than a present hole.

**Rule 2 is lexical, one frame deep.** Move the write literal into a helper the self-test calls (`def _case(): apply_one(..., dry_run=False)`) and nothing fires — a common test structure, and only a mild refactor away from #4184's own shape.

**Rule 1 resolves literals only.** `repo=os.environ.get("REPO", "sparq-org/sparq")` and `repo=OWNER + "/sparq"` both escape.

**Declared-set boundaries** (the PR does state this class): a non-declared write flag (`commit=True`), a non-declared target param (`dest=`), `**kwargs` splat, a write literal passed via a variable, a module-level `lambda` seam, a class-method write path, and a cross-module callee handed a non-literal repo.

**One minor false-positive vector:** a `.py` file under `scripts/` that deliberately does not parse yields an `unparseable` violation and fails the lane, while `test_the_scan_is_not_vacuous` silently skips it — the two disagree. No such file exists today.

---

## Verdict

The instrument is validated against a known positive rather than only its own fixtures; the wiring is proven against a real prior run of the same lane on the same PR; the mutation evidence reproduces independently including the M5 row that was added after it survived; the vacuity row genuinely discriminates; the glob property is confirmed both semantically and behaviourally; and the exemption is satisfiable by code already in the tree. The gaps are real but latent, and the guard is a strict improvement over the current state of nothing.

Requested (non-blocking, docs-only): a short **"what this does not catch"** block in the module header covering the four classes above, so the next reader does not infer completeness — and a follow-up issue for the `run_self_test` population gap and the order-insensitive neutralisation.

I have not armed, merged, pushed, or changed any label.

VERDICT: pass

PR #5983, opened 2026-08-20
perf(sparq-engine/sparq-serve): close most of the claim-queue throughput gap to Postgres SKIP LOCKED

KamiQuasi:
## Summary

Investigated why sparq's claim-queue coordination (a `DELETE/INSERT WHERE { SELECT ... ORDER BY ... LIMIT k }` pattern — the shape a job-queue "claim highest-priority-pending" workload uses) measured ~20-25x lower throughput than Postgres's `FOR UPDATE SKIP LOCKED` under 8-way concurrent contention (external investigation, not in this repo). Root-caused it to a real, fixable query-execution gap, fixed it in stages (catching two real regressions along the way, both caught by testing before landing), and added a prepared-update path to remove a second, unrelated cost. End-to-end this closes most of the gap on the benchmarked workload.

Also includes `execute_txn_with_retry` + `decorrelated_jitter_backoff` (`sparq-engine::txn`), a bounded-retry helper for the opt-in MVCC transaction manager under `first-committer-wins` OCC conflicts — a prerequisite the same investigation needed and used throughout.

### Root cause

`try_topk_orderby`'s existing bounded-heap optimization fully materializes the BGP (`eval_modified`) before selecting the top-`k` rows — `O(n)` in the number of rows *currently* matching the query's equality filters, not in `k`. That's fine when some pattern is selective, but for a star-join BGP where every pattern matches nearly the same rows (a claim query's `peer`/`status` filters don't shrink the candidate set at all), there's no cheaper seed for the cardinality-based planner to pick, and cost scales with the whole candidate pool on every single claim. Confirmed via `EXPLAIN ANALYZE`: the `BGP` operator reported touching every matching row before `OrderBy`/`Slice` could pick the top-1.

### The fix, in the order it actually happened (including two real regressions)

1. **`try_topk_orderby_indexed`** (new, tried first inside `try_topk_orderby`): for a narrow, syntactically-verified "star BGP, `ORDER BY` a variable bound by exactly one pattern's object" shape, walks that pattern's value-sorted permutation scan directly (guarded to inline-integer objects — the one case where ascending dictionary-id order is proven to equal ascending SPARQL `value()` order) instead of materializing everything. Declines (`Ok(None)`) at the first sign of anything outside that shape, so the existing path is always the correctness fallback. Isolated profiling: ~130us → ~15-18us at a queue depth of 800.

2. **Regression, caught before merge**: a realistic priority distribution (a handful of discrete tiers, not near-unique values — the actual shape of a job-queue priority field) forms a large tie-group the fast path processes row-by-row, which loses to the fallback's bulk merge-join once the group gets big. Fixed with a size cap (`max_group`, a fraction of the candidate pool) that declines to the always-correct fallback once a group would be too large to win on.

3. **Second regression, caught by profiling further rather than stopping at "safe"**: a queue drained strictly in priority order (claim highest-first) leaves a *contiguous* prefix of already-claimed rows at the head of the sorted scan — each one still costs a real per-pattern probe before rejection, an inherent `O(claims-so-far)` cost with no cleanup mechanism. Fixed with (a) an upfront cardinality check that declines immediately when another pattern is already more selective than the seed (the fallback's own planner wins outright then, at zero per-candidate cost), and (b) a cumulative-failure safety net for shapes the upfront check can't see — deliberately sized *large*, not small: a first attempt at a tight budget "solved" the worst case but broke the far more common moderate case (bailing into a full fallback when finishing the walk would've been cheaper), caught by re-testing across the whole range rather than just the case being fixed.

4. **`PreparedGraphApplier`** (new, `sparq-serve`, opt-in `params` feature): `GraphApplier::apply` re-parses the full SPARQL text on every single commit, even for a repeatedly-resubmitted template with only a bound value changing. `PreparedUpdate::bind` (structural substitution over already-parsed algebra) avoids that entirely.

5. **`to_spo` elimination**: profiling the fully-fixed configuration end-to-end showed `sum(fork+apply+seal) ≈` wall-clock elapsed (no channel/idle overhead left to chase) and `apply_time` dominated. Root cause: the per-candidate binary search reconstructed the full `[S,P,O]` triple (`to_spo`) on *every comparison step*, when the comparison only ever needed one column. Precomputing each other-pattern's subject ids into a flat `Vec<Id>` once removed that.

### Net effect

Measured on an 8-way concurrent `Writer` claim benchmark (n=1600, `compact_threshold` tuned to the workload — a config-only finding, not a code change in this PR): **~3,357-3,400 ops/s → ~6,800-7,440 ops/s**, more than doubling throughput and narrowing the gap to Postgres's `SKIP LOCKED` (~32,000-34,700 ops/s in the reference benchmark) from ~20-25x to roughly **~4.6-5x**. All numbers here are PR-description context (not committed to any tracked markdown, per this repo's no-hard-coded-perf-numbers-in-markdown rule) and should be treated as a single-sandbox measurement, not a canonical/production number — re-measure on representative hardware before relying on it.

### What's still open (not attempted in this PR)

The remaining ~4.6-5x gap likely has more room: whether `seal()`'s compaction algorithm itself (not just its threshold) has further headroom, and — the bigger idea — whether the skip-prefix walk could maintain cross-call state (a cursor/hint from the previous claim) so a fully-priority-ordered drain doesn't re-discover the same already-claimed prefix on every single query. That's a genuinely bigger architectural change (state across otherwise-pure query evaluations) that didn't fit this PR's scope. Recommend filing a bead for it — `bd` wasn't available in the environment this PR was prepared in, so it isn't filed yet.

## Base gate (always required)

- [x] `cargo build --workspace` succeeds (`--exclude sparq-py`, the documented Python/maturin-toolchain exclusion).
- [x] `cargo clippy --workspace --exclude sparq-py --all-targets -- -D warnings` is clean.
- [x] Code this PR touches is formatted, matching surrounding style (`cargo fmt --all --check` is informational per this repo's own policy — not run workspace-wide).
- [x] `cargo test` passes for every crate this PR touches, AND the full-workspace `cargo test --workspace --exclude sparq-py --no-fail-fast` (732 test-binary summaries) shows exactly one failure: `compressed_build_differential::one_pass_compressed_spill_build_matches_recompress`, confirmed pre-existing and reproducible identically on an unmodified checkout of this PR's base commit (dict-spill compressed-build, unrelated to anything touched here).

## Targeted re-evaluation

- [x] **Query execution / operators** (`sparq-engine` exec/optimizer): re-ran the full W3C SPARQL conformance suite (`cargo run -p sparq-conformance --release` against the pinned `w3c/rdf-tests`) — **1225 pass / 0 fail / 4 documented pre-existing divergences / 0 skip**, matching the existing floor exactly (the 4 divergences are unrelated spec-ambiguity cases already audited in the tool's own output, not introduced here). Also ran `cargo run -p sparq-conformance --bin sparq-conformance-scoreboard` (exit 0). Did not separately run the trend-only `operator-coverage` bench (marked `quiet_box_sensitive`, not part of the deterministic perf-gate, `featured = false`) — the change is a narrow, declining-by-default fast path with no plausible mechanism to affect the other 27 operator families it covers, and the differential suite + conformance suite already give strong coverage of the one family it does touch (ORDER BY+LIMIT).
- [x] **Public API**: `PreparedGraphApplier` is a new `pub` item in `sparq-serve`, a publishable crate with no dedicated `skills/<surface>/SKILL.md` (its own README states it "has no standalone public API surface of its own" — wrapped by `sparq-server`). Documented in `sparq-serve`'s own README in the same commit instead, per that crate's "keep this in sync with Cargo.toml" house rule. `execute_txn_with_retry` (inherited via merge, not authored in this PR's own commits) is documented in `skills/sparql-query/SKILL.md` in its own commit, satisfying the MAINTENANCE RULE directly.
- [x] **An opt-in cargo feature** (`params` on `sparq-serve`): wired into `.github/feature-matrix.d/sparq-serve.yml` + `scripts/tests/feature-matrix-legnames.golden.txt` (a verified two-file change — `diff <(python3 scripts/assemble-feature-matrix.py --names) scripts/tests/feature-matrix-legnames.golden.txt` matches). `python3 scripts/check-feature-test-execution.py --check` caught the gap before this fix (the new test file would have compiled empty in default-feature CI) and is now clean; `python3 scripts/feature-matrix-tiers.py --enforce` reports 187 legs, 0 violations.

## Ratchets and conventions

- [x] No conformance/perf/coverage ratchet lowered.
- [x] No hard-coded performance numbers added to markdown — caught one by `python3 scripts/preflight.py` (a specific measured microsecond figure in the `sparq-serve` README) and fixed it before this PR; the numbers in this PR *description* are not committed to the repo.
- [x] Follow-up work: the cross-call skip-prefix memoization idea above is not yet a bead (`bd` unavailable in this environment) — flagging here for a maintainer/agent with `bd` access to capture.
- [x] Doc accuracy: `python3 scripts/preflight.py` (G1/G2/G6/no-perf-numbers/readme-template/privacy-claims/guard-untested) is clean on this diff.
- [x] **Mutated the headline guard, per preflight's explicit instruction** (did not just reason about it): every existing differential test checks *correctness*, which the fallback path also satisfies on its own — none of them would fail if `try_topk_orderby_indexed` were deleted entirely. Added `fast_path_actually_engages_not_just_correct` (uses `EXPLAIN ANALYZE`'s execution-trace section, not the always-present static plan text — the first version of this test was itself vacuous for exactly that reason, caught by writing it and reading the output). Verified non-vacuous by temporarily forcing the new function to return `Ok(None)` unconditionally and re-running the suite: exactly the new test went red, all 14 other (correctness-only) tests stayed green. Reverted immediately; no residual diff.

## Security

- [x] No security-sensitive surface touched (`sparq-zk*`, `sparq-mpc`, `sparq-core` unsafe, `sparq-server` network) — this PR is confined to `sparq-engine`'s query executor and `sparq-serve`'s applier layer, both already-audited-shape changes (no new `unsafe`, no new external dependency, no new network/IO surface).


jeswr:
> 🤖 **SPARQ agent** — I am @jeswr's agent for the sparq-org/sparq RDF/SPARQL engine. @jeswr runs multiple agents; this was written by the SPARQ agent, not the PSS agent (prod-solid-server).

I reproduced two regressions in the indexed top-k delta from head `59bfc6c44b25a4e107d5d957ca3287ba0bc22c1a`, imported separately onto current main: a repeated-variable seed returns an invalid row, and an intermediate-row budget is bypassed by LIMIT. [Issue #6465](https://github.com/sparq-org/sparq/issues/6465) records the exact counterexample, paired main-fallback control, raw failures and repaired test evidence.

A narrow recovery is underway on an isolated local branch, preserving the existing contribution's Luke Dary / Sonnet5 attribution. It keeps PreparedGraphApplier and transaction retry helpers separate, and now declines repeated-variable, budgeted and multi-valued cases to the existing evaluator. Additional semantic/resource checks and actual independent Opus review remain ahead of publication; historical performance claims have not been re-established. This is a test-evidence update, not a merge approval.

PR #6440, opened 2026-09-07
fix(core): serialize predicate statistics in a stable order

jeswr:
> 🤖 **SPARQ agent** — I am @jeswr's agent for the sparq-org/sparq RDF/SPARQL engine. @jeswr runs multiple agents; this was written by the SPARQ agent, not the PSS agent (prod-solid-server).

An external compressed build and a reopened, recompressed store could contain identical predicate statistics but produce different `predstats.bin` bytes because the writer emitted hash-map iteration order. Sort the shared writer's entries by predicate ID. The binary record format and support for existing unordered files remain unchanged.

The regression checks exact bytes across different map capacities and insertion orders, reload/re-save stability, and reading an unordered legacy file. The existing engine differential test remains unchanged and now passes. The sort and temporary borrowed-entry vector apply only while saving statistics.

Validation: focused core statistics regressions, the engine compressed-build differential suite, full workspace tests, full workspace Clippy with warnings denied, all-feature workspace documentation with warnings denied, and Linux preflight passed against implementation commit `42147886f0e25ddc35958e40d04cbf1619e9c133`. Independent code review found no blocking issues.

The declaration-only follow-up records a manual review of paired WebAssembly artifacts: two stored panic line numbers move with the unchanged source statement; code and every other byte remain identical in that checked pair. The declaration distinguishes these inspected artifacts from CI's separate two-byte observation. The existing byte-comparison mechanism and size ratchet are unchanged, and final CI is rerunning.

### DIFF
--- a/bench/feature-off-declarations/6440.json
+++ b/bench/feature-off-declarations/6440.json
@@ -1,0 +1,75 @@
+{
+  "pr": 6440,
+  "date": "2026-09-07",
+  "reason": "[GPT-6] Canonical predicate-statistics serialization changes only the mmap-gated writer and test code. The added source lines move an unchanged indexing/unwrap statement and its embedded panic locations. An independently reviewed, same-host artifact comparison below identifies exactly two changed data bytes as line 1056 becoming 1065, at unchanged columns 29 and 39. Actual disassembly passes the corresponding records into non-returning bounds-check and Option::unwrap panic paths; every non-data section, including executable code, is byte-identical. This is an artifact-specific manual review of the retained aarch64 build pair, not the generic four-word-pattern heuristic or an automated derivation. The unavailable CI x86_64 binaries are a separate two-byte observation. No claim is made for all toolchains or for identical panic diagnostic line numbers. The byte comparator and wasm_bundle_bytes size ratchet remain unchanged.",
+  "derived": false,
+  "evidence": {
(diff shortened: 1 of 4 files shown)


jeswr:
> 🤖 **SPARQ agent** — I am @jeswr's agent for the sparq-org/sparq RDF/SPARQL engine. @jeswr runs multiple agents; this was written by the SPARQ agent, not the PSS agent (prod-solid-server).

Reviewed head 60eb72d3b690de8042029b7dd1f8736817cbee6f.

The shared writer at `crates/sparq-core/src/store.rs:699` sorts borrowed entries by their unique predicate IDs and preserves the record format and error handling. The regression at `crates/sparq-core/src/store.rs:1764` constructs expected bytes independently, varies map layout, checks reload/re-save stability, and reads an unordered legacy file. The unchanged engine differential test now passes. Full workspace tests, Clippy, all-feature documentation and Linux preflight passed for the implementation; the final follow-up adds only the declaration, and its exact-head Linux preflight also passes.

Independent review found no blocking correctness or compatibility issue. The extra allocation and sort occur only during persistence; no query-time speedup is claimed. The declaration at `bench/feature-off-declarations/6440.json:1` is supported by separately reviewed artifact hashes, exact source occurrence mapping, and actual panic callsites. It clearly separates the inspected ARM artifacts from the unavailable CI artifacts and preserves the comparator and size ratchet.

Final repository CI must pass before merge.

VERDICT: pass

